PR #15, opened 2024-02-20
kubeflow: fix go module

lampajr:
## Description

This PR aims to refactor the go module name and its references from `opendatahub-io` to `kubeflow`

## How Has This Been Tested?

```bash
make clean test
```
and
```bash
make FORCE_SERVER_GENERATION=pippo gen/openapi-server
```
to double check the server generation.

## Merge criteria:


.github/workflows/build-image-pr.yml
```diff
@@ -31,24 +31,3 @@ jobs:
...
-          echo "Deploying operator from model-registry-operator branch ${BRANCH}"
-          kubectl apply -k "https://github.com/opendatahub-io/model-registry-operator.git/config/default?ref=${BRANCH}"
-          kubectl set env -n model-registry-operator-system deployment/model-registry-operator-controller-manager REST_IMAGE="${IMG}"
-      - name: Create Test Registry
-        run: |
-          kubectl apply -k "https://github.com/opendatahub-io/model-registry-operator.git/config/samples?ref=${BRANCH}"
-          kubectl get mr
-      - name: Wait for Test Registry Deployment
-        run: |
-          kubectl wait --for=condition=Available=true modelregistries/modelregistry-sample --timeout=5m
```
lampajr, line 54: Removed the operator check as this is not required here in Kubeflow IIUC.

lampajr:
> The `Test container image build and deployment` failure will be fixed once merging #16 that would remove the operator check that is not required here.

Rebased after #16 merge

tarilabs:
/assign @rareddy 

@rareddy  (et. al), in my view, after we confirm "downstream" there are no longer pending items (we are verifying we did not "break anything" 🙃 ), this PR can be acted soon _and_ be ported down in ODH, so we start to account for proper go-module naming coordinates both in KF/ and ODH/ etc. already.

Other "debranding" such as removing these:
- https://github.com/search?q=repo%3Akubeflow%2Fmodel-registry%20%22odh.%22&type=code

instead, can be done at a later stage.

wdyt?
Hope this helps.

rareddy:
> after we confirm "downstream" there are no longer pending items
since we already have the forked into odh I do not see anymore risks here for this to go forward

>Other "debranding" such as removing these:
do we want to expand the current PR to have these included? also check the python library?

tarilabs:
> > Other "debranding" such as removing these:
> 
> do we want to expand the current PR to have these included? also check the python library?

I'd recommend not, as stated, any other "debranding":

> instead, can be done at a later stage.

rareddy:
>instead, can be done at a later stage.

imo this should be done before the first release, is there any reason in delaying this?

rareddy:
> imo this should be done before the first release, is there any reason in delaying this?

my thinking is once we have release changing the data types means invalidating any existing data or need to provide migration scripts etc. so I was thinking if it is wise to make changes right now and adjust any downstream changes on top of it.

I think any prefixes like  'kfmr` should do the trick

I am not sure how this affects the Python side of the code

tarilabs:
/lgtm
/approve

Thanks a lot @lampajr !! 🚀 💪 

Took the chances to address what mentioned in https://github.com/kubeflow/model-registry/pull/15#issuecomment-1954717645, added some overall "debranding" by replacing `odh.` prefixes in datatypes with `kfmr.` , and other minor adjustment.

Container image org/name changes can be done in a later stage once we have the ability/enable pushing to KF-identified container image registry.

PR #17, opened 2024-02-22
Fix metadata OpenAPI spec

lampajr:
## Description

This PR aims to improve the openapi spec for `MetadataValue` by properly using the `discriminator`.

## How Has This Been Tested?

Checking the generated models and server are correctly bootstrapped
```
make FORCE_SERVER_GENERATION=true clean build
```

Then by running both `unit` and  `robot` tests:

```bash
make test
```

and
```bash
# Start MLMD server
...

# Start model registry
make proxy

# In another terminal
robot test/robot/ && TEST_MODE=python robot test/robot/MRandLogicalModel.robot
```

## Merge criteria:

- [ ] The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #22, opened 2024-02-26
OAS: Fix discriminator field definition for Artifact

isinyaaa:
This fixes an improper polymorphic type definition by moving its discriminator to the composing types.

There's also a small fix for ignoring `coverage.html`, that is generated when running `make test-cover`.

## Description

This problem was spotted while using the Kiota generator, and can be traced back to [the following pattern](https://github.com/microsoft/kiota/issues/2304),
which is incompatible with JSONSchema.

## How Has This Been Tested?

Following the same steps laid out on the last PR referencing OAS #17:

```
make FORCE_SERVER_GENERATION=true clean build
make test
```

then,

1. start an MLMD server
2. run `make proxy`
3. run `robot test/robot/ && TEST_MODE=python robot test/robot/MRandLogicalModel.robot`

## Merge criteria:
* [ ]  The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.
* [x]  Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
* [x]  The developer has manually tested the changes and verified that the changes work

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -971,24 +971,26 @@ components:
     ModelArtifact:
       description: An ML model artifact.
       type: object
+      required:
+        - artifactType
+      properties:
+        artifactType:
+          type: string
+          default: "model-artifact"
       allOf:
         - $ref: "#/components/schemas/BaseArtifact"
         - $ref: "#/components/schemas/ModelArtifactCreate"
-        - type: object
-          properties:
-            artifactType:
-              type: string
-              default: "model-artifact"
     DocArtifact:
       description: A document.
       type: object
+      required:
+        - artifactType
+      properties:
+        artifactType:
+          type: string
+          default: "doc-artifact"
       allOf:
         - $ref: "#/components/schemas/BaseArtifact"
-        - type: object
-          properties:
-            artifactType:
-              type: string
-              default: "doc-artifact"
     RegisteredModel:
       description: A registered model in model registry. A registered model has ModelVersion children.
       allOf:
@@ -1105,12 +1107,12 @@ components:
       discriminator:
         propertyName: metadataType
         mapping:
-            MetadataBoolValue: '#/components/schemas/MetadataBoolValue'
-            MetadataDoubleValue: '#/components/schemas/MetadataDoubleValue'
-            MetadataIntValue: '#/components/schemas/MetadataIntValue'
-            MetadataProtoValue: '#/components/schemas/MetadataProtoValue'
-            MetadataStringValue: '#/components/schemas/MetadataStringValue'
-            MetadataStructValue: '#/components/schemas/MetadataStructValue'
+          MetadataBoolValue: "#/components/schemas/MetadataBoolValue"
+          MetadataDoubleValue: "#/components/schemas/MetadataDoubleValue"
+          MetadataIntValue: "#/components/schemas/MetadataIntValue"
+          MetadataProtoValue: "#/components/schemas/MetadataProtoValue"
+          MetadataStringValue: "#/components/schemas/MetadataStringValue"
+          MetadataStructValue: "#/components/schemas/MetadataStructValue"
       description: A value in properties.
     MetadataIntValue:
       description: An integer (int64) property value.
@@ -1345,12 +1347,6 @@ components:
       allOf:
         - $ref: "#/components/schemas/BaseArtifactCreate"
         - $ref: "#/components/schemas/BaseResource"
-        - required:
-            - artifactType
-          type: object
-          properties:
-            artifactType:
-              type: string
     ServingEnvironmentList:
       description: List of ServingEnvironments.
       type: object
--- a/internal/server/openapi/type_asserts.go
+++ b/internal/server/openapi/type_asserts.go
@@ -18,15 +18,6 @@ import (
 
 // AssertArtifactRequired checks if the required fields are not zero-ed
 func AssertArtifactRequired(obj model.Artifact) error {
-	// FIXME(manual): Artifact.ArtifactType is not present on client models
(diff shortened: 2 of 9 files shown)


dhirajsb:
@isinyaaa I noticed this PR is moving the discriminator to derived types. How does that help, or what is the difference?

api/openapi/model-registry.yaml
lampajr, line 1: I am still not sure if that is really enough or not because even with this change (which makes totally sense to me) I see:
```bash
[main] WARN  o.o.codegen.DefaultCodegen - 'Artifact' defines discriminator 'artifactType', but the referenced schema 'ModelArtifact' is incorrect. artifactType is missing from the schema, define it as required and type string
[main] WARN  o.o.codegen.DefaultCodegen - 'Artifact' defines discriminator 'artifactType', but the referenced schema 'DocArtifact' is incorrect. artifactType is missing from the schema, define it as required and type string
```
from the openapi generator logs, which seems to be fixed when adding `artifactType` as property before the `allOf` keyword:
```yaml
    ModelArtifact:
      description: An ML model artifact.
      type: object
      required:
        - artifactType
      properties:
        artifactType:
          type: string
          default: "model-artifact"
      allOf:
        - $ref: "#/components/schemas/BaseArtifact"
        - $ref: "#/components/schemas/ModelArtifactCreate"
        - type: object
          properties:
            artifactType:
              type: string
              default: "model-artifact"
...
    DocArtifact:
      description: A document.
      type: object
      required:
        - artifactType
      properties:
        artifactType:
          type: string
          default: "doc-artifact"
      allOf:
        - $ref: "#/components/schemas/BaseArtifact"
        - type: object
          properties:
            artifactType:
              type: string
              default: "doc-artifact"
```

TBH I am not that sure that having `artifactType` defined in multiple places is correct but just thinking loud here.

lampajr (commented):
Hi @isinyaaa , thanks a lot for this PR.

Overall the change looks good to me, I added a comment whether we should also add the `artifactProperty` i nthe default obj (outside the `allOf`) which seems making the WARN disappear on openapi generator side (but I don't know on Kiota side)

internal/server/openapi/type_asserts.go
```diff
@@ -83,15 +83,6 @@ func AssertBaseArtifactCreateConstraints(obj model.BaseArtifactCreate) error {
...
 func AssertBaseArtifactRequired(obj model.BaseArtifact) error {
-	elements := map[string]interface{}{
-		"artifactType": obj.ArtifactType,
-	}
-	for name, el := range elements {
-		if isZero := IsZeroValue(el); isZero {
-			return &RequiredError{Field: name}
-		}
-	}
-
```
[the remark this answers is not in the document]
tarilabs: hooray 🥳  for less patch to be managed

api/openapi/model-registry.yaml
lampajr: After some discussion with Peruffo we agreed that the following specs are equivalent each other, but keeping the `artifactType` to the leaf obj should make the `WARN` disappear on server side as well.

The spec defined in this PR where we still have the WARN on server side model generation (actually the spec is fine, we just have the WARN)
```yaml
    ModelArtifact:
      description: An ML model artifact.
      type: object
      required:
        - artifactType
      allOf:
        - $ref: "#/components/schemas/BaseArtifact"
        - $ref: "#/components/schemas/ModelArtifactCreate"
        - type: object
          properties:
            artifactType:
              type: string
              default: "model-artifact"
```

And the proposed one (here the actual final result is the same but the warning disappears)
```yaml
    ModelArtifact:
      description: An ML model artifact.
      type: object
      required:
        - artifactType
      properties:
        artifactType:
          type: string
          default: "model-artifact"
      allOf:
        - $ref: "#/components/schemas/BaseArtifact"
        - $ref: "#/components/schemas/ModelArtifactCreate"
```

To make the change I am proposing clearer I created a PR against this PR: https://github.com/isinyaaa/model-registry/pull/1

So I think it is up to us how to proceed here, if the WARN on server side is acceptable I think we can move ahead with this PR as it is.. otherwise I think the proposed change should solve the issue without changing the final code generation.

@isinyaaa can you check if this change works well with Kiota as well?

wdyt @isinyaaa @tarilabs @dhirajsb ?

tarilabs: thank you for following up on this OpenAPI topic, if the "proposing clearer" PR is acceptable sounds to me the best setup consolidating contributions from all

internal/server/openapi/type_asserts.go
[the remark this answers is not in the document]
isinyaaa: nice catch! updated.

api/openapi/model-registry.yaml
isinyaaa: that's interesting, thanks for the investigation! updated now.

isinyaaa:
> @isinyaaa I noticed this PR is moving the discriminator to derived types. How does that help, or what is the difference?

IIUC, having the discriminator field defined on `Artifact` doesn't imply that it's present on the composing types, as there's no such thing as "required properties" that have been inherited. Thus, `artifactType` should be required on all the "composing" classes.

Just updated the broken link on the description which I think also highlights the point.

lampajr (approved):
Thanks again @isinyaaa , lgtm!

tarilabs:
/lgtm

Thank you so much, this is great complementary work to make our OpenAPI contract more robust to be consumed by clients

tarilabs (approved):
/approve

PR #39, opened 2024-03-13
Construct proper S3 URIs on `register_model`

isinyaaa:
## Description
Provides an S3 URI builder util function with overloads that it can guarantee there's no metadata (e.g. from the environment) missing upon creating MR objects, either for UI or deployment purposes.

The usage of the helper is, although, optional. Proposed usage is presented on an example on the README file.

This work is based of off: https://issues.redhat.com/browse/RHOAIENG-3979

## How Has This Been Tested?

Inside the client folder, run

```py
poetry run pytest -vv --cov
```

## Merge criteria:

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -1,7 +1,7 @@
 """Standard client for the model registry."""
+
 from __future__ import annotations
 
-import os
 from typing import get_args
 from warnings import warn
 
@@ -75,16 +75,22 @@ class ModelRegistry:
         model_format_name: str,
         model_format_version: str,
         version: str,
-        author: str | None = None,
-        description: str | None = None,
         storage_key: str | None = None,
         storage_path: str | None = None,
         service_account_name: str | None = None,
+        author: str | None = None,
+        description: str | None = None,
         metadata: dict[str, ScalarType] | None = None,
     ) -> RegisteredModel:
         """Register a model.
 
-        Either `storage_key` and `storage_path`, or `service_account_name` must be provided.
+        This registers a model in the model registry. The model is not downloaded, and has to be stored prior to
+        registration.
+
+        Most models can be registered using their URI, along with optional connection-specific parameters, `storage_key`
+        and `storage_path` or, simply a `service_account_name`.
+        URI builder utilities are recommended when referring to specialized storage; for example `utils.s3_uri_from`
+        helper when using S3 object storage data connections.
 
         Args:
             name: Name of the model.
@@ -110,7 +116,7 @@ class ModelRegistry:
             version,
             author or self._author,
             description=description,
-            metadata=metadata or self.default_metadata(),
+            metadata=metadata or {},
         )
         self._register_model_artifact(
             mv,
@@ -123,19 +129,6 @@ class ModelRegistry:
         )
 
         return rm
-
-    def default_metadata(self) -> dict[str, ScalarType]:
-        """Default metadata valorisations.
-
-        When not explicitly supplied by the end users, these valorisations will be used
-        by default.
-
-        Returns:
-            default metadata valorisations.
-        """
-        return {
-            key: os.environ[key] for key in ["AWS_S3_ENDPOINT", "AWS_S3_BUCKET", "AWS_DEFAULT_REGION"] if key in os.environ
-        }
 
     def register_hf_model(
         self,
@@ -202,7 +195,6 @@ class ModelRegistry:
             model_author = author
         source_uri = hf_hub_url(repo, path, revision=git_ref)
         metadata = {
-            **self.default_metadata(),
             "repo": repo,
             "source_uri": source_uri,
             "model_origin": "huggingface_hub",
--- a/clients/python/src/model_registry/utils.py
+++ b/clients/python/src/model_registry/utils.py
@@ -1,0 +1,55 @@
+"""Utilities for the model registry."""
+
+from __future__ import annotations
+
+import os
+
+from .exceptions import MissingMetadata
+
+
+def s3_uri_from(
+    path: str,
+    bucket: str | None = None,
+    *,
+    endpoint: str | None = None,
+    region: str | None = None,
+) -> str:
+    """Build an S3 URI.
+
+    This helper function builds an S3 URI from a path and a bucket name, assuming you have a configured environment
+    with a default bucket, endpoint, and region set.
+    If you don't, you must provide all three optional arguments.
+    That is also the case for custom environments, where the default bucket is not set, or if you want to use a
+    different bucket.
+
+    Args:
+        path: Storage path.
+        bucket: Name of the S3 bucket. Defaults to AWS_S3_BUCKET.
+        endpoint: Endpoint of the S3 bucket. Defaults to AWS_S3_ENDPOINT.
+        region: Region of the S3 bucket. Defaults to AWS_DEFAULT_REGION.
+
+    Returns:
+        S3 URI.
+    """
+    default_bucket = os.environ.get("AWS_S3_BUCKET")
+    if not bucket:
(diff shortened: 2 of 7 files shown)


clients/python/src/model_registry/_client.py
```diff
@@ -75,16 +81,99 @@ def register_model(
...
+        ),
+        (  # custom env or non-default bucket
+            "storage_key",
+            "storage_path",
+            "bucket_name",
+            "bucket_endpoint",
+            "bucket_region",
+        ),
+    )
+    def register_model(
```
rareddy, line 148: Rather than having exploded version of each storage type in the overloaded method, could we do 
``` 
s3 = build_s3(); // builds a s3 object from ENV variables etc. or build manaually

register_model(self, name:str ...., s3{bucket_name, bucket_endpoint, ...})
register_model(self, name:str ...., URI)
register_model(self, name:str ...., somethingelse)

```

isinyaaa: Technically yes, but I personally think the proposed solution makes the API nicer on the user. What would motivate such a change?

rareddy: Respectfully, what you have IMO is very brittle API. Everytime if a new attribute for any reason need to be added for storage all the API will break(unless it has a default), or a new storage type added the API need to be expanded with more overloaded methods.

I can't say for python style, but If this was in java, I probably would have designed something like

```
interface storage {}
class StorageUri implements storage {}
class S3Storage implements storage {}
class OCIStorage implements storage {}

func storageBuilder():storage {
}

register_model(self, name:str ...., func storageBuilder{})

```
Another advantage is it can give way for functional programming. or no need for if/else logic in the method. above is just example, there can be better variations @tarilabs or @dhirajsb can suggest probably? since this is client facing API I want to make sure we have it correctly defined.

rareddy: >uri was required by design for register_model method
>there must NOT be "any magic" and everything need be explicit

Then we should have `URI` as a parameter in the method signature IMO. very explicit. How one builds the `URI` is irrelevant at this level. my suggestions from above taken as how to build a URI for various storage types

>An alternative to this PR current proposal could be to leave register_model untouched, as that provides 1:1 >mapping with the logical model, and then have specialized (wrapper) methods, ~like register_model_from_uri, >register_model_from_s3, etc, focusing on the actual "storage domains":

agree that this variation provides a more defined API, my view is this is another variation of the proliferation of the API, please note we need to support whatever we propose for a long time.

>less risk of having a property being specialized twice (bucket_path as part of uri and storage_path of kserve, etc)

This is KServe and S3 specific correct? can this be inferred from `URI` and query parameters? If not, force the user to provide them as query parameters by failing the validation check if it is S3?

>naturally one can always decide to consume purely the logical model -oriented register_model as is today

This I can certainly say no :) as we know abstract layers are designed to make developer's life easier.

isinyaaa: I would personally go with Matteo's suggestion as that would fit easily. However, the S3 URI builder still has at least two distinct overloads, namely for the cases where the user has:

1. default bucket on a pre-configured environment
2. custom bucket/custom environment

> I can't say for python style, but If this was in java, I probably would have designed something like
> 
> ```
> interface storage {}
> class StorageUri implements storage {}
> class S3Storage implements storage {}
> class OCIStorage implements storage {}
> 
> func storageBuilder():storage {
> }
> 
> register_model(self, name:str ...., func storageBuilder{})
> ```

Now, while this could work

> Another advantage is it can give way for functional programming. or no need for if/else logic in the method. above is just example, there can be better variations @tarilabs or @dhirajsb can suggest probably? since this is client facing API I want to make sure we have it correctly defined.

Functional programming is rather uncommon in python due to general and purposeful inefficiencies in the interpreter (https://stackabuse.com/functional-programming-in-python/#somethingstoconsider), so I wouldn't take that into consideration.

Rather, APIs in Python tend to be as declarative as possible, although I'd probably make more sense to not have such rich wrapper methods in that case.

isinyaaa: I've just updated the PR with the "simplest" version of such a builder for S3. I would say we go for it and leave API abstractions for later, wdyt?

rareddy: I like this variation much better. +1

if you want you can log "info" level messages when endpoint, path or any other optional parameters needed for automation is missing.

clients/python/src/model_registry/_client.py
```diff
@@ -75,16 +75,21 @@ def register_model(
...
     ) -> RegisteredModel:
         """Register a model.
 
-        Either `storage_key` and `storage_path`, or `service_account_name` must be provided.
+        This registers a model in the model registry. The model is not downloaded, and has to be stored prior to
+        registration.
+
+        Most models can be registered using their URI, along with optional connection-specific parameters, `storage_key`
+        and `storage_path` or, simply a `service_account_name`.
+        However it's advised to use the `utils.s3_uri_from` helper when using S3 object storage data connections.
```
tarilabs, line 92:
```suggestion
        URI builder utilities are recommended when referring to specialized storage; for example `utils.s3_uri_from` helper when using S3 object storage data connections.
```

what about something along this lines?

tarilabs (commented):
Thank you for the iterations on this @isinyaaa !

How about a complete example of using this in `clients/python/tests/test_client.py` ?

to me, already
/lgtm

isinyaaa:
> Thank you for the iterations on this @isinyaaa !
> 
> How about a complete example of using this in `clients/python/tests/test_client.py` ?

Sorry, I'll update the tests now, I missed this somehow.

clients/python/src/model_registry/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    endpoint = endpoint or os.getenv("AWS_S3_ENDPOINT")
+    region = region or os.getenv("AWS_DEFAULT_REGION")
+
+    if not (endpoint and region):
+        msg = "Missing environment variables: bucket_endpoint and bucket_region are required"
+        raise MissingMetadata(msg)
+
+    # https://alexwlchan.net/2020/s3-keys-are-not-file-paths/ nor do they resolve to valid URls
+    # FIXME: is this safe?
+    return f"s3://{bucket}/{path}?endpoint={endpoint}&defaultRegion={region}"
```
isinyaaa, line 92: bumping this note here on the last review, can you share your thoughts?

dhirajsb: Looks better for uri building. Did we also want to add the URI components as individual properties to the model metadata? Maybe we can pass a metadata object to the uri builders so that it can be reused in the register call. 
@tarilabs didn't UI say they would like the uri components individually, or are they happy to have the constructed uri?

clients/python/tests/test_client.py
```diff
@@ -31,6 +31,27 @@ def test_register_new(mr_client: ModelRegistry):
...
+    uri = utils.s3_uri_from(
+        "storage/path", "my-bucket", endpoint="my-endpoint", region="my-region"
+    )
+    rm = mr_client.register_model(
+        name,
+        uri,
+        model_format_name="test_format",
+        model_format_version="test_version",
+        version=version,
+    )
```
tarilabs, line 46: nit, but superflous; I believe here is missing storage_path, ie:
```suggestion
    rm = mr_client.register_model(
        name,
        uri,
        model_format_name="test_format",
        model_format_version="test_version",
        storage_path="storage/path",
        version=version,
    )
```

tarilabs (commented):
Thank you for having expanded on examples and tests!

per https://github.com/kubeflow/model-registry/pull/39#pullrequestreview-1948592642,
/lgtm

dhirajsb (approved):
/lgtm

PR #52, opened 2024-03-26
More controlled environment in the build

andreaTP:
Improvements to make the build self-contained.

## Description
Various minor improvements to make sure that we are using the same versions of the tools and more.

## How Has This Been Tested?
Building locally on Mac and building the container image with `podman`

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 PROJECT_BIN := $(PROJECT_PATH)/bin
+GO := $(PROJECT_BIN)/go1.19
 
 # add tools bin directory
 PATH := $(PROJECT_BIN):$(PATH)
@@ -56,7 +57,7 @@ api/grpc/ml_metadata/proto/metadata_store_service.proto:
 		sed -i 's#syntax = "proto[23]";#&\noption go_package = "github.com/kubeflow/model-registry/internal/ml_metadata/proto";#' metadata_store_service.proto
 
 internal/ml_metadata/proto/%.pb.go: api/grpc/ml_metadata/proto/%.proto
-	protoc -I./api/grpc --go_out=./internal --go_opt=paths=source_relative \
+	bin/protoc -I./api/grpc --go_out=./internal --go_opt=paths=source_relative \
 		--go-grpc_out=./internal --go-grpc_opt=paths=source_relative $<
 
 .PHONY: gen/grpc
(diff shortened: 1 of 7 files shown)


andreaTP:
Thanks to the help of @tarilabs we have verified that with those changes the build and the `docker-compose-local.yaml` are working as expected.

Makefile
```diff
@@ -204,7 +212,7 @@ endif
 # build docker image
 .PHONY: image/build
 image/build:
-	${DOCKER} build . -f ${DOCKERFILE} -t ${IMG}:$(IMG_VERSION)
+	${DOCKER} build . -f ${DOCKERFILE} -t ${IMG}:$(IMG_VERSION) --network host
```
tarilabs, line 215: tbh I'm not sure we need this change 🤔 , further it does _not_ work with Podman desktop on my end:
image

docker-compose-local.yaml
```diff
@@ -13,8 +16,13 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    command: ["proxy", "--mlmd-hostname", "localhost", "--mlmd-port", "9090"]
+      network: host
```
tarilabs, line 19: I'm not sure we need this `network: host` 🤔

docker-compose-local.yaml
```diff
@@ -13,8 +16,13 @@ services:
...
+    command: ["proxy", "--hostname", "0.0.0.0", "--mlmd-hostname", "mlmd-server", "--mlmd-port", "8080"]
```
tarilabs, line 20: confirming `"--hostname", "0.0.0.0",` is needed 😅

Makefile
```diff
@@ -147,23 +155,23 @@ clean/deps:
...
 
 .PHONY: build
 build: gen vet lint
-	go build
+	${GO} build -buildvcs=false
 
 .PHONY: build/odh
 build/odh: vet
-	go build
+	${GO} build -buildvcs=false
```
tarilabs, line 170: Was `-buildvcs=false` needed only for the build-in-docker? 🤔 Because on my end the `make build` works without issue as-is, without needing ` -buildvcs=false`

tarilabs (commented):
What I like about this PR:
- parametrize in bin/ the protoc toolchain, this was an issue with other contributors and seems this improves
- parametrize in bin/ the Go toolchain, as above
- avoid network mgt in docker-compose-local.yml

Things I haven't been able to confirm we need:
- buildvcs flag
- network: host settings/changes

per comments below.

Will add `docker-compose.yml` changes which I found beneficial when working with @andreaTP

Makefile
andreaTP: On podman Desktop on my Mac I get this error:

```
No routable interface for IPv6: IPv6 is disabled
Couldn't open network namespace /proc/4047/ns/net: Permission denied
```

docker-compose-local.yaml
andreaTP: it's used only during the build of the container and it was needed on my machine with podman to successfully build the image.

Makefile
andreaTP: it fails in CI without this flag ...

dhirajsb (approved):
/lgtm
I'm just not sure what's going on with the whole docker network thing. Is that necessary or not? Also, it's going to affect the DB as it'll have to be on the same network.

andreaTP:
> I'm just not sure what's going on with the whole docker network thing.

Using host network is not a "best practice" and with those changes, we are improving it.

> Also, it's going to affect the DB as it'll have to be on the same network.

The DB should be exposed on the docker compose network too, should we modify any additional files?

dhirajsb (approved):
/lgtm

lampajr (approved):

tarilabs:
- original PR discussed together with OP
- https://github.com/kubeflow/model-registry/pull/52#pullrequestreview-1977755359
- https://github.com/kubeflow/model-registry/pull/52#issuecomment-2049562975
- https://github.com/kubeflow/model-registry/pull/52#pullrequestreview-1994568416

/approve
/lgtm

PR #64, opened 2024-04-15
Enable TLS auth on py client

isinyaaa:
## Description
Now the python client will use TLS by default. Service account auth has been implemented only for secure routes, but will require channel intercepts otherwise, which is still pending.

## How Has This Been Tested?
Follow the [MR operator README instructions for MySQL with TLS](https://github.com/opendatahub-io/model-registry-operator/blob/main/README.md), then use the minimal example on the Python client's README.

## Merge criteria:


clients/python/src/model_registry/core.py
```diff
@@ -17,30 +21,56 @@ class ModelRegistryAPIClient:
...
-        self._store = MLMDStore(config)
+        if not user_token:
+            # /var/run/secrets/kubernetes.io/serviceaccount/token
+            sa_token = os.environ.get("KF_PIPELINES_SA_TOKEN_PATH")
+            if not sa_token:
+                msg = "Access token must be provided"
+                raise StoreException(msg)
+            token = Path(sa_token).read_bytes()
+        else:
+            token = user_token
```
tarilabs, line 44: Thank you for the update @isinyaaa !

Unless I miss something:
- this covers for sensible defaults, or user-provided, when the model registry python client is executed inside of a K8s pod/task of pipeline? 🤔 
- but this would make the `user_token` be _mandatory_ when:
  - running inside K8s Notebooks (unless I missed that `KF_PIPELINES_SA_TOKEN_PATH` is available when KF Notebook runs)
  - not running inside of K8s (I believe this is also why the Robotframework tests have been modified)

Wouldn't make more sense not to raise the exception on line 41, and maybe just emit a log line?

I'm a bit concerned user just using this client locally to connect model registry on vanilla KF would not know what to enter for `user_token`?
Can you clarify further about this please.

isinyaaa: > Wouldn't make more sense not to raise the exception on line 41, and maybe just emit a log line?

Works for me. I wasn't completely sure of the expected behavior on that case. Will update!

test/robot/ModelRegistry.py
```diff
@@ -10,7 +10,7 @@ def write_to_console(s):
 
 class ModelRegistry(mr.core.ModelRegistryAPIClient):
     def __init__(self, host: str = "localhost", port: int = 9090):
-        super().__init__(host, port)
+        super().__init__(host, port, "anything")
```
tarilabs, line 13: thank you for the updates, but pardon me if I'm missing anything, per [previous comment](https://github.com/kubeflow/model-registry/pull/64#discussion_r1579746364) I'm not understanding why this is required.

isinyaaa: it's not, I forgot to take it out.

clients/python/src/model_registry/_client.py
```diff
@@ -16,28 +16,24 @@ class ModelRegistry:
...
     def __init__(
         self,
-        server_address: str,
-        port: int,
         author: str,
-        client_key: str | None = None,
-        server_cert: str | None = None,
-        custom_ca: str | None = None,
+        server_address: str,
+        port: int = 443,
```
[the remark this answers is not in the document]
isinyaaa: Updated, but now the author needs to be a kwarg as it comes after an optional port.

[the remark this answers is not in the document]
tarilabs: understood, we could default it to some default value like `"unknown"` if any user complains, wdyt ?

clients/python/src/model_registry/core.py
```diff
@@ -17,30 +22,58 @@ class ModelRegistryAPIClient:
...
-        config.host = server_address
-        config.port = port
-        if client_key is not None:
-            config.ssl_config.client_key = client_key
-        if server_cert is not None:
-            config.ssl_config.server_cert = server_cert
-        if custom_ca is not None:
-            config.ssl_config.custom_ca = custom_ca
-        self._store = MLMDStore(config)
+        if not user_token:
```
rareddy, line 37: code under here seems read confusing to me. are we saying when port is 443 it is secure? or having a ca_cert is secure? some code reorg could be better imo

isinyaaa: I provided two separate builders in the last commit, wdyt?

clients/python/src/model_registry/core.py
```diff
@@ -2,45 +2,109 @@
...
+            custom_ca: The PEM-encoded root certificates as a byte string. Defaults to envvar CERT.
+        """
+        if not custom_ca:
+            ca_cert = os.environ.get("CERT")
+            if not ca_cert:
+                msg = "CA certificate must be provided"
+                raise StoreException(msg)
+            root_certs = Path(ca_cert).read_bytes()
+        else:
+            root_certs = custom_ca
```
rareddy, line 49: Since `secureconnection` or `insecureconnection` is based on this block, I would pull this up before the connection is determined. There is no need to check port is 443 or not IMO. so, such as

```
if root_certs:
   conn = buildSecureConnection
else:
  conn = insuecureConnection
```

clients/python/src/model_registry/core.py
```diff
@@ -2,45 +2,109 @@
...
+            else:
+                warn("User access token is missing", stacklevel=2)
+
+        if user_token:
+            chan_creds = grpc.composite_channel_credentials(
+                chan_creds,
+                grpc.access_token_call_credentials(user_token),
+            )
+
+        if port != 443:
```
rareddy, line 66: Based on the logic the code will never reach here!

clients/python/src/model_registry/core.py
```diff
@@ -2,45 +2,109 @@
...
-        if custom_ca is not None:
-            config.ssl_config.custom_ca = custom_ca
-        self._store = MLMDStore(config)
+        if not user_token:
+            # /var/run/secrets/kubernetes.io/serviceaccount/token
+            sa_token = os.environ.get("KF_PIPELINES_SA_TOKEN_PATH")
+            if sa_token:
+                user_token = Path(sa_token).read_bytes()
+            else:
+                warn("User access token is missing", stacklevel=2)
```
rareddy, line 96: You need to pull this code up, as it is duplicated in two different methods, into a single code block.  log should say, user_token missing so it is using the sa token which is what we are allowing.

rareddy (changes requested):

clients/python/src/model_registry/core.py
isinyaaa: the user can build an API service without using the high-level client. that's what I'm also doing on the Robot tests setup. Equally could have built a secure connection on another port.

clients/python/src/model_registry/_client.py
```diff
@@ -17,27 +19,55 @@ class ModelRegistry:
...
+        root_certs = None
+        if not custom_ca:
+            ca_cert = os.environ.get("CERT")
+            if ca_cert:
+                root_certs = Path(ca_cert).read_bytes()
+            elif port == 443:
+                warn(
+                    "missing CA certificate, which is required for a secure connection",
+                    stacklevel=2,
+                )
```
rareddy, line 59: this is the only code I do not understand. anyways, shouldn't this throw an exception here?

clients/python/README.md
```diff
@@ -10,7 +10,7 @@ This library provides a high level interface for interacting with a model regist
 ```py
 
-registry = ModelRegistry(server_address="server-address", port=9090, author="author")
+registry = ModelRegistry("server-address", author="Ada Lovelace")  # Defaults to port 443
```
dhirajsb, line 13: @isinyaaa can we document both TLS and non-TLS usage in the README?

clients/python/src/model_registry/_client.py
```diff
@@ -17,27 +19,55 @@ class ModelRegistry:
...
         self,
         server_address: str,
-        port: int,
+        port: int = 443,
+        *,
         author: str,
-        client_key: str | None = None,
-        server_cert: str | None = None,
-        custom_ca: str | None = None,
+        user_token: bytes | None = None,
```
dhirajsb, line 25: Not a requirement right now, but we should open another issue to add support for client certificate as well for mutual authentication.

clients/python/src/model_registry/_client.py
```diff
@@ -17,27 +19,55 @@ class ModelRegistry:
...
+                root_certs = Path(ca_cert).read_bytes()
+            elif port == 443:
+                warn(
+                    "missing CA certificate, which is required for a secure connection",
+                    stacklevel=2,
+                )
+        else:
+            root_certs = custom_ca
+
+        if root_certs:
```
dhirajsb, line 63: For POSIX systems, there is a default list of certificate authorities as well. This could be a use case where user creates a server cert using a public CA. 
So, there is a third use case that's supported by [gRPC authentication](https://grpc.io/docs/guides/auth/#using-client-side-ssltls) where a custom CA is not specified, but an auth token is provided.

dhirajsb (changes requested):

clients/python/src/model_registry/_client.py
```diff
@@ -17,27 +19,50 @@ class ModelRegistry:
...
+        if not user_token:
+            # /var/run/secrets/kubernetes.io/serviceaccount/token
+            sa_token = os.environ.get("KF_PIPELINES_SA_TOKEN_PATH")
+            if sa_token:
+                user_token = Path(sa_token).read_bytes()
+            else:
+                warn("User access token is missing", stacklevel=2)
+
+        root_ca = None
+        if not custom_ca and port == 443:
```
rareddy, line 51: why are you hard coding `port == 443` is secure, 443 is just a default port that supports encrypted channels. it is user driven it can be anything like 1443. so not hard code anything.

also what does "secure" mean in your code, is it having an encrypted channel considered here as secure or user authentication as secure, or a combination of both? make that intent clear in your code. imo, when ca is given it a `encrypted` channel. user-token should be available both ways, as we never leave the connection not authenticated. if user-token we should just fail. @dhirajsb

clients/python/src/model_registry/core.py
```diff
@@ -2,45 +2,82 @@
...
+        """
+        # custom_ca = None will get the default root certificates
+        chan_creds = grpc.ssl_channel_credentials(custom_ca)
+        if user_token:
+            chan_creds = grpc.composite_channel_credentials(
+                chan_creds, grpc.access_token_call_credentials(user_token)
+            )
+
+        if port != 443:
+            warn(f"Using non-standard port for TLS connection {port}", stacklevel=2)
```
rareddy, line 48: IMO now you can safely remove this warning.

clients/python/src/model_registry/_client.py
```diff
@@ -17,27 +19,55 @@ class ModelRegistry:
...
+            sa_token = os.environ.get("KF_PIPELINES_SA_TOKEN_PATH")
+            if sa_token:
+                user_token = Path(sa_token).read_bytes()
+            else:
+                warn("User access token is missing", stacklevel=2)
+
+        if is_secure:
+            root_ca = None
+            if not custom_ca:
+                if ca_path := os.getenv("CERT"):
```
rareddy, line 55: The way I see the code should flow is

```
if is_secure:
  if custom_ca:
     root_ca = custom_ca
  else
    root_ca = ...

  if not root_ca:
     throw exception; 

  create secure  conn
else 
  create insecure conn
```

isinyaaa: We can't throw an exception here because we have to let the Credentials object builder look for default CA paths on the system: https://github.com/kubeflow/model-registry/pull/64/files#diff-29690c8d2a6e6330eb8d28edc4eab0809f3a51fac3f0804578a09a51a7484cb7R46.

dhirajsb (approved):
/lgtm

rareddy (approved):
/lgtm

PR #74, opened 2024-04-29
Handle REST status codes

isinyaaa:
Closes: https://github.com/opendatahub-io/model-registry-bf4-kf/issues/305

## Description
This handles REST status codes coming from the server core layer using enums and error unwrapping as [introduced in GO 1.13](https://go.dev/blog/go1.13-errors).

## How Has This Been Tested?
Tested with httpie and go tests.

## Merge criteria:

### DIFF
--- a/internal/server/openapi/api_model_registry_service_service.go
+++ b/internal/server/openapi/api_model_registry_service_service.go
@@ -47,7 +47,7 @@ func (s *ModelRegistryServiceAPIService) CreateEnvironmentInferenceService(ctx c
 func (s *ModelRegistryServiceAPIService) CreateInferenceService(ctx context.Context, inferenceServiceCreate model.InferenceServiceCreate) (ImplResponse, error) {
 	entity, err := s.converter.ConvertInferenceServiceCreate(&inferenceServiceCreate)
 	if err != nil {
-		return Response(500, model.Error{Message: err.Error()}), nil
+		return Response(400, model.Error{Message: err.Error()}), nil
 	}
 
 	result, err := s.coreApi.UpsertInferenceService(entity)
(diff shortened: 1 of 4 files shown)


internal/server/openapi/error.go
```diff
@@ -15,10 +15,8 @@ import (
 	"net/http"
 )
 
-var (
-	// ErrTypeAssertionError is thrown when type an interface does not match the asserted type
-	ErrTypeAssertionError = errors.New("unable to assert type")
-)
+// ErrTypeAssertionError is thrown when type an interface does not match the asserted type
+var ErrTypeAssertionError = errors.New("unable to assert type")
```
tarilabs, line 19: I'm a bit perplexed why this generated file has been modified, and allegedly is not re-generated throwing error during GHA for uncommitted modifications; can you elaborate, please?

internal/server/openapi/api_model_registry_service_service.go
```diff
@@ -67,12 +70,17 @@ func (s *ModelRegistryServiceAPIService) CreateInferenceServiceServe(ctx context
 
 	result, err := s.coreApi.UpsertServeModel(entity, &inferenceserviceId)
 	if err != nil {
+		if errors.Is(err, api.ErrBadRequest) {
+			return Response(400, model.Error{Message: err.Error()}), nil
+		}
+		if errors.Is(err, api.ErrNotFound) {
+			return Response(404, model.Error{Message: err.Error()}), nil
+		}
```
tarilabs, line 78: eventually, there can be a mapping function as discussed which for known-errors will return a `Response` if mapped or nil, so we can refactor similar code here.

tarilabs (commented):
Thank you @isinyaaa !

I prefer for this PR to consider changing return error codes also for failed conversions like `s.converter ... ` in `api_model_registry_service_service.go` instead of 500 to return 400 for bad request, as discussed.

I'm not sure I understood `internal/server/openapi/error.go` modification, per comment below.

The mapping function, can be done in a subsequent PR.

internal/server/openapi/error.go
isinyaaa: I had previously tried to introduce the new error codes in there, as discussed on last Wednesday's call. Missed the autoformatter change, will remove it.

tarilabs (approved):
Thank you for incorporating required improvements and changes from https://github.com/kubeflow/model-registry/pull/74#pullrequestreview-2030284205 @isinyaaa 

/approve
/lgtm

PR #91, opened 2024-05-13
GHA: test python client connection

isinyaaa:
## Description
Test Python client connectivity with the latest client.

## How Has This Been Tested?
Tested by deploying a KinD cluster locally using the same setup.

## Merge criteria:


test/python/test_mr_conn.py
```diff
@@ -0,0 +1,26 @@
...
+        model_format_name="onnx",
+        model_format_version="1",
+        storage_key="my-data-connection",
+        storage_path="path/to/model",
+        metadata={
+            "day": 1,
+            "split": "train",
+        },
+    )
+    print(mr.get_registered_model("my-model"))
```
rareddy, line 20: shouldn't this be an assert of some kind?

isinyaaa: we can do that for sure. I just thought it'd be meaningless considering we're already testing that extensively on the unit tests -- so this one can only fail if the calls don't complete correctly. Either way we can double cover.

tarilabs (approved):
/lgtm

to start with, and we can always progressively iterate, for instance if QE folks could help us extend the RobotFramework testing which already incorporates some Python SDK calls

/approve

PR #93, opened 2024-05-15
Handle errors gracefully on routes

isinyaaa:
## Description
This fixes an unhandled error on the GetEnvironmentEnvironments route, as well as standardizes error value treatment using an error to http status converter. All status codes now use the standard net/http consts (enum).

## How Has This Been Tested?
Tested with `make test`.

## Merge criteria:

### DIFF
--- a/internal/server/openapi/api_model_registry_service_service.go
+++ b/internal/server/openapi/api_model_registry_service_service.go
@@ -12,6 +12,7 @@ package openapi
 import (
 	"context"
 	"errors"
+	"net/http"
 
 	"github.com/kubeflow/model-registry/internal/apiutils"
 	"github.com/kubeflow/model-registry/internal/converter"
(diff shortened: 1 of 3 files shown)


internal/server/openapi/api_model_registry_service_service.go
```diff
@@ -405,50 +339,41 @@ func (s *ModelRegistryServiceAPIService) GetModelVersionArtifacts(ctx context.Co
 	// TODO externalID unused
 	listOpts, err := apiutils.BuildListOption(pageSize, orderBy, sortOrder, nextPageToken)
 	if err != nil {
-		return Response(500, model.Error{Message: err.Error()}), nil
+		return Response(http.StatusInternalServerError, model.Error{Message: err.Error()}), nil
```
tarilabs, line 342: the only way `apiutils.BuildListOption` could err is user supplying a wrong content for pageSize,
so I would suggest this fn (`apiutils.BuildListOption`) should follow the same pattern,
i.e.: 
- make it return a semantic error,
- and use the conversion function. 

wdyt?
let me know if you want to do it as part of this PR, or in a following one (in which case on notification I'll proceed to merge this now)

tarilabs (commented):
thank you @isinyaaa for this PR,
let me know if the comment below you want to do it as part of this PR or future PR

internal/server/openapi/api_model_registry_service_service.go
isinyaaa: will update on this one, then. Thanks for the heads up!

rareddy:
Do we want to do anything with left over "todo" comments? if no further steps we are going to like 401 ones which get intercepted at filter level, can we remove them?

tarilabs:
> Do we want to do anything with left over "todo" comments? if no further steps we are going to like 401 ones which get intercepted at filter level, can we remove them?

I'd advise to manage the TODOs as follow-up PR, so to keep the scope of this PR to start consolidating repeated code, while "debt" to be managed separately

tarilabs (approved):
looks good to me @isinyaaa thank you for the quick loop, I'd say we merge this and we track the TODO for the auth in a single filter in a subsequent PR; reason is I would like to parallelize work since a reported issue (#95 ) could then investigated in parallel better with this one merged already so we set a baseline

/lgtm
/approve

PR #115, opened 2024-05-30
chore(deps): py bump docker ^7.1.0, requests ^2.32.2

tarilabs:
Resolves https://github.com/kubeflow/model-registry/security/dependabot/10

## Description
1. bump requests ^2.32,
2. at the same bump docker ^7.1.0 or else testcontainers would fail due to docker mgt error with `requests.exceptions.InvalidURL: Not supported URL scheme http+docker`; see more: `https://github.com/psf/requests/issues/6707#issuecomment-2121103400`

- pinning mlmd to 1.14.0 to avoid undesired updates (see more: https://github.com/kubeflow/model-registry/pull/83#pullrequestreview-2052052648)

- adding Mac OSX specific instruction since Testcontainers for Python make use of Ryuk

## How Has This Been Tested?
Example1:
Linux, no specific settings for docker engine
Screenshot from 2024-05-30 08-59-51

Example2:
Mac OSX, required settings for Testcontainers (need special permission for Ryuk, documented)
image

## Merge criteria:


CONTRIBUTING.md
```diff
@@ -71,6 +71,20 @@ and now you can substitute `gmake` every time the make command is mentioned in g
...
+You can install Podman from the Podman [website](https://podman.io/).
+
+We recommend setting up the Podman machine with root priviledges,
+and setting the environment variable
+
+```sh
+export TESTCONTAINERS_RYUK_PRIVILEGED=true
+```
+
+when running TestContainer-based Model Registry Python tests (for more information, see [here](https://pypi.org/project/testcontainers/#:~:text=TESTCONTAINERS_RYUK_PRIVILEGED)).
```
isinyaaa, line 86: IIRC by exporting the env variable we're already following the recommended usage, so I'd rewrite the recommendation, mainly taking out the "and" (as it sounds like there is something additional to be done when using the Podman machine in that sense). Also, nit (typo).

```suggestion
We recommend using a [privileged Podman machine](https://pypi.org/project/testcontainers/#:~:text=TESTCONTAINERS_RYUK_PRIVILEGED)) when running TestContainer-based Model Registry Python tests.
You can do that by setting the following environment variable

```sh
export TESTCONTAINERS_RYUK_PRIVILEGED=true
```

```

**Review by isinyaaa — APPROVED** [review]
Thanks for the update! Left some nits as suggestions, also a heads up to improve accessibility is to [put meaningful text on links](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Creating_hyperlinks#use_clear_link_wording).

File: CONTRIBUTING.md (hunk 1 above)
Comment (tarilabs) [reply CONTRIBUTING.md:86]: In my understanding they're not the same thing, as this proposed correction would make it look like (since the "podman machine" is now linking to Ryuk container privileges).

In my understanding, one thing is the emulation used to run containers in general on macOsx, for podman that is the "podman machine"; another thing is the Ryuk container, used by TestContainers as a ~"garbage collector" which need to opt-in to this privileges.

I have podman machine as root by default,
but if I miss to specify the environment variable for ryuk specifically it won't work for Testcontainers.

In conclusion, to me it's indeed an "and".

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -71,6 +71,19 @@ and now you can substitute `gmake` every time the make command is mentioned in g
 Several options of docker engines are available for Mac.
 Having Docker installed is also helpful for Testcontainers.
 
+### Podman
+
+[Podman](https://podman.io/) also supports Rosetta (Apple specific) emulation which is handy since the Google MLMD project dependency is x86 specific.
+
+We recommend setting up the Podman machine with root priviledges,
```
tarilabs, line 78:
```suggestion
We recommend setting up the Podman machine with root privileges,
```

CONTRIBUTING.md
isinyaaa: thanks for the heads up!

rareddy (approved):
/lgtm

PR #136, opened 2024-06-11
chore(py): avoid use of assert for business logic

tarilabs:
Avoid use of python `assert` use in application/runtime logic.

## Description
In a recent meetup ([Python Milano](https://www.youtube.com/live/-pHvc-xJE70?si=_Vhu6EVVqVAjgsqu&t=1146)) I've become aware use of `assert` in business logic is not robust.
In my understanding, this is the dual/analogous of the JVM: for Java is opt-in (`-ea`) while for Python is opt-out (`-O` or similar).
So it's not advisable to make use of `assert` in control-flow, as it's environment-dependent.

This PR rewrites with explicit checks using conditional instead of asserts.

This imho is relevant for environments we might not have control/visibility on, such as within a pipeline step (a container) or a custom/bespoke user environment.

## How Has This Been Tested?
- `nox -s lint-3.10`
- `pytest`
image

## Merge criteria:

cc: @isinyaaa wdyt?

### DIFF
--- a/clients/python/pyproject.toml
+++ b/clients/python/pyproject.toml
@@ -85,11 +85,11 @@ select = [
 ignore = [
     "D105", # missing docstring in magic method
     "E501", # line too long
-    "S101", # use of assert detected
 ]
 mccabe.max-complexity = 8
 per-file-ignores = { "tests/**/*.py" = [
     "D", # missing docstring in public module
+    "S101", # use of assert detected
 ] }
 
 [tool.ruff.lint.pydocstyle]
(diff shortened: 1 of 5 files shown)


isinyaaa:
Hey Matteo, thanks for the PR. I like to avoid raising errors on inside layers unless the user is supposed to see them. In that regard, there's (an attempt) at concealing code assumptions (as assertions) so that the final user should never see them either way. But if you prefer to have those guarantees shipped in the final product even on package builds, I can see your point. However I'd like to point out that we also have `AssertionError` in Python, to mimic exactly that use-case. Feel free to use the error types as appropriate.

isinyaaa (approved):

tarilabs:
superseeded by https://github.com/kubeflow/model-registry/pull/139

PR #137, opened 2024-06-11
Add readthedocs config

isinyaaa:
## Description
Adds support for using readthedocs for the Python client. To use the service a maintainer needs to import the project on readthedocs.io, and possibly also set up webhooks.

Note: it's possible to move the config to a subdirectory (e.g. `clients/python`), although that's a beta feature.

## How Has This Been Tested?

Pushed the documentation as a test on my fork: https://test-model-registry.readthedocs.io/en/latest/index.html

## Merge criteria:


isinyaaa:
@tarilabs should we remove docs build from the CI tests? It should only be helpful to know if a build failed.

tarilabs (commented):
Thank you for this PR @isinyaaa !

I have 2 questions please? 
 
q1.
wouldn't it be better to keep using poetry, so we don't have to keep aligned versions?
e.g.:
- `https://github.com/readthedocs/readthedocs.org/issues/4912#issuecomment-1992286540`
- https://browniebroke.com/blog/specify-docs-dependency-groups-with-poetry-and-read-the-docs/

q2.
this would still require (this) repo admin access to configure a webhook? 
I mean, this still doesn't make the external doc system periodically check for us, afaik. 
Is there a way we could trigger from a GHA? 🤔

isinyaaa:
I'll update for using poetry. The webhook setup seems to be the best way to fetch updates from GitHub, at least according to the RTD docs. We can set up automations rules via the platform, e.g. https://docs.readthedocs.io/en/stable/automation-rules.html#examples. But it doesn't seem to make sense to have a docs build GHA, because readthedocs always builds on their end (automation only helps it recognize the target). When importing the project from GitHub they already set up webhooks automagically, the same way they ask for the [manual setup to be executed](https://docs.readthedocs.io/en/stable/guides/setup/git-repo-manual.html).

We can however call in the [Generic API integration](https://docs.readthedocs.io/en/stable/guides/setup/git-repo-manual.html#webhook-integration-generic) from GHA.

FYI they have a GHA for previewing docs on PRs, which could also be useful, although it shouldn't work using the generic endpoint approach: https://github.com/readthedocs/actions/tree/v1/preview

There's also a way to set up [webhooks for build failure notifications](https://docs.readthedocs.io/en/stable/build-notifications.html)

tarilabs (approved):
/lgtm
/approve

PR #150, opened 2024-06-21
Dependency Update: bump go from 1.19 to 1.21@latest

tonyxrmdavidson:
This commit will update go from go1.19 to go1.21@latest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 PROJECT_BIN := $(PROJECT_PATH)/bin
-GO := $(PROJECT_BIN)/go1.19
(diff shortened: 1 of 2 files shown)


tarilabs (commented):
@tonyxrmdavidson thanks, looking here: https://go.dev/dl/
based on the error messages,
I suspect it shall actually specify `.0`, as in `1.21.0`

try that maybe? wdyt?

tarilabs:
(going forward, we should actually revert using bin toolchain and GOBIN, per other tracker)

tonyxrmdavidson:
> @tonyxrmdavidson thanks, looking here: https://go.dev/dl/ based on the error messages, I suspect it shall actually specify `.0`, as in `1.21.0`
> 
> try that maybe? wdyt?

should that be 1.21.9 and not 1.21.0?

tarilabs:
I'm okay with any microversion 👍

tarilabs (approved):
/lgtm

thanks @tonyxrmdavidson 

/approve

PR #152, opened 2024-06-21
Switch Python client backend to use REST

isinyaaa:
## Description
Use openapi-generator (the same used in the Go server) to create a sister library with REST bindings.
First, we get rid of the gRPC backend, remove attrs in favor of pydantic, then make the core layer async and wrap a sync client for the high-level interface.

## How Has This Been Tested?
The test suite was entirely updated to use the new backend.

Also tested with a vanilla kubeflow deployment following https://www.kubeflow.org/docs/components/model-registry/getting-started/.

## Merge criteria:


isinyaaa:
cc @tarilabs

tarilabs:
Thank you @isinyaaa , could you kindly check on your end for the Python tests in:
- Test container image build and deployment / build-and-test-image (pull_request) Failing after 4m
[Details](https://github.com/kubeflow/model-registry/actions/runs/9619911266/job/26537044258?pr=152)
- run-robot-tests / run-robot-tests (pull_request) Failing after 2m
[Details](https://github.com/kubeflow/model-registry/actions/runs/9619911260/job/26537043880?pr=152)

I suppose is only a matter of aligning the test to match the new impl, maybe just port settings?

tarilabs (commented):
I like very much the approach of using the git patch for fixing the openapi-codegen issues, thanks!

I given a first overall look and to me this is going in an awesome direction, I have only one concern about the use of `async`, I can understand this make it surface the ability to invoke operations asynchronously from callee, but most SDK in KF do not work this way. My preference here would be to make the "default user-facing methods" not coloured with `async` and if anything, provide a variant of the same for "advanced users/use-case" as an alternative. wdyt?

tarilabs:
Optional, but I would recommend to add `requests` to `tool.poetry.group.dev.dependencies`. I know it's a transitive to huggingface-hub anyway, but this latter dep is an extra, and if anyone (like me 😅 ) just did a banal poetry-install without the all-extras, you end up in a situation where you can't run the tests.

tarilabs:
Another note for this/immediately-subsequent-PR, I'm getting the following warning when creating the MR python client, while testing out on KF 1.9-rc.2:
```
[/opt/conda/lib/python3.11/site-packages/pydantic/_internal/_fields.py:161](http://localhost:8080/opt/conda/lib/python3.11/site-packages/pydantic/_internal/_fields.py#line=160): UserWarning: Field "model_format_name" has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
  warnings.warn(
[/opt/conda/lib/python3.11/site-packages/pydantic/_internal/_fields.py:161](http://localhost:8080/opt/conda/lib/python3.11/site-packages/pydantic/_internal/_fields.py#line=160): UserWarning: Field "model_format_version" has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
```

tarilabs:
Finally, we will need to update the readme + KF-website for an example that suggest:
```py
registry = ModelRegistry("http://server-address", author="Ada Lovelace")
```

i.e. including the `http://` prefix, because that is likely to get it wrong (or alternatively, figure out by heuristic of the port number)

tarilabs:
I've tested the poetry build both on Arm (Mac) and x86, and poetry to run the tests.

I've also tested a local built whl on KF `1.9.0-rc.2`, some screenshots:

Screenshot from 2024-07-02 14-09-49

Screenshot from 2024-07-02 14-10-19

Screenshot from 2024-07-02 14-10-32

Attached for details:
[test-20240702.zip](https://github.com/user-attachments/files/16067890/test-20240702.zip)

With the small remarks as mentioned earlier of:
- https://github.com/kubeflow/model-registry/pull/152#issuecomment-2202015759
- https://github.com/kubeflow/model-registry/pull/152#issuecomment-2202962838
- https://github.com/kubeflow/model-registry/pull/152#issuecomment-2202979395

To me, for as far as I can see and I could test, it's working as intended.

Thank you so much for the great work @isinyaaa and sorry this took me a while before I was in a position to thoroughly testing it :rocket: 

/lgtm

isinyaaa:
@tarilabs Thanks for the review!

Addressed all comments so far and updated some documentation on the Core layer that was out of date.
Hopefully the python release action is good as is, although we might want to verify that the generation is up to date in some CI step like we do with the go server.

rareddy:
I tested with your example zip wheel and scripts @tarilabs and all seem to work well. Thank you for the scripts. Does this work with secure cases as well @isinyaaa ??

Not related to this PR, but I found Python API for the model registry itself is limited and not expressive. For example, did not find a way to list all "registered models" or "artifacts". Maybe there should be a "deploy" method, that can wrap the kserve or give variance to deploy to other environments to wrap methods. latter one is a suggestion, the prior one I feel is needed in this API.

isinyaaa:
> I tested with your example zip wheel and scripts @tarilabs and all seem to work well. Thank you for the scripts. Does this work with secure cases as well @isinyaaa ??

Tested with the latest version using the mysql-tls samples, works fine.

> Not related to this PR, but I found Python API for the model registry itself is limited and not expressive. For example, did not find a way to list all "registered models" or "artifacts". Maybe there should be a "deploy" method, that can wrap the kserve or give variance to deploy to other environments to wrap methods. latter one is a suggestion, the prior one I feel is needed in this API.

I'm looking into that next, I agree those are much needed methods. We only have to think about how to handle pagination control to the user, as even the current `ListOptions` don't support it yet.

rareddy:
> about how to handle pagination
 
I would ignore handling the pagination to begin with providing the raw API, then you could add optional parameters later to handle any such extensions.

tarilabs:
re-instating 
/lgtm
per https://github.com/kubeflow/model-registry/pull/152#issuecomment-2203027362

rareddy (approved):
/lgtm
/approve

Issue #95, opened 2024-05-15
/api/model_registry/v1alpha3/registered_models responseBody for 400 is not following json schema


ederign:
**Describe the bug**
/api/model_registry/v1alpha3/registered_models responseBody for 401 is not following json schema

**To Reproduce**
1.  Run:
```
curl -i  -X POST "localhost:8080/api/model_registry/v1alpha3/registered_models" \
     -H "Content-Type: application/json" \
     -d '{
  "description": "dora description1",
  "externald": "9918",
  "name": "dora1",
  "owner": "eder",
  "stat1e": "LIVE", "lastUpdateTimeSinceEpoch": "1"
}'

You get the response: 

HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=UTF-8
Vary: Origin
Date: Wed, 15 May 2024 19:26:43 GMT
Content-Length: 36

"json: unknown field \"externald\""
```

### 
**Expected behavior**
```
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=UTF-8
Vary: Origin
Date: Wed, 15 May 2024 19:26:43 GMT
Content-Length: 36

{
  "code": "HTTP Code",
  "message": "Some error message"
}
```

**Additional context**

Also, for some errors, the "code" is being empty and not populated with error code. I'm able to reproduce this with error 500 (running the same post twice).

ederign:
If you confirm this is a valid issue, I can try to send a PR to fix it.

tarilabs:
> If you confirm this is a valid issue, I can try to send a PR to fix it.

I'm a bit perplexed why it didn't return you `openapi.Error` 🤔 

I'd advise for this _after_ https://github.com/kubeflow/model-registry/pull/93

tarilabs:
btw,

> Also, for some errors, the "code" is being empty and not populated with error code. I'm able to reproduce this with error 500 (running the same post twice).

the code in the body was not meant to reflect the http-status-code, but a placeholder if UI wanted to define their own codes.

isinyaaa:
Looking at the example there seem to be 2 typos. I'm also afraid `lastUpdateTimeSinceEpoch` isn't/wasn't a user-settable field, at least not in this call.

```diff
 curl -i  -X POST "localhost:8080/api/model_registry/v1alpha3/registered_models" \
      -H "Content-Type: application/json" \
      -d '{
   "description": "dora description1",
-  "externald": "9918",
+  "externalId": "9918",
[...]

PR #235, opened 2024-07-31
wrap default openapi error handler

isinyaaa:
## Description
Resolves: #95

## How Has This Been Tested?
Tested with making POST/PATCH API calls with random missing/mispelled fields.

## Merge criteria:

### DIFF
--- a/internal/server/openapi/helpers.go
+++ b/internal/server/openapi/helpers.go
@@ -11,6 +11,8 @@ package openapi
 
 import (
 	"reflect"
+
+	model "github.com/kubeflow/model-registry/pkg/openapi"
 )
 
 // Response return a ImplResponse struct filled
@@ -18,6 +20,13 @@ func Response(code int, body interface{}) ImplResponse {
 	return ImplResponse{
 		Code: code,
 		Body: body,
+	}
+}
+
+func ErrorResponse(code int, err error) ImplResponse {
+	return ImplResponse{
+		Code: code,
+		Body: model.Error{Message: err.Error()},
 	}
 }
 
--- a/.openapi-generator-ignore
+++ b/.openapi-generator-ignore
@@ -45,3 +45,5 @@ internal/server/openapi/.openapi-generator-ignore
 internal/server/openapi/api_model_registry_service_service.go
 internal/server/openapi/README.md
 internal/server/openapi/main.go
+internal/server/openapi/error.go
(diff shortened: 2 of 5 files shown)


internal/server/openapi/helpers.go
```diff
@@ -21,6 +23,13 @@ func Response(code int, body interface{}) ImplResponse {
 	}
 }
 
+func Error(code int, err error) ImplResponse {
```
tarilabs, line 26: to avoid confusing this, should we name this `ErrorResponse` ? wdyt?

tarilabs (commented):
Thank you @isinyaaa !

I believe you must include:
- `internal/server/openapi/error.go`
- `internal/server/openapi/helpers.go`

in the file `.openapi-generator-ignore`, or otherwise next generation if the OpenAPI file is changed would overwrite your changes.

Can you check this before we merge this PR please?

The other suggestion below is optional.

isinyaaa:
> Thank you @isinyaaa !
> 
> I believe you must include:
> 
> * `internal/server/openapi/error.go`
> * `internal/server/openapi/helpers.go`
> 
> in the file `.openapi-generator-ignore`, or otherwise next generation if the OpenAPI file is changed would overwrite your changes.

It doesn't seem to be the case that these files are updated at all. Trying to regen manually with `FORCE_SERVER_GENERATION=1 make` yields no diff, but I can include them in the ignore list to be on the safe side.

Will update the suggestion.

tarilabs:
> It doesn't seem to be the case that these files are updated at all. Trying to regen manually with FORCE_SERVER_GENERATION=1 make yields no diff,

You right, very strange 🤔 ; also tried on this end by removing the file, it is indeed not re-generated.
Could be it is one of the files which was only copy-pasted originally, not sure.

Thanks anyway for adding it to the ignore list, at least we have 1 source-of-truth for files which we have amended!

I just noticed now, this claims to resolve an issue but we didn't add any tests.
Do you think you can add it, please?

Since this is on the "REST surface" I believe a Robot test would be the simplest to do, but open to additional suggestions

tarilabs (approved):
I'm glad it's automatically picked up by https://github.com/kubeflow/model-registry/blob/main/.github/workflows/run-robot-tests.yaml#L57 without changes, thank you for having split into "negative test" file 👍 

/lgtm
/approve

PR #252, opened 2024-08-06
Add Dockerfile for serving containerised frontend

alexcreasy:
## Description

This PR allows the frontend to be served in a container using nginx and adds:
  * Dockerfile for serving the frontend and proxying calls to the bff.
  * An example docker-compose for running the stack.
  * Webpack config for proxying and url rewriting for the frontend in dev mode.
  * A frontend Makefile for building the docker image

## How Has This Been Tested?

* Run the example docker-compose file: `podman compose up -d` (swap podman for docker if that's your bag)
* Verify the frontend loads in a browser at: `localhost:8080`
* Verify the reverse proxy to the backend works via curl: `curl localhost:8080/api/v1/model-registry/` - this should return a valid response from the bff
* Run the frontend in dev mode using: `npm run start:dev` - run the bff directly using `make run PORT=4000 MOCK_K8S_CLIENT=true` in a separate terminal and verify the dashboard loads in a browser and the proxy functions by running: `curl localhost:9000/api/v1/model-registry/`.
* Run the frontend Makefile `make CONTAINER_TOOL=podman` (leave out the CONTAINER_TOOL param to use docker)

## Merge criteria:

### DIFF
--- a/clients/ui/docker-compose.yaml
+++ b/clients/ui/docker-compose.yaml
@@ -1,0 +1,23 @@
+services:
+  frontend:
+    build: ./frontend
+    container_name: model-registry-ui
+    ports:
+      - 8080:8080
+    environment:
+      API_URL: http://model-registry-bff:4001
+    networks:
+      - model_registry
+    depends_on:
+      - bff
+  bff:
+    build: ./bff
+    container_name: model-registry-bff
+    command:
+      - "--mock-k8s-client=true"
+    networks:
+      - model_registry
+
+networks:
+  model_registry:
+    name: model_registry
--- a/clients/ui/frontend/Dockerfile
+++ b/clients/ui/frontend/Dockerfile
@@ -1,0 +1,17 @@
+FROM node:18 AS build-stage
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+
+RUN npm cache clean --force
+RUN npm ci --omit=optional
+RUN npm run build
+
+FROM nginxinc/nginx-unprivileged
+
+ENV API_URL="http://localhost:4001"
+ENV NGINX_ENVSUBST_FILTER="API_URL"
+
+COPY --from=build-stage /usr/src/app/dist/ "/usr/share/nginx/html"
+COPY --from=build-stage /usr/src/app/nginx.conf "/etc/nginx/templates/default.conf.template"
--- a/clients/ui/frontend/config/webpack.dev.js
+++ b/clients/ui/frontend/config/webpack.dev.js
@@ -6,6 +6,9 @@ const common = require('./webpack.common.js');
 const { stylePaths } = require('./stylePaths');
 const HOST = process.env.HOST || 'localhost';
 const PORT = process.env.PORT || '9000';
+const PROXY_HOST = process.env.PROXY_HOST || 'localhost';
+const PROXY_PORT = process.env.PROXY_PORT || '4000';
+const PROXY_PROTOCOL = process.env.PROXY_PROTOCOL || 'http:';
 const relativeDir = path.resolve(__dirname, '..');
 
 module.exports = merge(common('development'), {
@@ -21,7 +24,16 @@ module.exports = merge(common('development'), {
     },
     client: {
       overlay: true
-    }
+    },
+    proxy: {
+      '/api': {
+        target: {
+          host: PROXY_HOST,
+          protocol: PROXY_PROTOCOL,
+          port: PROXY_PORT,
+        },
+      },
+    },
   },
   module: {
     rules: [
--- a/clients/ui/frontend/nginx.conf
+++ b/clients/ui/frontend/nginx.conf
@@ -1,0 +1,21 @@
+server {
+        listen       8080 default_server;
+        listen       [::]:8080 default_server;
+        server_name  _;
+        root         /usr/share/nginx/html;
+        gzip         on;
+        access_log   /dev/stdout main;
+
+        location / {
+            try_files $uri $uri/ /index.html;
+        }
+
+        location /api/ {
+            proxy_set_header Host $host;
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_pass ${API_URL};
+         }
+    }
+
--- a/clients/ui/frontend/.dockerignore
+++ b/clients/ui/frontend/.dockerignore
@@ -1,0 +1,2 @@
+dist/
+node_modules/
--- a/clients/ui/frontend/Makefile
+++ b/clients/ui/frontend/Makefile
@@ -1,0 +1,13 @@
+CONTAINER_TOOL ?= docker
+IMG ?= model-registry-ui:latest
+
+.PHONY: all
+all: docker-build
+
+.PHONY: help
+help: ## Display this help.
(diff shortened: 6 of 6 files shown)


Griffin-Sullivan (commented):
I must be doing something wrong. Everything in the testing works for me except when I run `docker compose up -d` then go to `localhost:8080` it's not working. I'm getting minified react errors so I can't see what the exact cause is. I'm going to keep digging.


Edit: Got it to work. I removed everything and ran `npm install` before docker compose again. Besides the two comments everything is good to me.

Edit 2: I've tested everything and it works for me. Note about `Run the frontend in dev mode using: npm run start:dev - verify this works with proxied backend URLs`. If you want to test yourself you need to run the backend separately in the `bff` dir with the command `make run PORT=4000 MOCK_K8S_CLIENT=true`.

clients/ui/frontend/config/webpack.dev.js
```diff
@@ -6,6 +6,9 @@ const common = require('./webpack.common.js');
 const { stylePaths } = require('./stylePaths');
 const HOST = process.env.HOST || 'localhost';
 const PORT = process.env.PORT || '9000';
+const PROXY_HOST = process.env.PROXY_HOST || 'localhost';
+const PROXY_PORT = process.env.PROXY_PORT || '4000';
+const PROXY_PROTOCOL = process.env.PROXY_PROTOCOL || 'http:';
```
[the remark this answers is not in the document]
alexcreasy: Good question, the main reason is this is a default value for local development - so we probably don't want the hassle of self signed certificates when developing against a local bff, this can be changed by setting the env var when developing remotely.

clients/ui/frontend/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:18 as build-stage
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
```
lucferbux, line 5: ok, a couple of annotations, this is just working cause you are copying your local `node_modules`, but that's a big nono, we should copy ONLY the files that are relevant to the project, we can either have `dockerignores` or something similar but this can cause some issues and won't work in a computer without a installation of the packages.

See the dockerignore we have here: https://github.com/opendatahub-io/odh-dashboard/blob/main/.dockerignore

And that we copy and do a ci before building: https://github.com/opendatahub-io/odh-dashboard/blob/main/Dockerfile#L15-L22

clients/ui/docker-compose.yaml
```diff
@@ -0,0 +1,21 @@
...
+  frontend:
+    build: ./frontend
+    container_name: model-registry-ui
+    ports:
+      - 8080:8080
+    environment:
+      API_URL: http://model-registry-bff:4001
+    networks:
+      - model_registry
+  bff:
```
lucferbux, line 11:
```
    depends_on:
      - bff
```

nit: it can be a good practice to start up the frontend after the backend to avoid reloads.

clients/ui/docker-compose.yaml
```diff
@@ -0,0 +1,21 @@
+services:
+  frontend:
+    build: ./frontend
```
lucferbux, line 3: This is only a suggestion/tip, I had a workaround for development (if we end up deciding some docker composes for dev, although we should work with clusters asap) that was the following:

```
  nginx:
    container_name: nginx_container
    restart: always
    image: nginx:XXXX
    volumes:
      - ./frontend/public:/usr/share/nginx/html
```
with this, there was a "hot reloading" method (which we added in makefile) so when executed it would build the frontend and since the volume was shared between the container and the machine you'll get the changes without restarting.

clients/ui/frontend/nginx.conf
```diff
@@ -0,0 +1,22 @@
...
+        server_name  _;
+        root         /usr/share/nginx/html;
+        gzip         on;
+        access_log   /dev/stdout main;
+
+        location / {
+            try_files $uri $uri/ /index.html;
+        }
+
+        location /p/ {
```
lucferbux, line 13: After giving it a thought, I think that this should be called `v1` and leave out that from the API, why?:

1. You can then switch services in case you have a v2 implementation and rereoute that to other differente service
2. If needed in isolation, you can create proxy pass rules in those services too or use an ingress with rules in deployment (so the notation v1, v2 shouldn't be in the implementation)
3. Leaves a clean path

lucferbux: That's just a suggestion btw 😅

lucferbux (changes requested):
couple of nits and suggestions

clients/ui/frontend/Dockerfile
alexcreasy: but this is only copied across during the build stage? Those artefacts are not copied across to the runtime container? Unless I'm misunderstanding something.

clients/ui/frontend/nginx.conf
alexcreasy: I'm not sure I understand the use case here? 

I think the versioning should be done directly on the API not on a reverse proxy that's only used in the frontend container. We have the ability to change the proxy address in the future since it's just an implementation detail - the backend API will become harder to change at some point down the line.

It's possible I've completely misunderstood what you're suggesting :)

clients/ui/docker-compose.yaml
alexcreasy: That makes sense, nothing is stopping someone from adding that line to a docker-compose, I guess we could have a couple of example files - a more "production" style example and a development one using the bind mount to hot reload the frontend?

lucferbux: nah don't worry, I just added as trivia, we can use the regular method which is more standard.

clients/ui/frontend/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN npm run build
+
+FROM nginxinc/nginx-unprivileged
+
+ENV API_URL="http://localhost:4001"
```
lucferbux, line 11: I would add this env param to the docker-compose and the future Deployment, to have ability to parametrize this in the future.

clients/ui/frontend/nginx.conf
lucferbux: wait, hold on, I don't understand this then, I'm sorry I was thinking about something else, but this is just a proxy rule, if we use for example `/api/v1` in our endpoint, why don't we add the rule to detect `/api/v1`, why do we need a `/p/` here?

lucferbux (changes requested):

clients/ui/frontend/Dockerfile
alexcreasy: This param is in the docker compose - it's being overridden to point to the bff via the hostname assigned by docker. This essentially just sets a default value.

alexcreasy: you're right, we could just match `/api/` and redirect that - that probably is cleaner. I used to prefer a visual clue that something's being proxied.

clients/ui/frontend/nginx.conf
alexcreasy: you're right, we could just match /api/ and redirect that - that probably is cleaner. I used to prefer a visual clue that something's being proxied.

alexcreasy:
I've addressed all the feedback given:
 - the proxy no longer uses the `/p/` prefix, so you can query using `/api/v1/model-registry/` directly. I've updated the instructions in the PR.
 - I've updated the build stage of the container build, it should no longe rely on you having the correct node_modules installed.

Griffin-Sullivan (approved):
Retested everything and it's working as expected

lucferbux (approved):
/lgtm

PR #256, opened 2024-08-12
fix: made install_protoc.sh compatible with wget2

Al-Pragliola:
fixes #254 

## Description

Switches paths from relative to absolute for wget and any other command that uses it, to make it work with wget and wget2.

## How Has This Been Tested?

I tested this on my local machine by running `make bin/protoc`, and also by building the docker image.

## Merge criteria:

- [ ] The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #265, opened 2024-08-12
Drop use of GOBIN

isinyaaa:
Resolves: https://issues.redhat.com/browse/RHOAIENG-8808

## Description
Enable multi-arch builds by dropping GOBIN.

## Merge criteria:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.21"
+      - name: install go deps
+        run: make go-deps
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,11 +30,11 @@ COPY patches/ patches/
 COPY templates/ templates/
 
 # Download tools
-RUN make deps
+RUN make deps go-deps
 
 # Build
 USER root
-RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean model-registry
+RUN PATH=$PATH:/opt/app-root/src/go/bin CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean model-registry
 
 # Use distroless as minimal base image to package the model-registry binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ internal/ml_metadata/proto/%.pb.go: api/grpc/ml_metadata/proto/%.proto
 gen/grpc: internal/ml_metadata/proto/metadata_store.pb.go internal/ml_metadata/proto/metadata_store_service.pb.go
 
 internal/converter/generated/converter.go: internal/converter/*.go
-	${GOVERTER} gen github.com/kubeflow/model-registry/internal/converter/
+	goverter gen github.com/kubeflow/model-registry/internal/converter/
 
 .PHONY: gen/converter
 gen/converter: gen/grpc internal/converter/generated/converter.go
@@ -104,29 +104,22 @@ vet:
 clean:
 	rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/converter/generated/*.go pkg/openapi
 
-.PHONY: clean/odh
-clean/odh:
+.PHONY: clean/docker
+clean/docker:
 	rm -Rf ./model-registry
 
 bin/protoc:
 	./scripts/install_protoc.sh
 
-bin/go-enum:
-	GOBIN=$(PROJECT_BIN) ${GO} install github.com/searKing/golang/tools/go-enum@v1.2.97
-
-bin/protoc-gen-go:
-	GOBIN=$(PROJECT_BIN) ${GO} install google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0
-
-bin/protoc-gen-go-grpc:
-	GOBIN=$(PROJECT_BIN) ${GO} install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
+go-deps:
+	${GO} install github.com/searKing/golang/tools/go-enum@v1.2.97
+	${GO} install google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0
+	${GO} install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
+	${GO} install github.com/jmattheis/goverter/cmd/goverter@v1.4.1
 
 GOLANGCI_LINT ?= ${PROJECT_BIN}/golangci-lint
 bin/golangci-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(PROJECT_BIN) v1.59.1
-
-GOVERTER ?= ${PROJECT_BIN}/goverter
-bin/goverter:
-	GOBIN=$(PROJECT_PATH)/bin ${GO} install github.com/jmattheis/goverter/cmd/goverter@v1.4.1
 
 OPENAPI_GENERATOR ?= ${PROJECT_BIN}/openapi-generator-cli
 NPM ?= "$(shell which npm)"
@@ -151,7 +144,7 @@ clean/deps:
 	rm -Rf bin/*
 
 .PHONY: deps
-deps: bin/protoc bin/go-enum bin/protoc-gen-go bin/protoc-gen-go-grpc bin/golangci-lint bin/goverter bin/openapi-generator-cli
+deps: bin/protoc bin/golangci-lint bin/openapi-generator-cli
 
 .PHONY: vendor
 vendor:
@@ -159,11 +152,11 @@ vendor:
 
 .PHONY: build
 build: gen vet lint
-	${GO} build -buildvcs=false
+	${GO} build
 
-.PHONY: build/odh
-build/odh: vet
-	${GO} build -buildvcs=false
+.PHONY: build/docker
+build/docker: vet
+	${GO} build
 
 .PHONY: gen
 gen: deps gen/grpc gen/openapi gen/openapi-server gen/converter
--- a/README.md
+++ b/README.md
@@ -10,10 +10,20 @@ Model registry provides a central repository for model developers to store and m
 
 ## Pre-requisites:
 - go >= 1.21
+- go-enum >= 1.2.97
+- protoc-gen-go >= 1.31.0
+- protoc-gen-go-grpc >= 1.3.0
+- goverter >= 1.4.1
 - protoc v24.3 - [Protocol Buffers v24.3 Release](https://github.com/protocolbuffers/protobuf/releases/tag/v24.3)
 - npm >= 10.2.0 - [Installing Node.js and npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 - Java >= 11.0
 - python 3.9
+
+To install the go dependencies you can use
+
+```sh
+make go-deps
+```
 
 ## OpenAPI Proxy Server
 
--- a/Dockerfile.odh
+++ b/Dockerfile.odh
@@ -21,7 +21,7 @@ COPY internal/ internal/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean/odh build/odh
+RUN PATH=$PATH:/opt/app-root/src/go/bin CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean/odh build/odh
 
 # Use distroless as minimal base image to package the model-registry binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
--- a/csi/Dockerfile
+++ b/csi/Dockerfile
@@ -22,7 +22,7 @@ COPY pkg/ pkg/
 
 # Build
 USER root
-RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make build
+RUN PATH=$PATH:/opt/app-root/src/go/bin CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make build
 
 # Use distroless as minimal base image to package the model-registry storage initializer binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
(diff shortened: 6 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -22,7 +22,14 @@ jobs:
...
         with:
-          go-version: '1.21'
+          go-version: "1.21"
+      - name: Install go dependencies
+        run: |
+          go install golang.org/dl/go1.21.9@latest
+          go install github.com/searKing/golang/tools/go-enum@v1.2.97
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
+          go install github.com/jmattheis/goverter/cmd/goverter@v1.4.1
```
tarilabs, line 32: I'm not convinced moving this from the makefile to the GHA is a good idea.

I like the avoidance of GOBIN, but having these "go install x" in the makefile avoided of the end-user manual action to ensure many prerequisites.
I can understand maybe a make target can no longer be identified easily, but in that case I'd move these go install x into a script file, which is invoked from make _and_ gha.

README.md
```diff
@@ -10,11 +10,21 @@ Model registry provides a central repository for model developers to store and m
...
 - npm >= 10.2.0 - [Installing Node.js and npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 - Java >= 11.0
 - python 3.9
 
+To install the go dependencies you can use
+
+```sh
+echo golang.org/dl/go1.21.9@latest github.com/searKing/golang/tools/go-enum@v1.2.97 google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0 google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0 github.com/jmattheis/goverter/cmd/goverter@v1.4.1 | xargs -n1 go install
+```
+
```
tarilabs, line 27: see above related comment; I'd rather have this taken care somehow by the makefile invocation.

I see some project further automate this for GoLang by having toolchains listed in go.mod too, but didn't dig deeper.

scripts/build_deploy.sh
```diff
@@ -1,13 +1,11 @@
...
+#!/usr/bin/env bash
 
 set -e
 
 # see Makefile for the IMG_ variables semantic
 IMG_REGISTRY=""
 IMG_ORG="${IMG_ORG:-kubeflow}"
 IMG_REPO="${IMG_REPO:-model-registry}"
-DOCKER_USER="${DOCKER_USER}"
-DOCKER_PWD="${DOCKER_PWD}"
```
tarilabs, line 10: isn't this conventional to make it more evident these are the env variables required when someone is invoking this script, for instance from make/gha ?

Dockerfile
```diff
@@ -9,32 +9,18 @@ COPY ["go.mod", "go.sum", "./"]
 RUN go mod download
 
 USER root
-# default NodeJS 14 is not enough for openapi-generator-cli, switch to Node JS currently supported
-RUN yum remove -y nodejs npm
-RUN yum module -y reset nodejs
-RUN yum module -y enable nodejs:18
-# install npm and java for openapi-generator-cli
-RUN yum install -y nodejs npm java-11
```
tarilabs, line 17: these in the past proved "painful" to be done in docker merely for producing the end image, and I'm so glad to see them removed _from here_ 🥹

if we wanted easier DX we can always have these in a [DevContainer](https://containers.dev/) instead in the future, wdyt ?

.github/workflows/build.yml
```diff
@@ -22,7 +22,14 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
+          go-version: "1.21"
+      - name: Install go dependencies
+        run: |
+          go install golang.org/dl/go1.21.9@latest
```
tarilabs, line 28:
```suggestion
```

I do not believe we still need this line, but kindly to confirm?

Makefile
```diff
@@ -165,8 +148,8 @@ vendor:
 build: gen vet lint
 	${GO} build -buildvcs=false
 
-.PHONY: build/odh
-build/odh: vet
+.PHONY: build/docker
+build/docker: vet
 	${GO} build -buildvcs=false
```
tarilabs, line 153: Can we do without `-buildvcs` now, or do we still need this option?

scripts/gen_openapi_server.sh
```diff
@@ -1,28 +1,28 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
 echo "Generating the OpenAPI server"
 
-ROOT_FOLDER="${ROOT_FOLDER:-..}"
+PROJECT_ROOT=$(dirname "$(realpath "$0")")/..
```
tarilabs, line 7: what are your thought about using this as an absolute path, to make it consistent with https://github.com/kubeflow/model-registry/pull/256/files ?

scripts/gen_type_asserts.sh
```diff
@@ -1,11 +1,11 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
-ROOT_FOLDER="${ROOT_FOLDER:-..}"
+PROJECT_ROOT=$(dirname "$(realpath "$0")")/..
```
tarilabs, line 5: ... same comment about absolute path here ...

scripts/install_protoc.sh
```diff
@@ -1,22 +1,22 @@
...
+
 ARCH="x86_64"
-if [[ "$(uname -m)" == "arm"* ]]; then
-  ARCH="aarch_64"
+if [[ $(uname -m) =~ arm ]]; then
+    ARCH="aarch_64"
 fi
 
-mkdir -p ${SCRIPT_PARENT_DIR}/bin
+PROJECT_ROOT=$(dirname "$(realpath "$0")")/..
```
tarilabs, line 15: ... the comment about absolute path is applying _mandatory_ here instead, for the reasons explained by @Al-Pragliola in https://github.com/kubeflow/model-registry/pull/256

scripts/build_deploy.sh
isinyaaa: I wasn't very sure tbh, we can keep them.

Makefile
isinyaaa: I managed to build it without this flag, let's see if that works on the CI as well.

scripts/gen_openapi_server.sh
isinyaaa: good catch!

isinyaaa:
> thank you so much @isinyaaa for this! 🚀
> 
> I left some comments, and in general I like how this simplifies the flow in general for the docker build too.
> 
> Can you kindly document as well, giving this a spin in producing an arm arch image, please?
> 
> To be specific, I'm _not_ asking to have the multi-arch done in this PR; I'm asking to give it a test so we're better confident with these modifications, we can go-compile for other archs _without_ the troubles we experienced in the past.

Simply changing the GOARCH variable works just fine :)

tarilabs:
thank you so much @isinyaaa this make it make leaner imho and I'm glad to hear didn't give any trouble trying to supply now a different target arch :)

I leave to you to consider one additional option:

```diff
diff --git a/Makefile b/Makefile
index a28e850..5659b62 100644
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ clean/deps:
        rm -Rf bin/*
 
 .PHONY: deps
-deps: bin/protoc bin/golangci-lint bin/openapi-generator-cli
+deps: go-deps bin/protoc bin/golangci-lint bin/openapi-generator-cli
 
 .PHONY: vendor
 vendor:
```

this would avoid having to launch the target manually as part of a build, or ensuring the versions we are intending for toolchains. wdyt? but again very minor.

I'm going to 
/lgtm
from my end 🚀 

I await others if they want try out this PR locally or if any remaining comments!

isinyaaa:
> thank you so much @isinyaaa this make it make leaner imho and I'm glad to hear didn't give any trouble trying to supply now a different target arch :)
> 
> I leave to you to consider one additional option:
> 
> ```diff
> diff --git a/Makefile b/Makefile
> index a28e850..5659b62 100644
> --- a/Makefile
> +++ b/Makefile
> @@ -144,7 +144,7 @@ clean/deps:
>         rm -Rf bin/*
>  
>  .PHONY: deps
> -deps: bin/protoc bin/golangci-lint bin/openapi-generator-cli
> +deps: go-deps bin/protoc bin/golangci-lint bin/openapi-generator-cli
>  
>  .PHONY: vendor
>  vendor:
> ```
> 
> this would avoid having to launch the target manually as part of a build, or ensuring the versions we are intending for toolchains. wdyt? but again very minor.

I'm not completely against it as it would be "easier" but keep in mind that this adds ~10s to the plain `make build`, which is the default cmd. This "negates" the supposed benefits of the go compiler having super-fast builds, so maybe we could create another rule closer to `build/docker`, or simply change its name, to facilitate iteration. WDYT?

tarilabs:
> I'm not completely against it as it would be "easier" but keep in mind that this adds ~10s to the plain `make build`, which is the default cmd. This "negates" the supposed benefits of the go compiler having super-fast builds, so maybe we could create another rule closer to `build/docker`, or simply change its name, to facilitate iteration. WDYT?

frankly 10s seems a bit exaggerated to me, I'm testing this one locally and if anything amounts to 2s, which in the scheme of the other generators time is negligible.

anyhow let's leave it off for now, since it's also documented prerequisite.
maybe time permitting for DX we could have some prechecks before issueing the `install`

isinyaaa:
> > I'm not completely against it as it would be "easier" but keep in mind that this adds ~10s to the plain `make build`, which is the default cmd. This "negates" the supposed benefits of the go compiler having super-fast builds, so maybe we could create another rule closer to `build/docker`, or simply change its name, to facilitate iteration. WDYT?
> 
> frankly 10s seems a bit exaggerated to me, I'm testing this one locally and if anything amounts to 2s, which in the scheme of the other generators time is negligible.
> 
> anyhow let's leave it off for now, since it's also documented prerequisite. maybe time permitting for DX we could have some prechecks before issueing the `install`

It looks like this varies depending on the system cache. Re-running after a while will take longer than "immediate" re-runs.
We can change the default cmd behaviour later after this PR, either way it could be helpful if we had such a precheck mechanism for (potentially) system deps.
Going without the suggestion as it might be unwanted to change user-system deps.

dhirajsb:
My biggest issue was the go binary in the bin directory. I thought that was an overkill. I was ok with having the other tool dependencies in `bin`. And, I've run into dependency compatibility issues with other projects, so having projects add their tool dependencies to my local go is a bit of an issue. 

Is the GOBIN really a hurdle for multi-platform builds? We can still download all the arch specific dependencies before building in the dockerfile for multiple platforms. 

I'm also concerned about the immutability of the Docker image build if it makes assumptions about what tools are available. That's the whole reason for a multi-stage build to avoid local env side affects. 

@lampajr wdyt?

dhirajsb:
About local go and project go versions, we can also look into using go [toolchain](https://go.dev/doc/toolchain)

Dockerfile
```diff
@@ -9,32 +9,18 @@ COPY ["go.mod", "go.sum", "./"]
...
 COPY pkg/ pkg/
-COPY patches/ patches/
-COPY templates/ templates/
-
-# Download tools
-RUN make deps
 
 # Build
-USER root
-RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean model-registry
```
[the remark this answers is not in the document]
isinyaaa: Using this option also makes multi-arch builds much more complex, as you have to specify all the toolchain components by hand. I also couldn't find a reason to keep it, do you have any objection?

isinyaaa:
> My biggest issue was the go binary in the bin directory. I thought that was an overkill. I was ok with having the other tool dependencies in `bin`. And, I've run into dependency compatibility issues with other projects, so having projects add their tool dependencies to my local go is a bit of an issue.
> 
> Is the GOBIN really a hurdle for multi-platform builds?

It is. While I'd recommend you try it yourself, feel free to take a look at this thread: https://github.com/golang/go/issues/57485.

> We can still download all the arch specific dependencies before building in the dockerfile for multiple platforms.

That would imply not using GOBIN inside the Dockerfile, actually. AFAIK you can still specify GOBIN when invoking commands locally, too.

> I'm also concerned about the immutability of the Docker image build if it makes assumptions about what tools are available. That's the whole reason for a multi-stage build to avoid local env side affects.

I'm not sure how to remediate that, unfortunately. I'll take a look at your suggestion for using the `toolchain` cmd.

Dockerfile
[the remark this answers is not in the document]
isinyaaa: We can discuss this separately, I'll push the change to a separate PR.

Dockerfile
```diff
@@ -9,32 +9,18 @@ COPY ["go.mod", "go.sum", "./"]
...
 USER root
-# default NodeJS 14 is not enough for openapi-generator-cli, switch to Node JS currently supported
-RUN yum remove -y nodejs npm
-RUN yum module -y reset nodejs
-RUN yum module -y enable nodejs:18
-# install npm and java for openapi-generator-cli
-RUN yum install -y nodejs npm java-11
 
 # Copy the go source
-COPY ["Makefile", "main.go", ".openapi-generator-ignore", "openapitools.json", "./"]
```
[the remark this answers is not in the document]
isinyaaa: Agreed, let's keep this. Reverted the change partially, wdyt?

tarilabs (commented):
We are checking-in generated source files inside the project.
I believe this is a good practice.

I believe the Dockerfile as it stands now is re-generating source files inside the docker build, with the end results that we're shipping binaries which potentially we don't have the source code anymore (as we don't have track of the sources generated inside the docker build itself).

I preferred the original modification, per https://github.com/kubeflow/model-registry/pull/265#discussion_r1714689546

tarilabs:
> My biggest issue was the go binary in the bin directory. I thought that was an overkill

 👉  https://github.com/kubeflow/model-registry/pull/288

isinyaaa:
I can rebase whenever we agree on which path to follow, the current failure stems from the current go bin path missing on the dockerfiles.

tarilabs:
> what is the resolution you guys found on this, as I see this is closed.

with more recent changes (ie as it stand currently on main default branch), we just discovered gobin looks currently not an impediment to crosscompilation, so we're now looking to keep it (hence not drop its current usage). More updates to follow

Issue #255, opened 2024-08-11
Context name containing characters requiring url quoting/escaping


tarilabs:
**Describe the bug**
Unable to index ModelVersion containing `:`.

**To Reproduce**
1. attempt to index a ModelVersion for "some:version"

following the [kubeflow tutorial](https://www.kubeflow.org/docs/components/model-registry/getting-started/#register-metadata), example: 
,
```py
rm = registry.register_model(
    "mnist",
    "https://github.com/tarilabs/demo20231212/raw/main/v1.nb20231206162408/mnist.onnx",
    model_format_name="onnx",
    model_format_version="1",
    version="some:version",    # <-- here
    description="lorem ipsum mnist",
    metadata={
        "accuracy": 3.14,
        "license": "apache-2.0",
    }
)
```

**Expected behavior**
Can use special characters without having to `urllib.parse.quote()` manually the version name.  

i.e. without requiring to:
```py
from urllib.parse import quote
# ...
    version=quote("some:version"),   
```

**Additional context**
This is likely deriving from the fact that in the Model Registry Python client, the logical model entity attributes are passed to the rest layer without quote where sometimes might be necessary; for example, when the model version name is used as a query parameter in a rest endpoint call.

It could be a helpful exercise to check if any limitations on MLMD side for Context names themselves (we are aware MLMD Context names are not updatable, but that is an unrelated limitation).

tarilabs:
thank you @isinyaaa

PR #295, opened 2024-08-22
restrict captures to prefix when mapping *FromOwned

isinyaaa:
Fixes: #255

## Description
This makes all subsequent `:` (colon) separators (i.e. those provided by the user) part of the ctx name.

## Merge criteria:

### DIFF
--- a/clients/python/Makefile
+++ b/clients/python/Makefile
@@ -13,7 +13,7 @@ clean:
 
 .PHONY: build-mr
 build-mr:
-	cd ../../ && make image/build
+	cd ../../ && IMG_REGISTRY=docker.io IMG_VERSION=latest make image/build
 
 .PHONY: test-e2e
 test-e2e: build-mr
--- a/clients/python/Makefile
+++ b/clients/python/Makefile
@@ -1,5 +1,8 @@
 all: install tidy
 
+IMG_REGISTRY ?= docker.io
+IMG_VERSION ?= latest 
+
 .PHONY: install
 install:
 	../../bin/openapi-generator-cli generate -i ../../api/openapi/model-registry.yaml -g python -o src/ --package-name mr_openapi --additional-properties=library=asyncio,generateSourceCodeOnly=true,useOneOfDiscriminatorLookup=true
@@ -13,7 +16,7 @@ clean:
 
 .PHONY: build-mr
 build-mr:
-	cd ../../ && IMG_REGISTRY=docker.io IMG_VERSION=latest make image/build
+	cd ../../ && IMG_REGISTRY=${IMG_REGISTRY} IMG_VERSION=${IMG_VERSION} make image/build
 
 .PHONY: test-e2e
 test-e2e: build-mr
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -10,6 +10,8 @@ from time import sleep
 import pytest
 import requests
 
+from model_registry import ModelRegistry
+
 
 def pytest_addoption(parser):
     parser.addoption("--e2e", action="store_true", help="run end-to-end tests")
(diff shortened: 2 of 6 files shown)


isinyaaa:
cc: @tarilabs

tarilabs:
thanks @isinyaaa , can you kindly check as well on the MR side, by adding a test (maybe a Robot framework test?) which creates a ModelVersion for something like `"my:version"` and analogously for ModelArtifact, please?

The reason I'm asking, I had the feeling the use of `:` character was not quoted when used as query parameter when making the REST request by the MR python client, in the check if a ModelVersion is existing.
Maybe I'm wrong here, but I would prefer for this to be double-checked at once since the topic is open.

tarilabs:
thanks Isabella, a0298d4cf8e145fba77e9cbed4b81bd9d20f2b0b is helpful, but I would still prefer for a test (even a robotframework test if easier) which leverages the Model Registry python client:

> The reason I'm asking, I had the feeling the use of : character was not quoted when used as query parameter when making the REST request by the MR python client, in the check if a ModelVersion is existing.
Maybe I'm wrong here, but I would prefer for this to be double-checked at once since the topic is open.

as far as I understand, the added robot test does not leverage the python client.

tarilabs (commented):
let's remove test/robot/MRandLogicalModel.robot
this was needed when the business logic of converting MR logical model entities to/from MLMD was done both in golang and also on python side
since the MR python client is now going via REST, now the business logic is only in one single place (better) and so this specific test/robot/MRandLogicalModel.robot to me is no longer needed

tarilabs:
also, we no longer really need "TEST_MODE" specifically... that was a decision when we had to do "the same" via both rest and grpc to ensure consistency, manually.
since the MR python client is now also rebased to go via REST, we can simply have separate .robot for separate concerns if that simplifies, ie one to check user stories, one to check for things specifically via rest (eg. non-regressions for dashboard ui) etc.
but this is also something could be dealt with, in separate PR.

lampajr:
> also, we no longer really need "TEST_MODE" specifically... that was a decision when we had to do "the same" via both rest and grpc to ensure consistency, manually. since the MR python client is now also rebased to go via REST, we can simply have separate .robot for separate concerns if that simplifies, ie one to check user stories, one to check for things specifically via rest (eg. non-regressions for dashboard ui) etc. but this is also something could be dealt with, in separate PR.

+1 and once we remove that we can simplify the CI as well by removing https://github.com/kubeflow/model-registry/blob/1dd9af5bf535cb5bdea8600427c5459f9a8e21d1/.github/workflows/run-robot-tests.yaml#L59-L61

> let's remove test/robot/MRandLogicalModel.robot

The PR checks are actually failing due to this robot test, so as @tarilabs was suggesting I think we can safely remove that test completely as there is no more reason to keep it and check the internal MLMD model as we don't have two core logics trying to do the same anymore

tarilabs (commented):
/lgtm

Per discussion https://github.com/kubeflow/model-registry/pull/295#issuecomment-2318155283 but will wait if it's possible to receive for a small "smoke test" locally using docker-compose-local up or analogous, and a small MR Python client tutorial feedback, because it would show confirmation against the original bug report, so to be sure it address indeed the same problem ..and not possibly another one which was also good to be found! :)

clients/python/Makefile
```diff
@@ -13,7 +13,7 @@ clean:
 
 .PHONY: build-mr
 build-mr:
-	cd ../../ && make image/build
+	cd ../../ && IMG_REGISTRY=docker.io IMG_VERSION=latest make image/build
```
tarilabs, line 16: can you kindly make this as a default?

ie:
```makefile
IMG_REGISTRY ?= docker.io
IMG_VERSION ?= latest
```

at the top? wdyt?

tarilabs (commented):
/lgtm

but one minor details as I believe would make it more explicit

clients/python/Makefile
isinyaaa: good one, will do.

tarilabs (approved):
thanks a lot @isinyaaa , with the other PRs, this makes writing new e2e tests way easier imho

/lgtm
/approve

PR #313, opened 2024-08-26
GHA: trigger CSI tests on go.mod update

isinyaaa:
This prevents CSI dependencies from being out-of-date.

## Merge criteria:

### DIFF
--- a/.github/workflows/csi-test.yml
+++ b/.github/workflows/csi-test.yml
@@ -4,18 +4,20 @@ on:
     branches:
       - "main"
     paths-ignore:
-      - 'LICENSE*'
-      - '**.gitignore'
-      - '**.md'
-      - '**.txt'
-      - '.github/ISSUE_TEMPLATE/**'
-      - '.github/dependabot.yml'
-      - 'docs/**'
+      - "LICENSE*"
+      - "**.gitignore"
+      - "**.md"
+      - "**.txt"
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/dependabot.yml"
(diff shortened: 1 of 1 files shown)


.github/workflows/csi-test.yml
```diff
@@ -4,18 +4,20 @@ on:
...
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/dependabot.yml"
+      - "docs/**"
   pull_request:
     paths:
       - "csi/**"
       - "internal/server/openapi/api_model_registry_service*"
       - "pkg/openapi/**"
+      # csi build depends on base go.mod https://github.com/kubeflow/model-registry/issues/311
+      - "go.mod"
```
tarilabs, line 20: @lampajr please be informed, as apparently there is some "further dependency ties" which we discovered (more in the linked issue)

tarilabs (commented):
/lgtm

tarilabs:
many thanks Isabella, will wait for another to go in then proceeding to merge this one

.github/workflows/csi-test.yml
lampajr: Good catch!!!

I think that makes totally sense, as the CSI test actually makes use of the local (thus relative) model registry dependency.. therefore the root go.mod is used (see https://github.com/kubeflow/model-registry/blob/d2312907025adbe83d3faafbecf1474824d055ed/csi/Dockerfile.dev#L23).

lampajr (commented):
/lgtm

tarilabs (approved):

PR #339, opened 2024-09-03
Adds an endpoint to get a registered model by id

alexcreasy:
## Description

Adds an endpoint to fetch registered models by id.

## How Has This Been Tested?

Test rig setup:
1. Deploy MR backend (you can follow the guide using kind in the docs)
2. Deploy BFF with e.g. `make run PORT=4000 MOCK_K8S_CLIENT=true`
3. Add a model to one of the registries: 
```
curl -i -X POST "http://localhost:4000/api/v1/model-registry/model-registry/registered_models" \
     -H "Content-Type: application/json" \
     -d '{
  "customProperties": {
    "my-label9": {
      "metadataType": "MetadataStringValue",
      "string_value": "val"
    }
  },
  "description": "bella description",
  "externalId": "9927",
  "name": "bella",
  "owner": "eder",
  "state": "LIVE"
}'
```

Test routine:
1. Verify the endpoint returns a 200 and the model from the endpoint: `curl -i localhost:4000/api/v1/model-registry/model-registry/registered_models/1`
2. Try with a different non existent id and verify that a 404 is returned: `curl -i localhost:4000/api/v1/model-registry/model-registry/registered_models/fluglebinder`

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/README.md
+++ b/clients/ui/bff/README.md
@@ -54,21 +54,22 @@ make docker-build
 
 ### Endpoints
 
-| URL Pattern                                                   | Handler                 | Action                                       |
-|---------------------------------------------------------------|-------------------------|----------------------------------------------|
-| GET /v1/healthcheck                                           | HealthcheckHandler      | Show application information.                |
-| GET /v1/model-registry/                                       | ModelRegistryHandler    | Get all model registries,                    |
-| GET /v1/model-registry/{model_registry_id}/registered_models  | RegisteredModelsHandler | Gets a list of all RegisteredModel entities. |
-| POST /v1/model-registry/{model_registry_id}/registered_models | RegisteredModelsHandler | Create a RegisteredModel entity.             |
+| URL Pattern                                                                        | Handler                  | Action                                       |
+|------------------------------------------------------------------------------------|--------------------------|----------------------------------------------|
+| GET /v1/healthcheck                                                                | HealthcheckHandler       | Show application information.                |
+| GET /v1/model-registry/                                                            | ModelRegistryHandler     | Get all model registries,                    |
+| GET /v1/model-registry/{model_registry_id}/registered_models                       | RegisteredModelsHandler  | Gets a list of all RegisteredModel entities. |
+| POST /v1/model-registry/{model_registry_id}/registered_models                      | RegisteredModelsHandler  | Create a RegisteredModel entity.             |
+| GET /v1/model-registry/{model_registry_id}/registered_models/{registered_model_id} | RegisteredModelHandler   | Get a RegisteredModel entity by ID           |
 
 ### Sample local calls
 ```
 # GET /v1/healthcheck
-curl -i localhost:4000/api/v1/healthcheck/
+curl -i localhost:4000/api/v1/healthcheck
 ```
 ```
-# GET /v1/model-registry/ 
-curl -i localhost:4000/api/v1/model-registry/
+# GET /v1/model-registry 
+curl -i localhost:4000/api/v1/model-registry
 ```
 ```
 # GET /v1/model-registry/{model_registry_id}/registered_models
@@ -91,4 +92,8 @@ curl -i -X POST "http://localhost:4000/api/v1/model-registry/model-registry/regi
   "owner": "eder",
   "state": "LIVE"
 }'
+```
+```
+# GET /v1/model-registry/{model_registry_id}/registered_models/{registered_model_id}
+curl -i localhost:4000/api/v1/model-registry/model-registry/registered_models/1
 ```
--- a/clients/ui/bff/api/app.go
+++ b/clients/ui/bff/api/app.go
@@ -2,23 +2,24 @@ package api
 
 import (
 	"fmt"
+	"github.com/julienschmidt/httprouter"
 	"github.com/kubeflow/model-registry/ui/bff/config"
 	"github.com/kubeflow/model-registry/ui/bff/data"
 	"github.com/kubeflow/model-registry/ui/bff/integrations"
 	"github.com/kubeflow/model-registry/ui/bff/internals/mocks"
 	"log/slog"
 	"net/http"
-
-	"github.com/julienschmidt/httprouter"
 )
 
 const (
 	Version              = "1.0.0"
 	PathPrefix           = "/api/v1"
 	ModelRegistryId      = "model_registry_id"
-	HealthCheckPath      = PathPrefix + "/healthcheck/"
-	ModelRegistry        = PathPrefix + "/model-registry/"
-	RegisteredModelsPath = ModelRegistry + ":" + ModelRegistryId + "/registered_models"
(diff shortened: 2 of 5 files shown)


Griffin-Sullivan:
FYI there's a type in the commit message. `Adds and endpoint to get individual registered models`

clients/ui/bff/data/registered_model.go
```diff
@@ -39,3 +39,18 @@ func CreateRegisteredModel(client integrations.HTTPClientInterface, jsonData []b
 
 	return &model, nil
 }
+
+func GetRegisteredModel(client integrations.HTTPClientInterface, id string) (*openapi.RegisteredModel, error) {
+	responseData, err := client.GET(registerModelPath + "/" + id)
```
Griffin-Sullivan, line 44: Maybe it would be nice to edit our `GET()` so that you can just pass a list for the path and not have to worry about adding `/` in between everything. The `GET()` could piece together our URL for us.

clients/ui/bff/api/registered_models_handler.go
```diff
@@ -83,3 +83,32 @@ func (app *App) CreateRegisteredModelHandler(w http.ResponseWriter, r *http.Requ
...
+		app.serverErrorResponse(w, r, errors.New("REST client not found"))
+		return
+	}
+
+	model, err := data.GetRegisteredModel(client, ps.ByName(RegisteredModelId))
+	if err != nil {
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+	fmt.Printf("%#v\n", model)
```
Griffin-Sullivan, line 99: Do we need this Printf?

Griffin-Sullivan (approved):
Tested the changes. Everything is working for me.

lucferbux:
Changes seem to work fine, great work here, I subscribe to @Griffin-Sullivan questions, once that's sorted out we can lgtm it!

clients/ui/bff/data/registered_model.go
alexcreasy: I'm not against this idea, but that's a reasonably sizeable refactor (in terms of code changes as opposed to effort) - you have to modify every call site to the GET func and IMO this kind of change belongs as a separate standalone PR.

alexcreasy: Now addressing your original issue, I'm not a massive fan of switching to taking a slice as it complicates the calling syntax a bit....

e.g. `responseData, err := client.GET([]string{registerModelPath, id})`

the slice literal isn't immediately intuitive to me, also it's rather annoying when you only have a single string.

Now a better solution might be to use a variadic `path ...string` param - this actually works really nicely - there's one gotcha here that varargs have to be the last param - so for the POST and other methods, we need to slip them to take other params (potentially optional) params first which isn't as nice.

WDYT?

Note: as an extra aside, if you're a big fan of syntactic sugar, go is 100% not your language, it's pretty much designed on the principal that syntactic sugar is bad :P

Griffin-Sullivan:
Changes are good to me

alexcreasy:
I mean, that's fair that I can't LGTM it myself technically, but Griffen just did :P

clients/ui/bff/api/app.go
```diff
@@ -16,9 +16,11 @@ const (
 	Version              = "1.0.0"
 	PathPrefix           = "/api/v1"
 	ModelRegistryId      = "model_registry_id"
+	RegisteredModelId    = "registered_model_id"
 	HealthCheckPath      = PathPrefix + "/healthcheck/"
 	ModelRegistry        = PathPrefix + "/model-registry/"
-	RegisteredModelsPath = ModelRegistry + ":" + ModelRegistryId + "/registered_models"
+	RegisteredModelsPath = ModelRegistry + ":" + ModelRegistryId + "/registered_models/"
```
ederign, line 22:
```suggestion
	RegisteredModelsPath = ModelRegistry + ":" + ModelRegistryId + "/registered_models"
```

clients/ui/bff/api/app.go
```diff
@@ -16,9 +16,11 @@ const (
...
+	RegisteredModelPath  = RegisteredModelsPath + ":" + RegisteredModelId
```
ederign, line 23:
```suggestion
	RegisteredModelPath  = RegisteredModelsPath + "/:" + RegisteredModelId
```
Without this change, we slightly change our API, requiring a trailing slash:

Screenshot 2024-09-04 at 4 29 53 PM

@alexcreasy, what do you think? Should we keep the current behavior by default?

clients/ui/bff/data/registered_model.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/kubeflow/model-registry/pkg/openapi"
 	"github.com/kubeflow/model-registry/ui/bff/integrations"
+	"net/url"
```
ederign, line 9: Not about this line, but L13:

As you are going to touch the code with the review, I wonder if we should change this for GetAllRegisteredModels

ederign (changes requested):
@alexcreasy just some small changes related to our API design. Another option is to use https://pkg.go.dev/github.com/julienschmidt/httprouter#Router.RedirectTrailingSlash , but we would need to investigate better if there is any side effect of using it.

Also, could you please add the example from the PR in our README?

clients/ui/bff/api/app.go
alexcreasy: Personally, I think it makes for an easier to use API if it functions with and without a trailing slash, rather than returning a 301. Some HTTP client libs I've used in the past, don't automatically follow the redirect - no idea if this is an issue we'd actually run into though :) 

This behaviour is perfectly valid within the HTTP spec, so I don't think this is a major issue?

alexcreasy:
@ederign think we should be gtg now

clients/ui/bff/README.md
```diff
@@ -54,12 +54,13 @@ make docker-build
...
-| GET /v1/model-registry/                                       | ModelRegistryHandler    | Get all model registries,                    |
-| GET /v1/model-registry/{model_registry_id}/registered_models  | RegisteredModelsHandler | Gets a list of all RegisteredModel entities. |
-| POST /v1/model-registry/{model_registry_id}/registered_models | RegisteredModelsHandler | Create a RegisteredModel entity.             |
+| URL Pattern                                                                        | Handler                  | Action                                       |
+|------------------------------------------------------------------------------------|--------------------------|----------------------------------------------|
+| GET /v1/healthcheck                                                                | HealthcheckHandler       | Show application information.                |
+| GET /v1/model-registry/                                                            | ModelRegistryHandler     | Get all model registries,                    |
```
ederign, line 60:
```suggestion
| GET /v1/model-registry                                                            | ModelRegistryHandler     | Get all model registries,                    |
```

ederign (changes requested):

ederign (approved):

PR #341, opened 2024-09-03
py: fix access token parsing

### LINKED ISSUE #340: TypeError in get_registered_model function when using model registry python client v0.2.5a1
**Describe the bug**

I have tried using the python client on a Jupyter notebook hosted inside the cluster, the function get_registered_model fails with the following error "TypeError: can only concatenate str (not "bytes") to str". I have checked the pod connectivity using the requests library and the model registry responds correctly.

**To Reproduce**

1. !pip install --pre model-registry=="0.2.5a1"
2. from model_registry import ModelRegistry
3. registry = ModelRegistry(
        server_address="http://test.namespace.svc.cluster.local",
        port=8080,
        author="John Doe",
        is_secure=False (True or False makes no difference)
    )
4. model = registry.get_registered_model("X")

**Expected behavior**

{'createTimeSinceEpoch': '1725286288951', 'customProperties': {}, 'id': '3', 'lastUpdateTimeSinceEpoch': '1725286288951', 'name': 'X', 'owner': 'apraglio', 'state': 'LIVE'}

**Additional context**

Error trace:

```
File /opt/app-root/lib64/python3.9/site-packages/mr_openapi/configuration.py:357, in Configuration.auth_settings(self)

350 auth = {}
351 if self.access_token is not None:
352 auth["Bearer"] = {
353 "type": "bearer",
354 "in": "header",
355 "format": "JWT",
356 "key": "Authorization", -->
357 "value": "Bearer " + self.access_token, <- affected line
358 }
359 return auth

TypeError: can only concatenate str (not "bytes") to str
```

Al-Pragliola:
## Description

fixes #340

After running `registry.get_registered_model("X")` without providing an user_token in ModelRegistry initialization, I got the following error:

```
File /opt/app-root/lib64/python3.9/site-packages/mr_openapi/configuration.py:357, in Configuration.auth_settings(self)

350 auth = {}
351 if self.access_token is not None:
352 auth["Bearer"] = {
353 "type": "bearer",
354 "in": "header",
355 "format": "JWT",
356 "key": "Authorization", -->
357 "value": "Bearer " + self.access_token, <- affected line
358 }
359 return auth

TypeError: can only concatenate str (not "bytes") to str
```

thanks to @isinyaaa , we figured out that there was a bug in the token parsing function in the client code and switched from reading it from "bytes mode" to "text mode"

## How Has This Been Tested?

Uploaded the wheel on the notebook and rerun.

## Merge criteria:

### DIFF
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -137,9 +137,8 @@ def client() -> ModelRegistry:
 
 @pytest.fixture(scope="module")
 def setup_env_user_token():
-    token_file = tempfile.NamedTemporaryFile(delete=False)
-    token_file.write(b"Token")
-    token_file.close()
+    with tempfile.NamedTemporaryFile(delete=False) as token_file:
+        token_file.write(b"Token")
     old_token_path = os.getenv("KF_PIPELINES_SA_TOKEN_PATH")
     os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = token_file.name
 
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -2,6 +2,7 @@ import asyncio
 import inspect
 import os
 import subprocess
+import tempfile
 import time
 from contextlib import asynccontextmanager
 from pathlib import Path
@@ -133,3 +134,19 @@ def event_loop():
 @cleanup
 def client() -> ModelRegistry:
     return ModelRegistry(REGISTRY_HOST, REGISTRY_PORT, author="author", is_secure=False)
+
+@pytest.fixture(scope="module")
+def setup_env_user_token():
+    token_file = tempfile.NamedTemporaryFile(delete=False)
+    token_file.write(b"Token")
+    token_file.close()
+    old_token_path = os.getenv("KF_PIPELINES_SA_TOKEN_PATH")
+    os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = token_file.name
+
+    yield token_file.name
+
+    if old_token_path is None:
+        del os.environ["KF_PIPELINES_SA_TOKEN_PATH"]
+    else:
+        os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = old_token_path
+    os.remove(token_file.name)
--- a/clients/python/tests/regression_test.py
+++ b/clients/python/tests/regression_test.py
@@ -22,3 +22,25 @@ def test_create_tagged_version(client: ModelRegistry):
     mv = client.get_model_version(name, version)
     assert mv
     assert mv.id
+
+@pytest.mark.e2e
+def test_get_model_without_user_token(setup_env_user_token, client):
+    """Test regression for using client methods without an user_token in the init arguments.
+
+    Reported on: https://github.com/kubeflow/model-registry/issues/340
+    """
+    assert setup_env_user_token != ""
+    name = "test_model"
+    version = "1.0.0"
+    metadata = {"a": 1, "b": "2"}
+    rm = client.register_model(
+        name,
+        "s3",
+        model_format_name="test_format",
+        model_format_version="test_version",
+        version=version,
+        metadata=metadata,
+    )
+    assert rm.id
+    assert (_rm := client.get_registered_model(name))
+    assert rm.id == _rm.id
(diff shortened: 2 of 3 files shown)


tarilabs:
nice catch, thanks @Al-Pragliola @isinyaaa ;

/lgtm

just before merging, can you comment on potential idea to create a non-regression testing for this?

It doesn't "have to be" in this PR, but I want to track it before we merge, thanks!

Al-Pragliola:
What I would I do is a test where there's no user_token defined when initializing the client (ModelRegistry), `KF_PIPELINES_SA_TOKEN_PATH` env var set to a tmp file path, where the content is a mocked token, register a new model and then trying `get_registered_model` to get it back

clients/python/tests/conftest.py
```diff
@@ -133,3 +134,19 @@ def event_loop():
 @cleanup
 def client() -> ModelRegistry:
     return ModelRegistry(REGISTRY_HOST, REGISTRY_PORT, author="author", is_secure=False)
+
+@pytest.fixture(scope="module")
+def setup_env_user_token():
+    token_file = tempfile.NamedTemporaryFile(delete=False)
+    token_file.write(b"Token")
+    token_file.close()
```
tarilabs, line 142:
```suggestion
    with tempfile.NamedTemporaryFile(delete=False) as token_file:
        token_file.write(b"Token")
```

possibly more pythonic

tarilabs (commented):
Awesome to have a non-regression along with the fix, thank you so much @Al-Pragliola ! 🚀 

One small suggestion, below.

The fixture looks to me following same pattern/style than other fixtures, and looks good to me.

Any further comment @isinyaaa ?
/lgtm

clients/python/tests/conftest.py
```diff
@@ -133,3 +134,18 @@ def event_loop():
...
+        token_file.write(b"Token")
+    old_token_path = os.getenv("KF_PIPELINES_SA_TOKEN_PATH")
+    os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = token_file.name
+
+    yield token_file.name
+
+    if old_token_path is None:
+        del os.environ["KF_PIPELINES_SA_TOKEN_PATH"]
+    else:
+        os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = old_token_path
```
isinyaaa, line 150: we don't actually need this, the envvar will be preserved outside the tests.

isinyaaa (approved):
lgtm, just a nit inline :)

clients/python/tests/regression_test.py
```diff
@@ -22,3 +22,25 @@ def test_create_tagged_version(client: ModelRegistry):
     mv = client.get_model_version(name, version)
     assert mv
     assert mv.id
+
+@pytest.mark.e2e
+def test_get_model_without_user_token(setup_env_user_token, client):
```
isinyaaa, line 27: I see here you're already using it. So the scope reduction is a better practice.

isinyaaa (changes requested):
I had misunderstood your intent before, sorry for the noise.

isinyaaa (approved):
On a third look this seems fine :) -- thanks @tarilabs for the reference:
https://stackoverflow.com/questions/25660064/in-which-order-are-pytest-fixtures-executed/64471427#64471427:~:text=while%20discouraging%20from%20relying%20on%20another%20factors%20(like%20order%20in%20test%20function%20arguments)%3A

clients/python/tests/conftest.py
```diff
@@ -133,3 +134,18 @@ def event_loop():
 @cleanup
 def client() -> ModelRegistry:
     return ModelRegistry(REGISTRY_HOST, REGISTRY_PORT, author="author", is_secure=False)
+
+@pytest.fixture(scope="module")
+def setup_env_user_token():
```
tarilabs, line 139: i.e. we rely on scope for fixture ordering, and not fn arg ordering in the actual test fn.

this fixture is to be consumed by the pytest fn as an argument, when the test wants to opt-in to the (mocked) token.
when that happens, this fixture here needs to be balled before the fixture producing the MR python client.
this is achieved by this fixture having wider scope

tarilabs (approved):
/lgtm
/approve

Issue #231, opened 2024-07-30
POST on /model_artifacts won't work


isinyaaa:
**Describe the bug**
Any POST call fails with

```
{
  "code": "",
  "message": "missing model version id, cannot create artifact without model version: bad request"
}
```
on the other hand, it's possible to create new model artifacts using the generic `/model_versions/{id}/artifacts` route.

**To Reproduce**
1. POST

  ```json
  {
    "externalId": "test"
  }
  ```
  to `/api/model_registry/v1alpha3/model_artifacts`
2. See error

**Expected behavior**
There should be either a way to create model artifacts unassociated from model versions, or to provide the required id in the request body (which is currently missing https://petstore.swagger.io/?url=https://raw.githubusercontent.com/kubeflow/model-registry/main/api/openapi/model-registry.yaml#/ModelRegistryService/createModelArtifact)

PR #376, opened 2024-09-10
enable standalone artifacts

isinyaaa:
Fixes: #231

## Description
To undo some confusion around https://github.com/opendatahub-io/model-registry-bf4-kf/pull/289 and re-enable standalone artifact support on both the `/model_artifacts` endpoint and the go core API.

TODO:
- [x] Test standalone artifacts through REST API (via Python)
- [x] Test late binding of standalone artifacts to model versions (in Go and via REST)
- [x] Update REST response status code

## Merge criteria:

If you have UI changes

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -391,8 +391,8 @@ paths:
           $ref: "#/components/responses/NotFound"
         "500":
           $ref: "#/components/responses/InternalServerError"
-      operationId: createModelVersionArtifact
-      summary: Create an Artifact in a ModelVersion
+      operationId: upsertModelVersionArtifact
+      summary: Upsert an Artifact in a ModelVersion
       description: Creates a new instance of an Artifact if needed and associates it with `ModelVersion`.
     parameters:
       - name: modelversionId
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -1224,9 +1224,8 @@ components:
           properties:
             id:
               format: int64
-              description: Output only. The unique server generated id of the resource.
+              description: The unique server generated id of the resource.
               type: string
-              readOnly: true
             createTimeSinceEpoch:
               format: int64
               description: Output only. Create time of the resource in millisecond since epoch.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -52,7 +52,9 @@ type ModelRegistryApi interface {
 
 	// ARTIFACT
 
-	UpsertArtifact(artifact *openapi.Artifact, modelVersionId *string) (*openapi.Artifact, error)
+	UpsertModelVersionArtifact(artifact *openapi.Artifact, modelVersionId string) (*openapi.Artifact, error)
+
+	UpsertArtifact(artifact *openapi.Artifact) (*openapi.Artifact, error)
 
 	GetArtifactById(id string) (*openapi.Artifact, error)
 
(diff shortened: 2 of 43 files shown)


api/openapi/model-registry.yaml
```diff
@@ -1224,9 +1224,8 @@ components:
           properties:
             id:
               format: int64
-              description: Output only. The unique server generated id of the resource.
+              description: The unique server generated id of the resource.
               type: string
-              readOnly: true
```
tarilabs, line 1229: I'm not sure I understood the reason for this change?
iirc we rely on unique id generated on MLMD side.

pkg/api/api.go
```diff
@@ -52,17 +52,18 @@ type ModelRegistryApi interface {
...
 
 	GetArtifacts(listOptions ListOptions, modelVersionId *string) (*openapi.ArtifactList, error)
 
 	// MODEL ARTIFACT
 
-	// UpsertModelArtifact create a new Artifact or update an Artifact associated to a specific
-	// ModelVersion identified by modelVersionId parameter
-	UpsertModelArtifact(modelArtifact *openapi.ModelArtifact, modelVersionId *string) (*openapi.ModelArtifact, error)
+	// UpsertModelArtifact creates or inserts an Artifact
+	UpsertModelArtifact(modelArtifact *openapi.ModelArtifact) (*openapi.ModelArtifact, error)
```
tarilabs, line 66: as this PR touches this "Core API" I would seize the chance to: 

1. revise if any changes required in doc: https://github.com/kubeflow/model-registry/blob/main/docs/mr_go_library.md

2. add a disclaimer warning in that doc that it currently expose direct connection to MLMD, and this is potentially subject of change in the future, hence the "mr go library" doc is intended for Developer purposes, while external clients are encouraged to use the REST API

tarilabs (commented):
the PR description mentions:

> TODO:
> 
>  Test standalone artifacts through REST API (via Python)
>  Test late binding of standalone artifacts to model versions (in Go and via REST)

as not yet fully done? 🤔 

Also, before merging this PR, I would suggest checking for non-regression by _also_ making use of the container image in a MR Deployment and check with the UI. (only after we're aligned on lgtm for this PR).

pkg/api/api.go
isinyaaa: oh yes, I forgot about it.

api/openapi/model-registry.yaml
isinyaaa: I tried to summarize in the commit description, but I can go into more detail:
Because we're using some vanilla models (i.e. not `*Create` or `*Update`) in routes that support `upsert`, we need to be able to send the model id in the request.
If we don't expose this field, we would need to incorrectly patch the Python autogen bindings, violating the openapi spec. Now, every client would need to do that, or we could create new models for the upsert routes that have this field exposed, but then the clients would lose on the simplicity of doing an upsert e.g. would have to convert to an intermediate model with no benefit.

tarilabs:
thank you for walking through the changes together @isinyaaa ,

well noted the modifications in https://github.com/kubeflow/model-registry/pull/376/commits/fdd02095dddf6b8a4b4b30bf0f665578a5b0ec33 are needed so that the POST modelversion/../artifact start to behave in upsert fashion

while none of the "base resource" are used in payload for std POST or PATCH, so the modification from that commit is compatible

nor impose any change on the goverter generated code

isinyaaa:
Before:
image

After:
image

Looks fine to me. And seems to be working the same.

Al-Pragliola (approved):
/lgtm

tarilabs (approved):
thank you @isinyaaa for all the changes, the discussions for clarifications, and [making a smoke-test](https://github.com/kubeflow/model-registry/pull/376#issuecomment-2356975130) with the deployed UI

/lgtm

as previously discussed and

/approve

Issue #357, opened 2024-09-06
Python Client: Issues with limit operation during iteration of models, versions 


rareddy:
**Describe the bug**
Issues with limit operation during iteration of models, versions 

**To Reproduce**
#### case 1:
If you run a query like below 

```
for model in registry.get_registered_models().limit(0):
    print(model)
```
or 
```
for model in registry.get_registered_models().limit(-1):
    print(model)
```
it ends with an error 
```
ServiceException: (500)
Reason: Internal Server Error
HTTP response headers: <CIMultiDictProxy('Content-Type': 'application/json; charset=UTF-8', 'Vary': 'Origin', 'Date': 'Fri, 06 Sep 2024 17:47:08 GMT', 'Content-Length': '169', 'x-envoy-upstream-service-time': '64', 'Server': 'istio-envoy')>
HTTP response body: code='' message='rpc error: code = InvalidArgument desc = max_result_size field value is required to be greater than 0 and less than or equal to 100. Set value:0'
```
#### case 2:
if run a query like 
```
for model in registry.get_registered_models().limit(1):
    print(model)
```
it displays the whole collection, not just limited to the first row.

**Expected behavior**
in case 1: it should return an empty collection
in case 2: it should be limited to outer bound specified

**Additional context**
Upon consulting, it looks like the Python library is treating `limit` as `page` semantics. In that case, I advise revising the API to clarify the iteration style and provide clear directions on its usage.

PR #379, opened 2024-09-11
Py: rename pager `limit` method to `page_size`

isinyaaa:
Closes: #357

This should sort out semantics issues.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -154,11 +154,15 @@ for version in latest_updates:
     ...
 ```
 
-By default, all queries well be `ascending`, but this method is also available for explicitness.
+By default, all queries will be `ascending`, but this method is also available for explicitness.
 
-> Note: You can also set the `page_size()` that you want to use, but the pager will automatically manage pages for you.
-> This is intended to aid in usability as currently the Model Registry backend treats model lists as a circular buffer,
-> and **will not end iteration** for you.
+> Note: You can also set the `page_size()` that you want the Pager to use when invoking the Model Registry backend.
+> When using it as an iterator, it will automatically manage pages for you.
+
+#### Implementation notes
+
+The pager will manage pages for you in order to prevent infinite looping.
+Currently, the Model Registry backend treats model lists as a circular buffer, and **will not end iteration** for you.
 
 ## Development
 
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -146,18 +146,19 @@ for version in registry.get_model_versions("my-model"):
     ...
 ```
 
-To customize sorting order or query limits you can also use
+You can also use `order_by_creation_time`, `order_by_update_time`, or `order_by_id` to change the sorting order
 
 ```py
-latest_updates = registry.get_model_versions("my-model").order_by_update_time().descending().limit(20)
+latest_updates = registry.get_model_versions("my-model").order_by_update_time().descending()
 for version in latest_updates:
     ...
 ```
 
-You can use `order_by_creation_time`, `order_by_update_time`, or `order_by_id` to change the sorting order.
+By default, all queries well be `ascending`, but this method is also available for explicitness.
 
-> Note that the `limit()` method only limits the query size, not the actual loop boundaries -- even if your limit is 1
-> you will still get all the models, with one query each.
+> Note: You can also set the `page_size()` that you want to use, but the pager will automatically manage pages for you.
+> This is intended to aid in usability as currently the Model Registry backend treats model lists as a circular buffer,
+> and **will not end iteration** for you.
 
 ## Development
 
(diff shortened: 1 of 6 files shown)


clients/python/tests/test_client.py
```diff
@@ -72,10 +72,10 @@ def test_register_existing_version(client: ModelRegistry):
         "model_format_version": "test_version",
         "version": "1.0.0",
     }
-    client.register_model(**params)
+    client.register_model(**params, metadata=None)
 
     with pytest.raises(StoreError):
-        client.register_model(**params)
+        client.register_model(**params, metadata=None)
```
tarilabs, line 78: question: is there a reason why we needed to be explicit of setting it to None here? 🤔

clients/python/src/model_registry/types/pager.py
```diff
@@ -73,12 +73,15 @@ def order_by_id(self) -> Pager[T]:
         self.options.order_by = OrderByField.ID
         return self.restart()
 
-    def limit(self, limit: int) -> Pager[T]:
-        """Limit the number of items to return.
+    def page_size(self, n: int) -> Pager[T]:
+        """Set the page size for each request.
```
tarilabs, line 77: +1

clients/python/README.md
```diff
@@ -146,18 +146,19 @@ for version in registry.get_model_versions("my-model"):
...
 ```
 
-You can use `order_by_creation_time`, `order_by_update_time`, or `order_by_id` to change the sorting order.
+By default, all queries well be `ascending`, but this method is also available for explicitness.
 
-> Note that the `limit()` method only limits the query size, not the actual loop boundaries -- even if your limit is 1
-> you will still get all the models, with one query each.
+> Note: You can also set the `page_size()` that you want to use, but the pager will automatically manage pages for you.
+> This is intended to aid in usability as currently the Model Registry backend treats model lists as a circular buffer,
+> and **will not end iteration** for you.
```
tarilabs, line 161: I would move this one into a separate box, title-ing it "implementation notes", because this is an implementation detail of the backend, which the end-user of MR Python client does not necessarily need to know.

clients/python/README.md
```diff
@@ -146,18 +146,19 @@ for version in registry.get_model_versions("my-model"):
...
 for version in latest_updates:
     ...
 ```
 
-You can use `order_by_creation_time`, `order_by_update_time`, or `order_by_id` to change the sorting order.
+By default, all queries well be `ascending`, but this method is also available for explicitness.
 
-> Note that the `limit()` method only limits the query size, not the actual loop boundaries -- even if your limit is 1
-> you will still get all the models, with one query each.
+> Note: You can also set the `page_size()` that you want to use, but the pager will automatically manage pages for you.
```
tarilabs, line 159:
```suggestion
> Note: You can also explicitly set the `page_size()` that you want this client to use when invoking the remote backend; in any case the pager will automatically manage pages for you.
```

wdyt?

tarilabs (commented):
thanks @isinyaaa 

/lgtm

and see some potential notes below?

clients/python/README.md
isinyaaa: it's not "in any case". It truly depends on the user using it as an iterator. If they use the internal methods it's up to them to make it not loop.

clients/python/README.md
isinyaaa: we can do that for sure. I just want to make it very clear that our limited interface has a reason to be.

clients/python/README.md
isinyaaa: I also wouldn't say "this client" because each iterator has its own options. So:

```
rm_iter = registry.get_registered_models()
rm_iter.page_size(1)
rm_iter2 = registry.get_registered_models() // new obj with default page size
```

clients/python/tests/test_client.py
tarilabs: sorry I'm not sure I follow 100%, isn't the default None already? 🤔 

https://github.com/kubeflow/model-registry/blob/7506eafbb18b60e52fcf63c27e09ee3c27f64be1/clients/python/src/model_registry/_client.py#L141

of which hints we're missing please? For awareness.

clients/python/README.md
tarilabs: s/this client/the iterator/ or s/this client/the pager/ ?

clients/python/tests/test_client.py
isinyaaa: The type checker complains that it doesn't know whether the `params` dict has an appropriate key for `metadata`. So if you specify that, instead of using the `params` dict, the type checker knows you're specifying a valid entry for `metadata`. simple as that :)

tarilabs: Aaa on this call side! Sorry I missed it

tarilabs (approved):
thank you for reflecting the changes @isinyaaa ,

/lgtm
/approve

PR #322, opened 2024-08-30
fix: fix Makefile and Dockerfile for idempotent images and to support multi-arch builds Fixes RHOAIENG-12078

dhirajsb:
## Description

Cleanup Makefile and Dockerfile to make them reproducable
Add multi-arch platform builds

## How Has This Been Tested?

CI will verify Makefile and Dockerfile changes
Dockerfile changes can be tested locally by updating model-registry.yaml

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [X] The commits have meaningful messages; the author will squash them after approval or in case of manual merges will ask to merge with squash.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).

If you have UI changes

- [ ] The developer has added tests or explained why testing cannot be added.
- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

PR #403, opened 2024-09-19
Fix type assert generation on MacOS

isinyaaa:
This makes type asserts in Python instead of bash, hopefully making it more portable.

## Description
Performing a `touch api/openapi/model-registry.yaml && make` currently won't work on MacOS, after #322. To fix this, we should drop the `-P` flag on `grep` as it's GNU specific.
As suggested by @tarilabs, the script could be significantly simplified by using a Python script to gather type asserts deterministically.

## How Has This Been Tested?
Ran various variations of the `make` command to ensure it won't break.

## Merge criteria:

If you have UI changes

cc: @tarilabs @dhirajsb


scripts/gen_type_asserts.py
```diff
@@ -0,0 +1,53 @@
...
+        with path.open() as f:
+            yield "".join(f.readlines()[14:])
+        path.unlink()
+
+
+if __name__ == "__main__":
+    if len(argv) > 1:
+        root = Path(argv[1])
+        assert root.is_dir(), f"{root} is not a directory"
+    else:
```
isinyaaa, line 18: on a second thought we might not even need this

scripts/gen_type_asserts.py
```diff
@@ -0,0 +1,70 @@
...
+            # skip first 14 lines of boilerplate
+            for raw in islice(f.readlines(), 14, None):
+                line = raw.rstrip()
+                buf.append(line)
+                if line.startswith("func"):
+                    in_func = True
+                elif line.startswith("}") and in_func:
+                    in_func = False
+                    yield "\n".join(buf)
+                    buf.clear()
```
tarilabs, line 22: I'd take the approach of re-ordering in the file by keeping the header, as shown here: https://github.com/kubeflow/model-registry/pull/404/files#diff-30505a65ba1ad63e91b0b363704b01d852f43625afea32db6355c7626679a7aaR8-R28

this way we don't have to rely on a fixed `14` my2c

tarilabs (commented):
what I lot about this approach is that it also take in charge of processing the model_ files

tarilabs (approved):
- https://github.com/kubeflow/model-registry/pull/403#pullrequestreview-2316680793
- https://github.com/kubeflow/model-registry/pull/404#issuecomment-2362926052

/lgtm
/approve

Issue #369, opened 2024-09-09
standardize artifact API


isinyaaa:
**Is your feature request related to a problem? Please describe.**
Currently there is no stable API for interacting with artifacts. For example:
1. To create a new artifact, you need to `POST /model_versions/{mvId}/artifacts`, but then you can only fetch from `GET /model_artifacts` (which is restricted to `ModelArtifact`s) or `GET /model_versions/{mvId}/artifacts` (which will fetch all of them).
2. To update a new artifact, you need to also `POST /model_versions/{mvId}/artifacts`, and there's no `PATCH` option.
This makes it impossible to work with generic artifacts, requiring extraneous work-arounds.
It also disallows standalone artifacts, which we support.

**Describe the solution you'd like**
Introduce a new CRUD API under `/artifacts` to enable generic `Artifact` model handling:

- `POST /artifacts` : Creates a new artifact.
- `PATCH /artifacts/{id}`: Updates the artifact model
- `GET /artifacts/{id}`: Get a single artifact
- `GET /artifacts`: Get all artifacts
- `GET /artifact`: Search for an artifact

> Note: An `Artifact` is a composed polymorphic type that can be "unwrapped" as one of (currently) `DocArtifact` or `ModelArtifact`.

As opposed to the existing `/model_versions/{mvId}/artifacts`, the new API will NOT associate `Artifacts` to a `ModelVersion`, instead allowing for a workflow with standalone models. 

The current APIs for `/model_versions/{mvId}/artifacts` and `/model_artifacts` will be maintained.

**Additional context**

This does not take into account #231 as it is a bug.

PR #406, opened 2024-09-20
OAS: add generic artifacts routes

isinyaaa:
Closes: #369 

## Description
Implements generic artifact routes as proposed.

There will be a follow-up to add new methods to the Python client as well.

## How Has This Been Tested?
Updated unit tests to cover new core methods.

## Merge criteria:

If you have UI changes


patches/type_asserts.patch
```diff
@@ -2,6 +2,75 @@ diff --git a/internal/server/openapi/type_asserts.go b/internal/server/openapi/t
...
+-		}
+-	}
++	// elements := map[string]interface{}{
++	// 	"artifactType": obj.ArtifactType,
++	// }
++	// for name, el := range elements {
++	// 	if isZero := IsZeroValue(el); isZero {
++	// 		return &RequiredError{Field: name}
++	// 	}
++	// }
```
isinyaaa, line 24: unfortunately the generation isn't perfect for these kinds of models. Patched like `MetadataValue`

tarilabs: I would just "file this" under another limitation of openapi-codegen

api/openapi/model-registry.yaml
```diff
@@ -970,36 +1093,39 @@ components:
...
-          type: string
-          default: "doc-artifact"
       allOf:
-        - $ref: "#/components/schemas/BaseArtifact"
+        - $ref: "#/components/schemas/BaseArtifactUpdate"
+        - type: object
+          properties:
+            artifactType:
+              type: string
+              default: "doc-artifact"
```
tarilabs, line 1121: Have you tried what happens if you don't make this changes, how does it impact the .patch below?

saying because iirc with @lampajr we tried several "places" where to put the discriminator, and that indeed impacted on the way the codegen behaved for this attribute

isinyaaa: the request fails, the current properties setup outside the AllOf block is invalid iiuc

lampajr: >the current properties setup outside the AllOf block is invalid iiuc

weird because that's exactly what was making the auto-generation working - but I don't know what was the culprit.. either the generator that is broken (not surprising) or the openapi spec that was wrong

tarilabs: > weird because that's exactly what was making the auto-generation working

I recall we had to move these discriminator around in a few places to indeed get the openapi-codegen working, and ultimately we settled for a "convention" on how to make the discriminator being generated as expected.

That's also to be added that new stuff is being added in this PR, so I would also not exclude that by different set of entities/endpoint, that "convention" no longer holds? 🤷 

I'm happy to hear however @lampajr we did recall the same thing!

isinyaaa: I'll review that because I remember just later yesterday I found an issue with the docker builder. So I might have tested stale versions.
TBH I don't remember the generation being different in any case. However this usage of `allOf` is what's listed on they swagger docs.

isinyaaa: Looks like this actually works, I'll update it soon.

internal/converter/openapi_reconciler.go
```diff
@@ -24,6 +24,11 @@ type OpenAPIReconciler interface {
 	// goverter:ignore Id CreateTimeSinceEpoch LastUpdateTimeSinceEpoch Name RegisteredModelId
 	UpdateExistingModelVersion(source OpenapiUpdateWrapper[openapi.ModelVersion]) (openapi.ModelVersion, error)
 
+	// Ignore all fields that can't be updated
+	// goverter:default InitWithExisting
+	// goverter:autoMap Update
+	UpdateExistingArtifact(source OpenapiUpdateWrapper[openapi.Artifact]) (openapi.Artifact, error)
```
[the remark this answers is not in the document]
tarilabs: I think the comment phrased in another way is possibly why differently than the other converter, this one seems to be missing the 

```go
// goverter:ignore Id CreateTimeSinceEpoch LastUpdateTimeSinceEpoch Name ArtifactType
```

?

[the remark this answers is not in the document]
lampajr: >this one seems to be missing the

I think that's because in this case the `Artifact` object, in according to how the model is auto-generated by openapi-generator, does not have the concrete fields (e.g., `id`..) but it does have `ModelArtifact` and `DocArtifact` and in according to which one is not `Nil` it will be of that type.

Given that, I think `UpdateExistingArtifact` should be somehow re-worked to consider this situation, which is different from all other entities we have right now.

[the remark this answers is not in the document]
tarilabs: possibly a way to ignore the same set, whether they are in the DocArtifact or ModelArtifact with a dot notation?

not sure if that would work with the pointers tho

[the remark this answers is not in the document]
isinyaaa: we can simply create our own helper IMO. It would call into the appropriate update func depending on the artifact type, wdyt?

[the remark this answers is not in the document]
tarilabs: sounds good with me, thanks @isinyaaa ; when you "wire it" in the comment, it would be helpful to "note down" this difference/motivation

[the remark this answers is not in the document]
isinyaaa: making a direct helper on the generator won't do, as `autoMap` won't take in custom mapping functions.
I'm going with a custom helper that takes in the reconciler to call the appropriate methods for each model.

tarilabs (commented):
https://github.com/kubeflow/model-registry/pull/406#issuecomment-2368313799
 +1

thank you again @isinyaaa 
/lgtm

focusing on the Go side, anything 👀 worth noting @Al-Pragliola ?

I see the new routes and I don't see changes to the path of the existing REST apis, beyond that 
```go
	if artifact == nil {
		return nil, fmt.Errorf("invalid artifact pointer, can't upsert nil: %w", api.ErrBadRequest)
	}
```

so I'm good with merging asap, I just wanted to double-check this understanding

Al-Pragliola:
> [#406 (comment)](https://github.com/kubeflow/model-registry/pull/406#issuecomment-2368313799) +1
> 
> thank you again @isinyaaa /lgtm
> 
> focusing on the Go side, anything 👀 worth noting @Al-Pragliola ?
> 
> I see the new routes and I don't see changes to the path of the existing REST apis, beyond that
> 
> ```go
> 	if artifact == nil {
> 		return nil, fmt.Errorf("invalid artifact pointer, can't upsert nil: %w", api.ErrBadRequest)
> 	}
> ```
> 
> so I'm good with merging asap, I just wanted to double-check this understanding

Looks Good To Me 👍

tarilabs (approved):
thanks again @isinyaaa for the great work and everyone else for 👀  on this

/approve

PR #421, opened 2024-09-23
py: add basic service URL resolver

isinyaaa:
## Description
Adds a simple constructor that resolves cluster service URLs.

## How Has This Been Tested?

Tested with the latest UI using a MR deployed directly through the settings panel.

The client then connects:
```py

mr = ModelRegistry.from_service("modelregistry-sample", "isinyaaa")
```
and creates a new MV:
```py
mr.register_model("test", "s3://catopia/meow.onnx", model_format_name="oooo", model_format_version="v1", version="123")
```

> ❗  IMPORTANT: to make this work with ODH deployments you need to create a clusterwide rolebinding for the current DSP service accounts, with the `odh-dashboard` role.

Change was reflected on the UI as expected:

image

## Merge criteria:

If you have UI changes


clients/python/src/model_registry/_client.py
```diff
@@ -87,11 +87,78 @@ def __init__(
...
+        from kubernetes import client, config
+
+        config.load_incluster_config()
+        if not ns:
+            kcustom = client.CustomObjectsApi()
+            try:
+                dsc_raw = kcustom.list_cluster_custom_object(
+                    group="datasciencecluster.opendatahub.io",
+                    version="v1",
+                    plural="datascienceclusters",
```
tarilabs, line 113: I would extract these maybe in a const tuple of sort, and ...

clients/python/src/model_registry/_client.py
```diff
@@ -87,11 +87,78 @@ def __init__(
...
+                ns = t.cast(
+                    dict[str, t.Any],
+                    dsc_raw["items"][0],
+                )["status"]["components"]["modelregistry"]["registriesNamespace"]
+
+        kcore = client.CoreV1Api()
+        serv = t.cast(client.V1Service, kcore.read_namespaced_service(name, ns))
+        meta = t.cast(client.V1ObjectMeta, serv.metadata)
+        ext_addr = t.cast(dict[str, str], meta.annotations).get(
+            "routing.opendatahub.io/external-address-rest"
```
tarilabs, line 130: ...and this in a const as well, so it might be customizable by someone using same contract of CRD ?

clients/python/src/model_registry/_client.py
```diff
@@ -87,11 +87,78 @@ def __init__(
...
+                    if port.app_protocol == "http"
+                ),
+                "8080",
+            )
+        else:
+            msg = "No external address found for secure connection"
+            raise StoreError(msg)
+
+        return cls(
+            f"https://{host}",
```
tarilabs, line 152: I might be missing something, but I think in the case of `is_secure` this ought to be switched to `http://` prefix? what am I missing? 🤔

tarilabs (commented):
thank you @isinyaaa , some early 👀 for your considerations

tarilabs:
@Al-Pragliola any thoughts on potential amendments which would not require a rolebinding or anyway any changes to the permission settings?

clients/python/src/model_registry/_client.py
isinyaaa: oh that's true.

Al-Pragliola:
> @Al-Pragliola any thoughts on potential amendments which would not require a rolebinding or anyway any changes to the permission settings?

mm no, I can't think of another way to let the notebook pod interact with the k8s api-server, but I think the odh-dashboard role is a bit too permissive, we should have an ad-hoc role with the only the required permissions

tarilabs:
Thanks @Al-Pragliola , adding to https://github.com/kubeflow/model-registry/pull/421#issuecomment-2371425492, what about having some "fallbacks" in case that permission is not there?  Do we believe we might have some options?

For example, attemping in case a list K8s resources in a set of known to be used namespaces, may require simpler/lower permissions? for example, listing MR CR instances in a set of known namespaces in case this require lower permission, or listing Services again in known namespaces, etc?

isinyaaa:
@tarilabs if we can't get DSC it falls back to looking for the instance in the specified namespace, or the default (`kubeflow` for upstream, should be changed to `odh-model-registries` when merging midstream)

tarilabs:
> should be changed to `odh-model-registries` when merging midstream

I don't think that is viable, since there will be only 1 "upstream" pypi of the MR python client?

Hence why I believe would have been preferable to adopt a fallback(s) approach

clients/python/src/model_registry/_utils.py
```diff
@@ -107,3 +110,163 @@ def wrapper(*args: object, **kwargs: object) -> object:
...
+class Address(t.NamedTuple):
+    protocol: str
+    host: str
+    port: int
+
+
+@dataclass
+class Kube:
+    user_token: str | None = None
+    # TODO: do we need to take care of custom CA config too?
```
isinyaaa, line 143: @dhirajsb is this necessary when talking to Kubernetes?

clients/python/src/model_registry/_utils.py
```diff
@@ -107,3 +110,163 @@ def wrapper(*args: object, **kwargs: object) -> object:
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.api_client.close()
+
+    def try_get(
+        self, op: t.Callable[[], t.Any], as_user: bool = False
+    ) -> Result[t.Any, client.ApiException]:
+        if as_user and self.user_token is not None:
+            # NOTE: even though this config is consumed by the RESTClient, auth is refresh on every request: https://github.com/kubernetes-client/python/blob/b7ccf179f1b0194a0ed18e39fb063ef8a963fc6b/kubernetes/client/api_client.py#L166
```
isinyaaa, line 167: Nit: typo
```suggestion
            # NOTE: even though this config is consumed by the RESTClient, auth is refreshed on every request: https://github.com/kubernetes-client/python/blob/b7ccf179f1b0194a0ed18e39fb063ef8a963fc6b/kubernetes/client/api_client.py#L166
```

clients/python/src/model_registry/_utils.py
isinyaaa: @dhirajsb:
> no, kubernetes calls use a kubernetes CA, which should be already configured in the pod

rareddy:
IMO the Python library should not use any OpenShift calls directly. That brings in additional dependency on where the MR deployed into a client library. I suggest we feed this information through some kind of config or env to keep it simple, then think through how this config can be supplied. I am thinking like `.kubeconfig` wdyt?

dhirajsb:
Yes, the DSC namespace discovery is an odh/rhoai feature, since there is no equivalent place to discover registries namespace in kubeflow. Although, the default MR manifests kind of make `kubeflow` namespace the default registry namespace. 

We should keep things simple and allow this capability to discover the presence of DSC CR in the client library. The Python client can simply ignore that service discovery path if DSC resource type is not present and try to lookup registries using the hostname provided by users. 

If we want to allow Python API to be service location agnostic, another way to do that would be by supporting an env variable for a default MR namespace, and also support specifying a direct service host. That way it can be also be injected into the notebook from a user supplied configmap (which could be cluster/env specific) if needed.

dhirajsb:
The `odh-dashboard` role referenced by @isinyaaa is a workaround, not a requirement once Notebooks grants the right permissions midstream. That's being worked on by the odh notebooks team.

rareddy:
Apologies, when I mentioned OpenShift in general I was referring to Kube. saying in general Python client should not make any assumptions on where the MR is hosted IMO. Thinking more in terms of what you said above

>If we want to allow Python API to be service location agnostic, another way to do that would be by supporting an env variable for a default MR namespace, and also support specifying a direct service host. That way it can be also be injected into the notebook from a user supplied configmap (which could be cluster/env specific) if needed.

then make some other automation happen to provide the config/env based on environment.

rareddy:
similar pattern by MLFlow https://mlflow.org/docs/latest/model-registry.html#id10

tarilabs:
> similar pattern by MLFlow https://mlflow.org/docs/latest/model-registry.html#id10

I believe this is the link you wanted: https://mlflow.org/docs/latest/model-registry.html#databricks-unity-catalog-model-registry (the original link is an entry in a TOC)

Sounds nice and following up from https://github.com/kubeflow/model-registry/pull/447#discussion_r1787293244 , what about using something analogous to Viper on the Go side, like https://pypi.org/project/12factor-configclasses/ that would support said injection?

This way:
- for upstream KF, we can provide them in the Manifest for example
- for midstream, folks can rely on their own Operators making the env or CM

Issue #398, opened 2024-09-17
[Proposal] Make Openapi(REST) client the only entrypoint


Al-Pragliola:
At the moment Model Registry has two client entrypoints in the codebase:

1) `pkg/openapi`: the auto generated client from the openapi schema
2) `pkg/core`: a grpc client that talks directly with the MLMD service

![image](https://github.com/user-attachments/assets/cd253165-3603-4449-a57e-71933b72499b)

This creates a strong dependency on the MLMD service and makes difficult changing the underlying service without major breaking changes in the future. My proposal is to move the package from `pkg/core` to `internal/core` and do a refactoring to make it a switchable datastore for the model registry. What I mean by switchable datastore is to change the config arguments that we take in the proxy:

- (change) mlmd_hostname -> datastore_hostname
- (change) mlmd_port -> datastore_port
- (create) datastore_service(type?) = mlmd / sqlite / whatever

then in model registry we can use a factory method to use the correct datastore implementation of the interface [ModelRegistryServiceAPIServicer](https://github.com/kubeflow/model-registry/blob/main/internal/server/openapi/api.go#L64)

I did a check on which projects import `pkg/core` and for upstream I didn't find any (https://pkg.go.dev/github.com/kubeflow/model-registry/pkg/core?tab=importedby), but we have a breaking change in the midstream (https://pkg.go.dev/github.com/opendatahub-io/model-registry/pkg/core?tab=importedby) we should make the team that works on [opendatahub-io/odh-model-controller](https://github.com/opendatahub-io/odh-model-controller) aware of this change.

lampajr:
Hey @Al-Pragliola, you have my +1 on this proposal for all the motivation you have already highlighted here!

I think that having a single entrypoint is always the best solution, especially here where we don't know if we will change the underlying "datastore" in the future.

Moreover, as you pointed out, I think this is the right timing to do this change as we don't have many external projects relying on the internal `pkg/core`.

>(change) mlmd_hostname -> datastore_hostname
(change) mlmd_port -> datastore_port
(create) datastore_service(type?) = mlmd / sqlite / whatever

I agree on this generalization and I also like the `datastore` name.

>then in model registry we can use a factory method to use the correct datastore implementation of the interface [ModelRegistryServiceAPIServicer](https://github.com/kubeflow/model-registry/blob/main/internal/server/openapi/api.go#L64)

[...]

PR #425, opened 2024-09-24
feat(proxy): introduce datastore abstraction

Al-Pragliola:
## Description
As proposed in issue #398 , I have introduced an abstraction to the underlying metadata storage service and added a `--datastore-type` flag to handle different types of datastore, each with a possible custom configuration. In this scenario MLMD is just one implementation of the abstraction and this opens the possibility to add more in the future, the next logical step should be to move the `pkg/core` to `internal`.

## How Has This Been Tested?

1. Tested with `make test-nocache`

2. Tested the python client with `make test-e2e`

3. Tested on a local kind cluster

```shell
kind create cluster

IMG_REGISTRY=localhost:{kind-registry-port} IMG_ORG=model-registry IMG_REPO=model-registry IMG_VERSION=local-1.0 make image/build

docker push localhost:{kind-registry-port}/model-registry/model-registry:local-1.0 (if needed --tls-verify=false)

kubectl create namespace kubeflow

kubectl apply -k "./manifests/kustomize/overlays/db"

kubectl set image -n kubeflow deployments/model-registry-deployment rest-container=localhost:{kind-registry-port}/model-registry/model-registry:local-1.0

kubectl port-forward svc/model-registry-service -n kubeflow 8081:8080 &

curl --request POST   --url http://localhost:8081/api/model_registry/v1alpha3/registered_models   --header 'content-type: application/json'   --data '{
  "name": "test",
  "description": "test"
}'

curl http://localhost:8081/api/model_registry/v1alpha3/registered_models
```

expected result:

```json
{"items":[{"createTimeSinceEpoch":"1727181207919","customProperties":{},"description":"test","id":"1","lastUpdateTimeSinceEpoch":"1727181207919","name":"test"}],"nextPageToken":"","pageSize":0,"size":1}
```

## Merge criteria:


tarilabs (commented):
thank you @Al-Pragliola especially for as well 15e3ec2

can you also reflect any changes required in the https://github.com/kubeflow/model-registry/blob/main/docs/mr_go_library.md dev doc, please?

overall
/lgtm

Al-Pragliola:
> thank you @Al-Pragliola especially for as well [15e3ec2](https://github.com/kubeflow/model-registry/commit/15e3ec238181f4c897dd9826a357563d18c9adc5)
> 
> can you also reflect any changes required in the https://github.com/kubeflow/model-registry/blob/main/docs/mr_go_library.md dev doc, please?
> 
> overall /lgtm

At the moment there are no changes to clients using `pkg/core`, the abstraction is for internal use only.

tarilabs (commented):
thanks for the clarification and the refactor for this factory @Al-Pragliola 

overall 
/lgtm

and as mentioned I'd
/hold
this until we have at least dry-run the release process to implement pinning so we efficiently update the manifests accordingly to KF/manifests

cmd/proxy.go
```diff
@@ -72,16 +50,29 @@ func init() {
...
+	DatastorePort     int
+	DatastoreType     string
 }
 
 var proxyCfg = ProxyConfig{
-	MLMDHostname: "localhost",
-	MLMDPort:     9090,
+	DatastoreHostname: "localhost",
+	DatastorePort:     9090,
+	DatastoreType:     "mlmd",
```
isinyaaa, line 77: should we have an enum for this?

cmd/proxy.go
```diff
@@ -134,16 +121,17 @@ func init() {
...
+	proxyCmd.Flags().IntVar(&proxyCfg.DatastorePort, "mlmd-port", proxyCfg.DatastorePort, "MLMD port")
+	if err := proxyCmd.Flags().MarkDeprecated("mlmd-port", "please use --datastore-port instead"); err != nil {
+		glog.Errorf("error marking flag as deprecated: %v", err)
+	}
 
-var proxyCfg = ProxyConfig{
-	MLMDHostname: "localhost",
-	MLMDPort:     9090,
+	proxyCmd.Flags().StringVar(&proxyCfg.DatastoreHostname, "datastore-hostname", proxyCfg.DatastoreHostname, "Datastore hostname")
+	proxyCmd.Flags().IntVar(&proxyCfg.DatastorePort, "datastore-port", proxyCfg.DatastorePort, "Datastore port")
```
pboyd, line 135: What do you think about combining these into a more generic single argument? Something like `--datastore-connection-string` (a less wordy name would be nice, that's all I can think of right now).

My thinking is that the MLMD address could be a single `host:port`, and a database connection can also be given as a single argument. But it would take a lot of arguments to split up the database connection params otherwise (`--username`, `--dbname`, among other vendor-specific options that someone will certainly want--like `sslmode` for PostgreSQL).

internal/datastore/builder.go
```diff
@@ -0,0 +1,36 @@
...
+	ErrCreatingDatastore    = errors.New("error creating datastore")
+	ErrUnsupportedDatastore = errors.New("unsupported datastore type")
+)
+
+type TeardownFunc func() error
+
+type Builder interface {
+	Build() (api.ModelRegistryApi, error)
+	Teardown() error
+}
```
pboyd, line 20: Is this used anywhere? I see that `MLMDService` implements it, but I don't see anywhere that uses this type.

pboyd (approved):
This looks good. I left a couple comments, but it's nothing major.

/lgtm

cmd/proxy.go
```diff
@@ -4,106 +4,92 @@ import (
...
 		}
 
-		ModelRegistryServiceAPIService := openapi.NewModelRegistryServiceAPIService(service)
+		ModelRegistryServiceAPIService := openapi.NewModelRegistryServiceAPIService(ds)
 		ModelRegistryServiceAPIController := openapi.NewModelRegistryServiceAPIController(ModelRegistryServiceAPIService)
 
 		router.SetRouter(openapi.NewRouter(ModelRegistryServiceAPIController))
 
-		glog.Infof("connected to MLMD server")
+		glog.Infof("connected to Datastore server")
```
rareddy, line 88: Log the data store details it is connected to with full identification; same in the error above. of log at the beginning it is trying connect the datastore, here just announce the success or failure.

cmd/proxy.go
rareddy: +1 for above

tarilabs:
I suppose this need a rebase and possibly account for https://github.com/kubeflow/model-registry/pull/425#pullrequestreview-2761196933, if misunderstand pls lmk

cmd/proxy.go
Al-Pragliola: Changed in https://github.com/kubeflow/model-registry/pull/425/commits/2486765c9ab30c09d2d673262096f14c220e60ce

cmd/proxy.go
Al-Pragliola: Building on this idea, there is now only one `--datastore-type` common to all datastore types, each datastore can have its own configuration settings, so we don't have to "squeeze" different needs into common flags (e.g. `--datastore-connection` in MLMD is the address of the service, in an embedded service there would be no external address to reach) https://github.com/kubeflow/model-registry/pull/425/commits/d4659a787ff2c236f470319e26b6006bb6b1ea59

If we see that there would be too many parameters, we already have viper in the project and the ability to use a config.yaml, so it's not really a problem

internal/datastore/builder.go
Al-Pragliola: nice catch, fixed thanks

cmd/proxy.go
pboyd: That sounds like a great plan.

pboyd (approved):
/lgtm

rareddy (approved):
/lgtm
/approve

PR #434, opened 2024-09-27
Fix disabled text color on DescriptionListDescription

Griffin-Sullivan:
## Description
In my last PR I was using `text.iconColorSubtle` to get around the disabled color being to low of a color ratio on a white background. It was causing our a11y testing to fail. Turns out you don't need to pass color contrast ratios on disabled elements so conditionally adding the `aria-disabled` attribute will cause the tests to pass.

The color change is probably not noticeable enough to anyone but can add an image if requested.

## How Has This Been Tested?
Running cypress mock tests

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/components/DashboardDescriptionListGroup.tsx
+++ b/clients/ui/frontend/src/components/DashboardDescriptionListGroup.tsx
@@ -110,9 +110,10 @@ const DashboardDescriptionListGroup: React.FC<DashboardDescriptionListGroupProps
           </Flex>
         </DescriptionListTerm>
       )}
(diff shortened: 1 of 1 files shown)


lucferbux (approved):
/lgtm

Ok this is great, I would imagine this is gonna happen again soon, we should track this and make sure we port it in other places.

Griffin-Sullivan:
@lucferbux I added a comment to the PF6 SPIKE so people will be aware when swapping to PF6.

ederign (approved):
/lgtm

Issue #380, opened 2024-09-12
refactor pytest e2e on top of k8s manifests


tarilabs:
Following the great work and refactor from @isinyaaa where the pytest are now being leveraged for `e2e`-testing, 

I would propose to move away from the use of `docker-compose.yaml`,
and rebase the e2e [on top of the KF manifests](https://github.com/kubeflow/model-registry/tree/main/manifests/kustomize#readme),
with the following proposal:

## Before all e2e tests

This is where we typically launch the docker compose up.

1. `kind create cluster`
2. `kubectl create namespace kubeflow`
3. from the [manifests directory](https://github.com/kubeflow/model-registry/tree/main/manifests/kustomize), `kubectl apply -k overlays/db`
4. [wait for Deployments like we do here](https://github.com/kubeflow/model-registry/blob/7506eafbb18b60e52fcf63c27e09ee3c27f64be1/.github/workflows/build-image-pr.yml#L61-L64)
5. launch `kubectl port-forward svc/model-registry-service -n kubeflow 8080:8080` in the background, maybe noting down pid
6. maybe await a couple seconds for the port fwd [like we do here](https://github.com/kubeflow/model-registry/blob/7506eafbb18b60e52fcf63c27e09ee3c27f64be1/.github/workflows/build-image-pr.yml#L84-L85)

## After each test

This is where we typically delete the sqlite file, and restart the mlmd service in docker compose.

Seems to me this:

```sh
kubectl exec -n kubeflow -it $(kubectl get pods -l component=db -o jsonpath="{.items[0].metadata.name}" -n kubeflow) -- mysql -u root -ptest -D metadb -e "START TRANSACTION; DELETE FROM Artifact; DELETE FROM ArtifactProperty; DELETE FROM Association; DELETE FROM Attribution; DELETE FROM Context; DELETE FROM ContextProperty; DELETE FROM Event; DELETE FROM EventPath; DELETE FROM Execution; DELETE FROM ExecutionProperty; DELETE FROM ParentContext; COMMIT;"
```

should do equivalently (yes it's a 1-liner).

I have run a couple scripts locally, for the same model name and version, which is passing (wouldn't otherwise) but not the full e2e test suite. I prefer to mention this in full transparency, because I don't have full confirmation beforehand this will be enough.

## After all e2e tests

This is where we typically do the docker compose down.

Should be enough to:
1. stop ps of port fwd from the previous step
2. `kubectl delete cluster`

# Additional considerations

[...]

PR #447, opened 2024-10-02
use deployed MR for python e2e tests

isinyaaa:
Fixes: #380

## Description
Python e2e tests will use an environment variable to MR, delegating clean-up to a separate script.
The script is made environment-aware using envvars.

The CI has been updated to account for the changes, separating the Python workflows into 3 types to make it clearer.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/Makefile
+++ b/clients/python/Makefile
@@ -1,6 +1,5 @@
 all: install tidy
 
-IMG_REGISTRY ?= docker.io
 IMG_VERSION ?= latest 
 
 .PHONY: install
@@ -14,12 +13,13 @@ install:
 clean:
 	rm -rf src/mr_openapi
 
-.PHONY: build-mr
-build-mr:
-	cd ../../ && IMG_REGISTRY=${IMG_REGISTRY} IMG_VERSION=${IMG_VERSION} make image/build
+.PHONY: deploy-latest-mr
+deploy-latest-mr:
+	cd ../../ && IMG_VERSION=${IMG_VERSION} make image/build && LOCAL=1 ./scripts/deploy_on_kind.sh
+	kubectl port-forward -n kubeflow services/model-registry-service 8080:8080 &
 
 .PHONY: test-e2e
-test-e2e: build-mr
+test-e2e: deploy-latest-mr
 	poetry run pytest --e2e -s
 
 .PHONY: test
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -6,7 +6,7 @@ import tempfile
 import time
 from contextlib import asynccontextmanager
 from pathlib import Path
-from time import sleep
+from urllib.parse import urlparse
 
 import pytest
 import requests
@@ -29,13 +29,14 @@ def pytest_collection_modifyitems(config, items):
             continue
 
 
-REGISTRY_HOST = "http://localhost"
-REGISTRY_PORT = 8080
-REGISTRY_URL = f"{REGISTRY_HOST}:{REGISTRY_PORT}"
-COMPOSE_FILE = "docker-compose.yaml"
-MAX_POLL_TIME = 1200  # the first build is extremely slow if using docker-compose-*local*.yaml for bootstrap of builder image
+REGISTRY_URL = os.environ.get("MR_URL", "http://localhost:8080")
+parsed = urlparse(REGISTRY_URL)
+host, port = parsed.netloc.split(":")
+REGISTRY_HOST = f"{parsed.scheme}://{host}"
+REGISTRY_PORT = int(port)
+
+MAX_POLL_TIME = 10
 POLL_INTERVAL = 1
-DOCKER = os.getenv("DOCKER", "docker")
 start_time = time.time()
 
 
@@ -64,38 +65,8 @@ def poll_for_ready():
         time.sleep(POLL_INTERVAL)
 
 
-@pytest.fixture(scope="session")
-def _compose_mr(root):
-    print("Assuming this is the Model Registry root directory:", root)
-    shared_volume = root / "test/config/ml-metadata"
-    sqlite_db_file = shared_volume / "metadata.sqlite.db"
-    if sqlite_db_file.exists():
-        msg = f"The file {sqlite_db_file} already exists; make sure to cancel it before running these tests."
-        raise FileExistsError(msg)
-    print(f" Starting Docker Compose in folder {root}")
-    p = subprocess.Popen(  # noqa: S602
-        f"{DOCKER} compose -f {COMPOSE_FILE} up",
-        shell=True,
-        cwd=root,
-    )
-    yield
-
-    p.kill()
-    print(f" Closing Docker Compose in folder {root}")
-    subprocess.call(  # noqa: S602
-        f"{DOCKER} compose -f {COMPOSE_FILE} down",
-        shell=True,
-        cwd=root,
-    )
-    try:
-        os.remove(sqlite_db_file)
-        print(f"Removed {sqlite_db_file} successfully.")
-    except Exception as e:
-        print(f"An error occurred while removing {sqlite_db_file}: {e}")
-
-
 def cleanup(client):
-    async def yield_and_restart(_compose_mr, root):
+    async def yield_and_restart(root):
         poll_for_ready()
         if inspect.iscoroutinefunction(client) or inspect.isasyncgenfunction(client):
             async with asynccontextmanager(client)() as async_client:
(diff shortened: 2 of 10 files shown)


tarilabs:
On brief look I like a lot the re-use of the kind startup part, the workflow-dispatch to trigger manually if desired and the industrialization of https://github.com/kubeflow/model-registry/issues/380#:~:text=we%20do%20here-,After%20each%20test,-This%20is%20where

thank you so far @isinyaaa @Al-Pragliola this is looking great already :)

clients/python/Makefile
```diff
@@ -14,12 +13,13 @@ install:
 clean:
 	rm -rf src/mr_openapi
 
-.PHONY: build-mr
-build-mr:
-	cd ../../ && IMG_REGISTRY=${IMG_REGISTRY} IMG_VERSION=${IMG_VERSION} make image/build
+.PHONY: deploy-latest-mr
+deploy-latest-mr:
+	cd ../../ && IMG_VERSION=${IMG_VERSION} make image/build && LOCAL=1 ./scripts/deploy_on_kind.sh
+	kubectl port-forward -n kubeflow services/model-registry-service 8080:8080 &
```
isinyaaa, line 19: should we kill this after running the tests?

tarilabs: I think a broader limitation I see atm, is that you cannot re-issue `make test-e2e` on the terminal after a first time: say some of the pytest failed, we amend something, and we want to try again.

I believe it would be awesome if scripts/deploy_on_kind.sh could handle those cases, ie:
- is there already the cluster? Don't try to create a kind cluster.
- is there already a namespace for the model registry? Don't return as-error, maybe leverage apply like in `kubectl create namespace "$MR_NAMESPACE" --dry-run=client -o yaml | kubectl apply -f -` ?
- is there already a port forward to 8080? I do not try to open a new one

wdyt?

scripts/deploy_on_kind.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+MR_NAMESPACE="${MR_NAMESPACE:-kubeflow}"
+
+if [[ -n "$LOCAL" ]]; then
```
tarilabs, line 7:
```suggestion
# modularity to allow re-use this script against a remote k8s cluster
if [[ -n "$LOCAL" ]]; then
```

I believe it would be easier for people to notice this which is great if we annotate with a small comment here :)

scripts/cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ -n "$LOCAL" ]]; then
+    echo 'Cleaning up local sqlite DB'
+
+    sqlite3 test/config/ml-metadata/metadata.sqlite.db <<<"BEGIN TRANSACTION; $SQL_CMD"
+else
+    echo 'Cleaning up kubernetes MySQL DB'
+
+    kubectl exec -n "$MR_NAMESPACE" -it "$(kubectl get pods -l component=db -o jsonpath="{.items[0].metadata.name}" -n "$MR_NAMESPACE")" \
+        -- mysql -u root -ptest -D "$TEST_DB_NAME" -e "START TRANSACTION; $SQL_CMD"
```
tarilabs, line 33:
```suggestion
    kubectl exec -n "$MR_NAMESPACE" -it "$(kubectl get pods -l component=db -o jsonpath="{.items[0].metadata.name}" -n "$MR_NAMESPACE")" \
        -- mysql -u root -ptest -D "$TEST_DB_NAME" -e "$SQL_CMD"
```



What about moving the `START TRANSACTION;` in the `SQL_CMD` variable, this way we handle as 1 single Tx script?

scripts/cleanup.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+
+MR_NAMESPACE="${MR_NAMESPACE:-kubeflow}"
+TEST_DB_NAME="${TEST_DB_NAME:-metadb}"
+
+SQL_CMD=$(
+    cat <<EOF
+DELETE FROM Artifact;
```
tarilabs, line 10:
```suggestion
START TRANSACTION;
DELETE FROM Artifact;
```

see other comment

clients/python/tests/conftest.py
```diff
@@ -29,13 +29,14 @@ def pytest_collection_modifyitems(config, items):
             continue
 
 
-REGISTRY_HOST = "http://localhost"
-REGISTRY_PORT = 8080
-REGISTRY_URL = f"{REGISTRY_HOST}:{REGISTRY_PORT}"
-COMPOSE_FILE = "docker-compose.yaml"
-MAX_POLL_TIME = 1200  # the first build is extremely slow if using docker-compose-*local*.yaml for bootstrap of builder image
+REGISTRY_URL = os.environ.get("MR_URL", "http://localhost:8080")
```
tarilabs, line 32: on a totally unrelated note, I would use _analogous_ "injection approach" also for the other capability about https://github.com/kubeflow/model-registry/pull/421

tarilabs (commented):
as mentioned in https://github.com/kubeflow/model-registry/pull/447#issuecomment-2391034092 I really like this, thank you so much

Overall

/lgtm

with some comments below:

scripts/cleanup.sh
isinyaaa: sqlite takes in a different command.

tarilabs: ah yes, like `BEGIN TRANSACTION`?

thanks for the reminder, I would note this as a comment :)

clients/python/tests/conftest.py
isinyaaa: I would prefer to use the serialization capabilities already available on Pydantic with the addition of Python-dotenv, would you prefer to have this change applied to this PR or later?

tarilabs: definitely for later.

I'm not opinionated on a library for bringing in the config values a-la Vyper or SmallRye Config, but we will need to support env variables, .env, and ConfigMaps, etc a-la 12Factor apps imho
@Al-Pragliola thoughts?

Al-Pragliola: Yes, we definitely need to support them as this approach makes everything more flexible 👍🏼

Al-Pragliola:
what about using `lsof -i :8080 -t ` to get the PID of a port-forward to 8080, if found we skip creating a new one? @isinyaaa @tarilabs

Al-Pragliola (approved):

scripts/deploy_on_kind.sh
```diff
@@ -0,0 +1,38 @@
...
+
+echo 'Deploying model registry to Kind cluster'
+if [[ $(kubectl get namespaces || false) =~ $MR_NAMESPACE ]]; then
+    echo 'Namespace already exists, skipping creation'
+else
+    kubectl create namespace "$MR_NAMESPACE"
+fi
+
+kubectl apply -k manifests/kustomize/overlays/db
+kubectl set image -n kubeflow deployment/model-registry-deployment rest-container="$IMG"
```
tarilabs, line 36: to double-check.

during release, a PR with the updated manifest kustomization image setting is applied.
So first, the manifests are deployed to kind with an image tag which will not be found (the tag does not exists yet in the container registry).

This line (36) actually set the deployment model registry image to the one locally built, hence existing and loaded in the kind cluster.

Can this be confirmed, please?

isinyaaa: the image is being loaded on local mode inside the script. and on the CI it's loaded with a different command, but by the same method. So two cases here:
1. LOCAL envvar set:
  create the cluster and load the pre-built image inside, then run line 36 eventually.
2. No envvar set:
  assume the cluster already exists and has the image loaded. then we only set it after deploying MR (this avoids using kustomize directly which would change the manifests)
  
does this answer? otherwise @Al-Pragliola might be able to answer better as he suggested this change :)

tarilabs: thanks @isinyaaa I see, in case of GHA we do not set the LOCAL and we invoke directly the pytest via nox (instead of make): 

https://github.com/kubeflow/model-registry/pull/447/files#diff-207f09dfa0bc604a684631a1c78be7d9aa00f9a6c778eec30afa287990be3da1R119-R143

that should do it. 
  
I will wait for confirmation from @Al-Pragliola and likely will add some comments into it.

(no need to reply over weekend 😅 but I wanted to mark it down before forgetting 🙃 thanks a lot)

Al-Pragliola: I confirm what @isinyaaa just said, the job of the deploy_on_kind script is to first fill all the requirements for deploying mr and then deploy it on the current's context cluster.

In case of CI all requirements:

- existing Kubernetes cluster
- mr docker image built
- mr docker image present on the cluster's registry

are already done in the previous steps of the action, so we can just create the namespace and deploy by not setting the LOCAL env.

We decided to replace kustomize edit with kubectl edit for two reasons:

1) remove a dependency
2) avoid creating drifts, since kustomize actually modifies the manifests, which is a bit of a pain when running the script locally because the user has to undo local changes. 

Hope this works well for everyone.

scripts/deploy_on_kind.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -e
+
+MR_NAMESPACE="${MR_NAMESPACE:-kubeflow}"
+
+# modularity to allow re-use this script against a remote k8s cluster
+if [[ -n "$LOCAL" ]]; then
+    CLUSTER_NAME="${CLUSTER_NAME:-kind}"
+    IMG="${IMG:-docker.io/kubeflow/model-registry:main}"
```
Al-Pragliola, line 10:
```suggestion
    IMG="${IMG:-localhost/kubeflow/model-registry:latest}"
```

I had to make this change locally to make e2e tests work, screenshot of the error:

image


after the change:

image

Al-Pragliola (changes requested):

Al-Pragliola (approved):
LGTM

tarilabs:
retested 

/lgtm

image

PR #450, opened 2024-10-04
Cleans up some unused function params in BFF

alexcreasy:
## Description

A small code clean up to better model some go idioms. 

## How Has This Been Tested?

`make test`
## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/api/model_registry_handler.go
+++ b/clients/ui/bff/internal/api/model_registry_handler.go
@@ -8,7 +8,7 @@ import (
 
 type ModelRegistryListEnvelope Envelope[[]data.ModelRegistryModel, None]
 
-func (app *App) ModelRegistryHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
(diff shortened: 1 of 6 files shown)


alexcreasy:
@Griffin-Sullivan @ederign I've raised this to better use the go idiom of using an underscore to intentionally signal to the compiler that you're not using a value.

In the case of our tests, I think this is actually less hacky than the linter override - it clearly signals "yes the function can return an error that will cause a panic if not handled but panicking is the correct behaviour in this instance".  

Also in the case of unused function params you're signalling "yes, I conform to this interface, but I don't need to actually use a given params value"

It looks a bit weird coming from other languages but it's quite common practice in go. Let me know what you think?

Griffin-Sullivan (approved):
That's what I originally did with the linter haha guess I should have asked before trying to ignore it. We also have the option to turn off that linter for `*_test.go` if you'd prefer.

ederign (approved):

PR #482, opened 2024-10-16
feat(bff): Use ctrl.NewManager from the controller-runtime as Kubernetes client

ederign:
The main goal of this PR is to move from our naive implementation of caching to using ctrl.NewManager from the controller runtime as a Kubernetes client.

## Description.

The main advantage of [ctrl.NewManager ](https://pkg.go.dev/sigs.k8s.io/controller-runtime#section-readme)offers us: 
- Automatic caching and efficient resource management;
- Unified client for cached reads and direct writes, simplifying interaction with Kubernetes resources.
- Graceful lifecycle management with built-in shutdown handling and cleanup;
- Customizable caching logic, allowing us to specify which resources to cache and sync. 

I've also done some cleanups in this PR:
- Start our HTTP server in a goroutine to enable us to gracefully shut down the manager during the shutdown process.
- Split data package into models (for our data structures) and repositories;

### Scenario 1: 
make run MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true
Everything keeps working as expected.

### Scenario 2: 
make run MOCK_K8S_CLIENT=false MOCK_MR_CLIENT=true
Apply one or two services to your cluster:
```yaml
kind: Service
apiVersion: v1
metadata:
  labels:
    app: metadata
  name: model-registry
spec:
  selector:
    component: model-registry-server
  type: ClusterIP
  ports:
  - port: 8083
    protocol: TCP
    appProtocol: http
    name: http-api
  - port: 9093
    protocol: TCP
    appProtocol: grpc
    name: grpc-api
```
Screenshot 2024-10-15 at 9 52 47 PM

Screenshot 2024-10-15 at 9 52 47 PM

Apply a new service without shutdown BFF, and the new service should appear.

Screenshot 2024-10-15 at 9 54 19 PM

## Merge criteria:

- [X] All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)


clients/ui/bff/internal/helpers/k8s.go
```diff
@@ -0,0 +1,28 @@
...
+
+// GetKubeconfig returns the current KUBECONFIG configuration based on the default loading rules.
+func GetKubeconfig() (*clientRest.Config, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	return kubeConfig.ClientConfig()
+}
+
+// BuildScheme returns builds a new runtime scheme with all the necessary types registered.
```
Griffin-Sullivan, line 20:
```suggestion
// BuildScheme builds a new runtime scheme with all the necessary types registered.
```

clients/ui/bff/internal/integrations/k8s.go
```diff
@@ -26,62 +32,157 @@ type ServiceDetails struct {
...
+}
+
+func (kc *KubernetesClient) GetServiceNames() ([]string, error) {
+	//TODO (ederign) when we develop the front-end, implement subject access review here
+	// and check if the username has actually permissions to access that server
+	// currently on kf dashboard, the user name comes in kubeflow-userid
+
+	//TODO (ederign) we should consider and rethinking listing all services on cluster
+	// what if we have thousand of those?
+	// we should consider label filtering for instance
```
Griffin-Sullivan, line 139: So does the code below return every service on the cluster and then our loop pulls out anything with component == "model-registry-server"?

Griffin-Sullivan (commented):
Code makes sense to me. Going to test locally now.

clients/ui/bff/internal/integrations/k8s.go
ederign: For now temporarly yes. I just started a slack thread with everyone to discuss the implications of this.

Griffin-Sullivan (approved):
Ok looks good to me! I have successfully added some MRs to my cluster and tested the mock mode works as expected.

image

alexcreasy (commented):
I've been through the code, especially the go routines and I can't see anything wrong with them, I like the way it works to try to bring everything down gracefully.

One suggested change that maybe we should trap SIGHUP as well as the other signals, it could make it cleaner if someone's developing from a remote terminal (hintL I do this quite often :P)

PR #508, opened 2024-10-22
feat(csi): support multiple model registries

Al-Pragliola:
## Description

As it stands, CSI is designed around the idea that there is only one model registry on the cluster and each model's metadata is registered here. In this PR I have modified the code to allow users to specify a different MR url in the InferenceService to have support for multiple model registries, again this change does not break the current behavior because the url from the env var `MODEL_REGISTRY_BASE_URL` is used as a fallback.

Here's an example of an InferenceService using the new option:

```yaml
    apiVersion: "serving.kserve.io/v1beta1"
    kind: "InferenceService"
    metadata:
      name: "iris-model"
    spec:
      predictor:
        model:
          modelFormat:
            name: sklearn
          storageUri: "model-registry://custom-mr-svc.test-namespace.svc.cluster.local:8080/iris/v1"
```

I have also added more scenarios to the e2e tests to check if every supported URI combination is working correctly:

- model-registry://{modelName}/{modelVersion}
- model-registry://{modelName}
- model-registry://{modelRegistryUrl}/{modelName}/{modelVersion}
- model-registry://{modelRegistryUrl}/{modelName}

## How Has This Been Tested?

```bash
cd csi
kind create cluster
docker build . -f Dockerfile -t docker.io/mr/mr-csi:0.1.0
kind load docker-image docker.io/mr/mr-csi:0.1.0
KSERVE_VERSION=0.14 MR_CSI_IMG=docker.io/mr/mr-csi:0.1.0 ./test/e2e_test.sh
```

## Merge criteria:

### DIFF
--- a/csi/pkg/storage/modelregistry_provider.go
+++ b/csi/pkg/storage/modelregistry_provider.go
@@ -52,6 +52,13 @@ func (p *ModelRegistryProvider) DownloadModel(modelDir string, modelName string,
 		return err
 	}
 
+	log.Printf("Parsed storageUri=%s as: modelRegistryUrl=%s, registeredModelName=%s, versionName=%v",
+		storageUri,
+		p.Client.GetConfig().Host,
+		registeredModelName,
+		versionName,
+	)
+
 	log.Printf("Fetching model: registeredModelName=%s, versionName=%v", registeredModelName, versionName)
 
 	// Fetch the registered model
--- a/csi/pkg/storage/modelregistry_provider.go
+++ b/csi/pkg/storage/modelregistry_provider.go
@@ -2,50 +2,57 @@ package storage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"regexp"
 	"strings"
 
 	kserve "github.com/kserve/kserve/pkg/agent/storage"
+	"github.com/kubeflow/model-registry/csi/pkg/constants"
 	"github.com/kubeflow/model-registry/pkg/openapi"
 )
 
-const MR kserve.Protocol = "model-registry://"
+var (
+	_                         kserve.Provider = (*ModelRegistryProvider)(nil)
+	ErrInvalidMRURI                           = errors.New("invalid model registry URI, use like model-registry://{dnsName}/{registeredModelName}/{versionName}")
+	ErrNoVersionAssociated                    = errors.New("no versions associated to registered model")
+	ErrNoArtifactAssociated                   = errors.New("no artifacts associated to model version")
+	ErrNoModelArtifact                        = errors.New("no model artifact found for model version")
+	ErrModelArtifactEmptyURI                  = errors.New("model artifact has empty URI")
+	ErrNoStorageURI                           = errors.New("there is no storageUri supplied")
+	ErrNoProtocolInSTorageURI                 = errors.New("there is no protocol specified for the storageUri")
+	ErrProtocolNotSupported                   = errors.New("protocol not supported for storageUri")
+	ErrFetchingModelVersion                   = errors.New("error fetching model version")
+	ErrFetchingModelVersions                  = errors.New("error fetching model versions")
+)
 
 type ModelRegistryProvider struct {
 	Client    *openapi.APIClient
 	Providers map[kserve.Protocol]kserve.Provider
 }
 
-func NewModelRegistryProvider(cfg *openapi.Configuration) (*ModelRegistryProvider, error) {
-	client := openapi.NewAPIClient(cfg)
-
+func NewModelRegistryProvider(client *openapi.APIClient) (*ModelRegistryProvider, error) {
 	return &ModelRegistryProvider{
 		Client:    client,
 		Providers: map[kserve.Protocol]kserve.Provider{},
 	}, nil
 }
 
-var _ kserve.Provider = (*ModelRegistryProvider)(nil)
-
-// storageUri formatted like model-registry://{registeredModelName}/{versionName}
+// storageUri formatted like model-registry://{modelRegistryUrl}/{registeredModelName}/{versionName}
 func (p *ModelRegistryProvider) DownloadModel(modelDir string, modelName string, storageUri string) error {
-	log.Printf("Download model indexed in model registry: modelName=%s, storageUri=%s, modelDir=%s", modelName, storageUri, modelDir)
+	log.Printf("Download model indexed in model registry: modelName=%s, storageUri=%s, modelDir=%s",
+		modelName,
+		storageUri,
+		modelDir,
+	)
 
-	// Parse the URI to retrieve the needed information to query model registry (modelArtifact)
-	mrUri := strings.TrimPrefix(storageUri, string(MR))
-	tokens := strings.SplitN(mrUri, "/", 2)
-
-	if len(tokens) == 0 || len(tokens) > 2 {
-		return fmt.Errorf("invalid model registry URI, use like model-registry://{registeredModelName}/{versionName}")
+	registeredModelName, versionName, err := p.parseModelVersion(storageUri)
+	if err != nil {
+		return err
 	}
 
-	registeredModelName := tokens[0]
-	var versionName *string
-	if len(tokens) == 2 {
-		versionName = &tokens[1]
-	}
+	log.Printf("Fetching model: registeredModelName=%s, versionName=%v", registeredModelName, versionName)
 
 	// Fetch the registered model
 	model, _, err := p.Client.ModelRegistryServiceAPI.FindRegisteredModel(context.Background()).Name(registeredModelName).Execute()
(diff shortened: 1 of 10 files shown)


Al-Pragliola (commented):
cc @tarilabs @lampajr

lampajr (changes requested):
>In this PR I have modified the code to allow users to specify a different MR url in the InferenceService to have support for multiple model registries, again this change does not break the current behavior because the url from the env var MODEL_REGISTRY_BASE_URL is used as a fallback.

I really like this approach and the fact that it is backward compatible by falling back to the default modelregistry.

I left just one comment that I think should be sorted out.

csi/pkg/storage/modelregistry_provider.go
```diff
@@ -110,19 +105,84 @@ func (p *ModelRegistryProvider) DownloadModel(modelDir string, modelName string,
 	return provider.DownloadModel(modelDir, "", *modelArtifact.Uri)
 }
 
-func extractProtocol(storageURI string) (kserve.Protocol, error) {
+// Possible URIs:
+// (1) model-registry://{modelName}
+// (2) model-registry://{modelName}/{modelVersion}
+// (3) model-registry://{modelRegistryUrl}/{modelName}
```
lampajr, line 111: I am not completely sure this is actually supported, because if you have just 2 tokens (after the trim) it will return the default apiClient (i.e., using the default env model registry).

I think here the problem would be, how do you know whether the user is providing option 2 or 3?

I tried running `make SOURCE_URI=model-registry://model-registry-url/model DEST_PATH=./ run` and in fact `model-registry-url` is interpreted as being the registered model, which is actually not what the user was trying to do.

I think that we have two options here:
1. If the `model-registry-url` is provided, users cannot omit the `version`
2. We use a different delimiter for the `model-registry-url`, e.g., `model-registry://{modelRegistryUrl}:{modelName}`

lampajr (changes requested):
>In this PR I have modified the code to allow users to specify a different MR url in the InferenceService to have support for multiple model registries, again this change does not break the current behavior because the url from the env var MODEL_REGISTRY_BASE_URL is used as a fallback.

I really like this approach and the fact that it is backward compatible by falling back to the default modelregistry.

I left just one comment that I think should be sorted out.

csi/pkg/storage/modelregistry_provider.go
Al-Pragliola: The way I approached this problem can be found here:

https://github.com/Al-Pragliola/model-registry/blob/feat/multi-mr-registries-csi-support/csi/pkg/modelregistry/api_client.go#L24

We try to reach (token 2) as a model registry and on failure we assume that it is a model name and not a valid mr url, it's not perfect but I think it might be a fair compromise

lampajr: Sorry missed that, I tried and I can confirm that if the "host" is reachable it works as expected:
```bash
$ make SOURCE_URI=model-registry://localhost:8080/mymodel DEST_PATH=./ run                                                                                             [17:59:56]
"/usr/bin/go" fmt ./...
"/usr/bin/go" vet ./...
"/usr/bin/go" run ./main.go model-registry://localhost:8080/mymodel ./
2024/10/24 18:00:02 Initializing, args: src_uri [model-registry://localhost:8080/mymodel] dest_path[ [./]
2024/10/24 18:00:02 Download model indexed in model registry: modelName=, storageUri=model-registry://localhost:8080/mymodel, modelDir=./
2024/10/24 18:00:02 Fetching model: registeredModelName=mymodel, versionName=<nil>
2024/10/24 18:00:02 404 Not Found
exit status 1
make: *** [Makefile:59: run] Error 1
```

With a model registry running at `localhost:8080`.

My main concern with this assumption is that, if there is a model registry running but for any reason it is not accessible/reachable we are going to make the wrong assumption that it is a registeredModel name , right? And the error would be misleading to the user as it will find in the logs.

What do you think?

Al-Pragliola: Yep that's true, the only hint that user have from the logs is this:

```golang
log.Printf("Falling back to base url %s for model registry service", cfg.Host)
```
We can improve this message telling the user that it failed to reach <url from uri> and it's going to use <fallback url>

---

I wanted this to be as retrocompatible as possible, otherwise there are many alternatives, like a query parameter approach:

`model-registry://url?modelName=x&modelVersion=y`
`model-registry://?modelName=x&modelVersion=y`

or the other delimiter you mentioned.

csi/test/kind_config.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
+nodes:
+  - role: control-plane
+  - role: worker
+  - role: worker
```
tarilabs, line 6: I've missed why this Kind context is needed/ introduced?

csi/pkg/storage/modelregistry_provider.go
```diff
@@ -110,19 +105,84 @@ func (p *ModelRegistryProvider) DownloadModel(modelDir string, modelName string,
...
+
+	if len(tokens) == 0 || len(tokens) > 3 {
+		return "", nil, ErrInvalidMRURI
+	}
+
+	// Check if the first token is the host and remove it so that we reduce cases (3) and (4) to (1) and (2)
+	if len(tokens) >= 2 && p.Client.GetConfig().Host == tokens[0] {
+		tokens = tokens[1:]
+	}
+
```
tarilabs, line 129: > My main concern with this assumption is that, if there is a model registry running but for any reason it is not accessible/reachable we are going to make the wrong assumption that it is a registeredModel name , right? And the error would be misleading to the user as it will find in the logs.

> We can improve this message telling the user that it failed to reach url from uri and it's going to use fallback url

Here (line 129) you have done all the storageURI parsing to determine in which case you are.
To me, add a Log info here that shows  ~circa

```
Parsed storageUri=... as: modelRegistryUrl=... modelName=... modelVersion=...
```

This way, we're being explicit on how we interpret what we have received based on the documentation provided. wdyt?

csi/pkg/storage/modelregistry_provider.go
tarilabs: Added comment below for line 129 my2c

tarilabs (commented):
/lgtm

with couple of comments; lmk your view?

csi/pkg/storage/modelregistry_provider.go
lampajr: I agree with you @tarilabs @Al-Pragliola , I think that providing a more meaningful and expecit log message would be enough for this use case!

csi/pkg/storage/modelregistry_provider.go
tarilabs: btw forgot to mention; options 3-4 also consistent with standard URLs and also KServe examples

```
    - regex: "https://(.+?).blob.core.windows.net/(.+)"
    - regex: "https://(.+?).file.core.windows.net/(.+)"
```

much appreciated @Al-Pragliola

csi/test/kind_config.yaml
Al-Pragliola: Sorry I forgot to mention it in the PR description, when adding more scenarios in the e2e tests there are now 4 inference services with the associated deployments/pods etcetera and the scheduler in the CI complained with errors like `Warning FailedScheduling 1s default-scheduler 0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes available: 1 No preemption victims found for incoming pod..`, so I had to add workers to the kind test cluster. Now that I think about it, I can also clean up inferenceservices after each scenario, wdyt @tarilabs ?

tarilabs: > Now that I think about it, I can also clean up inferenceservices after each scenario, wdyt @tarilabs ?

I believe this option is preferable in general, so to have a cleanup after each run, but it's nice to know in case one day if we want a more "extended" testings! my2c

Al-Pragliola: done

csi/pkg/storage/modelregistry_provider.go
Al-Pragliola: added
```golang
	log.Printf("Parsed storageUri=%s as: modelRegistryUrl=%s, registeredModelName=%s, versionName=%v",
		storageUri,
		p.Client.GetConfig().Host,
		registeredModelName,
		versionName,
	)
```

lampajr (approved):
Thanks a lot @Al-Pragliola 🚀 

/lgtm

csi/test/kind_config.yaml
```diff
@@ -2,5 +2,3 @@ apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
 nodes:
   - role: control-plane
-  - role: worker
-  - role: worker
```
tarilabs: Is this file left if one day we want to add feature gates, etc?

Because to my understanding, the way this file is right now, is a default setting by looking up in the doc. If confirmed, I'd suggest making a comment in this file (so we don't have one day to wonder "why is this config file necessary") or just leave the 2 workers even if the test scenarios can run efficiently without requiring 2 actual nodes. wdyt?

tarilabs (commented):
small comment then I think it's good to go; thanks a lot for keeping tabs on this

tarilabs (approved):
thanks a lot @Al-Pragliola 
and thanks to @lampajr for chiming in

/lgtm
/approve

csi/test/kind_config.yaml
Al-Pragliola: Third option, I forgot to remove it after successfully testing the e2e tests only with the control plane 😞 🙏🏼, I just removed the file in the last commit, I think it just brings confusion and adds nothing, if in the future we might need to add workers we'll recreate it at that time.

PR #524, opened 2024-10-28
Add NotificationContext and useNotification hook

Griffin-Sullivan:
## Description
This PR adds a new global context for Notifications. I also added a `useNotification` hook to allow for easier dispatching by calling the specific method you want to use `notification.success` or `notification.error`. I also looped in the removal of notifications into this hook. You can call it like: 
```javascript
notifications = useNotification() // returns state for notifications
...
notifications.remove(notification.id)
```

image

* I still not have tested restoring an archived version yet since I can't do it in mocked mode

There are a couple known issues at the moment:
1. The id count gets reset to 0 whenever you change pages meaning if you switch pages quickly and get another notification they will have the same id
2. The notification exit button looks out of place (I think it should be center aligned not in the top right corner)

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/pages/components/Notification.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/pages/components/Notification.ts
@@ -1,0 +1,7 @@
+export class ToastNotification {
+  constructor(private title: string) {}
+
+  find(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.findByText(this.title);
+  }
+}
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelVersionArchive.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelVersionArchive.cy.ts
@@ -4,13 +4,18 @@ import { mockModelVersionList } from '~/__mocks__/mockModelVersionList';
 import { mockModelVersion } from '~/__mocks__/mockModelVersion';
 import { mockRegisteredModel } from '~/__mocks__/mockRegisteredModel';
 import { verifyRelativeURL } from '~/__tests__/cypress/cypress/utils/url';
-import { labelModal } from '~/__tests__/cypress/cypress/pages/modelRegistry';
+import { labelModal, modelRegistry } from '~/__tests__/cypress/cypress/pages/modelRegistry';
 import type { ModelRegistry, ModelVersion } from '~/app/types';
 import { ModelState } from '~/app/types';
 import { mockModelRegistry } from '~/__mocks__/mockModelRegistry';
 import { mockBFFResponse } from '~/__mocks__/utils';
-import { modelVersionArchive } from '~/__tests__/cypress/cypress/pages/modelRegistryView/modelVersionArchive';
+import {
+  archiveVersionModal,
+  modelVersionArchive,
+  restoreVersionModal,
+} from '~/__tests__/cypress/cypress/pages/modelRegistryView/modelVersionArchive';
 import { MODEL_REGISTRY_API_VERSION } from '~/__tests__/cypress/cypress/support/commands/api';
+import { ToastNotification } from '~/__tests__/cypress/cypress/pages/components/Notification';
 
 type HandlersProps = {
   registeredModelsSize?: number;
@@ -129,12 +134,10 @@ describe('Model version archive list', () => {
     initIntercepts({ modelVersions: [mockModelVersion({ id: '3', name: 'model version 2' })] });
     modelVersionArchive.visitModelVersionList();
     verifyRelativeURL('/modelRegistry/modelregistry-sample/registeredModels/1/versions');
-    // TODO: Uncomment when dropdowns are fixed and remove the visit after the comments
-    // modelVersionArchive
-    //   .findModelVersionsTableKebab()
-    //   .findDropdownItem('View archived versions')
-    //   .click();
-    modelVersionArchive.visit();
+    modelVersionArchive
+      .findModelVersionsTableKebab()
+      .findDropdownItem('View archived versions')
+      .click();
     modelVersionArchive.shouldArchiveVersionsEmpty();
   });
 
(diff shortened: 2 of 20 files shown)


clients/ui/frontend/src/app/hooks/useNotification.ts
```diff
@@ -0,0 +1,110 @@
...
+
+type NotificationRemoveProps = (id: number | undefined) => void;
+
+type NotificationTypeFunc = {
+  [key in NotificationTypes]: NotificationProps
+}
+
+interface NotificationFunc extends NotificationTypeFunc {
+  remove: NotificationRemoveProps
+}
```
Griffin-Sullivan, line 23: This is a little hacky. Maybe we should just have a separate hook for when we want to return all notifications and do something with them like removing one.

clients/ui/frontend/src/components/ToastNotification.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <Alert
+      variant={asEnumMember(notification.status, AlertVariant) ?? undefined}
+      title={notification.title}
+      actionClose={
+        <AlertActionCloseButton onClose={() => notifications.remove(notification.id)} />
+      }
+      onMouseEnter={() => setMouseOver(true)}
+      onMouseLeave={() => setMouseOver(false)}
+    >
+      {notification.message}
```
Griffin-Sullivan, line 47: See the description: this is not being passed along properly.

clients/ui/frontend/src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { useContext } from 'react';
+import { AlertGroup } from '@patternfly/react-core';
+import ToastNotification from './ToastNotification';
+import { NotificationContext } from '~/app/context/NotificationContext';
+
+const ToastNotifications: React.FC = () => {
+  const [notifications] = useContext(NotificationContext);
```
Griffin-Sullivan, line 7: The current hook only gives access to notification actions. Should we expand it to also return the notifications list? This is why I'm using `useContext` here.

clients/ui/frontend/src/utilities/utils.ts
```diff
@@ -0,0 +1,17 @@
+export const asEnumMember = <T extends object>(
```
lucferbux, line 1: I think this already existed in the codebase. We can either keep one or if you feel this should be in the main utils I'm up for it.

clients/ui/frontend/src/app/context/NotificationContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  title: string;
+  message?: React.ReactNode;
+  hidden?: boolean;
+  read?: boolean;
+  timestamp: Date;
+};
+
+type NotificationAction =
+  | {
+      type: 'add_notification';
```
lucferbux, line 16: I would create a new enum to detail the types, it's a nice way to keep everything altogehter a part from the type itself.

clients/ui/frontend/src/app/context/NotificationContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  action,
+) => {
+  switch (action.type) {
+    case 'add_notification': {
+      return [
+        ...notifications,
+        {
+          status: action.payload.status,
+          title: action.payload.title,
+          timestamp: action.payload.timestamp,
```
lucferbux, line 46: You are not passing the message here, that's why the message is not displayed.

clients/ui/frontend/src/components/ToastNotification.tsx
lucferbux: Left a comment with the fix

lucferbux (changes requested):
Haven't tested it yet, but did a quick review, we can take a deeper look later today if you want.

Griffin-Sullivan:
Ok @lucferbux addressed your comments thanks!

Griffin-Sullivan:
Ok with the newest push I have added a bunch of testing which includes some for checking the notification pops up correctly.

clients/ui/frontend/src/__tests__/cypress/cypress/pages/components/Notification.ts
```diff
@@ -0,0 +1,7 @@
+export class Notification {
```
Griffin-Sullivan, line 1: Maybe we should rename `ToastNotification`?

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelVersionArchive.cy.ts
```diff
@@ -185,119 +188,124 @@ describe('Model version archive list', () => {
...
+    cy.interceptApi(
+      'PATCH /api/:apiVersion/model_registry/:modelRegistryName/model_versions/:modelVersionId',
+      {
+        path: {
+          modelRegistryName: 'modelregistry-sample',
+          apiVersion: MODEL_REGISTRY_API_VERSION,
+          modelVersionId: 2,
+        },
+      },
+      mockBFFResponse(mockModelVersion({})),
```
Griffin-Sullivan, line 202: Notice how we have to put a bunch of `mockBFFResponse` everywhere. We should figure out a way for all of our `mockObject` calls to inherently know about the BFF response so we don't have to be so explicit.

clients/ui/frontend/src/__tests__/cypress/cypress/pages/components/Notification.ts
lucferbux: Yes, we are not using Notification, we might even refactor this in the future if this is not part of the project but reusable code, we'll see

lucferbux: At the moment, just rename it and we can call it a day

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelVersionArchive.cy.ts
lucferbux: This is a great point, my first guess is to just modifify the mock objects and add a wrapper there, can you open a ticket in enhancemets to cover this?

clients/ui/frontend/src/app/context/NotificationContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type NotificationAction =
+  | {
+      type: NotificationActionTypes.ADD_NOTIFICATION;
+      payload: Notification;
+    }
+  | {
+      type: NotificationActionTypes.DELETE_NOTIFICATION;
+      payload: { id: Notification['id'] };
+    };
```
lucferbux, line 27: Can you move this to the `types.ts` file under `app`?

clients/ui/frontend/src/app/hooks/useNotification.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+interface NotificationFunc extends NotificationTypeFunc {
+  remove: NotificationRemoveProps;
+}
+
+export const useNotification = (): NotificationFunc => {
+  const { dispatch } = useContext(NotificationContext);
+  // need to move this count somewhere else since it will reset on every new useNotification instance (like switching pages)
+  let notificationCount = 0;
```
lucferbux, line 27: Why don't you move the notification count to the context itself?

lucferbux (approved):
Everything looks great, just a couple of nits, really great work here!

clients/ui/frontend/src/app/hooks/useNotification.ts
Griffin-Sullivan: Why don't I......

ederign (approved):

PR #540, opened 2024-11-06
Fixing slog.Logger.Error arg err should be a string or a slog.Attr (possible missing key or value)

ederign:
## Description
On go 23, this error occurs:

internal/mocks/k8s_mock.go:53:52: slog.Logger.Error arg "err" should be a string or a slog.Attr (possible missing key or value)
internal/mocks/k8s_mock.go:60:54: slog.Logger.Error arg "err" should be a string or a slog.Attr (possible missing key or value)
internal/mocks/k8s_mock.go:67:40: slog.Logger.Error arg "err" should be a string or a slog.Attr (possible missing key or value)

## How Has This Been Tested?
Ran the BFF and ran the FE.

## Merge criteria:

### DIFF
--- a/clients/ui/bff/internal/mocks/k8s_mock.go
+++ b/clients/ui/bff/internal/mocks/k8s_mock.go
@@ -3,14 +3,15 @@ package mocks
 import (
 	"context"
 	"fmt"
-	k8s "github.com/kubeflow/model-registry/ui/bff/internal/integrations"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes/scheme"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"runtime"
+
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
Should we update the go.mod version or add documentation about what version of go to develop with?

ederign:
@Griffin-Sullivan, that is indeed one of the option or we would stay with our:

https://github.com/kubeflow/model-registry/tree/main/clients/ui/bff#dependencies

Dependencies
Go >= 1.22.2


I would prefer to keep using 1.22 so as not to be 'too far' ahead from https://github.com/kubeflow/model-registry/blob/main/go.mod#L3 and keep this PR with the purpose of 'fixes built in 1.23.x'

alexcreasy:
Yeah I say we stick close to the go version used in the rest of the project. We can decide on docs separately, I've tested and it works with 1.23 so lgtm

alexcreasy (approved):

PR #558, opened 2024-11-11
feat(bff): load x-forwarded-access-token in cluster setup

ederign:
## Description
In this PR:
- middleware.go: load the bearerToken from x-forwarded-access-token and did a small refactoring to avoid leaking k8s client api;
- k8s.go: move the rest.inclusterConfig logic here for better encapsulation
- apiUtils.go: add the token parameter to the header also in calls where content type is set,
 
## How Has This Been Tested?
1. Built UI and BFF images locally and pushed to kind.
2. Used new 'make kind-deployment' to create and deploy everything to kind.

Screenshot 2024-11-11 at 11 32 44 AM

## Merge criteria:

### DIFF
--- a/clients/ui/frontend/src/shared/api/apiUtils.ts
+++ b/clients/ui/frontend/src/shared/api/apiUtils.ts
@@ -76,7 +76,10 @@ const callRestJSON = <T>(
 
   return fetch(`${host}${path}${searchParams ? `?${searchParams}` : ''}`, {
     ...otherOptions,
-    ...(contentType && { headers: { 'Content-Type': contentType } }),
+    headers: {
+      ...otherOptions.headers,
+      ...(contentType && { 'Content-Type': contentType }),
+    },
     method,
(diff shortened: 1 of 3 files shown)


lucferbux:
Ok, tested and it works flawlesly! Great addition, we now have a working version of the deployment.

clients/ui/frontend/src/shared/api/apiUtils.ts
```diff
@@ -76,7 +76,10 @@ const callRestJSON = <T>(
 
   return fetch(`${host}${path}${searchParams ? `?${searchParams}` : ''}`, {
```
alexcreasy, line 77: I think it's generally better to use the URL api to construct this rather than a string template. You get URL encoding / escaping for free that way.

That said this is a nit so I won't hold the PR up for it :)

alexcreasy (approved):
Great PR, thanks Eder!

PR #560, opened 2024-11-12
Add BFF dev mode for local K8s testing

Griffin-Sullivan:
## Description
This PR allows developers to run a local MR cluster in kind and port forward their service to work with the BFF running locally.

## How Has This Been Tested?
Follow the [dev guide](https://github.com/kubeflow/model-registry/blob/main/clients/ui/bff/docs/dev-guide.md) all the way to step 4.

Port forward with:
```
kubectl port-forward svc/model-registry-service 8081:8080 -n kubeflow
```

Run the BFF with:
```
make run DEV_MODE=true DEV_MODE_PORT=8081
```
and the UI in a separate terminal.

You should see an empty MR that you can interact with (make models, versions, archive, restore, etc)

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/api/middleware.go
+++ b/clients/ui/bff/internal/api/middleware.go
@@ -3,9 +3,11 @@ package api
 import (
 	"context"
 	"fmt"
-	"github.com/julienschmidt/httprouter"
-	"github.com/kubeflow/model-registry/ui/bff/internal/integrations"
 	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/kubeflow/model-registry/ui/bff/internal/config"
+	"github.com/kubeflow/model-registry/ui/bff/internal/integrations"
 )
 
 type contextKey string
@@ -41,7 +43,7 @@ func (app *App) AttachRESTClient(handler func(http.ResponseWriter, *http.Request
 
 		modelRegistryID := ps.ByName(ModelRegistryId)
 
-		modelRegistryBaseURL, err := resolveModelRegistryURL(modelRegistryID, app.kubernetesClient)
+		modelRegistryBaseURL, err := resolveModelRegistryURL(modelRegistryID, app.kubernetesClient, app.config)
 		if err != nil {
 			app.serverErrorResponse(w, r, fmt.Errorf("failed to resolve model registry base URL): %v", err))
 			return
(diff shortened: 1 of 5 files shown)


clients/ui/bff/internal/api/middleware.go
```diff
@@ -83,11 +84,16 @@ func resolveBearerToken(k8s integrations.KubernetesClientInterface, header http.
...
-func resolveModelRegistryURL(id string, client integrations.KubernetesClientInterface) (string, error) {
-	serviceDetails, err := client.GetServiceDetailsByName(id)
+func resolveModelRegistryURL(id string, app *App) (string, error) {
+	serviceDetails, err := app.kubernetesClient.GetServiceDetailsByName(id)
 	if err != nil {
 		return "", err
 	}
+
+	if app.config.DevMode {
+		serviceDetails.ClusterIP = "localhost"
```
lucferbux, line 94: Just to bring the internal conversation here, it would be really interesting to parametrize the port too, that way we can get whatever port users might have used for port-forwarding.

lucferbux (changes requested):
Everything looks great, only nit would be parametrization of ports.

clients/ui/bff/internal/api/middleware.go
```diff
@@ -83,11 +84,17 @@ func resolveBearerToken(k8s integrations.KubernetesClientInterface, header http.
 	return bearerToken, nil
 }
 
-func resolveModelRegistryURL(id string, client integrations.KubernetesClientInterface) (string, error) {
-	serviceDetails, err := client.GetServiceDetailsByName(id)
+func resolveModelRegistryURL(id string, app *App) (string, error) {
```
alexcreasy, line 87: I'm not a fan of passing a pointer to the app struct into this function. It's meant as a utility function so I would stick to value semantics and add params for each value it requires.

An alternative would be to convert this function into a method of the app struct and use a receiver param to access the methods.

alexcreasy (changes requested):
Looks good, just one thing to take a look at!

clients/ui/bff/internal/api/middleware.go
ederign: Agree with Alex here.

ederign (changes requested):
just to a second review after my comment above ^

alexcreasy (approved):
Great work, thanks Griffin!

PR #563, opened 2024-11-15
py: client: add logging

isinyaaa:
Adds logging and additional keyword args to the Python client.
This also makes the custom_ca totally optional and not sourced by default by any means.

Logging settings can be further extended (e.g. with a prefix path for file logging) using other constructor flags.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -91,26 +91,28 @@ class ModelRegistry:
 
         if not user_token and user_token_envvar:
             logger.info("Reading user token from %s", user_token_envvar)
-            if user_token_envvar == DEFAULT_USER_TOKEN_ENVVAR:
-                logger.warning(
-                    "Using default user token envvar, if you want to use a different one, pass it as `user_token`"
-                )
             # /var/run/secrets/kubernetes.io/serviceaccount/token
-            sa_token = os.environ.get(user_token_envvar)
-            if sa_token:
+            if sa_token := os.environ.get(user_token_envvar):
+                if user_token_envvar == DEFAULT_USER_TOKEN_ENVVAR:
+                    logger.warning(
+                        f"Sourcing user token from default envvar: {DEFAULT_USER_TOKEN_ENVVAR}"
+                    )
                 user_token = Path(sa_token).read_text()
             else:
                 warn("User access token is missing", stacklevel=2)
 
         if is_secure:
-            if not custom_ca and custom_ca_envvar:
+            if (
+                not custom_ca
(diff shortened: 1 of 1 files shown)


clients/python/src/model_registry/_client.py
```diff
@@ -45,47 +73,57 @@ def __init__(
...
         # TODO: get remaining args from env
         self._author = author
 
-        if not user_token:
+        if not user_token and user_token_envvar:
+            logger.info("Reading user token from %s", user_token_envvar)
+            if user_token_envvar == DEFAULT_USER_TOKEN_ENVVAR:
+                logger.warning(
+                    "Using default user token envvar, if you want to use a different one, pass it as `user_token`"
+                )
```
tarilabs, line 97: I believe this needs to be moved below, as this warning ("Using ...") will be printed even in the cases that `DEFAULT_USER_TOKEN_ENVVAR` is actually not in the current env. wdyt?

We can _use it_ only if `os.environ.get(user_token_envvar)` is not None.

clients/python/src/model_registry/_client.py
```diff
@@ -45,47 +73,57 @@ def __init__(
...
             # /var/run/secrets/kubernetes.io/serviceaccount/token
-            sa_token = os.environ.get("KF_PIPELINES_SA_TOKEN_PATH")
+            sa_token = os.environ.get(user_token_envvar)
             if sa_token:
                 user_token = Path(sa_token).read_text()
```
tarilabs, line 101: I would move it below, ...here at line 101.

clients/python/src/model_registry/_client.py
```diff
@@ -45,47 +73,57 @@ def __init__(
...
         if is_secure:
-            root_ca = None
-            if not custom_ca:
-                if cert := os.getenv("CERT"):
-                    root_ca = cert
+            if not custom_ca and custom_ca_envvar:
+                logger.info(
+                    "Custom CA not provided, checking for envvar %s",
+                    custom_ca_envvar,
+                )
```
tarilabs, line 110: Analogous. I would move this below, because in the most cases as custom_ca is not provided (relying on the system defined one) and custom_ca_envvar empty None, this would print out.

I would move this into line 112 saying that we assume custom_ca_envvar contains the intended CA.

tarilabs (commented):
Thank you very much @isinyaaa !

/lgtm

I've added some comments for considerations; cc @lugi0

tarilabs (approved):
as discussed

/lgtm
/approve

PR #564, opened 2024-11-15
Add enum and env var to conditionally render MUI theme

jenny-s51:
## Description

Adds env var to support toggling MUI theme on / off. 
Screenshot 2024-11-15 at 10 51 56 AM
Screenshot 2024-11-15 at 10 52 52 AM

## How Has This Been Tested?

Verify that the MUI theme is toggled on / off based on the value of `STYLE_THEME`. 

Tests N/A, visual updates only. 

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/App.tsx
+++ b/clients/ui/frontend/src/app/App.tsx
@@ -21,6 +21,7 @@ import {
 import { BarsIcon } from '@patternfly/react-icons';
 import ToastNotifications from '~/shared/components/ToastNotifications';
 import { useSettings } from '~/shared/hooks/useSettings';
+import { isMUITheme, Theme } from '~/shared/utilities/const';
 import NavSidebar from './NavSidebar';
 import AppRoutes from './AppRoutes';
 import { AppContext } from './AppContext';
@@ -33,6 +34,15 @@ const App: React.FC = () => {
     loaded: configLoaded,
     loadError: configError,
   } = useSettings();
+
+  React.useEffect(() => {
+    // Apply the theme based on the value of STYLE_THEME
+    if (isMUITheme()) {
+      document.documentElement.classList.add(Theme.MUI);
+    } else {
+      document.documentElement.classList.remove(Theme.MUI);
+    }
+  }, []);
 
   const contextValue = React.useMemo(
     () =>
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelPropertiesTableRow.tsx
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelPropertiesTableRow.tsx
@@ -14,6 +14,7 @@ import { CheckIcon, TimesIcon } from '@patternfly/react-icons';
 import { KeyValuePair } from '~/shared/types';
 import { EitherNotBoth } from '~/shared/typeHelpers';
 import FormFieldset from '~/app/pages/modelRegistry/screens/components/FormFieldset';
+import { isMUITheme } from '~/shared/utilities/const';
 
 type ModelPropertiesTableRowProps = {
   allExistingKeys: string[];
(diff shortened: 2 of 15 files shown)


clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelPropertiesTableRow.tsx
```diff
@@ -49,6 +49,23 @@ const ModelPropertiesTableRow: React.FC<ModelPropertiesTableRowProps> = ({
...
+        setIsSwitched(event.detail); // Update state with the new value
+      }
+    };
+
+    window.addEventListener('isSwitchedChanged', handleSwitchChange);
+
+    return () => {
+      window.removeEventListener('isSwitchedChanged', handleSwitchChange);
+    };
+  }, []);
```
Griffin-Sullivan, line 67: Since this gets reused a lot I would guess a custom hook like `useMUI` would be a good idea

lucferbux (changes requested):
Ok changes look reaaly great, but overall, we would wanna avoid this global declarations, apologies for not detailing too much in the description the conditional rendering, but the main goal for this is just to turn it on or off at **build** time, for that:

* Instead of having a global window value, let's just control this with a constant
* That constant could be either a boolean, but I would suggest moving this into an ENUM and then have some utility to control (we only want MUI but it's a good practice to get extensibility)
* We might not want a switcher in the UI, upstream should stick with MUI, this is just for us to be able to contribute both upstream with MUI and downstream with regular PF6 look and feel
* The global const should be an env variable, I'm working right now on a PR to enhance our env variable handling but so far it's gonna be similar on [what we have in odh](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/utilities/const.ts#L20), so you can follow that pattern in the `const.ts` file
* Since we don't have proper env support (I'm working on it right now) you can test that in a similar way as we have [here](https://github.com/kubeflow/model-registry/blob/main/clients/ui/frontend/package.json#L30) just add the env variable you decided to use (maybe STYLE_THEME) to the Kubeflow value (if it's a boolean call it other way) and add that to the dev command, I can tweak it later in my PR

jenny-s51 (commented):
Thank you @Griffin-Sullivan @lucferbux for your feedback - updated the PR to reflect your comments.

clients/ui/frontend/src/app/App.tsx
```diff
@@ -21,11 +21,18 @@ import {
...
 import { useSettings } from '~/shared/hooks/useSettings';
+import { STYLE_THEME, Theme } from '~/shared/utilities/const';
 import NavSidebar from './NavSidebar';
 import AppRoutes from './AppRoutes';
 import { AppContext } from './AppContext';
 import { ModelRegistrySelectorContextProvider } from './context/ModelRegistrySelectorContext';
 
+declare global {
+  export interface Window {
+    isSwitched?: boolean;
```
lucferbux, line 32: Do we still need `isSwitched`?

lucferbux (changes requested):
Ok, just a couple of nits, everything else working as expected!

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelVersionListView.tsx
```diff
@@ -148,22 +150,35 @@ const ModelVersionListView: React.FC<ModelVersionListViewProps> = ({
...
-                          setSearch(searchValue);
-                        }}
-                        style={{ minWidth: '200px' }}
-                        data-testid="model-versions-table-search"
-                        aria-label="Search"
-                      />
-                    }
-                    field={`Find by ${searchType.toLowerCase()}`}
-                  />
+                  {STYLE_THEME === Theme.MUI ? (
```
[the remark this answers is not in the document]
jenny-s51: Great idea, thanks! Updated

lucferbux (approved):
/lgtm

ederign (approved):

PR #574, opened 2024-11-19
feat(inferenceservice): added integration tests to inferenceservice reconcile function

Al-Pragliola:
## Description

This pull request introduces integration tests for the new inferenceservice reconcile feature, and I chose to use `envtest` to test the functionality, as it is the de facto standard for testing controller-like behavior. Since envtest is not a real Kubernetes cluster, I had to mock responses from the model registry, as you can see in the suite_test.go file, and check that the inference-service-id label is added to the kserve inferenceservice CR. I also found some bugs in the reconciler thanks to the tests, and fixed them in this PR.

NOTE: The large chunk of code in this PR comes from the CRD that needs to be added to the envtest cluster to make it handle inference services.

## How Has This Been Tested?

`make test`

## Merge criteria:


pkg/inferenceservice-controller/testdata/crd/serving.kserve.io_inferenceservices.yaml
```diff
@@ -0,0 +1,16330 @@
+---
```
tarilabs, line 1: question: instead of having this file locally, can we use from kserve repo like:
```
https://github.com/kserve/kserve/blob/release-0.13/charts/kserve-crd/templates/serving.kserve.io_inferenceservices.yaml
```

by indicating a link of sort to envtest?

pkg/inferenceservice-controller/controller.go
```diff
@@ -242,7 +244,7 @@ func (r *InferenceServiceController) getMRUrlFromService(ctx context.Context, na
 		return "", fmt.Errorf("unable to find the http port in the Model Registry Service")
 	}
 
-	return fmt.Sprintf("%s.%s.svc.cluster.local:%d", name, namespace, *restApiPort), nil
+	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", name, namespace, *restApiPort), nil
```
tarilabs, line 247: wouldn't hardcoding `http://` be a problem midstream/downstream of Kubeflow where Distributions may default to mTLS? 🤔

tarilabs (commented):
thank you very much @Al-Pragliola some minor comments, otherwise 

/lgtm

already!

pkg/inferenceservice-controller/controller.go
Al-Pragliola: I would leave it like that for a default configuration and let the user decide using the `modelregistry.kubeflow.org/url` for every other scenario mLTS/external route/etc

tarilabs (approved):
thank you for following up on the comments @Al-Pragliola 

/lgtm
/approve

PR #594, opened 2024-11-25
Adds ability to change logging verbosity and adds more debug telemetry

alexcreasy:
## Description

Adds several logging related features:
* Ability to change logging verbosity through a cli arg or environment variable
* Adds a unique trace id to all requests via context to link them together
* Adds debug logging for upstream requests / responses with a unique request id   

Note: at some point in the future it would be good to add headers to the logging, however we have to be careful about logging confidential credentials etc, so I've intentionally left that part out for now.

## How Has This Been Tested?

Try this out with a real MR backend deployed to kind and a mock k8s client, e.g.
```make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=false LOG_LEVEL=debug```

Make some GET / POST / PATCH requests (the ones from the README are fine)

Verify you get the debug logging.

Also verify different log levels work, e.g. try info and verify the debug messages aren't there. Also try running without the LOG_LEVEL param to ensure it defaults to INFO.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/Makefile
+++ b/clients/ui/bff/Makefile
@@ -8,6 +8,7 @@ STANDALONE_MODE ?= true
 STATIC_ASSETS_DIR ?= ./static
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.29.0
+LOG_LEVEL ?= info
 
 .PHONY: all
 all: build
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: clients/ui/bff/Makefile [hunk 1]
```diff
@@ -47,7 +48,7 @@ build: fmt vet test ## Builds the project to produce a binary executable.
 .PHONY: run
 run: fmt vet envtest ## Runs the project.
 	ENVTEST_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-	go run ./cmd/main.go  --port=$(PORT) --mock-k8s-client=$(MOCK_K8S_CLIENT) --mock-mr-client=$(MOCK_MR_CLIENT) --dev-mode=$(DEV_MODE) --dev-mode-port=$(DEV_MODE_PORT)
+	go run ./cmd/main.go  --port=$(PORT) --mock-k8s-client=$(MOCK_K8S_CLIENT) --mock-mr-client=$(MOCK_MR_CLIENT) --dev-mode=$(DEV_MODE) --dev-mode-port=$(DEV_MODE_PORT) --log-level=$(LOG_LEVEL)
```
lucferbux, line 51: Shouldn't we change as well the dockerfile or other manifest files to reflect the log level we want?

lucferbux (changes requested):
Ok, so far it looks good, and when enabling the app we got log levels, I have a nit in for other build files and other question:

Should we get the logger in the k8s client too? Apologies if I don't fully understand bff still, but the http.go will cover al the MR and swagger routes, but what happens with k8s?

alexcreasy:
Yeah, I did think about this - I agree it would be really useful to have the k8s stuff logged too. The hard part is right now there's only a single, global k8s client, I guess if I can pass the trace id through at least we have a way of matching things up to a specific "transaction". 

I'll take a look this afternoon, but will just have a quick chat with Eder first to see how the k8s client is likely to change.

alexcreasy:
I should probably add some documentation too!!!

clients/ui/bff/Makefile
alexcreasy: it defaults to INFO which is probably the level we want deployed so we don't slow things down... You can just set an env var on the docker container to change the log level, I should probably alter this so it's more explicit and also add some documentation.

lucferbux:
/hold

We are gonna wait until https://github.com/kubeflow/model-registry/pull/599 is merged.

lucferbux:
@alexcreasy I think there's still some changes left but you can rebase this safely

Griffin-Sullivan (approved):
Nice work! Everything is working. Tested in dev mode pointing the BFF to a Kubeflow install of Model Registry and tested in mock mode.

Issue #338, opened 2024-09-02
Refactor CSI as different command of the model registry module


lampajr:
**Is your feature request related to a problem? Please describe.**
At the moment the CSI heavily relies on the model registry project, this is because the model-registry dependency is imported in order to interact with model registry by making use of the auto-generated and provided client:
https://github.com/kubeflow/model-registry/blob/332c0851544c525a68f8a467712b4b1542c25cf6/csi/pkg/storage/modelregistry_provider.go#L22-L27 

This implementation could lead to some issues, especially from a maintenance point of view, see https://github.com/kubeflow/model-registry/pull/328#issuecomment-2324712508 as an example.

**Describe the solution you'd like**
Decouple CSI and Model Registry by removing the `model-registry` dependency from the CSI dependencies.

Then update the CSI to make use of a custom and auto-generated REST client (or whatever could be valuable and easy-to-maintain approach) relying on the versioned model-registry OpenAPI specification.

**Describe alternatives you've considered**
n/a

**Additional context**
If we go for a custom auto-generated client to interact with the MR REST interface consider that we just need a couple of endpoints to invoke so I think it is worth to evaluate whether we can auto-generate just a subset of the OpenAPI spec.

rareddy:
Is there reason why CSI needs to be designed as extension to MR vs have it part of the MR code directly to avoid such issues?

tarilabs:
> Describe alternatives you've considered

an alternative I didn't have chance to investigate is whether [workspaces](https://go.dev/ref/mod#workspaces) would make this easier also to be properly managed from dependabot automated toolings.

lampajr:
> Is there reason why CSI needs to be designed as extension to MR vs have it part of the MR code directly to avoid such issues?

@rareddy  TBH I've never considered this because I always tried to make CSI as isolated and independent project as possible (at least this was the initial goal) but now that you pointed out again I am starting thinking it is worth to evaluate whether adding another go cmd https://github.com/kubeflow/model-registry/tree/main/cmd just for the CSI. Obviously generating the executable for the CSI will rebuild the model-registry source code as well (but this is actually already happening with the relative import).

Don't know if that could introduce other issues but I think it is worth to investigate.

> an alternative I didn't have chance to investigate is whether [workspaces](https://go.dev/ref/mod#workspaces) would make this easier also to be properly managed from dependabot automated toolings.

Good point, this is another thing I did not try yet too.

lampajr:
[...]

PR #598, opened 2024-11-27
fix(csi): make csi use mr version from go.mod in ci

Al-Pragliola:
## Description

When building/testing CSI in the CI we are using "local" mr as a dependency instead of the version pinned in the go.mod file, this bring an inconsistency between what we are shipping as a docker image and what is actually in the codebase.

Side effect: This should fix all the problems that dependabot gives us when touching mr go.mod.

## How Has This Been Tested?

no local testing is needed

## Merge criteria:


tarilabs (commented):
I understand this re-uses the standard docker-build target from the main makefile, and 

/lgtm

thanks a lot @Al-Pragliola

.github/workflows/csi-test.yml
lampajr, line 1: If we do this, we are making use of the model-registry version that has pinned in the go.mod as you said, but this makes impossible to test the current CSI with the dev (not yet released) version of model-registry.

lampajr (commented):
Hi @Al-Pragliola,

Overall I don't think this is going t owork as expected for some reasons:
- If we change `.github/workflows/csi-build-and-push-image.yml` we might be in a chicken-egg situation as we would like to release CSI together with model-registry but we don't have yet the tag released
- If we change `.github/workflows/csi-test.yml` we might be not able to test csi with a dev version of model-registry, i.e., we might be not able to detect any breaking change in model-registry that would break csi

wdyt?

Al-Pragliola:
> I understand this re-uses (*in github action) the standard docker-build target from the main makefile, and
> 
> /lgtm
> 
> thanks a lot @Al-Pragliola

yep, with `docker-build-dev` we had

```docker
# Copy the Go Modules manifests
COPY csi/go.mod go.mod
COPY csi/go.sum go.sum
RUN echo "replace github.com/kubeflow/model-registry => ../" >> go.mod
```

overriding the version of model-registry from the csi/go.mod file

lampajr:
Expanding on my previous comment, with `docker-build-dev` I had to do
```dockerfile
# Copy the Go Modules manifests
COPY csi/go.mod go.mod
COPY csi/go.sum go.sum
RUN echo "replace github.com/kubeflow/model-registry => ../" >> go.mod
```

exactly because we would like to test the `csi` with the local/dev model-registry in order to be able to detect any breaking change earlier.

Consider the scenario where you change something, by mistake or on purpose, you would not be able to detect that if you use a pinned version of model registry.

On the other hand, we could try to change `.github/workflows/csi-build-and-push-image.yml` to make use of a pinned version of the model-registry but either:
- We change/set that version during the workflow (as it is triggered on tag push) but how do we manage the `main` branch push?
- We decide the keep csi and model registry **NOT** in sync and keep using a pinned version, in this case we should not release csi on tag push

.github/workflows/csi-test.yml
Al-Pragliola: Yeah, this is not the solution to the bigger problem of the relationship between CSI and MR and it's not aimed to be, for that we should consider what you guys proposed in the related issue https://github.com/kubeflow/model-registry/issues/338

But to test CSI with the current unreleased version of MR we could make use of:

`go get github.com/kubeflow/model-registry@<commit-hash>`

so that when we are releasing a new version of both we take the latest commit hash and update the go.mod (if it's needed at all, because if there's no change to the rest client it should be compatible)

wdyt?

Al-Pragliola:
> Expanding on my previous comment, with `docker-build-dev` I had to do
> 
> ```dockerfile
> # Copy the Go Modules manifests
> COPY csi/go.mod go.mod
> COPY csi/go.sum go.sum
> RUN echo "replace github.com/kubeflow/model-registry => ../" >> go.mod
> ```
> 
> exactly because we would like to test the `csi` with the local/dev model-registry in order to be able to detect any breaking change earlier.
> 
> Consider the scenario where you change something, by mistake or on purpose, you would not be able to detect that if you use a pinned version of model registry.
> 
> On the other hand, we could try to change `.github/workflows/csi-build-and-push-image.yml` to make use of a pinned version of the model-registry but either:
> 
>     * We change/set that version during the workflow (as it is triggered on tag push) but how do we manage the `main` branch push?
> 
>     * We decide the keep csi and model registry **NOT** in sync and keep using a pinned version, in this case we should not release csi on tag push

Let's analyze this scenario: 

- breaking changes from MR and pinned version in go.mod

(from this PR) CI will build using the pinned version from go.mod so the build will be successful - then the e2e tests will fail because we are testing against the local built version of MR (and there's a breaking change)

pro:
- we avoid false negatives from go.mod bumps in MR
- docker image and the codebase are in sync

cons:
- we rely on the robustness of our e2e tests to be sure that everything works
- we might have to bump using commit hashes for dev releases of MR


(previous) CI will fail to build and stop here

pro:
- ci will fail earlier
- no need to bump the go.mod file when releasing

cons:
- what we push on the image hub does not reflect what is in the codebase
- there will be false negatives when we modify the go.mod in MR


neither is a perfect solution, as mentioned earlier we can improve the "fix" from this PR by adding a GHA that actually checks if something breaks when there's a change in the mr pkg folder

Al-Pragliola:
> I think we can remove this `cons` by clearly setting that CSI uses the locally referenced model-registry codebase, without pinning any version, I think this is actully possible (i.e., keeping `"replace github.com/kubeflow/model-registry => ../"` by default)
> 
> > there will be false negatives if we change the go.mod in MR
> 
> I agree this is a problem, but I am not sure this is going to be fixed with this PR, happy to be wrong though

You are right, I did not provide any evidence 🙏🏼  I opened a PR for a recent example of a false negative, here you can see the previous behavior:

https://github.com/kubeflow/model-registry/pull/548
https://github.com/kubeflow/model-registry/actions/runs/12032677420/job/33545058641?pr=548

and post-"fix":

https://github.com/kubeflow/model-registry/pull/600

lampajr:
> > I think we can remove this `cons` by clearly setting that CSI uses the locally referenced model-registry codebase, without pinning any version, I think this is actully possible (i.e., keeping `"replace github.com/kubeflow/model-registry => ../"` by default)
> > > there will be false negatives if we change the go.mod in MR
> > 
> > 
> > I agree this is a problem, but I am not sure this is going to be fixed with this PR, happy to be wrong though
> 
> You are right, I did not provide any evidence 🙏🏼 I opened a PR for a recent example of a false negative, here you can see the previous behavior:
> 
> #548 https://github.com/kubeflow/model-registry/actions/runs/12032677420/job/33545058641?pr=548
> 
> and post-"fix":
> 
> #600
> 

Thanks for the examples 🙏🏻 

> I agree with you that we should implement a solution from #338 , I am a fan of the sub command approach 👍🏼

I am planning to propose something sooner, then I think we can discuss whether that approach is cleaner or not and if we want to proceed that way, wdyt?

Al-Pragliola:
> I am planning to propose something sooner, then I think we can discuss whether that approach is cleaner or not and if we want to proceed that way, wdyt?

Sure, that sounds good to me. Let's discuss it further once you've proposed the approach

PR #603, opened 2024-12-02
fix(isvc): post testing

Al-Pragliola:
## Merge criteria:


pkg/inferenceservice-controller/suite_test.go
```diff
@@ -41,7 +41,8 @@ const (
 	modelVersionIDLabel     = "modelregistry.kubeflow.org/model-version-id"
 	namespaceLabel          = "modelregistry.kubeflow.org/namespace"
 	nameLabel               = "modelregistry.kubeflow.org/name"
-	urlLabel                = "modelregistry.kubeflow.org/url"
+	skipTLSVerifyAnnotation = "modelregistry.kubeflow.org/skip-tls-verify"
```
tarilabs, line 44: open question: this seems a cross-cutting concern of the environment, wouldn't make more sense to make it configurable at the test level/controller level, rather than per-k8s-resource?

I'm wondering if this doesn't open some concern in "prod" as it would allow to define a per-Isvc insecure connection (truly, it's opt-in, but I thought worthy in raising this)

wdyt?

tarilabs (commented):
thank you @Al-Pragliola !

small comment for quick check, otherwise 

/lgtm

pkg/inferenceservice-controller/suite_test.go
Al-Pragliola: yeah you are right, seems like the wrong place to set this field

tarilabs (approved):
looks more intentional to me, thanks a lot for incorporating the feedback and it's always easy to reconsider this choice eventually if needed

/lgtm
/approve

PR #683, opened 2025-01-03
feat(isvc): made name and url optional - added support to svc url annotation

Al-Pragliola:
## Description

This PR aims to add backwards compatibility to existing inferenceservices that are missing the new labels/annotations, so the following changes have been made:

- made the name label and URL annotation optional, the reconciler will try to connect to a model registry service in the namespace specified by the label or the default (if missing), if there's more than one MR service it will stop and fail.
- The reconciler will try to get the URL from the service's annotation if it's specified and it's present on the service, if it's missing it will build the URL using the name/namespace/rest-api port.

## How Has This Been Tested?

`make test`

## Merge criteria:

### DIFF
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -29,6 +29,7 @@ type InferenceServiceController struct {
 	modelRegistryNameLabel        string
 	modelRegistryURLAnnotation    string
 	modelRegistryFinalizer        string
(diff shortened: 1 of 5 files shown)


pboyd (approved):
Looks good to me @Al-Pragliola

tarilabs (approved):
thanks @Al-Pragliola and @pboyd !

/approve

PR #690, opened 2025-01-08
feat(bff): serving static assets on bff

ederign:
## Description
This PR is a preparation for the BFF and FE being served in production on the same container.

At "/" we are now serving static files:

Screenshot 2025-01-08 at 5 23 57 PM

As we used to do, we keep serving the BFF API on "/api/v1/*".

It is important also to highlight that on production (when using binaries), we need to copy the assets to the target containers. I've decided not to pack the static assets in the binary and serve them in the filesystem. I believe this makes it more coherent with our architecture (but I'm open to discussion and suggestions). If you would like more details on this, please look at MakeFile and Docker file changes.
 
Changes on this PR:

- Dockerfile: copying the static assets during build process and passing the target asset dir as parameter to binary
- Makefile: also added the STATIC_ASSETS_DIR property, pointing to root static folder
- main.go: added the new configuraton for static assets
- app.go: in order to avoid class, I've to split the router in two. One for the /api and another (server mux) to also handle "/" for static assets.
- app_Test.go: test file to make sure that we don't break this
- middleware.go: I've ignored the enforecer to use headers when we are using static files
- clients/ui/bff/internal/api/test_utils.go: as we can run the tests on the ide, it's annoying to hardcode the parameters of every test location to find the static folder. So I write this small traversal to find assets dir based on go.mod, not matter where the tests are located. THIS IS NOT INTEND FOR PRODUCTION BECAUSE WE SHOULD USE PARAMETERS there.
- clients/ui/bff/internal/config/environment.go: new configurations
- clients/ui/bff/static/index.html and clients/ui/bff/static/sub/test.html STUB files

## How Has This Been Tested?
 
- Loading the static assets: 
Screenshot 2025-01-08 at 5 23 57 PM

- Loading the API to make sure that keep working
Screenshot 2025-01-08 at 5 24 51 PM

```shell
making sure that if I pass a bad header keep blocking requests
curl -i -H "fsdfsdfs-userid: user@example.com" "localhost:4000/api/v1/healthcheck"                      (kind-up-running/default)
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Content-Type: application/json
Date: Wed, 08 Jan 2025 22:32:38 GMT
Content-Length: 92

{
	"error": {
		"code": "400",
		"message": "missing required header: kubeflow-userid"
	}
}
```

I've also started to test this with our production assets of frontend.

Screenshot 2025-01-08 at 4 59 28 PM

Kind of work, but I was not able to figure out how to fix the routing. I've asked for FE guys help to debug the proxy calls..

## Merge criteria:

### DIFF
--- a/clients/ui/bff/Dockerfile
+++ b/clients/ui/bff/Dockerfile
@@ -15,7 +15,8 @@ RUN go mod download
 COPY cmd/main.go cmd/main.go
 COPY internal/ internal/
 
-
+# Copy the static assets
+COPY static/ static/
 
 # Build the Go application
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bff ./cmd/main.go
@@ -29,4 +30,4 @@ USER 65532:65532
 # Expose port 4000
 EXPOSE 4000
 
-ENTRYPOINT ["/bff"]
+ENTRYPOINT ["/bff", "--static-assets-dir=/static"]
--- a/clients/ui/bff/Dockerfile
+++ b/clients/ui/bff/Dockerfile
@@ -16,7 +16,7 @@ COPY cmd/main.go cmd/main.go
 COPY internal/ internal/
 
 # Copy the static assets
-COPY static/ static/
+COPY $STATIC_ASSETS_DIR static/
 
 # Build the Go application
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bff ./cmd/main.go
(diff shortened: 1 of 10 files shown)


Griffin-Sullivan (approved):
I tested locally after rebasing and rebuilding the frontend with the proper env variables you can run the BFF with
```
make run MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true STATIC_ASSETS_DIR="../frontend/dist"
```
and you will get: 
screenshot_from_2025-01-13_10-51-07_720

ederign:
nice catch @Griffin-Sullivan

PR #710, opened 2025-01-17
Adds BFF support for CORS

alexcreasy:
NOTE: This PR is stacked on top of my logging PR, that will need to be merged first.

## Description

Adds W3C standards compliant support for CORS for the BFF

## How Has This Been Tested?

This can be tested by running the BFF in full mock mode as the headers are generated by a middleware that will run in all scenarios. The tests involve merely querying the health check endpoint with a different combination of headers. 

This to note: 
* CORS is only enforced by browsers, so all requests should be successful, the headers returned will vary and the browser will either accept or deny the request based on this.
* When the frontend is served by the BFF, CORS will not be required as it will be served from the same origin - as such the default behaviour is to not return any CORS headers, this disallows cross origin requests and is the most secure.

Scenario 1: Testing with CORS disabled
Start the server:
```
 make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true
```

Run a curl against the health check endpoint:
```
curl -i -H "kubeflow-userid: user@example.com" "localhost:4000/api/v1/healthcheck"
```

Expected outcome: The endpoint should function as normal and there should be no CORS headers in the response.

Scenario 2: Test with a single allowed origin
Start the server:
```
 make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true ALLOWED_ORIGINS="http://example.com"
```

2.1 Make a request with an allowed origin header
```
curl -i -H "Origin: http://example.com" -H "kubeflow-userid: user@example.com" "localhost:4000/api/v1/healthcheck"
```

Expected outcome: Successful request with the following CORS specific headers:
```
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://example.com
Vary: Origin
```

2.2 Make a request with a disallowed origin header
```
curl -i -H "Origin: http://shady-as-duck.com" -H "kubeflow-userid: user@example.com" "localhost:4000/api/v1/healthcheck"
```
Expected outcome: The request should be successful, but the only CORS header in the response should be:
```
Vary: Origin
```
If there are any other Access-Control-* headers present the test has failed!

2.3 Simulate a CORS pre-flight request for an allowed origin:
```
curl -i -X OPTIONS -H "Origin: http://example.com" -H "Access-Control-Request-Method: POST" -H "Access-Control-Request-Headers: kubeflow-userid" "localhost:4000/api/v1/healthcheck"
```
Expected outcome: A response like below - note the CORS headers.
```
HTTP/1.1 204 No Content
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: kubeflow-userid
Access-Control-Allow-Methods: POST
Access-Control-Allow-Origin: http://example.com
Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
Date: Fri, 17 Jan 2025 18:12:13 GMT
```

2.4 Simulate a CORS pre-flight request for a disallowed origin
```
curl -i -X OPTIONS -H "Origin: http://shady-as-duck.com" -H "Access-Control-Request-Method: POST" -H "Access-Control-Request-Headers: kubeflow-userid" "localhost:4000/api/v1/healthcheck"
```

Expected outcome: A response like below - note the lack of CORS headers aside from the `Vary` header
```
HTTP/1.1 204 No Content
Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
Date: Fri, 17 Jan 2025 18:15:10 GMT
```

Additional tests:
You can repeat the above tests using an allow list of multiple origins, simply add comma separated ones to the make command e.g.

```
 make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true ALLOWED_ORIGINS="http://example1.com,http://example2.com"
```
NOTE: Only a single origin is allowed in the `Access-Control-Allow-Origin` header, so the BFF will return only a single origin if the `Origin` request header matches the allow list.

Also you can test that allowing all origins (wildcard) works as expected:
```
 make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true ALLOWED_ORIGINS="*"
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/api/middleware.go
+++ b/clients/ui/bff/internal/api/middleware.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubeflow/model-registry/ui/bff/internal/config"
 	"github.com/kubeflow/model-registry/ui/bff/internal/constants"
 	"github.com/kubeflow/model-registry/ui/bff/internal/integrations"
+	"github.com/rs/cors"
 	"log/slog"
 	"net/http"
 	"runtime/debug"
@@ -66,14 +67,23 @@ func (app *App) InjectUserHeaders(next http.Handler) http.Handler {
 	})
 }
 
-func (app *App) enableCORS(next http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// TODO(ederign) restrict CORS to a much smaller set of trusted origins.
-		// TODO(ederign) deal with preflight requests
-		w.Header().Set("Access-Control-Allow-Origin", "*")
+func (app *App) EnableCORS(next http.Handler) http.Handler {
+	allowedOrigins, ok := ParseOriginList(app.config.AllowedOrigins)
 
-		next.ServeHTTP(w, r)
+	if !ok {
+		return next
+	}
+
+	c := cors.New(cors.Options{
+		AllowedOrigins:     allowedOrigins,
+		AllowCredentials:   true,
+		AllowedMethods:     []string{"GET", "PUT", "POST", "PATCH", "DELETE"},
+		AllowedHeaders:     []string{constants.KubeflowUserIDHeader, constants.KubeflowUserGroupsIdHeader},
+		Debug:              strings.ToLower(app.config.LogLevel) == "debug",
+		OptionsPassthrough: false,
 	})
+
+	return c.Handler(next)
 }
 
 func (app *App) EnableTelemetry(next http.Handler) http.Handler {
--- a/clients/ui/bff/Makefile
+++ b/clients/ui/bff/Makefile
@@ -9,6 +9,7 @@ STATIC_ASSETS_DIR ?= ./static
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.29.0
 LOG_LEVEL ?= info
+ALLOWED_ORIGINS ?= ""
 
 .PHONY: all
 all: build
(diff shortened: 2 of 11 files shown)


clients/ui/bff/internal/api/middleware.go
```diff
@@ -66,14 +67,23 @@ func (app *App) InjectUserHeaders(next http.Handler) http.Handler {
 	})
 }
 
-func (app *App) enableCORS(next http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// TODO(ederign) restrict CORS to a much smaller set of trusted origins.
-		// TODO(ederign) deal with preflight requests
-		w.Header().Set("Access-Control-Allow-Origin", "*")
+func (app *App) EnableCORS(next http.Handler) http.Handler {
+	allowedOrigins, ok := ParseOriginList(app.config.AllowedOrigins)
```
ederign, line 71: It's a nitpick and can be done in a FUP PR. 

As we are not planning to change AllowedOrigins values in runtime, we can parse them once during initialization and avoid dois this call on every request.

clients/ui/bff/internal/api/middleware.go
```diff
@@ -66,14 +67,23 @@ func (app *App) InjectUserHeaders(next http.Handler) http.Handler {
...
 
-		next.ServeHTTP(w, r)
+	if !ok {
+		return next
```
ederign, line 74: It's a nitpick and can be done in a FUP PR. Can  we log a warning (error) for invalid entries instead of silently failing?

ederign (approved):
Alex great PR and I loved the explanation in the description! Just did two small suggestions that we can do in a FUP.

ederign:
@alexcreasy what do you prefer? Merge this and do a FUP?

clients/ui/bff/internal/api/middleware.go
alexcreasy: Yeah that's a really good point actually - I'll put a follow up in today just to fix these couple

clients/ui/bff/internal/api/middleware.go
alexcreasy: I think maybe my naming isn't so good here - "not ok" in this instance means that the allowlist is empty - as such we don't need a CORS middleware since not having CORS headers is the default - I'll make this more explicit in the follow up.

alexcreasy: As this is the "normal" use case (yeah, "not ok" being the normal case is weird naming!!) we'd be getting a WARN on probably 99% of deployed use cases.

ederign:
just talked with @alexcreasy and he will send a FUP PR.

PR #761, opened 2025-02-04
Add `upload_artifact_and_register_model` method to python client

Crazyglue:
## Description

Adds a convenience method to perform a common operation within the python client, to both upload and register a model that exists locally on the filesystem.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -188,13 +188,12 @@ In order to utilize this method you must instantiate an `upload_params` object w
 
 #### S3 based external storage
 
+Common S3 env vars will be automatically read, such ass the access_key_id, etc. It can also be provided explicitly in the `S3Params` object if desired.
+
 ```python
 s3_upload_params = S3Params(
     bucket="my-bucket",
-    s3_prefix="hello_world_model",
-    access_key_id="my-key-id",
-    access_key_secret="super-secret",
-    region="us-east-1"
+    s3_prefix="models/my_fraud_model",
 )
 
 registered_model = client.upload_artifact_and_register_model(
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -178,6 +178,92 @@ By default, all queries will be `ascending`, but this method is also available f
 > Advanced usage note: You can also set the `page_size()` that you want the Pager to use when invoking the Model Registry backend.
 > When using it as an iterator, it will automatically manage pages for you.
 
+### Uploading local models to external storage and registering them
+
+To both upload and register a model, use the convenience method `upload_artifact_and_register_model`.
+
+This method supports both s3-based storage (via [boto3](https://github.com/boto/boto3)) as well as OCI-based image registries (via [olot](https://github.com/containers/olot), using either of the CLI tools [skopeo](https://github.com/containers/skopeo) or [oras](https://github.com/oras-project/oras))
+
+In order to utilize this method you must instantiate an `upload_params` object which contains the necessary locations and credentials needed to perform the upload to that storage provider.
+
+#### S3 based external storage
+
+```python
+s3_upload_params = S3Params(
+    bucket="my-bucket",
+    s3_prefix="hello_world_model",
+    access_key_id="my-key-id",
+    access_key_secret="super-secret",
+    region="us-east-1"
+)
+
+registered_model = client.upload_artifact_and_register_model(
+    name="hello_world_model",
+    model_fiels_path="/home/user-01/models/model_training_01"
+    # If the model consists of a single file, such as a .onnx file, you can specify that as well
+    # model_fiels_path="/home/user-01/models/model_training_01.onnx"
+    author="Mr. Trainer",
+    version="0.0.1",
+    upload_params=s3_upload_params
+)
+```
+
+#### OCI-registry based storage
+
+```python
+oci_upload_params = OCIParams(
+    base_image="busybox",
+    oci_ref="registry.example.com/acme_org/hello_world_model:0.0.1"
+)
+
+registered_model = client.upload_artifact_and_register_model(
+    name="hello_world_model",
+    model_fiels_path="/home/user-01/models/model_training_01"
+    # If the model consists of a single file, such as a .onnx file, you can specify that as well
+    # model_fiels_path="/home/user-01/models/model_training_01.onnx"
+    author="Mr. Trainer",
+    version="0.0.1",
+    upload_params=oci_upload_params
+)
+```
+
+Additionally, OCI-based storage supports multiple CLI clients to perform the upload. However, one of these clients must be available in the hosts PATH. **Ensure your host has either [skopeo](https://github.com/containers/skopeo) or [oras](https://github.com/oras-project/oras) installed and available.** 
+
+By default, `skopeo` is used to perform the OCI image download/upload.
+
+If you prefer to use `oras` instead, you can specify it like so:
+
+```python
+oci_upload_params = OCIParams(
+    base_image="busybox",
+    oci_ref="registry.example.com/acme_org/hello_world_model:0.0.1",
+    backend="oras"
+)
+```
+
+Additionally, if neither of these CLI clients are sufficient for you, you can provide a `custom_oci_backend` in the `OCIParams` and specify the appropriate methods
+
+```python
+def is_available():
+    pass
+def pull():
+    pass
+def push():
+    pass
+
+custom_oci_backend = {
+    "is_available": is_available,
+    "pull": pull,
+    "push": push,
+}
+
+oci_upload_params = OCIParams(
+    base_image="busybox",
+    oci_ref="registry.example.com/acme_org/hello_world_model:0.0.1",
+    custom_oci_backend=custom_oci_backend,
+)
+```
+
 #### Implementation notes
 
 The pager will manage pages for you in order to prevent infinite looping.
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -182,6 +182,8 @@ class ModelRegistry:
         name: str,
         model_files_path: str,
         *,
+        # Upload/client Params
+        upload_params: OCIParams | S3Params,
         # Artifact/Model Params
         version: str,
         model_format_name: str,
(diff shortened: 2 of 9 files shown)


clients/python/src/model_registry/_client.py
```diff
@@ -164,6 +166,73 @@ async def _register_model_artifact(
...
+        author: str | None = None,
+        owner: str | None = None,
+        description: str | None = None,
+        metadata: Mapping[str, SupportedTypes] | None = None,
+        upload_client_params: Mapping[str, str] | None = None,
+    ) -> RegisteredModel:
+        if is_s3_uri(destination_uri):
+            self._upload_to_s3(artifact_local_path, destination_uri, upload_client_params['region_name'])
+        elif is_oci_uri(destination_uri):
+            self._upload_to_oci(artifact_local_path, destination_uri)
```
tarilabs, line 211: I concur the flow is decided based on destination; just thinking we could leverage here functions to check the required configuration parameters (ie Env variables) are present

tarilabs (commented):
thank you for sharing this as a draft @Crazyglue !

I think this will be helpful to ease the work of the MLOps eng, DS persona, who typically need to operate these capabilities otherwise manually.

Any general thought of a plug-in mechanism if people want to override an implementation?
ie we are likely to offer a default with `boto3` as an extra for the `_upload_to_s3()`, what mechanism do we want to suggest to other Devs who may want to implement their own tooling for S3 ? (for future extensibility, doens't have to be in _this_ PR)

clients/python/src/model_registry/_client.py
Crazyglue: I was thinking that the individual upload methods would be checking for those, as they would instantiate their respective clients (boto3 vs skopeo). 

I haven't done it yet in this PR, but the `upload_client_params` mapping will actually need to contain either the boto3 params OR the skopeo params, and that mapping is passed to both upload methods. Then those methods can determine whether it has enough information to instantiate a client and connect/perform the upload

Crazyglue:
> Any general thought of a plug-in mechanism if people want to override an implementation? ie we are likely to offer a default with `boto3` as an extra for the `_upload_to_s3()`, what mechanism do we want to suggest to other Devs who may want to implement their own tooling for S3 ? (for future extensibility, doens't have to be in _this_ PR)

To be honest I'm not familiar enough with Python to know the "pythonic" way of doing this, haha. but perhaps we can have a kwarg to provide a boto3 or skopeo client as an optional param to this method. if provided, use it, else instantiate a client. after client creation/resolution we can perform the upload.

it might be worth considering another option, lower level to the above, which is to create a separate `StorageClient` class which provides an interface ontop of the boto3/skopeo client to perform a `.upload()`

@tarilabs any thoughts on what the right level of abstraction might be in this case? if the two storage options are just s3 and oci, then im not sure the added complexity is worth it

clients/python/src/model_registry/_client.py
```diff
@@ -164,6 +166,73 @@ async def _register_model_artifact(
...
+        region_name: str | None = None,
+    ) -> str:
+        """
+        Uploads a file to an S3 bucket.
+        """
+        pass
+
+    def _upload_to_oci(
+        self,
+        artifact_local_path: str,
```
tarilabs, line 183: One thing to consider here, is how to pass/customize the "base image" of the modelcar being built. Do we want that to be an explicit parameter, or a default from a config which could be overridden if needed ?

tarilabs:
> @tarilabs any thoughts on what the right level of abstraction might be in this case? if the two storage options are just s3 and oci, then im not sure the added complexity is worth it

considering this is to support an internal implementation which likely will need further iterations/refinement, I wouldn't "prematurely overengineer it" :)
Until is confided within a module, we can always refactor later as we discover more details on use-cases.

clients/python/src/model_registry/utils.py
```diff
@@ -135,6 +131,37 @@ def _get_oras_backend() -> BackendDefinition:
...
+class S3Params:
+    """Parameters for the S3 Client (boto3) to perform the upload.
+
+    Allows for some amount of customization when performing an upload, such as providing a custom endpoint url, access keys, etc.
+    """
+
+    bucket_name: str
+    s3_prefix: str
+    endpoint_url: str | None = None
+    access_key_id: str | None = (None,)
```
syntaxsdev, line 160: why is the default a `None` tuple?

clients/python/src/model_registry/utils.py
```diff
@@ -201,9 +231,72 @@ def save_to_oci_registry(
...
+    files = [file[0] for file in get_files_from_path(model_files_path)]
+    oci_layers_on_top(local_image_path, files, modelcard)
     backend_def["push"](local_image_path, oci_ref)
 
     # Return the OCI URI
 
     return f"oci://{oci_ref}"
+
+
+s3_prefix = "s3://"
```
syntaxsdev, line 244: Did you mean to put this here

clients/python/tests/test_client.py
```diff
@@ -708,7 +708,7 @@ def test_singular_store_in_s3(get_model_file, patch_s3_env, client: ModelRegistr
     assert model_name_pfx in objects_by_name
 
     # Test file not exists
-    with pytest.raises(StoreError) as e:
+    with pytest.raises(ValueError, match="Please ensure path is correct.") as e:
```
syntaxsdev, line 711: nice! I didn't know this was possible 👍

clients/python/tests/test_utils.py
```diff
@@ -136,39 +140,72 @@ def pull_mock_imple(base_image, dest_dir):
...
+
+def test_get_files_from_path_single_file(get_model_file):
+    file = get_files_from_path(get_model_file)
+    # It returns only 1 file in the list, and it is a tuple of (absolute_path, filename)
+    assert len(file) == 1
+    assert file[0] == (get_model_file, os.path.basename(get_model_file))
+
+
+def test_get_files_from_path_multiple_files(get_temp_dir_with_models):
+    path, generated_files = get_temp_dir_with_models
```
syntaxsdev, line 205: nice reuse of `get_temp_dir_with_models`

syntaxsdev (approved):
Functionally looks good, some minor questions/feedback

clients/python/src/model_registry/utils.py
Crazyglue: Whoops, removed

tarilabs (commented):
To me this is fantastic, thanks a lot @Crazyglue and great job you both @syntaxsdev for the underlying methods.

/lgtm

Leaving this open a little more for comments, in the meantime I recommend building locally and testing it out (for the extent possible) in multiple environments!

clients/python/README.md
```diff
@@ -178,6 +178,92 @@ By default, all queries will be `ascending`, but this method is also available f
...
+This method supports both s3-based storage (via [boto3](https://github.com/boto/boto3)) as well as OCI-based image registries (via [olot](https://github.com/containers/olot), using either of the CLI tools [skopeo](https://github.com/containers/skopeo) or [oras](https://github.com/oras-project/oras))
+
+In order to utilize this method you must instantiate an `upload_params` object which contains the necessary locations and credentials needed to perform the upload to that storage provider.
+
+#### S3 based external storage
+
+```python
+s3_upload_params = S3Params(
+    bucket="my-bucket",
+    s3_prefix="hello_world_model",
```
tarilabs, line 194:
```suggestion
    s3_prefix="models/fraud",
```

so we hint how the underlying function organize them in a bucket imho 😉

clients/python/README.md
```diff
@@ -178,6 +178,92 @@ By default, all queries will be `ascending`, but this method is also available f
...
+    access_key_id="my-key-id",
+    access_key_secret="super-secret",
+    region="us-east-1"
```
tarilabs, line 197: this is great to show, but imho we can omit and making reference to the fact the underlying function picks from 12factor style Env variables, my2c

clients/python/README.md
```diff
@@ -178,6 +178,92 @@ By default, all queries will be `ascending`, but this method is also available f
...
+    upload_params=s3_upload_params
+)
+```
+
+#### OCI-registry based storage
+
+```python
+oci_upload_params = OCIParams(
+    base_image="busybox",
+    oci_ref="registry.example.com/acme_org/hello_world_model:0.0.1"
```
tarilabs, line 216: very fair use of Tags, when we will have permanent Tags in some OCI registry we will update here as well 🚀

tarilabs (commented):
thanks a lot for these updates!

clients/python/README.md
```diff
@@ -178,6 +178,91 @@ By default, all queries will be `ascending`, but this method is also available f
...
+
+To both upload and register a model, use the convenience method `upload_artifact_and_register_model`.
+
+This method supports both s3-based storage (via [boto3](https://github.com/boto/boto3)) as well as OCI-based image registries (via [olot](https://github.com/containers/olot), using either of the CLI tools [skopeo](https://github.com/containers/skopeo) or [oras](https://github.com/oras-project/oras))
+
+In order to utilize this method you must instantiate an `upload_params` object which contains the necessary locations and credentials needed to perform the upload to that storage provider.
+
+#### S3 based external storage
+
+Common S3 env vars will be automatically read, such ass the access_key_id, etc. It can also be provided explicitly in the `S3Params` object if desired.
```
tarilabs, line 191: 👍

clients/python/README.md
```diff
@@ -178,6 +178,91 @@ By default, all queries will be `ascending`, but this method is also available f
...
+
+```python
+s3_upload_params = S3Params(
+    bucket="my-bucket",
+    s3_prefix="models/my_fraud_model",
```
tarilabs, line 196: 👍

tarilabs (commented):
/lgtm

tarilabs:
thank you for keeping tabs on this over this past month @Crazyglue and all contributors

- https://github.com/kubeflow/model-registry/pull/761#pullrequestreview-2662024343
- https://github.com/kubeflow/model-registry/pull/761#pullrequestreview-2663759539

/approve

PR #765, opened 2025-02-05
feat(py): Add support for storing models in S3

syntaxsdev:
Within the Python client, users will be able to directly store models to an S3 compatible object storage.
AWS S3 connections will automatically be consumed.

## Description

The bulk of the changes were done in `clients/python/src/_client.py`
With E2E tests that were added to test functionality by creating a Minio instance

This implementation is within the `ModelRegistry` class

Example usage:
```
mr = ModelRegistry(...)
mr.save_to_s3(
  path="models",
  bucket_name="default",
  endpoint_url="xxx",
  access_key_id="xxx",
  secret_access_key="xxx"
)
```
You can path both files and paths to the `path` parameter.
If a path is nested (ex. `data/models`) then an `s3_prefix` variable will be required to allow for proper folder naming.
Environment variables are picked up, but variables supplied via direct params are preferred over environment variables that are mounted.

## How Has This Been Tested?
A local and remote version of Minio S3

1. Changes to the Makefile target were made to add a local Minio instance for the same Kind cluster than the Makefile target `deploy-latest-mr` uses.
2. An `.env` file is generated so that vars can be used for the added tests
3. Three new tests were added that single for storing a singular model to S3 and for storing a path of models, recursively

Tests are passing when completely rebuilding and testing the new e2e workflow
image

To test, run `make test-e2e` as per usual.

## Merge criteria:

If you have UI changes


clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +441,101 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+            msg = "Failed to upload to S3"
+            raise StoreError(msg) from e
+
+    def __connect_to_s3(
+        self,
+        endpoint_url: str | None = None,
+        access_key_id: str | None = None,
+        secret_access_key: str | None = None,
+        region: str | None = None,
+    ) -> "BaseClient":  # <--- here is a question I have
```
syntaxsdev, line 487: do we want to forward reference whatever is returned from boto3 Client or does it not matter since its private?

Boto3 does not return actually types for the resource, they are generated at runtime
so I can't set the type to "botocore.client.S3" for example, but I can set it to a base object. @tarilabs wdyt?

tarilabs: I would strike for the simplest maintenance solution, indeed as you point out being a private method.
How would this change if we consider also olot or other libraries in the future, still as extra?

tarilabs (commented):
thank you @syntaxsdev for this!

some initial comments below.
which type of test can we consider to make sure the functionality is covered?

I'm thinking we could have some dedicated e2e test by extending the current opt-in pytest mechanism and deploy minio in that "scenario" of e2e testing. Do you have some additional ideas?

clients/python/src/model_registry/_client.py
syntaxsdev: @tarilabs Eric suggested https://youtype.github.io/types_boto3_docs/types_boto3_s3/

not sure if we want to go this route however

tarilabs: if we're unsure, let's defer this for later, provided we will get chances to get back to it (maybe log it as a separate jira)

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +441,101 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
             return self.async_runner(self._api.get_model_versions(rm.id, options))
 
         return Pager[ModelVersion](rm_versions)
+
+    def save_to_s3(
+        self,
+        file: str,
```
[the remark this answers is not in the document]
syntaxsdev: No, it only uploads a singular file due to how `upload_file` works. Are you suggesting writing a wrapper to achieve recursive path uploads?

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +441,101 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+        name: str,
+        bucket_name: str,
+        *,
+        endpoint_url: str | None = None,
+        access_key_id: str | None = None,
+        secret_access_key: str | None = None,
+    ) -> None:
```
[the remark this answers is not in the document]
syntaxsdev: I have changed to return an s3_uri however this highly depends on your input for adding recursive paths, since we'd have to return a list of the s3 uris if so, or some other structure we'd need to conjure

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +441,101 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+        Raises:
+            StoreError: If there was an issue uploading to S3.
+        """
+        s3 = self.__connect_to_s3(
+            endpoint_url=endpoint_url,
+            access_key_id=access_key_id,
+            secret_access_key=secret_access_key,
+        )
+        try:
+            s3.upload_file(file, bucket_name, name)
```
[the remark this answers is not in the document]
tarilabs: it was in relation if it uploads recursively or need an explicit orchestration

clients/python/src/model_registry/_client.py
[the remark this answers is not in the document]
tarilabs: I would use the S3 uri format we're using in the MR readthedocs tutorial for building an S3 URI to a bucket

clients/python/src/model_registry/_client.py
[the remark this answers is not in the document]
syntaxsdev: what exactly are you referring to?
if you are referring to [this](https://model-registry.readthedocs.io/en/latest/#listing-models) then that's not exactly what I was talking about.

Afaik, boto3 S3 does not have a multiple upload definition or allow recursive. uploads, we'd have to build that.

that's not a problem - the issue is do we want to add that built into this method and if so, see https://github.com/kubeflow/model-registry/pull/765#discussion_r1945054393

clients/python/src/model_registry/_client.py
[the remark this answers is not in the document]
syntaxsdev: It does not. See here: https://github.com/kubeflow/model-registry/pull/765#discussion_r1945192515

clients/python/tests/conftest.py
```diff
@@ -112,3 +112,27 @@ def setup_env_user_token():
...
+def get_model_file():
+    with tempfile.NamedTemporaryFile(delete=False) as model_file:
+        pass
+
+    yield model_file.name
+
+    os.remove(model_file.name)
+
+
+@pytest.fixture
```
syntaxsdev, line 127: no `scope` added here because MonkeyPatch needs to use a `function` scope and default is function so its omitted

clients/python/tests/test_client.py
```diff
@@ -623,3 +629,44 @@ def test_hf_import_default_env(client: ModelRegistry):
 
     for k in env_values:
         os.environ.pop(k)
+
+
+@pytest.mark.dd
```
syntaxsdev, line 634: ignore for now, will change

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +442,200 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+            raise StoreError(msg)
+
+        if os.path.isfile(path):
+            filename = os.path.basename(path)
+            if not filename:
+                msg = "An error occured determining if the supplied path was file."
+                raise StoreError(msg)
+
+            s3.upload_file(path, bucket, filename)
+            uri = s3_uri_from(
```
tarilabs, line 532: for context, this is to provide (mainly for KServe) the "path" of the S3 bucket.

see example in: https://kserve.github.io/website/latest/modelserving/storage/s3/s3/#deploy-the-model-on-s3-with-inferenceservice

and also out tutorial with: https://www.kubeflow.org/docs/components/model-registry/getting-started/#setup

so I believe wee need to provide the parent directory in the S3 (or create one if missing)

clients/python/tests/conftest.py
tarilabs: this is a great introduction to the way we should handle this in pytest, thanks @syntaxsdev

tarilabs (commented):
I would like to make sure the end results works as expected for KServe (not necessarily as a test in the PR itself, making the check manually can also do)

otherwise 
/lgtm
already thanks @syntaxsdev

tarilabs:
( I also believe this PR need rebasing to account for more recent openapi and generated changes )

clients/python/src/model_registry/_client.py
syntaxsdev: I'm assuming this is what you meant:
image

If so, I have a fix coming up. The S3 upload did however save it like such.

tarilabs: see for example this tutorial leveraging KServe: https://rh-aiservices-bu.github.io/fraud-detection/fraud-detection-workshop/deploying-a-model-single-model-server.html#:~:text=Type%20the%20path%20that%20leads%20to%20the%20version%20folder%20that%20contains%20your%20model%20file%3A%20models/fraud

it points to the directory in the bucket containing the ML model file(s)

syntaxsdev: the concerns *should* be addressed in latest commit that allows an s3_prefix due that kserve might not work with models directly to bucket root. I will also test with fraud detection demo to verify

tarilabs: thanks @syntaxsdev much appreciated

syntaxsdev:
Confirmed functionality works similarly by going through other kubeflow saving model demos

image

scripts/deploy_minio_on_kind.sh
```diff
@@ -0,0 +1,56 @@
...
+if [[ $(kubectl get namespaces || false) =~ $MINIO_NAMESPACE ]]; then
+    echo 'Namespace already exists, skipping creation'
+else
+    kubectl create namespace "$MINIO_NAMESPACE"
+fi
+
+kubectl apply -f $DIR/manifests/minio/deployment.yaml -n $MINIO_NAMESPACE
+kubectl wait --for=condition=available deployment/minio -n $MINIO_NAMESPACE --timeout=2m
+kubectl apply -f $DIR/manifests/minio/create_bucket.yaml -n $MINIO_NAMESPACE
+kubectl wait --for=condition=complete job/minio-init -n $MINIO_NAMESPACE --timeout=2m
```
tarilabs, line 40: my recommendation would be to wrap these waits with something like

https://github.com/kubeflow/model-registry/blob/7eb57c769098798359d9fb7ef96292643e95be02/scripts/deploy_on_kind.sh#L42-L47

this way if one day a failure arises on GHA we have some helpful hints, wdyt?

scripts/deploy_minio_on_kind.sh
```diff
@@ -0,0 +1,56 @@
...
+KF_MR_TEST_ACCESS_KEY_ID=$(kubectl get secret minio-secret -n minio -o jsonpath="{.data.ACCESS_KEY_ID}" | base64 --decode)
+KF_MR_TEST_SECRET_ACCESS_KEY=$(kubectl get secret minio-secret -n minio -o jsonpath="{.data.SECRET_KEY}" | base64 --decode)
+
+
+if [[ -z "$KF_MR_TEST_ACCESS_KEY_ID" || -z "$KF_MR_TEST_SECRET_ACCESS_KEY" ]]; then
+    echo "Error: Failed to retrieve MinIO credentials. Exiting."
+    exit 1
+fi
+
+cat <<EOF > $DIR/manifests/minio/.env
```
tarilabs, line 51: do you want to add this to the `.gitignore` just for safety to avoid undesired commits by other contributors?

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +442,228 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+            path: The path to where the models or artifacts are.
+            path_prefix: The folder prefix to store under the root of the bucket.
+            bucket: The name of the S3 bucket.
+            s3: The S3 Client object.
+
+        Keyword Args:
+            endpoint_url: The endpoint url for the S3 bucket.
+            region: The region name for the S3 bucket.
+
+        Returns: Int of how files were uploaded.
```
tarilabs, line 522: not sure if `Int` is a typo

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +442,228 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+            endpoint_url, access_key_id, secret_access_key, region
+        )
+
+        s3 = self.__connect_to_s3(
+            endpoint_url=endpoint_url,
+            access_key_id=access_key_id,
+            secret_access_key=secret_access_key,
+        )
+        try:
+            return self.__upload_to_s3(
```
tarilabs, line 489: just to double-check: in the case of a folder, this will return an array list, and we will need to determine from the other PR here:
- https://github.com/kubeflow/model-registry/pull/761
the _single_ URI to register in MR, to be later used to populate the Isvc.
Correct?
cc @Crazyglue

clients/python/Makefile
```diff
@@ -18,9 +18,15 @@ deploy-latest-mr:
...
+.PHONY: deploy-test-minio
+deploy-test-minio:
+	cd ../../ && ./scripts/deploy_minio_on_kind.sh
+	kubectl port-forward -n minio svc/minio 9000:9000 &
+	
 .PHONY: test-e2e
-test-e2e: deploy-latest-mr
-	poetry run pytest --e2e -s -rA
+test-e2e: deploy-latest-mr deploy-test-minio
+	bash -c 'set -a && source ../../scripts/manifests/minio/.env && set +a && poetry run pytest --e2e -s -rA'
```
tarilabs, line 28: is there a way to avoid relying exclusively on bash, which may not be the default shell on some system?

tarilabs (commented):
minor comments + question, otherwise

/lgtm

as a first iteration, thanks a lot @syntaxsdev

scripts/deploy_minio_on_kind.sh
syntaxsdev: This is deleted here after the tests are ran.

https://github.com/kubeflow/model-registry/blob/677688c5e388c9c92e42b9f8f2d198e78e7e7404/clients/python/Makefile#L29C6-L29C43

However in the case of a unsuccessful test, there could be a chance someone does have the .env but I don't know how prominent that would be. should we add?

scripts/deploy_minio_on_kind.sh
syntaxsdev: Good idea

clients/python/Makefile
syntaxsdev: Initially I had this solution:
https://github.com/kubeflow/model-registry/blob/b7f28fb034495302e1e56ea5944abbc75ca46d4e/clients/python/Makefile#L28

but it did not seem to work, because `source` does not work in Makefile's shell instances because source only works in interactive shells (afaik). wdyt we can do?

dhirajsb: You should be able to use Makefile [include](https://www.gnu.org/software/make/manual/html_node/Include.html) to pull in a file with environment variables.

scripts/deploy_minio_on_kind.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+
+set -e
```
rareddy, line 3: @tarilabs @syntaxsdev one question I have is that I was assuming we are using pytest libraries for any cluster management too, is that a wrong assumption?

tarilabs: no we are not "orchestrating cluster setup from pytest" if that was the question, as the request was to stick to makefile and scripting; the downside is netting on a basic common ground, but I see the advantage as reuse from Go or else in the future if needed.

rareddy: Correct for reuse and manageability, this was the topic in shift left.  @lugi0 might have some comments on this. I am not asking to change this PR but maybe we can capture a follow-up if that is the overall direction we need to be writing the tests with, if not we can ignore.

tarilabs: I'm favourable of revising the setup, as you know I prsonally prefer to reduce Makefile and scripts to the minimum as possible, but we need indeed to account for multiple factors.

rareddy:
@lucferbux can you look into the issues with the UI tests here?

lucferbux:
Agh, this might happen due to latest config PR cc @christianvogt I'm gonna take a look

christianvogt:
A fix for the tests has been posted here https://github.com/kubeflow/model-registry/pull/832

clients/python/Makefile
```diff
@@ -24,8 +24,9 @@ deploy-test-minio:
 	kubectl port-forward -n minio svc/minio 9000:9000 &
 	
 .PHONY: test-e2e
-test-e2e: deploy-latest-mr deploy-test-minio
-	bash -c 'set -a && source ../../scripts/manifests/minio/.env && set +a && poetry run pytest --e2e -s -rA'
+test-e2e: deploy-latest-mr deploy-local-registry deploy-test-minio
+	@set -a; . ../../scripts/manifests/minio/.env; set +a; \
```
dhirajsb, line 28: @tarilabs can you verify whether this will work in non linux envs too?

dhirajsb (commented):
API looks good to me, we should verify Makefile changes on all platforms.

clients/python/Makefile
tarilabs: it does 
image

clients/python/src/model_registry/_client.py
```diff
@@ -437,3 +442,228 @@ def rm_versions(options: ListOptions) -> list[ModelVersion]:
...
+        """Saves a model to an S3 compatible storage.
+
+        Args:
+            path: Location to where the model(s) or artifact(s) are located. \
+                Can recursively upload nested files in folders.
+            bucket_name: The bucket to use for the S3 compatible object storage.
+
+        Keyword Args:
+            s3_prefix: The path to prefix with under root of bucket. If unset, will default to `path` if \
+                `path` is not a nested folder.
```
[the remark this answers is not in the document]
tarilabs: done with 9883236

tarilabs (approved):
thank you @syntaxsdev , and all for this PR, this is a good foundation step to make life easier to the DS + MLOps Eng to orchestrate the Store-then-Register flow! 

/lgtm
/approve

PR #767, opened 2025-02-06
change uname check in scripts/install_protoc.sh for compatibility with macos

lugi0:
## Description

The current version of the install_protoc.sh fails when running `make build` on MacOS due to a "bad substitution" error
```
./scripts/install_protoc.sh: line 6: ${OSTYPE,,}: bad substitution
make: *** [bin/protoc] Error 1
```
The change gets the value of `$OSTYPE` directly.

Furthermore, `uname -s` returns `Darwin`, which does not match against `=~ darwin`, causing the linux executable to be downloaded instead. 
The change looks for both capitalized and non-capitalized versions of the name (`[Dd]arwin`).

The proposed changes have been tested both in MacOS as well as Fedora

## How Has This Been Tested?

ran `make build`, correct version of protoc downloaded to bin folder

## Merge criteria:

If you have UI changes

### DIFF
--- a/scripts/install_protoc.sh
+++ b/scripts/install_protoc.sh
@@ -3,7 +3,7 @@ set -euxo pipefail
 
 OSTYPE=$(uname -s)
 OS="linux"
-if [[ ${OSTYPE,,} =~ darwin ]]; then
+if [[ $OSTYPE =~ [Dd]arwin ]]; then
     OS="osx"
 fi
 


tarilabs:
> Furthermore, `uname -s` returns `Darwin`, which does not match against `=~ darwin`, causing the linux executable to be downloaded instead.

I believe the check is intended to make lowercase the return value from uname -s

See `${parameter,,pattern}` from:
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion:~:text=%24%7Bparameter%2C%2Cpattern%7D

so regardless if it's coming from uname as `Darwin`, `darwin`, `dArWiN`, etc it was intended to be lowercase'd and checked to be the literal `darwin` (use of regexp for future extension)

I believe however current PR proposal makes it less Bash-centric and more explicit, so +1

Thanks @lugi0

tarilabs (approved):
per comments

/lgtm
/approve

PR #768, opened 2025-02-06
feat(manifests): add securityContext to deployments

### LINKED ISSUE #760: Add security context to model-registry manifests
Hello, you are missing a security context for most of your deployments. Please add

https://github.com/kubeflow/manifests/blob/550e0b3511bfcd4001bdab180ec1904bd06326a2/common/dex/base/deployment.yaml#L25-L32

to all of your deployments.

_Originally posted by @juliusvonkohout in https://github.com/kubeflow/manifests/pull/2973#pullrequestreview-2592150562_

pboyd:
## Description

Set `seccompProfile`, forbid containers to run as root, and disable
unnecessary system calls. This applies to:

- Model registry itself
- Example database (MySQL and PostgreSQL)
- Model registry UI

Fixes #760 

## How Has This Been Tested?
Applying the manifests in a local cluster.

## Merge criteria:

If you have UI changes

### DIFF
--- a/manifests/kustomize/overlays/postgres/kustomization.yaml
+++ b/manifests/kustomize/overlays/postgres/kustomization.yaml
@@ -39,7 +39,7 @@ vars:
 - name: POSTGRES_PORT
   objref:
     kind: ConfigMap
-    name: model-registry-db-parameters
+    name: metadata-postgres-db-parameters
     apiVersion: v1
   fieldref:
     fieldpath: data.POSTGRES_PORT
--- a/manifests/kustomize/overlays/postgres/kustomization.yaml
+++ b/manifests/kustomize/overlays/postgres/kustomization.yaml
@@ -13,11 +13,11 @@ patchesStrategicMerge:
 - patches/model-registry-deployment.yaml
 
 configMapGenerator:
-- name: metadata-postgres-db-parameters
+- name: metadata-registry-db-parameters
   envs:
   - params.env
 secretGenerator:
-- name: metadata-postgres-db-secrets
+- name: metadata-registry-db-secrets
   envs:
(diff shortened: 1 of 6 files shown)


manifests/kustomize/options/ui/overlays/standalone/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+namespace: kubeflow
```
lucferbux, line 3: Why adding namespace here? that would create an opinionated installation right? are we ok with having `kubeflow` as the default upstream?

lucferbux (changes requested):

manifests/kustomize/overlays/postgres/kustomization.yaml
```diff
@@ -39,7 +39,7 @@ vars:
 - name: POSTGRES_PORT
   objref:
     kind: ConfigMap
-    name: model-registry-db-parameters
+    name: metadata-postgres-db-parameters
```
dhirajsb, line 42:
```suggestion
    name: metadata-registry-db-parameters
```
Let's keep it agnostic of the DB type if we can since we can't use multiple DBs and types at the same time

dhirajsb (changes requested):

manifests/kustomize/overlays/postgres/kustomization.yaml
pboyd: I pushed another commit to rename the configmap and secret.

tarilabs (approved):
thank you @pboyd and all

/approve

PR #769, opened 2025-02-06
fix: high cpu load when idle

pboyd:
## Description

Model Registry was using a full CPU when idle. I traced it back to some
recent error handling code that had a select with a normally closed
channel.

The problem is that `errMLMDChan` closes quickly and afterwards the `select`
never blocks. Since it's in a loop, it's constantly re-running the select. I
don't think it affected the behavior, but it does hit the CPU pretty hard (to
say nothing of my laptop battery, which is really why I noticed).

## How Has This Been Tested?

Sanity check on model registry behavior.

## Merge criteria:

If you have UI changes

### DIFF
--- a/cmd/proxy.go
+++ b/cmd/proxy.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


Al-Pragliola (approved):
nice catch @pboyd !
I tested it locally and the behavior is as expected, also I confirm that the cpu load issue is fixed 💪🏼 

before
image

after
image


/lgtm
/approve

tarilabs (approved):
Thank you for the quick turnaround @pboyd @Al-Pragliola 

/lgtm
/approve

PR #771, opened 2025-02-07
feat(gha): added gha to check if python client autogen code is in sync

Al-Pragliola:
## Description

In this PR, I added a GHA to check if the autogenerated code from the python client is up to date with the changes made to the openapi schema, so that we don't have divergences when trying to build the python client locally, as already discussed with @tarilabs and @lugi0. 

example of failed run: https://github.com/kubeflow/model-registry/actions/runs/13198154981/job/36843945528

## How Has This Been Tested?

on the CI

## Merge criteria:


lugi0 (approved):
LGTM!

.github/workflows/python-check-autogen.yml
```diff
@@ -0,0 +1,60 @@
...
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/dependabot.yml"
+      - "docs/**"
+  pull_request:
+    paths:
+      - "clients/python/src/mr_openapi/**"
+      - "api/openapi/model-registry.yaml"
+
+jobs:
+  check-autogen:
```
tarilabs, line 20: I would recommend moving this job here:

https://github.com/kubeflow/model-registry/blob/366243346e606e3104586d81f0f54880d9787e32/.github/workflows/python-tests.yml#L17

wdyt?

tarilabs (commented):
thanks @Al-Pragliola , why not adding it to the existing suite of test jobs to avoid proliferation of workflow files?

.github/workflows/python-check-autogen.yml
Al-Pragliola: Thanks, it makes perfect sense to include this in the Python testing workflow 🙏🏼

tarilabs:
thank @Al-Pragliola !

/lgtm
/approve

PR #797, opened 2025-02-12
feat(openapi): make items a required field in List requests

Al-Pragliola:
## Description

Slight improvement in declaring the items field in the list request as required, so that the autogen does not set it to omitempty, which causes json umarshal to remove the field.

## How Has This Been Tested?

local testing

## Merge criteria:

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -1225,6 +1225,8 @@ components:
               type: array
               items:
                 $ref: "#/components/schemas/ModelVersion"
+          required:
+            - items
         - $ref: "#/components/schemas/BaseResourceList"
     ModelArtifactList:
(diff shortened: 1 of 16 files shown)


tarilabs (commented):
thanks @Al-Pragliola overall LGTM but I believe while we're at it we should do this as well for `InferenceServiceList`, wdyt?

Al-Pragliola:
> thanks @Al-Pragliola overall LGTM but I believe while we're at it we should do this as well for `InferenceServiceList`, wdyt?

🙏🏼  nice catch, I have added also InferenceServiceList in the latest commit

tarilabs (approved):
Thank you again @Al-Pragliola 

/lgtm
/approve

PR #799, opened 2025-02-13
Updates model registry client lib

alexcreasy:
## Description

Updates model registry client lib to fix non-existant property when array is empty

## Merge criteria:

If you have UI changes


clients/ui/bff/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/brianvoe/gofakeit/v7 v7.1.2
 	github.com/google/uuid v1.6.0
 	github.com/julienschmidt/httprouter v1.3.0
-	github.com/kubeflow/model-registry v0.2.9
+	github.com/kubeflow/model-registry v0.2.7-alpha.0.20250213094235-2b1b4d6325bf
```
lucferbux, line 9: Just one quick question, cause I lack all the context of the async conversation, why are we reverting to this version?

ederign: @alexcreasy @lucferbux if we are going to wait the release to merge this, we can use the next model registry version isn't?

ederign: @alexcreasy do you think we can point now for the recent release and avoid the hash here?

ederign (changes requested):
Just a quick question about version bump

clients/ui/bff/go.mod
alexcreasy: @lucferbux @ederign this wasn't a revert (I had to ask the same question, apparently main was still on 0.2.7-alpha, hence why getting the bleeding edge looks like a revert). As what we need has now been released there's no need to have a temporary commit using the hash. I'll update that now!

alexcreasy:
@ederign @lucferbux this one should be ready to go now!

ederign (approved):

PR #48, opened 2024-03-20
docs: initial proposal for OCI artifact registry

rchincha:
Partially addresses https://github.com/kubeflow/community/pull/682

## Merge criteria:

- [ ] The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #800, opened 2025-02-13
Add save_to_oci_registry python client method

Crazyglue:
## Description

This PR aims to add a helper method to the python client to append a set of files to a given oci-compatible container image. In the future this method will be used as a part of the PR found here: https://github.com/kubeflow/model-registry/pull/761 which aims to provide yet another helper method that will both upload and register a model to the MR.

With this PR and the one mentioned above, the use case for the client (after training) is to simply call 1 method to both push the newly trained model to a storage location AND register that newly trained model in the MR. Removing a number of manual steps that would have to be done separately as a part of a standard DS flow.

## How Has This Been Tested?

Added an E2E test and validated that they run. The E2E test uses a local registry to upload test-images to.

The local environment will need to at a minimum install skopeo on the system (for MacOS and Brew, this means running `brew install skopeo`).

## Merge criteria:

If you have UI changes


scripts/deploy_local_kind_registry.sh
Crazyglue, line 1: This was added as a way to have the e2e tests include image download/upload. A registry is needed to perform the upload step, and so a local registry should be good enough to perform this test.

clients/python/src/model_registry/utils.py
```diff
@@ -90,3 +92,69 @@ def s3_uri_from(
...
+        skopeo_push(dest_dir, oci_ref)
+        
+    elif backend == 'oras':
+        from olot.backend.oras_cp import is_oras, oras_pull, oras_push
+        if not is_oras():
+            raise ValueError('oras is selected, but it is not present on the machine. Please validate the oras cli is installed and available in the PATH')
+    
+        oras_pull(base_image, local_image_path)
+        oci_layers_on_top(local_image_path, model_files, modelcard)
+        oras_push(local_image_path, oci_ref)
```
Crazyglue, line 156: I chose not to abstract this away into an interface/facade because the steps are very straightforward, and i think it would over-engineer what is effectively calling 3 separate functions.

If `olot` supports more backends, this will probably need to be refactored to include an interface of some kind.

clients/python/pyproject.toml
```diff
@@ -27,9 +27,11 @@ nest-asyncio = "^1.6.0"
 eval-type-backport = "^0.2.0"
 
 huggingface-hub = { version = ">=0.20.1,<0.29.0", optional = true }
+olot = { version = "^0.1.2", optional = true }
```
Crazyglue, line 30: This is following the same pattern by `huggingface-hub` above. This particular dependency is optional because if you choose not to use this new "save" method, you would never need this dependency. The README was updated to reflect this

tarilabs: +1

clients/python/src/model_registry/utils.py
```diff
@@ -90,3 +91,71 @@ def s3_uri_from(
     # https://alexwlchan.net/2020/s3-keys-are-not-file-paths/ nor do they resolve to valid URls
     # FIXME: is this safe?
     return f"s3://{bucket}/{path}?endpoint={endpoint}&defaultRegion={region}"
+
+def save_to_oci_registry(
+        base_image: str,
+        dest_dir: str | os.PathLike,
```
tarilabs, line 97: I would actually make this a default to a temp dir, wdyt?

clients/python/src/model_registry/utils.py
```diff
@@ -90,3 +91,71 @@ def s3_uri_from(
...
+            msg = "oras is selected, but it is not present on the machine. Please validate the oras cli is installed and available in the PATH"
+            raise ValueError(msg)
+
+        oras_pull(base_image, local_image_path)
+        oci_layers_on_top(local_image_path, model_files, modelcard)
+        oras_push(local_image_path, oci_ref)
+
+    else:
+        msg = f"Invalid backend chosen: '{backend}'"
+        raise StoreError(msg)
```
tarilabs, line 161: overall I believe fine for now, but I would consider having a dict where we can plug this "providers" maybe instread of just failing?

clients/python/tests/test_utils.py
```diff
@@ -71,3 +72,21 @@ def test_s3_uri_builder_with_complete_env():
     os.environ["AWS_S3_ENDPOINT"] = "test-endpoint"
     os.environ["AWS_DEFAULT_REGION"] = "test-region"
     assert s3_uri_from("test-path") == s3_uri_from("test-path", "test-bucket")
+
+@pytest.mark.e2e(type="oci")
+def test_save_to_oci_registry_with_skopeo():
+    # TODO: We need a good source registry which is oci-compliant and very small in size
```
tarilabs, line 78:
```suggestion
    # TODO: We need a good base image which is oci-compliant and very small in size
```

scripts/deploy_on_kind.sh
```diff
@@ -10,7 +10,7 @@ source ./${DIR}/utils.sh
 
 # modularity to allow re-use this script against a remote k8s cluster
 if [[ -n "$LOCAL" ]]; then
-    CLUSTER_NAME="${CLUSTER_NAME:-kind}"
+    CLUSTER_NAME="${CLUSTER_NAME:-mr-e2e}"
```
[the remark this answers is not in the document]
tarilabs: @Al-Pragliola wdyt? lgtm but I don't recall if we rely on this name elsewhere?

scripts/deploy_local_kind_registry.sh
```diff
@@ -0,0 +1,13 @@
...
+# Some of this copy-pasted from https://kind.sigs.k8s.io/docs/user/local-registry/
+
+# 1. Create registry container unless it already exists
+reg_name='local-mr-registry'
+reg_port='5001'
+if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
+  docker run \
+    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --network bridge --name "${reg_name}" \
+    registry:2
+fi
```
tarilabs, line 13: interesting approach, thank you for sharing; which pro/cons if using again distribution registry but as a bespoke deployment in olot?

clients/python/Makefile
```diff
@@ -18,8 +18,12 @@ deploy-latest-mr:
 	cd ../../ && IMG_VERSION=${IMG_VERSION} make image/build && LOCAL=1 ./scripts/deploy_on_kind.sh
 	kubectl port-forward -n kubeflow services/model-registry-service 8080:8080 &
 
+.PHONY: deploy-local-registry
+deploy-local-registry:
+	cd ../../ && ./scripts/deploy_local_kind_registry.sh
+
 .PHONY: test-e2e
-test-e2e: deploy-latest-mr
+test-e2e: deploy-latest-mr deploy-local-registry
```
tarilabs, line 26: can you kindly doublecheck this is consistent with the approach from https://github.com/kubeflow/model-registry/pull/765/files#diff-e01c397ab995b0ee8633b9be78064e1cff0b691f10166df64614dc0bb2b5395eR27 please?

tarilabs (commented):
some early comment, thank you @Crazyglue

clients/python/src/model_registry/utils.py
tarilabs: yeah, but as in https://github.com/kubeflow/model-registry/pull/800#discussion_r1956267679
I would consider passing to a monkeypatched method all params, so that at least we steer users to a provider mechanism of sort

scripts/deploy_local_kind_registry.sh
tarilabs: btw, not sure if the e2e failure is related: 

```
----------------------------- Captured stderr call -----------------------------
time="2025-02-14T14:55:54Z" level=fatal msg="copying image 1/2 from manifest list: trying to reuse blob sha256:1933e30a3373776d5c7155591a6dacbc205cf6a2665b6dced682c6d2ea7b000f at destination: pinging container registry localhost:5001: Get \"http://localhost:5001/v2/\": dial tcp [::1]:5001: connect: connection refused"
```

Crazyglue: Yes that failure could be related, I'll look into it. 

>which pro/cons if using again distribution registry but as a bespoke deployment in olot?

Could you elaborate on this? I'm not quite following what you mean

tarilabs: > Could you elaborate on this? I'm not quite following what you mean

in olot/omlmd, we use a CNCF Distribution Registry deployed as a K8s workload.

- https://github.com/containers/olot/blob/main/e2e/deploy_distribution_registry.sh
- https://github.com/containers/omlmd/blob/a1b2030ed39355dfa6ff93cdc118c47f1e210b70/.github/workflows/e2e.yaml#L10-L33

here it seems to me you're running it "barebone" outside the cluster (so not sure which defaults are applied).

clients/python/src/model_registry/utils.py
Crazyglue: What would make sense as a temp dir? just `/tmp`? or somethign autogenerated in the current `cwd`?

scripts/deploy_local_kind_registry.sh
Crazyglue: Understood! I decided to run it outside of the kind cluster just for simplicity, but I can take a look at how its done in olot and match it here. That will provide some consistency

clients/python/src/model_registry/utils.py
syntaxsdev: you can use `tempfile.mkdtemp()` if you chose

scripts/deploy_on_kind.sh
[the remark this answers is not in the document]
syntaxsdev: I also had this issue a few weeks ago with assuming the local kind or rosa is immediately the name to use

dhirajsb (approved):
/lgtm
reviewed at a high level, and I don't see any red flags
@tarilabs will have to approve to merge this initial version

scripts/deploy_on_kind.sh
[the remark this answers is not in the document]
Al-Pragliola: I don't think so, looks also good to me

tarilabs (approved):
thank you @Crazyglue this will make way easier for users to leverage this MR py client to orchestrate/coordinate the Store-then-Register flow for Model Registry, in this case by storing it to a OCI registry a ModelCar, which partially address our needs but as also highlighted in 
- https://github.com/kubeflow/model-registry/pull/48

as more details will emerge for the production of the ModelCar, we can incorporate in subsequent PRs, but this PR is already a good foundational element [along with this other PR](https://github.com/kubeflow/model-registry/pull/765), to the work for:
- https://github.com/kubeflow/model-registry/pull/761

/lgtm
/approve

Issue #774, opened 2025-02-07
Usage of nested-asyncio in the Python client


blublinsky:
**Describe the bug**
The current implementation of the model registry client uses nested async ([registry/blob/main/clients/python/src/model_registry/_client.py#L130](https://github.com/kubeflow/model-registry/blob/main/clients/python/src/model_registry/_client.py#L130), which works fine for standalone client usage, but fails when running inside Ray or any other framework using uviloop.

**To Reproduce**
Here is a simple failing Ray sample:
```
import ray
import asyncio
from model_registry import ModelRegistry

runtime_env = {
    "pip": ["model-registry==0.2.10"],
}


@ray.remote(runtime_env=runtime_env)
class ModelRegistryClient:
    def __init__(
            self,
            url: str="http://localhost",
            port: int=8081,
            user: str="boris",
            secure: bool=False
    ):
        self.registry = ModelRegistry(
            server_address=url, port=port, author=user, is_secure=secure
        )

    async def list(self) -> tuple[list, list]:
        lh_list = []
        hf_list = []
        rm_iter = self.registry.get_registered_models().page_size(20)
        for rm in rm_iter:
            versions_iter = (
                self.registry.get_model_versions(rm.name).page_size(20).order_by_id().descending()
            )
            # we use the latest version
            rv = next(versions_iter)
            try:
                origin = rv.custom_properties["model_origin"]
            except:
                print(f"Custom properties {rv.custom_properties} does not have model_origin")
                continue
            if origin == "lakehouse":
                # lakehouse
                lh_list.append(
                    {
                        "name": rm.name,
                        "env": rv.custom_properties["env"],
                        "namespace": rv.custom_properties["namespace"],
                        "revision": rv.custom_properties["revision"],
                    }
                )
            else:
                # hugging face
                hf_list.append({"name": rm.name})
        return lh_list, hf_list

async def main():
    rg = ModelRegistryClient.remote()
    lh, hf = await rg.list.remote()
    print(lh)
    print(hf)

if __name__ == "__main__":
    asyncio.run(main())
```

**Expected behavior**
This should of produce:

``` 
2025-02-09 19:28:26,810	INFO worker.py:1777 -- Started a local Ray instance. View the dashboard at 127.0.0.1:8265 
(ModelRegistryClient pid=4232) /Users/borisl/Projects/vllm_actuator/model_registry/ray_version.py:19: UserWarning: User access token is missing
[...]

PR #802, opened 2025-02-14
removing nested-asyncio from Model Registry client

blublinsky:
## Description
Replaced async-runner by a new implementation to avoid usage of nested-async to address https://github.com/kubeflow/model-registry/issues/774

## How Has This Been Tested?
Unit tests and running in our Ray environment

## Merge criteria:

- [ x] The commits have meaningful messages; the author will squash them after approval or in case of manual merges will ask to merge with squash.

If you have UI changes


clients/python/src/model_registry/_async_task_runner.py
```diff
@@ -0,0 +1,102 @@
+# from https://gist.github.com/blink1073/969aeba85f32c285235750626f2eadd8
```
tarilabs, line 1: We should verify if alternative libraries exists on pypi to avoid having to introduce multiple licensing in the project itseld

blublinsky: @tarilabs, We had the same need in our internal project, and this was the only solution that I found after more than a week of searches and conversations with my colleagues. If you can find a better solution for this I will be very interested to hear it

tarilabs: Sure, but maybe there are ways we could allow user to plug their own executor/runner strategies, without requiring necessarily to opt for one or the other ourselves, maybe picking one

(this is because seems to me at the end of the day it's the same signature/contract)

blublinsky: Do you mean introducing a base class and using AsyncTaskRunner as a default? To make ModelRegistry independent of it we will also need a factory class, doing instantiation.

So add 2 additional classes. So yes, this can be done. Is it worth it?
If the answer is yes, we can do it

syntaxsdev:
@blublinsky Trying to replicate your issue - could you provide more context? https://github.com/kubeflow/model-registry/issues/774#issuecomment-2695246942

blublinsky:
There is an example here https://github.com/kubeflow/model-registry/issues/774

syntaxsdev:
> There is an example here #774

That is the example I used. Did you do anything else besides run in a standalone script?

blublinsky:
What can I say, it does not work in Ray

syntaxsdev:
@blublinsky thanks for submitting this. I created a PR around yours that reimplements an async task runner, but abstracts it away from the core part of the code - as it seems to only be an issue with Uvloop event loops rather than Asyncio event loops - which a majority of runners use.

It also avoids incorporating that license of the AsyncTaskRunner into the ModelRegistry.

I tested with Ray and Uvloop and also added tests cases for it.

https://github.com/kubeflow/model-registry/pull/849

Could you verify that this solution also works for you?

tarilabs:
we can close this PR, based on 
- https://github.com/kubeflow/model-registry/issues/774#issuecomment-2705799379

PR #820, opened 2025-02-18
Run container image build of UI on PR

alexcreasy:
## Description

Runs an image build of the UI container on PR. This would have caught a recent case where a merged PR broke the container image build.
 
## How Has This Been Tested?

- See if CI job passes!

## Merge criteria:

If you have UI changes

### DIFF
--- a/.github/workflows/build-image-ui-pr.yml
+++ b/.github/workflows/build-image-ui-pr.yml
@@ -1,0 +1,25 @@
+name: Test UI container image build and deployment
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "clients/ui/**"
+      - "!LICENSE*"
+      - "!DOCKERFILE*"
+      - "!**.gitignore"
+      - "!**.md"
+      - "!**.txt"
+env:
+  IMG_ORG: kubeflow
+  IMG_REPO: model-registry-ui
+  PUSH_IMAGE: false
+  BRANCH: ${{ github.base_ref }}
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    steps:
+      # checkout branch
+      - uses: actions/checkout@v4
+      - name: Build UI Image
+        shell: bash
+        run: ./scripts/build_deploy.sh


.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+      # checkout branch
+      - uses: actions/checkout@v4
+      # set image version
+      - name: Set main-branch environment
+        if: env.BUILD_CONTEXT == 'main'
+        run: |
+          commit_sha=${{ github.event.after }}
+          tag=main-${commit_sha:0:7}
+          echo "VERSION=${tag}" >> $GITHUB_ENV
+      - name: Set tag environment
```
lucferbux, line 40: I don't think we need this for tags, since it's not gonna be for releases I think

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+        if: env.BUILD_CONTEXT == 'tag'
+        run: |
+          echo "VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
+      - name: Build and Push UI Image
```
lucferbux, line 44: We should disable the push here right?

lucferbux (changes requested):

.github/workflows/build-image-ui-pr.yml
alexcreasy: I thought it was disabled by setting PUSH to false in the env block... I'll update the name though and double check the logic!

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      # Assign context variable for various action contexts (tag, main, CI)
+      - name: Assigning tag context
+        if: github.head_ref == '' && startsWith(github.ref, 'refs/tags/v')
+        run: echo "BUILD_CONTEXT=tag" >> $GITHUB_ENV
+      # Assign context variable for various action contexts (main, CI)
+      - name: Assigning main context
+        if: github.head_ref == '' && github.ref == 'refs/heads/main'
+        run: echo "BUILD_CONTEXT=main" >> $GITHUB_ENV
```
lucferbux, line 30: To be honest I dont think this si needed since the primary goal is only to call `build_deploy.sh`, and hence we don't need to know the tag itself since we are not pushing it, I would just remove this.

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+      # checkout branch
+      - uses: actions/checkout@v4
+      # set image version
+      - name: Set main-branch environment
+        if: env.BUILD_CONTEXT == 'main'
+        run: |
+          commit_sha=${{ github.event.after }}
+          tag=main-${commit_sha:0:7}
+          echo "VERSION=${tag}" >> $GITHUB_ENV
```
lucferbux, line 39: Same thing with this, we can harcode the `VERSION` variable and just call it a day.

lucferbux (changes requested):

tarilabs:
did you wanted to merge this one?

alexcreasy:
@tarilabs I think so, I'm just going to make the changes Lucas suggested

.github/workflows/build-image-ui-pr.yml
alexcreasy: Actually, I don't think we even need the VERSION variable as the script sets a default based on the commit hash if one isn't provided.

alexcreasy:
@lucferbux changes made, would you mind re-reviewing so @tarilabs can merge if you're happy?

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,27 @@
+name: Test UI container image build and deployment
+on:
+  workflow_dispatch:
+  pull_request:
+    paths-ignore:
```
lucferbux, line 5: agh, one last change, I'm really sorry @alexcreasy I didn't ping you there, we had a conversation in which we were blocking other branches if we did messed up the UI, so we've agreed on just running ui builds when changing something inside `clients/ui`, can you change it like this?: https://github.com/kubeflow/model-registry/pull/833/files#diff-5320b64457303894487247143c3a15570f6dca28abcb74fc4989e511c5b6aacbR6

lucferbux (changes requested):

.github/workflows/build-image-ui-pr.yml
alexcreasy: No worries, I've updated to use the agreed upon paths!

tarilabs:
thanks all, should I push the button? 🙃

alexcreasy:
@tarilabs yeah if you wouldn't mind, I've fixed everything Lucas asked for... if there are any other tweaks we can make them in another PR :D

lucferbux (approved):
/lgtm

@rareddy @tarilabs can you guys take a look at this please?

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,25 @@
...
+  BRANCH: ${{ github.base_ref }}
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    steps:
+      # checkout branch
+      - uses: actions/checkout@v4
+      - name: Build UI Image
+        shell: bash
+        run: ./scripts/build_deploy.sh
```
tarilabs, line 25: I'm a bit confused, the PR body specs:

> Runs an image build of the UI container on PR.

this invokes `./scripts/build_deploy.sh`

which invokes:

https://github.com/kubeflow/model-registry/blob/c27acc22d4ab77e38b0bf5495f09b1dba4930624/scripts/build_deploy.sh#L49-L57

which invokes:

https://github.com/kubeflow/model-registry/blob/c27acc22d4ab77e38b0bf5495f09b1dba4930624/Makefile#L263-L266

which is

https://github.com/kubeflow/model-registry/blob/main/Dockerfile

which is the Model Registry Go REST server.

What am I missing, please?

tarilabs (commented):
I have a doubt, below @alexcreasy @lucferbux

.github/workflows/build-image-ui-pr.yml
tarilabs: bump @alexcreasy @lucferbux

lucferbux: Oh wow, I'm sorry @tarilabs good catch, I swear at some point it had `IMG_REPO: model-registry` and I blanked out readiing that this time.
Yes, @alexcreasy this should be `model-registry-ui`, if not, you are building model registry, not the UI.
Great catch @tarilabs

tarilabs:
relies on

.github/workflows/build-image-ui-pr.yml
```diff
@@ -0,0 +1,25 @@
...
+    paths:
+      - "clients/ui/**"
+      - "!LICENSE*"
+      - "!DOCKERFILE*"
+      - "!**.gitignore"
+      - "!**.md"
+      - "!**.txt"
+env:
+  IMG_ORG: kubeflow
+  IMG_REPO: model-registry-ui
```
tarilabs, line 14: relies on mechanism from :

https://github.com/kubeflow/model-registry/blob/21b578e1256a7d2a157d89fcd17b950a65a2e3f8/Makefile#L40-L41

tarilabs (commented):
/lgtm

@Al-Pragliola thoughts?

.github/workflows/build-image-ui-pr.yml
Al-Pragliola: thanks for highlighting this, I would have missed that easily 😁 

/lgtm

tarilabs: glad to help, I did also forgot. maybe we could consider ways to refactor this too going fwd, separately from this PR, wdyt?

tarilabs:
- https://github.com/kubeflow/model-registry/pull/820#discussion_r1993209050

thanks all again :) 

/approve

.github/workflows/build-image-ui-pr.yml
Al-Pragliola: I agree

PR #821, opened 2025-02-19
change resp status from 500 to 409 for dup POST reqs

adysenrothman:
## Description
Small change to return a 400 error if a user attempts to add a duplicate value. Previously, this was defaulting to a 500 internal service error which is not appropriate. 

## How Has This Been Tested?
https://issues.redhat.com/browse/RHOAIENG-19196
Locally reproduced this error to confirm that the resp status is 400 instead of 500.

## Merge criteria:

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -317,6 +317,8 @@ paths:
           $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
+        "409":
+          $ref: "#/components/responses/Conflict"
         "500":
           $ref: "#/components/responses/InternalServerError"
         "503":
@@ -443,6 +445,8 @@ paths:
           $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
+        "409":
+          $ref: "#/components/responses/Conflict"
         "500":
           $ref: "#/components/responses/InternalServerError"
         "503":
@@ -552,6 +556,8 @@ paths:
           $ref: "#/components/responses/Unauthorized"
         "404":
           $ref: "#/components/responses/NotFound"
+        "409":
+          $ref: "#/components/responses/Conflict"
         "500":
           $ref: "#/components/responses/InternalServerError"
         "503":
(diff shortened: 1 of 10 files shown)


tarilabs (commented):
thanks @adysenrothman for catching this! 🚀 

/lgtm

but I would encourage if possible to "pin this" with some non-regression testing.

I believe this should be doable somewhere in the Go code, if otherwise since this surfaces up to the REST client, maybe the ideal would be to write a non-regression [testing here](https://github.com/kubeflow/model-registry/blob/7eb57c769098798359d9fb7ef96292643e95be02/clients/python/tests/REST_bindings_test.py#L76-L78).

wdyt?

Al-Pragliola (approved):
Grats on your first contribution!

tested locally

main:
image

pr:
image

/lgtm

adysenrothman:
> If we want to be more precise we can opt to use the 409 http status code, wdyt? @tarilabs @adysenrothman

this is a good point, I will make that change. thanks for the feedback!

Al-Pragliola:
> > If we want to be more precise we can opt to use the 409 http status code, wdyt? @tarilabs @adysenrothman
> 
> this is a good point, I will make that change. thanks for the feedback!

You are welcome! Now that we have added a new status code in the response, we should also update the openapi schema with the new response for every request that supports 409 (e.g. https://github.com/adysenrothman/model-registry/blob/change-duplicate-resp-status/api/openapi/model-registry.yaml#L439-L449) and then run the autogenerator

adysenrothman:
@Al-Pragliola @tarilabs Thank you both for the feedback and recommendations! I have updated the code to reflect the 409 error package updates across all impacted requests and their related test packages. Please let me know if you have any other thoughts on this PR.

tarilabs:
thank you for acting on @Al-Pragliola 's comment and _also_ expanding the non-regression testing for more coverage of other logical model entities.

as Alessio mentioned, I believe we need to re-run the code generation for both [Go](https://github.com/kubeflow/model-registry/blob/aa18cce4b935fd3b2f27cb8827c77804fd755ea0/Makefile#L215C32-L215C50) and [Python](https://github.com/kubeflow/model-registry/blob/aa18cce4b935fd3b2f27cb8827c77804fd755ea0/clients/python/Makefile#L7), then it should be good to go

Al-Pragliola (approved):
/lgtm

tarilabs (approved):
thank you @adysenrothman and all 

/lgtm
/approve

PR #822, opened 2025-02-19
fix: Update openapi-server codegen to re-create manual changes

pboyd:
## Description

I needed to re-generate the openapi server code, but found that the generated
code had several differences from what's checked in. This overrides the
templates and codegen scripts so that we can re-generate it again.

## How Has This Been Tested?
Mostly comparing diffs between the generated code and the checked in code. Verified that everything still builds.

## Merge criteria:


tarilabs (commented):
this is definitely helpful, thank you @pboyd 

I have a question however: when the Go files were modified, was to support or circumvent or adapts to limitation in the original openapi-codegen (templates).

If we *now have adjusted templates, will imply any further modification need to be done in the template, or we modify the Go generated code and once happy we need to ask the committer to modify the template.

What is your perspective on the ease of use of this development flow?

edit: typo

tarilabs:
thank you for the updates @pboyd ; do I gather correctly, some of these changes could/should be proposed in the upstream ootb openapi-codegen templates? 🤔

rareddy:
+1 pushing the `cors` functionality upstream. But I would not hold this PR for that.

Al-Pragliola (commented):
Thanks for the work @pboyd, in my opinion the more we can automate without messing with the generated files the better and this PR is a good step forward in that direction, next I think we should see if there's a way to have a more "interface" approach that I have seen in other openapi gen products

/lgtm

tarilabs (approved):
thanks all

/approve

PR #725, opened 2025-01-22
Initial version of Catalog Service

dhirajsb:
Initial Version of Model Catalog Service

## Description

Add a new Model Catalog service to model registry API. The API shares types with Model Registry service
to maximize the reuse of common properties.

## How Has This Been Tested?

WIP

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-   off-commits)  (To pass the `DCO` check)

- [X] The commits have meaningful messages; the author will squash them after approval or in case of manual merges will  ask to merge with squash.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work.
- [X] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).

If you have UI changes

- [ ] The developer has added tests or explained why testing cannot be added.
- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

Issue #702, opened 2025-01-15
Add Model Catalog API to Model Registry


dhirajsb:
**Is your feature request related to a problem? Please describe.**
Several public ML and LLM model catalogs such as [Hugging Face](https://huggingface.co/) are now available with easily accessible opensource models. 
At present Kubeflow Model Registry has an API for registering and managing locally trained and published _Registered Models_. Also, users have to use a variety of different websites, UIs, or APIs to browse and discover foundation models in various catalogs and manually register them for deployment. 

**Describe the solution you'd like**
There is a need for a uniform and simple way to access various Model Catalogs hosting foundation models to allow users to easily discover and register models in a Model Registry for local training, enhancement, and serving. 
The implementation could start simple by allowing users to create a simple curated **model catalog source** that is backed by a yaml file. The yaml file could contain a list of high level foundation models metadata, some README text, and information about catalog model versions. 
In the future other catalog source implementations can be created to allow browsing Hugging Face, OpenAPI, etc. 

**Describe alternatives you've considered**
As an alternative, a common UI could be built that has adhoc client code to access different catalogs to browse and register models to a Kubeflow Model Registry. However, having a consolidated/common simple backend will make developing such a catalog browsing UI simpler in the future. 

**Additional context**
As an example, a simple API that exposes information about models such as the information provided in HuggingFace [modelcard](https://huggingface.co/docs/hub/en/model-cards), and also supports simple catalog model search by names, tags, etc. would be incredibly useful for Kubeflow users.

dhirajsb:
@andreyvelich @thesuperzapper we're working on a design for a simple model catalog API wrapper for local and remote catalogs like Hugging Face. I'll add more details with diagram, etc. to this feature request tomorrow. 

We're hoping it's a valuable addition to Model Registry and helps make the entry point and user experience easier into Kubeflow Model Registry, especially for foundation models and LLMs.

dhirajsb:
A proposed high level architecture would like below. 

A configmap can be used to configure a small amount of high level metadata about a collection of catalog sources, such as name, type, description, etc. Every catalog source may also include a secretName if credentials are required to connect to the catalog. E.g. HuggingFace could store API secret/token in a Kubernetes secret. 

Model registry would implement the logic to handle different catalog types, and process the catalog source information to fetch the catalog or create a remote connection to it. When Catalog API operations to list CatalogModels and CatalogModelRevisions are received, model registry will query the catalog source implementation logic to fetch the requested information. 

[...]

PR #823, opened 2025-02-19
feat(openapi): Model catalog API spec

pboyd:
## Description

Proposed model catalog API spec to implement #702. Draws heavily from the spec in PR #725, and builds on the (presently unmerged) changes PR #822.

This is a draft right now, because there's no implementation. But it would be helpful to get some feedback on the spec.

## How Has This Been Tested?
Nothing to test right now, as it's just an API spec.

## Merge criteria:


pboyd:
@dhirajsb @dtrifiro could I get your eyes on this? It's not nearly as big as it looks, it's mostly an update to the [openapi spec](https://github.com/kubeflow/model-registry/pull/823/files#diff-76e892700d0c69e6da9e12e5f92028b640e18a02d359d3e7fc4a32854a10dc9d) and re-generated code.

rareddy:
I read through a high-level API; +1 for generating the code. In the spirit, this looks closer to what @dhirajsb had in his PR (https://github.com/kubeflow/model-registry/pull/725), where the definition of models was separated between the MR and MC.  As I understood the ask from @danielezonca his ask is if we could make common schema for model. I understand they do not match currently due the model we adopted to support with ml-metadata.

dhirajsb:
@pboyd instead of adding the catalog types right away, can we create a PR with the base types as we discussed along with @adysenrothman? 
I know there aren't a lot of common fields based on the catalog types schema in this PR, but I still think we will benefit in the long run from having common BaseModel* types so that we can keep shared properties in the base types.

rareddy:
@dhirajsb when we know a model comes with certain top-level properties as showcased in the CatalogModel why can't that become the base model and derive the RegisteredModel from it with additional `custom properties` and `properties` to add to it to make it generic for user so that they can use that for arbitary properties? maybe we went too conservative in the first round to define model properties as we were adopting MLMD types as is.

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,126 @@ servers:
...
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+      operationId: getAllCatalogModels
+  /api/model_catalog/v1alpha3/sources/{source_id}/models/{model_id}:
```
danielezonca, line 75: I think this is a duplication of 
`/api/model_catalog/v1alpha3/models` 
where `source` is one of the parameters

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,126 @@ servers:
...
+          type: string
+        in: path
+        required: true
+      - name: model_id
+        description: A unique identifier for a `CatalogModel`.
+        schema:
+          type: string
+        in: path
+        required: true
+  /api/model_catalog/v1alpha3/sources/{source_id}/models/{model_id}/readme:
```
danielezonca, line 105: What is the key to identify a model in this context? 
`model_id` or the combination of `source_id` and `model_id`?

api/openapi/model-registry.yaml
```diff
@@ -1736,6 +1860,166 @@ components:
...
+          properties:
+            items:
+              description: Array of `CatalogModel` entities.
+              type: array
+              items:
+                $ref: "#/components/schemas/CatalogModel"
+          required:
+            - items
+        - $ref: "#/components/schemas/BaseResourceList"
+    CatalogModel:
```
danielezonca, line 1896: > can we create a PR with the base types

I agree, we can use `CatalModel` as base type if it has all the common shared properties.
If we prefer to have a different base entity that's fine too but we should really limit the hierarchy of model metadata to avoid possible incompatibilities/gaps when we convert one "model type" (i.e. catalog model) in another "model type" (i.e. registered model)

api/openapi/model-registry.yaml
```diff
@@ -1736,6 +1860,166 @@ components:
...
+                - text-generation
+            artifacts:
+              type: array
+              description: |-
+                If a model has multiple versions, each version should have a
+                separate artifact.
+              items:
+                $ref: "#/components/schemas/CatalogModelArtifact"
+        - $ref: "#/components/schemas/BaseResourceDates"
+    CatalogModelArtifact:
```
danielezonca, line 1999: How does it differ with existing `ModelArtifact`?
If it is a similar concept I would apply the same comment as before: define a common type or extend one type to implement the other

pboyd:
@rareddy Yes, it is much closer to @dhirajsb's #725.

@dhirajsb @rareddy Perhaps I have misunderstood the request, but there aren't many fields in common. Registered models have the generic MLMD `customProperties` and there aren't many fields in common with the catalog model. I did pull the date fields into a common base, but there weren't any other common fields. I can make a PR with just that, if you want.

api/openapi/model-registry.yaml
pboyd: They are different. `/api/model_catalog/v1alpha3/models` returns a list of models (optionally filtered by source), and `/api/model_catalog/v1alpha3/sources/{source_id}/models/{model_id}` returns a single model.

api/openapi/model-registry.yaml
pboyd: `source_id` and `model_id` are intended to be a composite primary key.

api/openapi/model-registry.yaml
danielezonca: My bad, I missed the `{model_id}` part :)

api/openapi/model-registry.yaml
danielezonca: I would expect a `CatalolgModelArtifact` to have `name` and `description` at least.
At the same time `artifactType`, `modelFormatName`, `storageKey`, `storagePath`, `modelFormatVersion` can become common and maybe replace `protocol`.

PR #838, opened 2025-02-27
feat: add model source properties to store metadata about origin of a model artifact, fixes RHOAIENG-19885

dhirajsb:
## Description

Add the following model source properties to store metadata about origin of a model artifact
ModelSourceKind
ModelSourceClass
ModelSourceGroup
ModelSourceId
ModelSourceName

Note that these properties are used to reference an external source where a model artifact was created.
A fuller lineage model for tracking ML `Experiments` and `Runs` will be added in the future as model registry API resource types. 
These properties also support referencing arbitrary model artifact sources, without constraining it to ML models only. 

Fixes RHOAIENG-19885

## How Has This Been Tested?

Added new model source properties to existing unit tests for model artifacts in backend service as well as Python client API and tests. 

## Merge criteria:

If you have UI changes

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -1550,6 +1550,35 @@ components:
             serviceAccountName:
               description: Name of the service account with storage secret.
               type: string
+            modelSourceKind:
+              type: string
+              description: |-
+                A string identifier describing the source kind. It differentiates various sources of model artifacts.
+                This identifier should be agreed upon by producers and consumers of source model metadata.
+                It is not an enumeration to keep the source of model metadata open ended. 
+                E.g. Kubeflow pipelines could use `pipelines` to identify models it produces.
+            modelSourceClass:
+              type: string
+              description: |-
+                A subgroup within the source kind. It is a specific sub-component or instance within the source kind.
+                E.g. `pipelinerun` for a Kubeflow pipeline run.
+            modelSourceGroup:
+              type: string
+              description: |-
+                Unique identifier for a source group for models from source class. 
+                It maps to a physical group of source models. 
+                E.g. a Kubernetes namespace where the pipeline run was executed.
+            modelSourceId:
+              type: string
+              description: |-
+                A unique identifier for a source model within kind, class, and group.
+                It should be a url friendly string if source supports using URLs to locate source models.
+                E.g. a pipeline run ID.
+            modelSourceName:
+              type: string
+              description: |-
+                A human-readable name for the source model. 
+                E.g. `my-project/1`, `ibm-granite/granite-3.1-8b-base:2.1.2`.
     ModelArtifactCreate:
       description: An ML model artifact.
       properties:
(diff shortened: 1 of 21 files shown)


tarilabs (commented):
thank you @dhirajsb , overall is a good information to add to the logical model.

- do we really need to denormalize between (Kind, Class) and (Id, Name)? unless I'm missing anythin in the example in the case of a PipelineRun one field would have been enough and it's up to the producer+consumer (as you rightly say) to manage this field to their need, that is if they want to make it even more explicit they could use the "FQDN" of the resource
- why on ModelArtifact and not on ModelVersion as discussed? A ML model is an ensemble of assets with one "trainer" source.

dhirajsb:
> do we really need to denormalize between (Kind, Class) and (Id, Name)? unless I'm missing anythin in the example in the case of a PipelineRun one field would have been enough and it's up to the producer+consumer (as you rightly say) to manage this field to their need, that is if they want to make it even more explicit they could use the "FQDN" of the resource

The idea is to allow clients to search by increasing smaller groups of model sources. So, we can support search by kind, class, group, etc. 

> why on ModelArtifact and not on ModelVersion as discussed? A ML model is an ensemble of assets with one "trainer" source.

A version could be a collection of multiple model artifacts all coming from separate sources, huggingface, pipelines, etc. in to create a composite model like an agentic system.

dhirajsb:
@tarilabs can we merge this in if there aren't any more questions?

Al-Pragliola:
A suggestion to lower the tech debt:

Seen that we have a similar mental model of the GVK + metadata.name to identify unique entities can we slim down the number of fields?

ModelSourceKind               |
                                               | =>          ModelSourceKindClass (or a better name)                   
ModelSourceClass              |

we set a {lowercase RFC 1123 validation}/{lowercase RFC 1123 validation} that seems reasonable.

Also, if the fields

ModelSourceKind
ModelSourceClass
ModelSourceGroup
ModelSourceId

are needed to uniquely identify an external resource, we do not have ownership of what the ModelSourceName can be at any time, as it can be modified over time, so I propose to leave that to whoever consumes our API.

So in the end this leaves us with:

ModelSourceKindClass
ModelSourceGroup
ModelSourceId

those are just m2c, let me know what you think about that

dhirajsb:
I don't think we should combine kind and class, since it would make searching by just kind or class harder if not impossible. MR users would be forced to adopt a string concatenation convention for the combination, which doesn't sound good either and something Dashboard is explicitly against. 

The `name` property is just for display as a human readable property. As the other properties could have ids and machine generated values it's good to at least support something that human consumers can read. 

The lowercase RFC 1123 validation would add the assumption that kind or class could be hostnames or somehow be used in a URL. The properties don't explicitly say anything about enforcing being able to create external urls from these properties. 

I understand that it means no validation of these field values, but that's the con of being open ended for external source models. 
Forcing restrictions such as case on these property values also forces such constraints on external systems, which may already exist and not be able to change to suit MR API. imho, MR should be flexible enough to adapt to external systems instead of the other way around.

tarilabs:
thanks @dhirajsb for this contribution and all

/lgtm
/approve

PR #849, opened 2025-03-04
feat(py): Implement Modular Async Runner

syntaxsdev:
## Description
Implements a modular way to pass in you own async scheduler/runner to solve issues where `ModelRegistry` has issues using a nested_asyncio when ran on an environment with a locked event loop (such as uvloop being used in Ray).

This issue was noted by https://github.com/blublinsky with their issue: https://github.com/kubeflow/model-registry/issues/774

and they also gracefully submitted a PR as well for this solution: https://github.com/kubeflow/model-registry/pull/802

In order to preserve licensing and also reduce complexity, we opted for a hybrid approach in theirs, where a runner can be specified externally, rather than shipping the code with the licensed snippet that works for situation. 

## How Has This Been Tested?
The tests have been updated in `conftest.py` to include:
- Uvloop fixture
- Dict of ModelRegistry params to rebuild from scratch within the Ray test

Tests in `test_client.py` to include:
- Testing of a slimmed example the user provided of their use case for issue replication

Additional tests:
- Tested in Jupyter notebook (with standard nested_asyncio - original functionality)
- Tested in Jupyter notebok (with custom async runner)
- Tested Ray in Jupyter notebook (with customer async runner)
- Tested in standalone script with Ray (custom async runner)
- Tested in standalone script (with standard nested_asyncio - original functionality)

Testing is standard
- Make sure local MR is up
- `pytest -m e2e` (or rename the specific test case to another tag to avoid running the others)

## Merge criteria:

- [x] **For first time contributors**: Please reach out to the [Reviewers](../OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes


syntaxsdev:
> This is pretty much the same PR. I have 2 concerns:
> 
> 1. It does not have an abstract runner class, that you can overwrite at any point with your own implementation
> 2. It does not have an abstract class as a parameter to make it even simpler
>    Also, I do not quite see why is it a different PR? It could have been an extension of the one that I have started

1. I removed the abstraction to reduce additional requirements/files added to the client, as async runner swapping would not be super common throughout the lifecycle of MR as most async envs use asyncio which is compatible with original implementation. Also, many implementations might seek more flexibility (like in the initialization), so this approach is simple.

 However, adding the abstraction is still possible. Please let me know what you think?
 
 2. It doesn't have an abstract runner class `parameter` but it does have a runner callable (bound method or function) `parameter`
 https://github.com/kubeflow/model-registry/blob/23dc78631f4b369815dc6dfc575a426e8c394a64/clients/python/src/model_registry/_client.py#L72
 
 which is still compatible with a class: 
 https://github.com/kubeflow/model-registry/blob/23dc78631f4b369815dc6dfc575a426e8c394a64/clients/python/tests/test_client.py#L856-L863
 
 however on line 856, my idea is that if a user wanted additional flexibility in their init, we'd have to take in the instantiated object, which means the code example above would still look similar, except `atr` (the instantiated object) would be passed instead of `atr.run` (ref to the method). What do you think on this as well?
 
 Also, unfortunately I cannot edit your PR directly since I'm not a maintainer so I cherrypicked onto my branch also to add additional testing. Your attribution and commit is still preserved, which will still count as a contribution as you are listed in the commits as an author. Who opened the PR does not have an effect on that 😄

syntaxsdev:
@blublinsky Thanks for the added context. I also did some research and Ray actors keep a single instantiated object for the lifetime of the actor. This could be problematic as well for Ray not with serializing - but all tasks share the same singleton event loop running in a background thread.

The current approach might be safer, as an MR (for the lifecycle of the remote function) has its own dedicated task runner, not a singleton.

clients/python/src/model_registry/_client.py
```diff
@@ -90,16 +92,23 @@ def __init__(
...
+        if async_runner:
+            if not (inspect.ismethod(async_runner) or inspect.isfunction(async_runner)):
+                msg = "`async_runner` must be a bound method or a function that takes in a coroutine to run."
+                raise ValueError(msg)
+            self._user_async_runner = async_runner
+        else:
+            import nest_asyncio
+
+            logger.debug("Setting up reentrant async event loop")
+            nest_asyncio.apply()
```
tarilabs, line 111: preserves original implementation, 👍

clients/python/src/model_registry/_client.py
```diff
@@ -140,6 +149,9 @@ def __init__(
         self.get_registered_models().page_size(1)._next_page()
 
     def async_runner(self, coro: Any) -> Any:
+        if hasattr(self, "_user_async_runner"):
+            return self._user_async_runner(coro)
+
```
tarilabs, line 154: optionally defer to user-plug-ged runner 👍

clients/python/src/model_registry/_client.py
```diff
@@ -140,6 +149,9 @@ def __init__(
...
         import asyncio
```
tarilabs, line 155: line 155 onwards, as before 👍

tarilabs (approved):
thanks a lot @syntaxsdev for tuning the refactoring, incorporating also @blublinsky remainder comments.

- https://github.com/kubeflow/model-registry/pull/849#issuecomment-2704706875
- https://github.com/kubeflow/model-registry/pull/849#issuecomment-2704775010

/lgtm
/approve

PR #914, opened 2025-03-26
500 custom prop type

adysenrothman:
## Description
We currently return a 500 error if a user requests to register a model with a customProperties field that does not include a metadataType parameter. This is required **conditionally** since customProperties is an optional field. 

We need to return a 400 bad request if a user attempts to pass a malformed customProperties parameter and give a more informed error message. 

Linked Jira: https://issues.redhat.com/browse/RHOAIENG-19192

## How Has This Been Tested?
Locally tested using the following request:

`curl -v -X POST \
-H 'Authorization: [Filtered]' \
-H 'Content-Type: application/json' -d \
'{"name": "Name4","customProperties": {"additionalProp1": {"metadataType": "MetadataStringValue", "string_value": "teststr"}}}' \
--insecure http://localhost:8080/api/model_registry/v1alpha3/registered_models`

Screenshot 2025-03-26 at 3 55 26 PM

## Merge criteria:

- [ x] The commits have meaningful messages; the author will squash them after approval or in case of manual merges will ask to merge with squash.
- [x ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/internal/converter/openapi_mlmd_converter_util.go
+++ b/internal/converter/openapi_mlmd_converter_util.go
@@ -9,6 +9,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/kubeflow/model-registry/internal/defaults"
 	"github.com/kubeflow/model-registry/internal/ml_metadata/proto"
+	"github.com/kubeflow/model-registry/pkg/api"
 	"github.com/kubeflow/model-registry/pkg/openapi"
 	"google.golang.org/protobuf/types/known/structpb"
 )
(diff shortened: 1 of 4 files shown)


internal/converter/openapi_mlmd_converter_util.go
```diff
@@ -93,7 +94,7 @@ func MapOpenAPICustomProperties(source *map[string]openapi.MetadataValue) (map[s
 					StructValue: asStruct,
 				}
 			default:
-				return nil, fmt.Errorf("type mapping not found for %s:%v", key, v)
+				return nil, fmt.Errorf("%w: metadataType not found for %s: %v", api.ErrBadRequest, key, v)
```
pboyd, line 97: This seems like a good fix, one question though. Do the other error returns in this function need the same change?

adysenrothman: good callout, we might as well register a bad request if the metadatatype is not present or if the value for the specified type is invalid

tarilabs:
Thanks a lot @adysenrothman for this DX improvement, and for the review Paul

/approve

PR #939, opened 2025-04-04
feat(openapi): Catalog API spec

pboyd:
## Description
This is a proposed spec for the catalog API. It is separate a separate spec from the existing model-registry.yaml.

## How Has This Been Tested?
There are no functional changes to test, but the spec can be previewed: https://editor.swagger.io/?url=https://raw.githubusercontent.com/pboyd/model-registry/refs/heads/catalog-types-2/api/openapi/model-catalog.yaml

## Merge criteria:
- [x] All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)


rareddy:
@pboyd what is your opinion on taking out the generated files from this PR? if possible I want to evaluate this just for the API then we can come back with generated as add-on?

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,135 @@ servers:
...
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+      operationId: getCatalogSources
+      summary: List All CatalogSources
+      description: Gets a list of all `CatalogSource` entities.
+  /api/model_catalog/v1alpha3/sources/{source_id}/models/{repository}/{model_name}:
```
dhirajsb, line 73: This path ends up making repositories a mandatory feature of all catalogs. We should either add endpoints to make it optional, e.g.
```
/api/model_catalog/v1alpha3/sources/{source_id}/models/{model_name}
```
Or, move repository to be a property to allow users to search by repository (if the catalog provider supports it) using a query param.

dhirajsb (changes requested):

rareddy:
I am still of the opinion that we should move all even 3 existing properties of `BaseModel` to `BaseModelVersion` and rename the `BaseModelVersion` to `BaseModel` or simply `Model`. IMO that will reduce the clutter in the API.

pboyd:
> I am still of the opinion that we should move all even 3 existing properties of `BaseModel` to `BaseModelVersion` and rename the `BaseModelVersion` to `BaseModel` or simply `Model`. IMO that will reduce the clutter in the API.

@rareddy, there's a new commit on this PR which does that.

api/openapi/model-registry.yaml
```diff
@@ -1215,27 +1362,30 @@ components:
...
-              readOnly: true
-            lastUpdateTimeSinceEpoch:
-              format: int64
-              description: |-
-                Output only. Last update time of the resource since epoch in millisecond
-                since epoch.
-              type: string
-              readOnly: true
+        - $ref: "#/components/schemas/BaseResourceDates"
     BaseResourceCreate:
```
rareddy, line 1389: I looked twice around trying to understand schema structure, in general, I believe we can merge `BaseResourceCreate` into `BaseResource` and keep the `BaseResourceUpdate` as separate entity. I am not just saying `BaseResource`, but same can be said about also `RegisterModelCreate`, `ModelArtifactCreate` etc. I do not think It will not change the contract but  API can be read with much clarity then. We do not have to do that as part of this PR but can be done pre-cursor to this PR to reduce the complexity of the API. wdyt?

rareddy (dismissed):

dhirajsb:
*Create types are only for user provided properties at creation time used
in POST.

So, we can't merge them into anything else.

rareddy:
>*Create types are only for user provided properties at creation time used
in POST.

I believe THE primary objective was not to let fields on the `create` object be not updatable during a `patch`.  A `post` is free to manipulate any fields. For example see below:

```    BaseExecution:
      allOf:
        - $ref: '#/components/schemas/BaseExecutionCreate'
        - $ref: '#/components/schemas/BaseResource'
    BaseExecutionCreate:
      allOf:
        - $ref: '#/components/schemas/BaseExecutionUpdate'
        - $ref: '#/components/schemas/BaseResourceCreate'
    BaseExecutionUpdate:
      allOf:
        - type: object
          properties:
            lastKnownState:
              $ref: '#/components/schemas/ExecutionState'
        - $ref: '#/components/schemas/BaseResourceUpdate'
```
or see this

```
    RegisteredModel:
      description: A registered model in model registry. A registered model has ModelVersion children.
      allOf:
        - $ref: '#/components/schemas/BaseResource'
        - $ref: '#/components/schemas/RegisteredModelCreate'
        - $ref: '#/components/schemas/Model'
    RegisteredModelCreate:
      description: A registered model in model registry. A registered model has ModelVersion children.
      required:
        - name
      allOf:
        - $ref: '#/components/schemas/BaseResourceCreate'
        - $ref: '#/components/schemas/RegisteredModelUpdate'
        - type: object
          properties:
            name:
              description: |-
                The client provided name of the model. It must be unique among all the RegisteredModels of the same
                type within a Model Registry instance and cannot be changed once set.
              type: string
    RegisteredModelUpdate:
      description: A registered model in model registry. A registered model has ModelVersion children.
      allOf:
        - $ref: '#/components/schemas/BaseResourceUpdate'
        - $ref: '#/components/schemas/Model'
        - type: object
          properties:
            owner:
              type: string
            state:
              $ref: '#/components/schemas/RegisteredModelState'
```
can u help what I am missing to see here?  why is advantageous to split apart into 3 different objects?

api/openapi/model-registry.yaml
pboyd: Those are separate because some fields are immutable, and others can be set initially but not changed on update. I find it much easier to see in the [swagger editor](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pboyd/model-registry/refs/heads/catalog-types-2/api/openapi/model-registry.yaml).

`BaseResource` has all the fields returned when viewing a resource:
base-resource

`BaseResourceCreate` has all the fields that can be set during creation:
base-resource-create

`BaseResourceUpdate` has all the fields that can be changed during update:
base-resource-update

If we merge these, then the schema will indicate that timestamps can be set and that name can be changed after creation.

api/openapi/model-registry.yaml
```diff
@@ -1247,6 +1397,21 @@ components:
...
+          description: Output only. Create time of the resource in millisecond since epoch.
+          type: string
+          readOnly: true
+        lastUpdateTimeSinceEpoch:
+          format: int64
+          description: |-
+            Output only. Last update time of the resource since epoch in millisecond
+            since epoch.
+          type: string
+          readOnly: true
```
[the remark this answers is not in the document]
pboyd: It's separate so that I could reuse the date fields without also including fields from `BaseResource` that didn't apply. For instance, artifacts don't need `id` and `name`.

pboyd:
@danielezonca, this is a re-attempt at PR #823. Can you take a look?

api/openapi/model-registry.yaml
rareddy: Thank you for the explanation; makes sense. I did not realize the immutable fields. Could they be set readonly? for example:

```
BaseResource
      type: object
      $ref: BaseResourceUpdate
      properties:
        name:
          type: string
          example: foobar
       lastUpdatedTime:
           type: Time
           readOnly: true

BaseResourceUpdate
      type: object
      properties:
         name:
          readOnly: true
         description:
            type: string
            example: Will this work?     
```

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,153 @@ servers:
...
 paths:
+  /api/model_catalog/v1alpha3/models:
+    summary: Path used to list and search catalog models accross all sources.
+    description: >-
+      The REST endpoint/path used to list zero or more `CatalogModel` entities from all `CatalogSources`.
+    get:
+      tags:
+        - ModelCatalogService
+      parameters:
+        - name: source
```
rareddy, line 21: if this is supposed to be used for selection of `source`, then maybe we need to add an additional parameter called `search` that can be used to filter results. otherwise call this `search` and then we need to  add special DSL for it

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,153 @@ servers:
...
+            are returned. If multiple sources are provided, models from any of
+            the sources are returned.
+          schema:
+            type: string
+          in: query
+          required: false
+        - $ref: "#/components/parameters/pageSize"
+        - $ref: "#/components/parameters/orderBy"
+        - $ref: "#/components/parameters/sortOrder"
+        - $ref: "#/components/parameters/offset"
```
rareddy, line 33: On `Inference` offset defined as `nextPageToken` are seeing this as different here; why not reuse same? Also could all the `pageSize` `orderBy` `sortOrder` and `offset` be combined into single `$ref` say `Paging` or `Pagination`

api/openapi/model-registry.yaml
```diff
@@ -10,6 +10,153 @@ servers:
   - url: "https://localhost:8080"
```
rareddy, line 10: we talked about splitting the file into multiple files, but did not due to reduced usage when using tools like Swagger. Can we consider creating multiple file but stitching at build time?

api/openapi/model-registry.yaml
```diff
@@ -1516,6 +1756,60 @@ components:
       example:
         string_value: my_value
         metadataType: MetadataStringValue
+    Model:
```
danielezonca, line 1759: As far as I understand this `Model` entity is the shared schema for the catalog but also for all the other model but this is just an abstract class so all concrete entities have additional fields.
Can you please confirm it?

Overall it looks good to me, how do we expect to manage the difference/mapping when I take a model from a catalog and I want to import it as a registered model?
Is the idea to implement and maintain the mapping/conversion logic in all sdk clients (python/ts/etc)?

pboyd: Yes, that is correct. It has fields common between `RegisteredModel` and `CatalogModel`.

It shouldn't require very much conversion code to take a catalog model and register it, so I think it should be fine for the client to handle that. We could add a convenience endpoint for it pretty easily too.

api/openapi/model-registry.yaml
pboyd: Yes, this parameter is only for filtering by `source`. I'll add one for a keyword search.

api/openapi/model-registry.yaml
tarilabs: > We could add a convenience endpoint for it pretty easily too

What I like about this possibility, is that is not an XOR, but a complementary option to the different clients handling that conversion (which I concur is fine, but I believe realistic to expect may drift in mapping logic eventually).

api/openapi/model-registry.yaml
rareddy: >But I'm not sure what a read-only name on RegisteredModelUpdate is supposed to indicate.

The POST will use the `RegisteredModel`, where you can set the `name`; when you do a PATCH, RegisteredModelUpdate it will not allow you to update the `name`. That was my assumption

api/openapi/model-registry.yaml
```diff
@@ -1290,6 +1461,81 @@ components:
             The external id that come from the clients’ system. This field is optional.
             If set, it must be unique among all resources within a database instance.
           type: string
+    CatalogModel:
```
rareddy, line 1464: Is the absence of a `ModelVersion` in the `CatalogModel` deliberate? Are we certain that the catalog model will not have any versions, or that each version will be represented by a different model?

api/openapi/model-registry.yaml
pboyd: I was following the convention. If we want to change it, can it be in another PR?

api/openapi/model-registry.yaml
pboyd: I'll work on it.

api/openapi/model-registry.yaml
rareddy: Yes, any simplifications we do should be a different PR, even pre-cursor to PR IMO; but will that keep the functionality AS IS or am I missing any other points, is what I was seeking.

api/openapi/model-registry.yaml
```diff
@@ -1516,6 +1756,60 @@ components:
...
+          description: Short name of the model's license.
+          example: apache-2.0
+        licenseLink:
+          type: string
+          format: uri
+          description: URL to the license text.
+        libraryName:
+          type: string
+          example: transformers
     ModelArtifact:
```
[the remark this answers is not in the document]
rareddy: @dhirajsb do you recollect why we used them on Artifact?

My main thought is to see if we can "flatten" the Artifacts into a single component.

api/openapi/model-registry.yaml
rareddy: Also, I think this will help us if we decide to build a different container just for the model catalog.

api/openapi/model-registry.yaml
rareddy: I am trying to keep the model close to the same on both sides, otherwise the original intent is not there IMO. wdyt?

api/openapi/model-registry.yaml
rareddy: thank you.

api/openapi/model-registry.yaml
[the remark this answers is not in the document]
rareddy: see #1009

api/openapi/model-registry.yaml
pboyd: I can go either way on `CatalogModelVersion`. We don't absolutely need it because the catalog sources don't have the concept (at most HF has a link to a previous version). But I can see the value in parity with `RegisteredModel`/`ModelVersion`, which is why I added in the first iteration of the PR.

rareddy: @danielezonca any insight here?

pboyd: I'm just getting back to this.. so I had originally created model version for catalog models, but we decided to remove it because it made everything a little more complicated. In particular, the primary key for a catalog model artifact has to be a 4-field composite key (`source_id`, `model_id`, `model_version_id`, `artifact_id`), and that requires a long URL path (i.e. `/sources/{source_id}/models/{model_id}/versions/{model_version_id}/artifacts`). That said, catalog model version is only one more layer in the hierarchy, so it won't make a huge difference either way. I'm more than happy to add it back, if that's the direction we're going.

rareddy: ok; if we made any changes on MR side for this, lets revert to how it was originally designed model --> has* --> Modelversion. Then keep the this one as you designed CatalogModel.

pboyd:
This PR has been overhauled after some recent discussions and the re-worked base types:

- Separate the model catalog from the model registry spec (based on #1012, which can be merged independently).
- Add versions to catalog models.
    - Not all sources and models need versions, so there are endpoints to retrieve the latest version.
- Most of the specific model and artifact fields were removed with the expectations that clients will use `customProperties` for these.

api/openapi/src/catalog.yaml
```diff
@@ -0,0 +1,375 @@
...
+        in: path
+        required: true
+      - name: model_name
+        description: A unique identifier for the model.
+        schema:
+          type: string
+        in: path
+        required: true
+  /api/model_catalog/v1alpha1/sources/{source_id}/models/{model_name}/versions/{version_id}:
+    description: >-
```
rareddy, line 170: I thought we had agreed not to include versions in the catalog, did I misunderstand? I'm a bit confused by this design. Based on our conversation during the LLS discussion, I assumed we would keep the Model schema flexible, similar to what we have in LLS and Model Registry. Then we’d use a common model definition as the base, as originally planned.

pboyd: I think I misinterpreted this message in regard to model versions:

> ok; if we made any changes on MR side for this, lets revert to how it was originally designed model --> has* --> Modelversion. Then keep the this one as you designed CatalogModel.

I thought that was saying go back to the original design for CatalogModel, which had a version.

Anyway, I'm happy to remove the versions again if we don't need them at all, but the latest update might be a better alternative. The versions are available for sources/models where it makes sense, but they can largely be ignored where they don't make sense.

The latest version has fields we need for functionality, but fields that we don't specifically need are left up to the client to put in `customProperties`. This was my interpretation of last week's discussion to make it more flexible like LLS, although it was a departure from earlier designs where model registry and model catalog largely shared a schema.

pboyd: Versions are gone now, and the fields from the earlier version of this PR were added back to `CatalogModel` and `RegisteredModel`.

pboyd:
The latest update removes catalog model versions and adds the specific fields back. We can't merge this until the registry code is updated to handle the new fields, but the spec should be complete.

rareddy:
@pboyd this looks very good. Encapsulates all the discussions so far. Do you want to bring back all the generated files so that I can LGTM this? I suspect that in the Catalog `artifacts` you are expecting the list to provide a URL that can be used to download correctly?

tarilabs:
thank you @pboyd ,

- I've reviewed this using the swagger editor and importing from `https://raw.githubusercontent.com/kubeflow/model-registry/c27c3f76fe82d6c0d9cc10e423aefe6db468a931/api/openapi/catalog.yaml`
- I understand you have endpoints as required to list Sources, Models (in a Source) and overall Models.
- I understand bespoke fields will be mapped so to avoid having to replicate mapping logic in the (several) clients (e.g. UI, Py client, etc)

I have one question please.
For a Model, "uri" is only available in the several Artifact(s), as the relationship is Model 1:N Artifacts.
I'm not following then _which_ (of the potential N) would be the URI for Deployment? ie we need to identify 1 (one) single URI for:
- KServe Isvc `spec.predictor.model.storageUri`
- LLM-d application `model.modelArtifactURI`
(and others)

I'm wondering if your semantic is:
- if it's truly intended to be Model 1:N Artifacts and Artifacts having URI(s) the "first one" or one particularly annotated is the one intended for this scope
- it's an oversight, and `CatalogModel` was intended to have an `uri` too.

pboyd:
> @pboyd Maybe I'm missing something, but how is common.yaml being used?

[`scripts/merge_opernapi.sh``](https://github.com/kubeflow/model-registry/blob/main/scripts/merge_openapi.sh) merges `common.yaml` with either `api/openapi/src/catalog.yaml` or `api/openapi/src/model-registry.yaml`.

pboyd:
Thanks for the review, @tarilabs.

> I have one question please. For a Model, "uri" is only available in the several Artifact(s), as the relationship is Model 1:N Artifacts. I'm not following then _which_ (of the potential N) would be the URI for Deployment? ie we need to identify 1 (one) single URI for:
> 
> * KServe Isvc `spec.predictor.model.storageUri`
> * LLM-d application `model.modelArtifactURI`
>   (and others)
> 
> I'm wondering if your semantic is:
> 
> * if it's truly intended to be Model 1:N Artifacts and Artifacts having URI(s) the "first one" or one particularly annotated is the one intended for this scope
> * it's an oversight, and `CatalogModel` was intended to have an `uri` too.

The intention was to have multiple artifacts per model to handle different formats or locations. For example, a source could publish model files in an S3 bucket and a modelcar image in an OCI registry. We could probably do a better job annotating those, but I was hoping it would be clear from the URI (e.g. `s3://` vs. `oci://`).

tarilabs:
> The intention was to have multiple artifacts per model to handle different formats or locations

I see, thanks for this clarification,; so based on this definition, I understand **_any_** CatalogModelArtifact is considered a valid Deployment/Inference artifact, just alternative locations or "packaging format" of the same weights.

For example:
```
CatalogModel "neuralmagic/granite-3.1-8b-base-quantized.w4a16"
|
+--- CatalogModelArtifact.uri = "hf://neuralmagic/granite-3.1-8b-base-quantized.w4a16"
+--- CatalogModelArtifact.uri = "oci://registry.redhat.io/rhelai1/modelcar-granite-3-1-8b-base-quantized-w4a16:1.5"
```
that is, depending on user choice, they might select one or the other format to be deployed.

/lgtm

api/openapi/model-registry.yaml
```diff
@@ -1718,6 +1781,8 @@ components:
       description: A registered model in model registry. A registered model has ModelVersion children.
       allOf:
         - $ref: "#/components/schemas/BaseResourceUpdate"
+        # TODO: uncomment this when the BaseModel fields are implemented
+        #- $ref: "#/components/schemas/BaseModel"
```
pboyd, line 1785: I'm commenting this out for now to avoid adding unimplemented fields to the model registry spec. I intend to uncomment this and implement these fields in an upcoming PR.

tarilabs:
a follow-up question @pboyd if you don't mind.

I just recall that actually CatalogModel+CatalogModelArtifact(s) are always on a per-CatalogSource basis, ie

```
/api/model_catalog/v1alpha1/sources/{source_id}/models/{model_name}/artifacts
```

therefore I'm asking if the semantic from https://github.com/kubeflow/model-registry/pull/939#issuecomment-2898542777 is still holding? 🤔 

I could expect the example I provided would be more accurate with:

```
CatalogSource "Getting from HF"
|
+-- CatalogModel "neuralmagic/granite-3.1-8b-base-quantized.w4a16"
    |
    +--- CatalogModelArtifact.uri = "hf://neuralmagic/granite-3.1-8b-base-quantized.w4a16"
    +--- CatalogModelArtifact.uri = "hf://RedHatAI/granite-3.1-8b-base-quantized.w4a16"
    +--- CatalogModelArtifact.uri = "unknown://undeployable"

CatalogSource "Getting from OCI"
|
+-- CatalogModel "neuralmagic/granite-3.1-8b-base-quantized.w4a16"
    |
    +--- CatalogModelArtifact.uri = "oci://registry.redhat.io/rhelai1/granite-3-1-8b-base-quantized-w4a16:1.5" # artifact
    +--- CatalogModelArtifact.uri = "oci://registry.redhat.io/rhelai1/modelcar-granite-3-1-8b-base-quantized-w4a16:1.5" # modelcar
    +--- CatalogModelArtifact.uri = "unknown://undeployable"
```

ie in which cases, we have a CatalogSource that for a CatalogModel can return several CatalogModelArtifact of different "format" as described in https://github.com/kubeflow/model-registry/pull/939#issuecomment-2898542777 ?

Just trying to understand all edge-cases :) thanks in advance

pboyd:
@tarilabs, yes, we could have the same model from multiple sources. I'm not sure what we could do to merge the records though.

tarilabs:
> we could have the same model from multiple sources

agree

> I'm not sure what we could do to merge the records though

I don't think we can or should, it's a perfectly valid use-case that different sources might provide the same models.

My question was another, I understand different "forms" (in your example `e.g. s3:// vs. oci://` from https://github.com/kubeflow/model-registry/pull/939#issuecomment-2898542777) are unlikely to happen for the same CatalogModel instance. 

I understand that, if ever the case, the most complex example could be the one presented in https://github.com/kubeflow/model-registry/pull/939#issuecomment-2898895322

I also suspect the most common scenario could be:

```
CatalogSource "Getting from HF"
|
+-- CatalogModel "neuralmagic/granite-3.1-8b-base-quantized.w4a16"
    |
    +--- CatalogModelArtifact.uri = "hf://RedHatAI/granite-3.1-8b-base-quantized.w4a16"

CatalogSource "Getting from OCI"
|
+-- CatalogModel "neuralmagic/granite-3.1-8b-base-quantized.w4a16"
    |
    +--- CatalogModelArtifact.uri = "oci://registry.redhat.io/rhelai1/modelcar-granite-3-1-8b-base-quantized-w4a16:1.5" 
```

would you agree or do you see errors in my understanding please?

pboyd:
@dhirajsb, that's pretty much it. I've made the necessary changes in https://github.com/kubeflow/model-registry/pull/1146. I'm happy to add them to this PR if it makes sense.

tarilabs (commented):
per previous https://github.com/kubeflow/model-registry/pull/939#issuecomment-2899115333 and follow-ed up comments, this is still
/lgtm
with the update that we need only to manage 1 "description"; thanks @pboyd

rareddy (commented):
/lgtm

Al-Pragliola (commented):
/lgtm

PR #960, opened 2025-04-10
feat(controller): added reconciliation of inferenceservice url

Al-Pragliola:
needs #917 

## Description

In this PR, I have added a new field to the InferenceService customProperties, which is the 'url' field.
This field is used to reconcile the url from the status of the cluster InferenceService, and update the url in the customProperties.
Also, I have also added new unit tests to test the new functionality.

*Why a customProperty and not a first class field?*

Because the url is part of the InferenceService status, and not the spec so it is ephemeral and should be treated as such in my opinion.

## How Has This Been Tested?

```shell
make test
```

And also, I have created a mock controller that periodically updates the URL in the InferenceService status, and the URL in the customProperties is updated accordingly.

```go
package controllers

import (
	"context"
	"fmt"
	"time"

	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
	"k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/runtime"
	"knative.dev/pkg/apis"
	ctrl "sigs.k8s.io/controller-runtime"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/log"
)

type InferenceServiceMockReconciler struct {
	client.Client
	Scheme *runtime.Scheme
}

// +kubebuilder:rbac:groups=serving.kserve.io,resources=inferenceservices,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups=serving.kserve.io,resources=inferenceservices/status,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups=serving.kserve.io,resources=inferenceservices/finalizers,verbs=get;list;watch;update;create;patch;delete
// +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch

func (r *InferenceServiceMockReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	_ = log.FromContext(ctx)

	isvc := &kservev1beta1.InferenceService{}
	err := r.Get(ctx, req.NamespacedName, isvc)
	if err != nil {
		if errors.IsNotFound(err) {
			// Resource not found, nothing to do
			return ctrl.Result{}, nil
		}
		// Error fetching resource
		return ctrl.Result{}, err
	}

	// Update the status.url field
	newUrl, err := apis.ParseURL(fmt.Sprintf("https://example.com/%d", time.Now().Unix()))
	if err != nil {
		return ctrl.Result{}, err
	}

	isvc.Status.URL = newUrl
	err = r.Status().Update(ctx, isvc)
	if err != nil {
		return ctrl.Result{}, err
	}

	// Requeue the reconcile loop after 30 seconds
	return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
}

func (r *InferenceServiceMockReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&kservev1beta1.InferenceService{}).
		Named("mockinferenceservice").
		Complete(r)
}
```

## Merge criteria:

### DIFF
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -169,8 +169,8 @@ func (r *InferenceServiceController) Reconcile(ctx context.Context, req ctrl.Req
 
 		mrCurrentIvcUrl := ""
 
-		if mrIs.CustomProperties != nil && (*mrIs.CustomProperties)["url"].MetadataStringValue != nil {
-			mrCurrentIvcUrl = (*mrIs.CustomProperties)["url"].MetadataStringValue.StringValue
+		if mrIs.CustomProperties != nil {
+			mrCurrentIvcUrl = (*mrIs.CustomProperties)["url"].MetadataStringValue.GetStringValue()
 		}
 
 		urlAreDiff := r.checkURLDiff(isvc, mrCurrentIvcUrl)
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -377,7 +377,11 @@ func (r *InferenceServiceController) createMRInferenceService(
 		}
 
 		if isvc.Status.URL != nil {
-			(*isCreate.CustomProperties)["url"].MetadataStringValue.StringValue = isvc.Status.URL.String()
+			isCreate.CustomProperties = &map[string]openapi.MetadataValue{}
+
+			(*isCreate.CustomProperties)["url"] = openapi.MetadataValue{
+				MetadataStringValue: openapi.NewMetadataStringValue(isvc.Status.URL.String(), "MetadataStringValue"),
+			}
 		}
 
 		is, _, err = mr.ModelRegistryServiceAPI.CreateInferenceService(ctx).InferenceServiceCreate(isCreate).Execute()
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -166,6 +166,27 @@ func (r *InferenceServiceController) Reconcile(ctx context.Context, req ctrl.Req
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to find InferenceService with id %s in model registry: %w", mrIsvcId, err)
 		}
+
+		mrCurrentIvcUrl := ""
+
+		if mrIs.CustomProperties != nil && (*mrIs.CustomProperties)["url"].MetadataStringValue != nil {
+			mrCurrentIvcUrl = (*mrIs.CustomProperties)["url"].MetadataStringValue.StringValue
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/inferenceservice-controller/controller.go
```diff
@@ -334,19 +367,55 @@ func (r *InferenceServiceController) createMRInferenceService(
...
 			ModelVersionId:       modelVersionIdPtr,
 			Name:                 &isName,
 			RegisteredModelId:    registeredModelId,
 			Runtime:              isvc.Spec.Predictor.Model.Runtime,
 			ServingEnvironmentId: servingEnvironmentId,
-		}).Execute()
+		}
+
+		if isvc.Status.URL != nil {
+			(*isCreate.CustomProperties)["url"].MetadataStringValue.StringValue = isvc.Status.URL.String()
```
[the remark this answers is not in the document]
dhirajsb: Need test cases for all combinations of url and customProperties set/unset, etc. in `controller_test.go` to verify.

pkg/inferenceservice-controller/suite_test.go
```diff
@@ -177,6 +180,9 @@ var _ = AfterSuite(func() {
 func ModelRegistryDefaultMockServer() *httptest.Server {
 	handler := http.NewServeMux()
 
+	servingEnvironments := new(sync.Map)
+	inferenceServices := new(sync.Map)
```
dhirajsb, line 184: Maybe I missed it, but it doesn't look like these sync maps are needed, does it?

dhirajsb (changes requested):

pkg/inferenceservice-controller/controller.go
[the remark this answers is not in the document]
Al-Pragliola: @pboyd Good catch, it didn't happen in my tests because the .Status.URL always gets populated _after_ the basic creation, so CustomProperties was empty but not nil

@dhirajsb thanks, I'll add more tests

pkg/inferenceservice-controller/suite_test.go
Al-Pragliola: I was trying them while debugging a weird bug in tests and I forgot to switch back to regular maps, good catch reverted in https://github.com/kubeflow/model-registry/pull/960/commits/5847b3c5aa7744781107557bad05390d501deaa9

dhirajsb: If I'm not mistaken, there is only one hardcoded id `1` being used. If so, does it even need a map?

Al-Pragliola: Not for the time being, I'm going to keep it in case of more complex scenarios

dhirajsb (approved):
/lgtm

PR #984, opened 2025-04-15
feat: improve dev QoL

Al-Pragliola:
## Description

Refactor of the mr Dockerfile, now there's a common stage that prepares the data, a dev-build stage that skips all the autogen/lint/vet and just builds the binary, and the normal stage that is built by default if no `--target` is specified, this greatly improves the build times when testing locally.

To take advantage of this new feature, you can use the `ARGS="--target dev-build"` when running the make image/build target, or use the `DEV_BUILD=true` when running the make test-e2e target in the clients/python folder.

I also increased the timeout when waiting for the mr deployment in the deploy_on_kind.sh script, as it occasionally failed and gave false negatives.

## How Has This Been Tested?

```shell
time make image/build .

...

real    2m41.152s
user    9m18.038s
sys     2m10.551s
```
vs

```shell
time ARGS="--target dev-build" make image/build .

...

real    0m14.952s
user    0m39.747s
sys     0m11.835s
```
-----------------------------------
```shell
time make test-e2e

...

real    5m49.049s
user    9m41.787s
sys     2m18.887s
```
vs

```shell
time DEV_BUILD=true make test-e2

...

real    3m17.977s
user    1m3.871s
sys     0m19.981s
```

## Merge criteria:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the model-registry binary
-FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi8/go-toolset:1.23 AS builder
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi8/go-toolset:1.23 AS common
 ARG TARGETOS
 ARG TARGETARCH
 
@@ -10,20 +10,8 @@ COPY ["go.mod", "go.sum", "./"]
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
-USER root
-# default NodeJS 14 is not enough for openapi-generator-cli, switch to Node JS currently supported
-RUN yum remove -y nodejs npm
-RUN yum module -y reset nodejs
-RUN yum module -y enable nodejs:18
-# install npm and java for openapi-generator-cli
-RUN yum install -y nodejs npm java-11 python3
-
 # Download tools
 COPY Makefile .
-COPY scripts/install_protoc.sh scripts/
-RUN make deps
-
-# Copy rest of the source
 COPY ["main.go", ".openapi-generator-ignore", "openapitools.json", "./"]
 COPY cmd/ cmd/
 COPY api/ api/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -33,9 +21,36 @@ COPY pkg/ pkg/
...
 
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o model-registry
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as dev-build
+
+WORKDIR /
+COPY --from=dev /workspace/model-registry .
+USER 65532:65532
+
+ENTRYPOINT ["/model-registry"]
```
tarilabs, line 38: So when the `--target=dev-build` is active, it will stop completing this section, to my understanding.

(and when no specific `--target` is provided, eg building the full image, it progresses)

tarilabs (commented):
Thank you so much @Al-Pragliola , I believe this is quite clever use of [target](https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage) builds actually!!

As of today the Containerfile is re-generating some sources internally during one stage of the multi-stage build.
In my view, as I've expressed in the past, I would have preferred for the Containerfile to just limit itself to compile sources from the repo, organic sources and code-generated sources, since we're committing _both_ to the repo already; so to also make it simpler to demonstrate we're shipping what is committed.
(as a reminder, we're checking with CI/CD that all code-generated files are indeed committed so the result of the build being idempotent).

Not that being pursued, I believe this is a great improvements for a vast majority of local DX cases without changing the _current_ Containerfile intent 

/approve

Dockerfile
Al-Pragliola: @tarilabs It's actually better that! container tools, when no target is specified, take the latest (from the bottom) by default and build a dependency graph from there, since `dev-build` is not required from the latest (unnamed target), it gets skipped 😁 If you specify `dev-build` as the target instead, it starts from there and builds its dependency graph, skipping the rest of the container file

tarilabs: you mean like taking the "branch" here : 

```mermaid
flowchart TD
    common[registry.access.redhat.com/ubi8/go-toolset:1.23 AS common]
    dev[common AS dev]
    dev-build[registry.access.redhat.com/ubi8/ubi-minimal:latest as dev-build]
    builder[common as builder]
    default[registry.access.redhat.com/ubi8/ubi-minimal:latest]
    common --> dev
    dev --> dev-build
    common --> builder
    builder --> default
```

Dockerfile
```diff
@@ -33,9 +21,36 @@ COPY pkg/ pkg/
...
+
+###### Dev stage - end ######
+
+FROM common as builder
```
Crazyglue, line 42: running this locally, `as` on this line and line 32 show a few errors in the console

```
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 32)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 42)
 ```

Crazyglue (approved):
Tested this locally and confirm an improvement in the docker image build time.

In my testing, it doesn't seem to shave off any time if working with a totally clean docker image/build cache, but on subsequent executions, things definitely seem faster and more cached.

PR #1007, opened 2025-04-21
remove todo 401

adysenrothman:
I'm cleaning up these TODOs for now as I do not believe that they are super useful. We do need to setup a system for role based authorization that will eventually be applied to these routes. Since that is a decent sized lift, I don't think that it is necessary to bloat the service with comments right now. Additionally, as new functionality is added and we don't want to have to keep adding TODOs.


Al-Pragliola (approved):
/lgtm

tarilabs (approved):
thanks @adysenrothman , I think these were originally created by the one-off openapi-codegen, not sure if this require some considaration with the more recent work on the mustache template by @pboyd too?

- https://github.com/kubeflow/model-registry/pull/1007#pullrequestreview-2783512301
 
/approve

PR #1009, opened 2025-04-22
Simplify the API file for model registry

rareddy:
Simplify the OpenAPI specification for the Model Registry without losing any functionality.

## Description
The current OpenAPI schema for the Model Registry is overly complex, with excessive nesting and hierarchy that makes it difficult to read and auto-generate code from.

My focus has been on the following:

1) Consolidate xxxxCreate components
Removed the separate `xxxxCreate` definitions and merged their functionality into the main `xxxx` entities. To preserve update semantics, properties that are not meant to be modified are marked as readOnly in the `xxxxUpdate` objects.

2) Eliminate unnecessary base classes
Removed redundant base classes like `BaseExecution` and `BaseArtifact` that added extra nesting without clear value. Now, `BaseResource` is the only remaining base entity used for shared attributes.

3) [TODO] Unify Artifact entities
There’s an opportunity to simplify further by consolidating DocArtifact and ModelArtifact into a single Artifact entity. However, this would change current system behavior, so it's deferred for now.

## How Has This Been Tested?
My test containers is not working; so not tested YET. This work in progress. I am putting out there for comments first.

## Merge criteria:


rareddy:
for some reason, I could not invite you asa  reviewer @dhirajsb

api/openapi/model-registry.yaml
```diff
@@ -71,7 +71,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/ArtifactCreate"
+              $ref: "#/components/schemas/Artifact"
```
Crazyglue, line 74: Will this renaming change the generated names for the downstream clients? Ie, in the python client would this change the data-types that the end-user needs to import from the package?

rareddy:
@tarilabs @Al-Pragliola @dhirajsb @pboyd Now that the PR is compiling and running test failure-free, do you want me to remove all the generated files to show the actual changes to the code without a lot of clutter for review? LMK.

tarilabs:
> do you want me to remove all the generated files to show the actual changes to the code without a lot of clutter for review? LMK.

I can see the diff by clicking here; https://github.com/kubeflow/model-registry/pull/1009/files#diff-76e892700d0c69e6da9e12e5f92028b640e18a02d359d3e7fc4a32854a10dc9d

I think that shows what you described in the 1st message:

> Consolidate xxxxCreate components
> Removed the separate xxxxCreate definitions and merged their functionality into the main xxxx entities. To preserve update semantics, properties that are not meant to be modified are marked as readOnly in the xxxxUpdate objects.
> Eliminate unnecessary base classes
> Removed redundant base classes like BaseExecution and BaseArtifact that added extra nesting without clear value. Now, BaseResource is the only remaining base entity used for shared attributes.

this, and 
_by ..._
surfacing the attributes `uri, state` or similarly, to the remaining one in the hierarchy

api/openapi/model-registry.yaml
rareddy: This is addressed now!

clients/python/Makefile
```diff
@@ -64,7 +64,7 @@ tidy:
 	poetry run black src/mr_openapi
 
 .PHONY: build
-build: install tidy
+build: install tidy
```
pboyd, line 67:
```suggestion
build: install tidy
```

api/openapi/model-registry.yaml
```diff
@@ -1287,24 +1240,14 @@ components:
           type: string
         externalId:
           description: |-
-            The external id that come from the clients’ system. This field is optional.
+            The external id that come from the clients' system. This field is optional.
```
pboyd, line 1243:
```suggestion
            The external id that comes from the client's system. This field is optional.
```

pboyd (commented):
I'm still working through the field changes, but a couple things stand out.

- `RegisteredModel` has a mutable `id` field. `id` wasn't present at all in `RegisteredModelCreate`, so I don't believe we support users setting the `id`.
- `RegisteredModel`'s `name` field is read-only, but user's should be allow to set it when registering a model.

api/openapi/model-registry.yaml
```diff
@@ -1244,17 +1205,6 @@ components:
                 since epoch.
               type: string
               readOnly: true
```
dhirajsb, line 1207: While it seems like a good idea, readOnly is only for documentation:
```
JSON Schema's readOnly field is primarily for documentation and API contract purposes and is not inherently enforced by Go's encoding/json package during serialization or deserialization. While the readOnly keyword can be included in a JSON schema definition, Go will not automatically prevent modification or ignore the field. Handling readOnly fields requires custom logic.
```
That's why the generated code has no enforcement or safeguard anywhere. We had investigated this in RedHat managed services and discovered this issue a few years ago.

While it seems like a good idea, users are going to completely miss the readOnly _tag_ and try to send those properties. Service code will also have to manually hand written to ignore/handle them appropriately. What should the service do when readonly fields are sent by the client when they shouldn't, should it throw an error/ignore them? 

This will make the API ambiguous and the implementation complicated. The current separate types schema gives us a fool proof separation of concerns, and simplifies our service implementation which is able to leverage generation of goverter helpers that automatically filter fields for us.

dhirajsb (changes requested):
We can't drop *Create types by using `readOnly` as it's just for documentation without any enforcement in implementation.

api/openapi/model-registry.yaml
rareddy: Thank you @dhirajsb for the blurb from the OpenAPI spec. I did not know this before.

I agree that `readOnly` does not provide strict enforcement. However, I respectfully disagree with the concern that this change makes the API ambiguous or complicated. On the contrary, it improves both the API’s readability and the experience of code generation and simplicity in extending the API.

Given that clients are accidentally sending readOnly properties, it indicates that API clients are not adhering to the API specification as intended. The enforcement behavior on the backend is an implementation detail, and I don't see a strong justification to complicate the API design because of it.

On the backend, we should explore
1. Silently ignore them if they were present in a request and/or
2. Find a way to generate the enforcement through our extensions for the code generation

While I appreciate the enforcement provided by the current approach, I also feel the API design is too closely tied to the domain model. IMO, prioritizing overall readability and making the API easier to extend in the future would add more long-term value.

Al-Pragliola:
as @pboyd said, when creating resources users shouldn't specify the `id` but the `name`

image

image

source: https://editor.swagger.io/?url=https://raw.githubusercontent.com/kubeflow/model-registry/37dcc466d329e97cc354dcbaeb78630387fd7769/api/openapi/model-registry.yaml

api/openapi/model-registry.yaml
Al-Pragliola: there's the possibility of using "not" , as a way to exclude a field from a verb like enable "name" in create and remove it in update

sources:

https://github.com/OAI/OpenAPI-Specification/issues/1497#issuecomment-370951318

https://github.com/OAI/OpenAPI-Specification/issues/1497#issuecomment-371249646

api/openapi/model-registry.yaml
rareddy: Corrected the `id` to be marked as `readonly` field, next I will see where I can add the validation to omit/reject the `readOnly` fields in `GoLang` generated code. The Python client code generates the `readOnly` fields validation correctly to reject the `readOnly` fields to omit.

rareddy:
> as @pboyd said, when creating resources users shouldn't specify the `id` but the `name`

The Swagger editor is not showing any object with `allOf` correctly in my investigation. is there a trick to it?

tarilabs:
> is there a trick to it?

not that I'm aware of, this was also the experience and findings we reported back then with @lampajr , hence why there was some back/forth, while required to maintain the established hierarchy.

In _some_ situations, we found that using composition with "anonymous" objects, helped the case to display in Swagger and have proper codegen ([link](https://github.com/opendatahub-io/model-registry-bf4-kf/pull/66#issuecomment-1759630182)) another reason why a not-deeply-nested hierarchy was recommended.

rareddy:
>why a not-deeply-nested hierarchy was recommended.

Being so popular, with only one level of nesting in this PR, the Swagger Editor behaving so poorly surprises me. Does nesting help at all beyond the design time? I was thinking, whatif, if there is a clever way to flatten the schema before generation of the code, but that may not be worth the time. :(

Al-Pragliola:
@rareddy found the problem

In allOf any subesequent ref overwrites the fields that are in common -- so:
name in BaseResource is correct
name in RegisteredModelUpdate gets the readOnly:true (so POST requests don't see anymore this field)

the way to fix this issue is to set in the anonymous object readOnly to false

```yaml
    RegisteredModel:
      description: A registered model in model registry. A registered model has ModelVersion children.
      required:
        - name
      allOf:
        - $ref: "#/components/schemas/RegisteredModelUpdate"
        - $ref: "#/components/schemas/BaseResource"
        - type: object
          properties:
            name:
              description: |-
                The client provided name of the model. It must be unique among all the RegisteredModels of the same
                type within a Model Registry instance and cannot be changed once set.
              type: string
              readOnly: false
```

image

image

dhirajsb:
The amount of work being put in by multiple devs to try to make this work correctly at runtime on the server side proves my argument about this approach being needlessly complex. 
Not to mention all the extra testing that'll be required to validate all this logic being added to jump unnecessary hoops.

It can also be argued that json schema readonly property itself is an anti-pattern. It's a property of a schema type, but makes *assumptions* about context in which it is used. It's adding behavior to what should otherwise be a data transfer object. 
E.g. if the code generation tooling supports ignoring such fields on the client side when marshaling it to the server, the generated type cannot be used to generate a json string to log a resource or print it on the console as it will skip readonly fields. So, that won't work for an `mrctl` like cli. 

IMHO, with my decades of experience with API design and use, no body has ever argued to deliberately make an API ambiguous and less clear and introduce assumptions that manifest at runtime. 

Even the supposed pros of reducing duplication and number of types are non-issues for MR API because we use json schema composition, and we also have distinct `*Update` types, which users will see as a simple naming pattern for resources. 
So, users are not going to be extraordinarily burdened by going from 2 request response types per resource to 3. Type drift between `*Create` and resource types is not an issue either as we follow a boilerplate pattern for creating these types for resources. 

So, I asked chatgpt to compare these API design options and this was the response:

### **🔍 Why Separate Schemas Are Preferable Long-Term**

#### **1\. Enhanced Clarity and Correctness**

Defining distinct schemas for requests and responses allows for precise specification of the data structures expected in each context. This separation minimizes ambiguity, ensuring that clients and servers have a clear contract for data exchange. As noted in the Unleash ADR:[Unleash Docs](https://docs.getunleash.io/contributing/ADRs/overarching/separation-request-response-schemas?utm_source=chatgpt.com)

"With dedicated schemas for requests and responses, we can define the precise structure and constraints for each interaction... ensuring that both parties are aligned in terms of data structures and expectations." [Unleash Docs](https://docs.getunleash.io/contributing/ADRs/overarching/separation-request-response-schemas?utm_source=chatgpt.com)

#### **2\. Improved Maintainability and Flexibility**

Separate schemas allow for independent evolution of request and response structures. This decoupling simplifies maintenance and reduces the risk of unintended side effects when modifying one part of the API. The same ADR highlights:

"By avoiding the reuse of schemas between requests and responses, we can modify and update them independently." [Unleash Docs](https://docs.getunleash.io/contributing/ADRs/overarching/separation-request-response-schemas?utm_source=chatgpt.com)

#### **3\. Alignment with RESTful Principles**

RESTful API design emphasizes clear, self-descriptive messages. Using separate schemas supports this by providing distinct representations for different operations, enhancing the API's expressiveness and adherence to REST constraints. The Richardson Maturity Model, for instance, advocates for such clarity in API interactions. [Wikipedia](https://en.wikipedia.org/wiki/Richardson_Maturity_Model?utm_source=chatgpt.com)

### **⚖️ Comparison: Single Schema with readOnly/writeOnly vs. Separate Schemas**

| **Aspect** | **Single Schema (readOnly/writeOnly)** | **Separate Schemas** |
| --- | --- | --- |
| **Clarity** | Moderate; relies on annotations | High; explicit definitions |
| --- | --- | --- |
| **Maintainability** | Lower; changes can have broader impact | Higher; changes are localized |
| --- | --- | --- |
| **Tooling Support** | Variable; depends on tool capabilities | Broad; aligns with standard practices |
| --- | --- | --- |
| **Flexibility** | Limited; shared structure constraints | Greater; tailored to specific use cases |
| --- | --- | --- |
| **REST Alignment** | Partial; may blur operation semantics | Strong; supports distinct representations |
| --- | --- | --- |

### **📚 Supporting Resources**

- **Unleash ADR on Schema Separation**: Discusses the rationale and benefits of separating request and response schemas. [Unleash Docs](https://docs.getunleash.io/contributing/ADRs/overarching/separation-request-response-schemas?utm_source=chatgpt.com)
- **OpenAPI Design Strategies**: Explores patterns for creating distinct request and response objects in OpenAPI. [Medium](https://medium.com/%40wtr/openapi-design-strategies-8a626e09e309?utm_source=chatgpt.com)
- **Swagger's Best Practices in API Design**: Emphasizes the importance of self-descriptive messages in RESTful APIs. [Wikipedia+1Swagger+1](https://en.wikipedia.org/wiki/REST?utm_source=chatgpt.com)

### **✅ Conclusion**

Adopting separate request and response schemas in your OpenAPI design promotes clearer contracts, easier maintenance, and better alignment with RESTful principles. While using readOnly and writeOnly properties can be suitable for simple cases, separating schemas provides a more robust and scalable foundation for complex APIs.

pboyd:
Just my $0.02, but I prefer separate types for read, create and update. My biggest problem dealing with the spec file (and I just mean working in the YAML file itself) isn't the types but trying to untangle the inheritance. For instance, I made this tree for myself when I was trying to modify `RegisteredModel` fields:

`RegisteredModel`:
- `BaseResource`:
	- `id`
	- `createTimeSinceEpoch`
	- `lastUpdateTimeSinceEpoch`
	- `BaseResourceCreate`:
		- `description`
		- `BaseResourceUpdate`:
			- `customProperties`
			- `description`
			- `externalId`
- `RegisteredModelCreate`:
	- `properties`
	- `BaseResourceCreate`: (above)
	- `RegisteredModelUpdate`:
		- `owner`
		- `state`
		- `BaseResourceUpdate`: (above)

It's easy to see the fields in the swagger docs or editor, so I doubt readers of the docs are bothered by this. But if I'm modifying this file, I have to dig through a few to layers to find its definition. And once I find the field I'm interested in, the only I can be sure that what I have is complete is by checking every node in the tree to see what might be overriding it.

I'd rather have a flatter structure that's easier to work with, even if that leads to more duplicated field definitions. But I think it's possible to flatten the structure without much duplication by composing the types from individual fields instead of using base types. Here's a concrete example: https://github.com/pboyd/model-registry/commit/1fd5f32934ee047940138e27906502e5b7df287b

That example defines some common fields up-front:

```yaml
  schemas:
    _common:
      type: object
      properties:
        id: &property-id
          format: int64
          description: The unique server generated id of the resource.
          type: string
        createTimeSinceEpoch: &property-createTimeSinceEpoch
          format: int64
          description: Output only. Create time of the resource in millisecond since epoch.
          type: string
          readOnly: true
...
```

Then composes types from those fields, overriding properties as needed:

```yaml
    RegisteredModel:
      description: A registered model in model registry. A registered model has ModelVersion children.
      type: object
      required:
        - name
      properties:
        customProperties: *property-customProperties
        description: *property-description
        externalId: *property-externalId
        name:
          <<: *property-name
          description: |-
            The client provided name of the model. It must be unique among all the RegisteredModels of the same
            type within a Model Registry instance and cannot be changed once set.
        id: *property-id
        createTimeSinceEpoch: *property-createTimeSinceEpoch
        lastUpdateTimeSinceEpoch: *property-lastUpdateTimeSinceEpoch
        owner: *property-owner
          type: string
        state: *property-state
          $ref: "#/components/schemas/RegisteredModelState"
    RegisteredModelCreate:
      description: A registered model in model registry. A registered model has ModelVersion children.
      required:
        - name
      type: object
      properties:
        customProperties: *property-customProperties
        description: *property-description
        externalId: *property-externalId
        name:
          <<: *property-name
          description: |-
            The client provided name of the model. It must be unique among all the RegisteredModels of the same
            type within a Model Registry instance and cannot be changed once set.
        owner: *property-owner
        state: *property-state
```

templates/go-server/model.mustache
```diff
@@ -109,4 +109,14 @@ func Assert{{classname}}Constraints(obj model.{{classname}}) error {
 {{/maximum}}
 {{/Vars}}
 	return nil
-}{{/model}}{{/models}}
+}
+
+// RemoveReadOnly{{classname}}Properties sets all readonly properties to nil
+func RemoveReadOnly{{classname}}Properties(obj *model.{{classname}}) {
```
pboyd, line 115: I don't think we want to expose this function outside the package.

```suggestion
// removeReadOnly{{classname}}Properties sets all readonly properties to nil
func removeReadOnly{{classname}}Properties(obj *model.{{classname}}) {
```

templates/go-server/model.mustache
```diff
@@ -109,4 +109,14 @@ func Assert{{classname}}Constraints(obj model.{{classname}}) error {
...
+{{#Vars}}
+{{#isReadOnly}}
+	obj.{{name}} = nil
```
pboyd, line 118: This won't work for all types unfortunately, you'll need to assign `""` to strings and `0` to numeric types.

rareddy:
The latest commit removes all the uses of `readOnly` except for in the `BaseResource` it adheres to most of original design minus the hierarchy and create types. And moves validation of readOnly flags serverside.

The current failures are due to mixed use of `POST` vs `PATCH` and use of `Artifact` Vs `ArtifactUpdate` method [here](https://github.com/kubeflow/model-registry/blob/de5f225d96a4daeca77506d233082b1c4ea5afa3/api/openapi/model-registry.yaml#L699) 

All the other parts of API in this spec are what I see as edge/leaf level calls at to a domain model, but this is particular `upsertModelVersionArtifact` call I find it as odd ball, trying to a `upsert` using POST (POST is always create new object) and then for the payload instead of using `ArtifactCreate` vs `ArtifactUpdate` object it uses `Artifact` object

I

templates/go-server/model.mustache
rareddy: sicne this is in a seperate file I thought I use capital; but right that is has package level scope; need to relearn a lot of Go :)

templates/go-server/model.mustache
rareddy: Good catch; will take a look

templates/go-server/model.mustache
rareddy: this is fixed now

templates/go-server/model.mustache
rareddy: It looks like what I have works as the Go structs define all the parameters as pointers; I tried configuring for the data type but that results in Go errors. so this good for now.

api/openapi/model-registry.yaml
rareddy: The current version generates the server side code to ignore the `readOnly` values when they are encountered in the `POST` and `PATCH` requests, thus keeping existing functionality as is.

pboyd:
@rareddy, all of my nit-picky comments have been addressed, thank you. I'll let you and @dhirajsb sort out the larger debate over read/create/update types.

rareddy:
>@rareddy, all of my nit-picky comments have been addressed, thank you. I'll let you and @dhirajsb sort out the larger debate over read/create/update types.

rareddy:
>@rareddy, all of my nit-picky comments have been addressed, thank you. I'll let you and @dhirajsb sort out the larger debate over read/create/update types.

Thank you, @pboyd, for your review. 

BTW, this was a simple change, thanks to Cursor helping me Python code. The quirkiness is only from handling the `type_assert.patch` that was bad before I understood how that worked.

syntaxsdev:
@rareddy As @dhirajsb  mentioned, my main concern is with the flexibility of those types/schemas. I have ran into this decision before - ultimately while it might "increase" readability, it decreases flexibility. I prefer leaner models as well, but sometimes it's not possible for scalability reasons.

Some solutions I have seen implemented before to aid with your 1st point of consolidation is to create a unified create/update type? If enough attrs are shared/updated evenly then it could work.

However to your second point it would still add those nested base models, but it would reduce file count.

lugi0:
I will not get into the overall design discussion, but @rareddy if you want to simplify the schema please ensure that the existing `links` that have been defined still make sense

rareddy:
> please ensure that the existing links that have been defined still make sense

@lugi0, please see #1055 and let me know your opinion.

rareddy:
As there is no consensus on this PR, closing to pursue what has been suggested.

Issue #719, opened 2025-01-20
Ensure Model Registry Achieves 100% Score for OpenSSF Best Practices Badge


milosjava:
As part of the CNCF Graduation Checklist, we need to ensure that the Model Registry achieves a 100% score in order to get OpenSSF Best Practices Badge. This is crucial for meeting the CNCF’s graduation requirements and demonstrating the maturity and readiness of the project.

Here's the current status and score: https://www.bestpractices.dev/en/projects/9937

for visibility: @tarilabs @akgraner

tarilabs:
thank you for sharing about this important topic in today's KF MR biweekly! we appreciate your role as liaison and making us aware of all these KF community topics 🙏

milosjava:
During the MR bi-weekly call today, a brief overview of the Best Practices Badge initiative was provided, and the team was asked if anyone would be willing to serve as a contact person to assist with the self-evaluation process.

tarilabs:
> willing to serve as a contact person to assist with the self-evaluation process

count on me as well @milosjava as mentioned 👍

tarilabs:
for the records, we introduced CNCF instance of FOSSA with:
- https://github.com/kubeflow/model-registry/issues/323

tarilabs:
today at the time of writing, we "moved the needle" of the badge from 19% to 34%, then sub-issues have been created for divide-et-impera and promote community contributions

tarilabs:
at the time of writing, we "moved the needle" of the badge to 63%

rareddy:
very nice!!! @tarilabs

tarilabs:
at the time of writing, we "moved the needle" of the badge to 81%

tarilabs:
at the time of writing, we "moved the needle" of the badge to 94%

tarilabs:
I'm very pleased to note that at the time of writing, we achieved 122% on the [badge](https://www.bestpractices.dev/en/projects/9937)! 🎉 🚀 🪐 

<img width="1195" alt="Image" src="https://github.com/user-attachments/assets/16591ec6-6761-49fa-85e5-e38b0861077d" />

[...]

PR #1011, opened 2025-04-22
chore: add OpenSSF Best Practices badge

pboyd:
Related to #719.

Signed-off-by: Paul Boyd <pboyd@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 # Model Registry
+
+![build checks status](https://github.com/kubeflow/model-registry/actions/workflows/build.yml/badge.svg?branch=main)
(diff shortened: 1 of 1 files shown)


tarilabs:
cc @milosjava wdyt?

thanks a lot @pboyd, I think this is quite helpful

we will need to identify how we can fill the questionnaire, @milosjava any indication how we can collectively work towards that?

tarilabs:
> we will need to identify how we can fill the questionnaire, @milosjava any indication how we can collectively work towards that?

answering my own question here, per KF CNCF Graduation call, KSC will create OpenSSF account and then share it with each WGs to fill autonomously

/approve
thanks again @pboyd

Issue #1027, opened 2025-04-24
Python client does not set description field of the model for register_model() call


dbasunag:
**Describe the bug**
Model Registry python client does not set description field of the model for register_model() call
**To Reproduce**
1. mr_temp = ModelRegistry(server_address="my own url", port="443",author="me", user_token="my own token", is_secure=False)
2.  mr_temp.register_model(name="temp", description="test desc", owner="me", uri="s3", model_format_version="v1",version="1.1.1", model_format_name="temp")
3. This is the response: RegisteredModel(id='4', description=None, external_id=None, create_time_since_epoch='1745525310940', last_update_time_since_epoch='1745525310940', custom_properties=None, name='temp', owner='me', state=<RegisteredModelState.LIVE: 'LIVE'>)

**Expected behavior**
description field should not be dropped.
A post call via rest api:
 % curl --silent -X 'POST' "https://<MR_REST_URL>:443/api/model_registry/v1alpha3/registered_models" -H 'accept: application/json' -H 'Content-Type: application/json' -d '{"name": "mymodel", "owner": "me", "state": "LIVE", "description" : "my desc"}' -H 'Content-Type: application/json'  -H "Authorization: Bearer <my token>" -k

{"createTimeSinceEpoch":"1745524876193","customProperties":{},"description":"my desc","id":"3","lastUpdateTimeSinceEpoch":"1745524876193","name":"mymodel","owner":"me","state":"LIVE"}

Subsequent get call also shows this:
 {
      "createTimeSinceEpoch": "1745524876193",
      "customProperties": {},
      "description": "my desc",
      "id": "3",
      "lastUpdateTimeSinceEpoch": "1745524876193",
      "name": "mymodel",
      "owner": "me",
      "state": "LIVE"
    },
**Additional context**

dbasunag:
Currently the description field seems to be aligned with model version description. Even though artifacts, register model, and version all supports description fields.

tarilabs:
per [other discussion](https://github.com/kubeflow/model-registry/pull/1029#issuecomment-2835727254), please check if the other functions, tailored to each logical model entity (eg: function to create/update a specific RegisteredModel) can fit for your use-case. if not, we can augment or create new ones as necessary, thanks!

tarilabs:
thank you for taking care that the attribute was properly managed in all needed functions, @dbasunag

PR #1029, opened 2025-04-25
Add support for description field for ModelVersion and ModelArtifact in python client

dbasunag:
https://github.com/kubeflow/model-registry/issues/1027

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -290,6 +290,8 @@ class ModelRegistry:
         owner: str | None = None,
         description: str | None = None,
         metadata: Mapping[str, SupportedTypes] | None = None,
+        model_version_description: str | None = None,
+        model_source_description: str | None = None,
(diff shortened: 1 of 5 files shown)


tarilabs (changes requested):
Thanks @dbasunag for this patch!

I'm not very favourable of this change however, since the intent of `register_model` is to offer a simple one-shot call to register a triplet of RegisterModel, ModelVersion and ModelArtifact, and it was decided the Description (like similarly customProperties) belonged to a ModelVersion, so that subsequent one-shot calls would NOT overwrite the RegisterModel attributes, but only for the _new_ ModelVersion.

A user shall/can always use detailed methods focused on RM, MV, MA if they want to change specifically a description any of the entities.

I would recommend instead of adding more parameters to this function, to check the domain model oriented function support this intent, or instead patch those as needed.

Hope this explanation make sense, and happy to hear your thoughts?

PR #1031, opened 2025-04-25
feat(py): Add support for OCI Auth/Login via MR Client

syntaxsdev:
Support to login to an OCI Registry (both src and dest registries) in MR client, and automatic detection of default environment variables to facilitate logins.

## Description
This PR adds support for OCI Auth Login in both:
1. environment variables
2. parameters supplied to `OCIParams` object

The default environment variable used is `.dockerconfigjson`
The `.dockerconfigjson` expects a "pull secret" style auth file as such:

```json
    {
        "auths": {
            "localhost:xxxx": {
            "auth": "base64 encoded auth here",
            "email": ""
            }
        }
    }
```

When passing `oci_username` and `oci_password` to the `OCIParams` you do not need to use base64.

The function is then wrapped in a secure error catcher so any failures will filter out sensitive data to keep the password/auth JSON from ending up in logs.

## How Has This Been Tested?

These changes were tested by use of `test_save_to_oci_registry_auth_params`
where a real skopeo BackendDefinition was used but the PullFn and PushFn were mocked and replaced with functions that check for the correct param usage (eg. `--src-password=passwordhere`) since the local registry in the E2E was not configured to have authentication.

This was also tested locally to confirm it worked for Quay.io and docker.io
## Merge criteria:

If you have UI changes


clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
+    """Pull function definition."""
+
+    def __call__(self, base: str, dest: Path, **kwargs) -> T: ...  # noqa: D102
+
+
+class PushFn(Protocol):
+    """Push function definition."""
+
+    def __call__(self, src: Path, oci_ref: str, **kwargs) -> T: ...  # noqa: D102
+
```
syntaxsdev, line 114: Converted the TypedDict of callables to linter-IDE and developer type hint friendly definition.. (also this way can eliminate calling via indexing (`["pull"]()` -> `.pull()`)

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
-    - pull(base_image: str, dest_dir: Path) -> None
-    - push(local_image_path: Path, oci_ref: str) -> None.
+    - pull(base_image: str, dest_dir: Path, **kwargs) -> T
+    - push(local_image_path: Path, oci_ref: str, **kwargs) -> T.
     """
 
-    available: Callable[[], bool]
-    pull: Callable[[str, Path], None]
-    push: Callable[[Path, str], None]
+    is_available: Callable[[], bool]
```
syntaxsdev, line 125: previous typo

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
+    pull: PullFn
+    push: PushFn
 
 
-def _get_skopeo_backend() -> BackendDefinition:
+def _kwargs_to_params(kwargs: dict[str, str]) -> list[str]:
```
syntaxsdev, line 130: converts kwargs to CLI style args (--arg=x)

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
+        params.extend(pull_args or [])
+
+        return _scrub_errors(lambda: skopeo_pull(base_image, dest, params))
+
+    def wrapped_push(src: Path, oci_ref: str, **kwargs) -> T:
+        kwargs = _backend_specific_params("skopeo", "push", **kwargs)
+        params = _kwargs_to_params(kwargs)
+        params.extend(push_args or [])
 
+        return _scrub_errors(lambda: skopeo_push(src, oci_ref, params))
```
syntaxsdev, line 160: wrap the `olot` skopeo functions so we can neatly pass them in

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
+        prefix = "--from" if type == "pull" else "--to"
+
+    # Actual param specifications
+    if username := kwargs.pop("username", None):
+        kwargs[f"{prefix}-username"] = username
+
+    if password := kwargs.pop("password", None):
+        kwargs[f"{prefix}-password"] = password
+
+    return kwargs
```
syntaxsdev, line 224: pretty much DRY. not too hard coded and expandable if olot adds other backends

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
+            msg = """Sensitive operation when running a command line process failed.
+                (secrets were detected and scrubbed)"""
+            raise RuntimeError(msg) from None
+        raise e
+    except Exception as e:
+        if determine_sensitivity(e):
+            msg = """Sensitive operation failed before running the command line process.
+                (secrets were detected and scrubbed)"""
+            raise RuntimeError(msg) from None
+        raise e
```
syntaxsdev, line 252: self explanatory, takes in callable and scrubs any errors that are sensitive.

clients/python/tests/test_utils.py
```diff
@@ -114,12 +118,38 @@ def test_save_to_oci_registry_with_custom_backend(
         custom_oci_backend=get_mock_custom_oci_backend,
     )
     # Ensure our mocked backend was called
-    get_mock_custom_oci_backend["is_available"].assert_called_once()
-    get_mock_custom_oci_backend["pull"].assert_called_once()
-    get_mock_custom_oci_backend["push"].assert_called_once()
+    get_mock_custom_oci_backend.is_available.assert_called_once()
+    get_mock_custom_oci_backend.pull.assert_called_once()
+    get_mock_custom_oci_backend.push.assert_called_once()
```
syntaxsdev, line 123: like mentioned, much nicer now internally

clients/python/src/model_registry/utils.py
tarilabs: super handy, thanks!

clients/python/src/model_registry/utils.py
tarilabs: helpful, I would have even separated earlier, ie avoid first determining prefix, and for each backend spell-out the full argument, but even in current form is very okay and we can always come back later 👍

clients/python/tests/conftest.py
```diff
@@ -237,13 +238,44 @@ def get_mock_custom_oci_backend():
...
+
+
+@pytest.fixture
+def get_mock_skopeo_backend_for_auth(monkeypatch):
+    # ZmFrZV91c2VyOmFwYXNzd29yZGhlcmUK = fake_user:passwordhere (in base64)
+    auth_json = """
+    {
+        "auths": {
+            "localhost:5001": {
+            "auth": "ZmFrZV91c2VyOmFwYXNzd29yZGhlcmUK",
```
tarilabs, line 258: I would recommend to build the base64 encoded string in code, so to avoid "security scan" false positive on the repo when a base64 literal is detected?


ie something ~like:

```py
value = base64.b64encode(b'fake_user:passwordhere')
json = f' ... {value.decode("utf-8")} ...'
```

or similar, wdyt?

syntaxsdev: I agree on this.

syntaxsdev:
thanks for review :D I am still working on README.

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,155 @@ def s3_uri_from(
...
 
 
-def _get_skopeo_backend() -> BackendDefinition:
+def _kwargs_to_params(kwargs: dict[str, str]) -> list[str]:
+    """Convert kwargs to list of params.
+
+    Args:
+        kwargs: The keyword args dict.
+    """
+    return [f"{k}={v}" for k, v in kwargs.items()]
```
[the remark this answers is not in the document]
tarilabs: to complement, something like:

```py
import subprocess

def go():
    variable1 = "asd"
    variable2 = "asdf"
    value = [ "--key1", variable1, "--key2", variable2 ]

    try:
        subprocess.run(["skopeo", "copy", *value, "unknown"], check=True)
    except subprocess.CalledProcessError as e:
        raise ValueError("Problem with command", e.returncode, e.output, e.stderr) from None

if __name__=="__main__":
    go()
```

imho would keep it consistent to expectations, without the need of scrubbing?

```
% python test.py
FATA[0000] unknown flag: --key1                         
Traceback (most recent call last):
  File "/Users/mmortari/git/test.py", line 14, in <module>
    go()
  File "/Users/mmortari/git/test.py", line 11, in go
    raise ValueError("Problem with command", e.returncode, e.output, e.stderr) from None
ValueError: ('Problem with command', 1, None, None)
```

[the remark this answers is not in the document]
syntaxsdev: I like this approach, thanks

clients/python/tests/test_utils.py
```diff
@@ -84,15 +84,19 @@ def test_save_to_oci_registry_with_skopeo(get_temp_dir_with_models, get_temp_dir
...
+    save_to_oci_registry(
+        base_image=base_image,
+        oci_ref=oci_ref,
+        model_files_path=dest_dir,
+        dest_dir=get_temp_dir,
+        custom_oci_backend=utils._get_skopeo_backend(
+            push_args=[
+                "--dest-tls-verify=false",
+                "--src-username=non_user",
+                "--src-password=nonpassword",
```
syntaxsdev, line 96: both approaches work by default.

```python
push_args=[
  "--src-username="x",
]
# or
push_args=[
  "--src-username", "x"
]
```

clients/python/src/model_registry/utils.py
tarilabs: I think this is more related to how we need to pass down args to the "backends" (ie skopeo) but Gila is imho a good suggestion in general for config managment for MR py client as a whole, +1

tarilabs (commented):
thanks a lot @syntaxsdev 

/assign @Crazyglue 
given this changes some part of the orchestration to support this use-case

/lgtm
waiting for chime-in

clients/python/src/model_registry/utils.py
dhirajsb: It'll give us more options for how to pass args via either configmaps, env variables, or cmdline for the future copy model jobs feature.

clients/python/src/model_registry/_client.py
```diff
@@ -248,8 +248,12 @@ def upload_artifact_and_register_model(
                 **asdict(upload_params), path=model_files_path
             )
         elif isinstance(upload_params, OCIParams):
+            dict_params = asdict(upload_params)
+            del dict_params["custom_oci_backend"]
             destination_uri = save_to_oci_registry(
-                **asdict(upload_params), model_files_path=model_files_path
+                **dict_params,
+                custom_oci_backend=upload_params.custom_oci_backend,
+                model_files_path=model_files_path,
```
Crazyglue, line 256: I'm not sure i follow the purpose behind this change, isnt the old code equivalent to the new code?

clients/python/src/model_registry/utils.py
```diff
@@ -2,17 +2,23 @@
...
 
-from typing_extensions import overload
+from typing_extensions import Literal, overload
 
 from ._utils import required_args
 from .exceptions import MissingMetadata, StoreError
 
+# Generic return type
+T = TypeVar("T")
+
```
Crazyglue, line 21: There is a `types/` folder, should this live there?

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,144 @@ def s3_uri_from(
...
+    - pull(base_image: str, dest_dir: Path, **kwargs) -> T
+    - push(local_image_path: Path, oci_ref: str, **kwargs) -> T.
     """
 
-    available: Callable[[], bool]
-    pull: Callable[[str, Path], None]
-    push: Callable[[Path, str], None]
+    is_available: Callable[[], bool]
+    pull: PullFn
+    push: PushFn
```
Crazyglue, line 127: Why do this? doesn't this mean the `PullFn` definition above must be instantiated before it is called? Wouldn't that impact consumption of the `custom_oci_backend` as a result?

ie, usage would look like this

```
pull = PullFn()
pull()
```

clients/python/src/model_registry/utils.py
```diff
@@ -95,41 +101,144 @@ def s3_uri_from(
...
+
+    Keyword Args:
+        username: the usrername of the registry.
+        password: the password of the registry.
+    """
+    # Determine backend
+    if backend == "skopeo":
+        prefix = "--src" if type == "pull" else "--dest"
+    elif backend == "oras":
+        prefix = "--from" if type == "pull" else "--to"
```
Crazyglue, line 219: Not urgent, but this might have scaling issues if we add additional backends or want to support things other than 'push' and 'pull'. It might be better to encode these params as data rather than branching logic, however that is of course additional work that may never actually be needed. Up to you!

clients/python/src/model_registry/utils.py
```diff
@@ -278,3 +421,44 @@ def get_files_from_path(path: str) -> list[tuple[str, str]]:
...
+    try:
+        auth_json = json.loads(auth_data)
+        if type(auth_json) is not dict:
+            msg = ""
+            raise TypeError(msg)
+        registries = auth_json["auths"]
+        reg_keys = list(registries.keys())
+        if len(reg_keys) > 1:
+            msg = f"Auth JSON has multiple registries ({', '.join(reg_keys)}). This is not supported."
+            raise ValueError(msg)
```
Crazyglue, line 453: Why would this not be supported? Shouldn't it just be a matter of checking the hostname of the OCI registry the user is attempting to upload to against the entries in the dockerconfig file?

afaik the shape of this particular part of the json is like so

```
"auths": {
    "quay.io": {
      "auth": "base64-encoded-credentials"
    },
    "another-registry.com": { ... }
```

clients/python/src/model_registry/utils.py
syntaxsdev: I wouldn't think it's necessary to put there, it's essentially "instantiating" the generic (`<T>` in other languages), so its odd to import the generic, it's typeless and has no definition.

clients/python/src/model_registry/utils.py
syntaxsdev: I see your concern, it might be the way I named.
`PullFn` is just a definition of what the function should be like.

It's essentially the same as what you did here:
https://github.com/kubeflow/model-registry/blob/ffa006bc7043d3286b8697fda0e0392d15c82e08/clients/python/src/model_registry/utils.py#L107

except I changed the type to be a `Protocol`, this helps the IDE for type-hinting for developers or those who want to expand the functionality and define their own custom definition, because they will know immediately if their `PullFn` or `PushFn` meets the right criteria.

PullFn does not need to be instantiated, it is just a callable reference of the function.  It just needs to be the same definitions of the protocol.

Ex: I pass this function
https://github.com/kubeflow/model-registry/blob/79fb84db45b173d2dc69ffcdd992b1f21e277c9d/clients/python/src/model_registry/utils.py#L152

to `BackendDefinition` here:
https://github.com/kubeflow/model-registry/blob/79fb84db45b173d2dc69ffcdd992b1f21e277c9d/clients/python/src/model_registry/utils.py#L166-L168

clients/python/src/model_registry/utils.py
syntaxsdev: If we ever need a third branch, I'll swap for a lookup tree 🌳😉

clients/python/src/model_registry/utils.py
Crazyglue: Understood! Thanks for the explanation

clients/python/src/model_registry/utils.py
syntaxsdev: yes.. I've thought on that too. the issue I found was that too much custom logic is too much extra logic is needed for that..

for example:
1. user supplies domainless image+tag. do we just assume it's going to be from docker.io?
2. (rare*) what if user is pushing and pulling from different registries, how would we know which one is going to be authenticated 
3. it's for user safety, avoiding guessing or automatically using (odh off/topic discussion)
4. can also not use the same env var, you just supply a different `OCIParams` and set the env var to use based on use case

happy to take more feedback on this @tarilabs @Crazyglue but I think for now we can suffice once we have more data and use cases? don't want to overbuild

Crazyglue: Understood and agree, when we have use cases fleshed out more we can always adjust things!

Crazyglue (approved):
/lgtm

clients/python/src/model_registry/_client.py
syntaxsdev: yes but because I changed the type of BackendDefinition from a TypedDict to an object that holds functions 

using `asdict` will verbatim try to convert those the protocol definitions to dicts


what I'm doing here is converting ahead of time and then replacing the definition as is again

clients/python/src/model_registry/utils.py
syntaxsdev: Good idea @dhirajsb, maybe we implement that in another feature later?

clients/python/src/model_registry/utils.py
tarilabs: I'd recommend in a followup PR, since this type of Secret (ODH Connection, etc) are per-registry anyway

tarilabs (approved):
- https://github.com/kubeflow/model-registry/pull/1031#pullrequestreview-2800430001
 
we can indeed follow-up in subsequent PR for any broader scope changes

/approve

Issue #1018, opened 2025-04-23
[MLMD Removal] Write DB migrations to recreate MLMD schemas and types using golang-migrate


Al-Pragliola:
### Description

The goal of this task is to implement database migration scripts using the golang-migrate tool in order to recreate the MLMD schemas and types directly within the Model Registry’s MySQL database. All necessary tables, indexes, constraints, and type mappings provided by MLMD should be replicated to ensure that the Model Registry API remains unchanged.

The migration scripts must support forward migrations to apply the new schema and rollback operations to revert if necessary. This ensures a safe transition process that maintains performance and data integrity throughout the deployment cycle.

### Acceptance Criteria

- Schema Accuracy

  - The migration scripts must accurately recreate all MLMD schemas including tables, columns, indexes, constraints, and custom types as currently exposed by MLMD.

- Tool Integration

  - Migrations must be written for and executed using the golang-migrate tool. They must successfully run on a fresh MySQL instance, as well as in integration testing environments.

- Rollback Capability

  - For every forward migration, a corresponding rollback (down migration) script must be provided to allow for a safe reversal without data loss or corruption.

- API and Performance Consistency

  - Integration tests must verify that the new internal metadata service interacting with the migrated database is fully compliant with the existing Model Registry API, and that performance does not degrade compared to the legacy MLMD-driven setup.

- Documentation and Testing

  - Complete documentation outlining the execution and rollback process for the migrations must be provided.
  - integration testing results should confirm that the recreated schema supports all required CRUD operations as expected.

Al-Pragliola:
closed by https://github.com/kubeflow/model-registry/pull/1107 and https://github.com/kubeflow/model-registry/pull/1148

PR #1107, opened 2025-05-13
feat: create the embedmd datastore type

Al-Pragliola:
part of #1018

## Description

Please note that EmbedMD is a temporary name and is up to debate. :sweat_smile:

In this PR, we add a new datastore type called EmbedMD. It is a simple, embedded metadata service that (in the future) will be used to store metadata for the Model Registry.

This PR includes all the migrations needed to recreate the MLMD database schema,
this should make the new datastore retro-compatible with the MLMD database.

Two new flags have been added to the proxy command:
- `embedmd-database-type`: The type of database to use for the EmbedMD datastore.
- `embedmd-database-dsn`: The DSN for the EmbedMD database.

So for example, to run the proxy with an EmbedMD datastore, on a MySQL database, you can run the following command:

```shell
./model-registry proxy --hostname=localhost --port=8080 --datastore-type=embedmd --embedmd-database-dsn="root:test@tcp(127.0.0.1:3306)/metadb?charset=utf8mb4&parseTime=True&loc=Local" --embedmd-database-type=mysql
```

## How Has This Been Tested?

I have tried in a local cluster two things:

1) deploy the usual model registry with MLMD, set the deployment replicas to 0 and run the command above. Migrations have been applied successfully, then set the deployment replicas to 1 and the model registry has been started successfully.

2) start with an empty database and run the command above. Migrations have been applied successfully and then use the new database with the usual model registry deployment with MLMD, it worked as expected.

also I added two new integration tests for the migrations (up and down) with the usual command:

```shell
make test
```

## Merge criteria:


internal/datastore/embedmd/service.go
```diff
@@ -0,0 +1,65 @@
...
+
+	migrator, err := db.NewDBMigrator(s.DatabaseType, connectedDB)
+	if err != nil {
+		return nil, err
+	}
+
+	err = migrator.Migrate()
+	if err != nil {
+		return nil, err
+	}
```
pboyd, line 58: Will this still work if we have more than one pod?

internal/datastore/embedmd/mysql/migrations/000011_create_mlmdenv_table.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS `MLMDEnv` (
+  `schema_version` int NOT NULL,
+  PRIMARY KEY (`schema_version`)
+);
```
pboyd, line 4: Doesn't golang migrate keep track of the schema version itself (in the `schema_migrations` table, iirc)?

internal/datastore/embedmd/mysql/migrate.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if *steps < 0 {
+		return fmt.Errorf("steps cannot be negative")
+	}
+
+	return m.migrator.Steps(*steps)
+}
+
+func (m *MySQLMigrator) Down(steps *int) error {
```
pboyd, line 76: Does anything call this (except for the tests)? Is there a way to recover from a failed migration?

pboyd (commented):
I have a couple questions about the migrations, but it looks good.

internal/datastore/embedmd/service.go
Al-Pragliola: It should be safe, `golang-migrate` uses the DB driver locking mechanism when the DB supports it, I tried it locally with three replicas, and it went smoothly. However, this is something we can include in the e2e tests once the rest of the feature is ready

internal/datastore/embedmd/mysql/migrations/000011_create_mlmdenv_table.up.sql
Al-Pragliola: Yes, in this phase, I am trying to support switching from MLMD to our system and vice versa. MLMD needs this table; otherwise, it will crash. In phase two, when we can drop compatibility with MLMD, we can also remove this table

internal/datastore/embedmd/mysql/migrate.go
Al-Pragliola: Not for now. My idea was to add a command to the model registry to handle migrations manually in case of failure. However, it's something that I would rather put in another pull request

syntaxsdev:
so is this like a temporary-migration holding period just to be off of MLMD?

Al-Pragliola:
> so is this like a temporary-migration holding period just to be off of MLMD?

Hey @syntaxsdev,

This is more than a temporary solution. It's the second step towards getting off MLMD. The plan is as follows:

1) Create an abstraction layer for metadata stores (because the code was bound to MLMD). This has been done in #425
2) Create a new metadata store type (included in this PR)
3) Migrate an existing or empty DB to the same schema as MLMD for backwards compatibility reasons (included in this PR)
4) Replace MLMD's functionality by implementing the datastore interface using GORM
5) Test/document
6) Remove MLMD

Al-Pragliola:
> Why start with SQL directly, why not start with tagged ORM structs that we'll use with Gorm anyway?

:wave: @dhirajsb,
 
Two reasons (in my opinion)

1) **Decoupling**: I wanted to keep the dependencies from the migration and ORM libraries to a minimum so that, if one or both fail to meet our expectations, they can be easily replaced independently of each other. Also, ORM structs should only reflect the currently used schema version. If we need to add columns to one of those tables in the future, we would need a struct for the older type in the older migration and a newer one for the current schema.

2) **Control**: I wanted more control over the migrations because it is often sensible to run them with direct SQL. This ensures that complex data changes are handled exactly right, and we can use any specific database command we need.

dhirajsb:
Thanks @Al-Pragliola for the detailed response. See comments below:

> Decoupling: I wanted to keep the dependencies from the migration and ORM libraries to a minimum so that, if one or both fail to meet our expectations, they can be easily replaced independently of each other. 

That's what the DB interface is for, to abstract the DB implementation. We don't need to further abstract out the ORM layer  and feels like a YAGNI situation. 

> Also, ORM structs should only reflect the currently used schema version. If we need to add columns to one of those tables in the future, we would need a struct for the older type in the older migration and a newer one for the current schema.

That is sort of correct. The reason for using the ORM is to avoid writing DB specific DDL/SQL to make it as portable as possible to all Gorm supported DB types. 
We can actually handle schema changes and older/newwe structs without much overhead. See this [code example](https://github.com/dhirajsb/kas-fleet-manager/blob/d375e9edc139eec04cb44e3e4515fb6c6eaf08a0/internal/connector/internal/migrations/migrations.go) for a RH control plane. 
There is a starting DB schema, which is in sync with the main DB schema to start. Only changes for updated types need to be in migrations from that point on. We can also use raw SQL if needed. 
IMHO, since the mlmd DB schema is so generic, there _shouldn't_ :crossed_fingers: be that many changes to the schema in the future. 

> Control: I wanted more control over the migrations because it is often sensible to run them with direct SQL. This ensures that complex data changes are handled exactly right, and we can use any specific database command we need.

As shown in the sample migrations strategy above, we still retain control over types and have access to raw SQL if needed. So, we aren't necessarily sacrificing anything by using Gorm types to create the initial schema.

dhirajsb:
BTW, the sample migration implementation I shared was for a continuously released SaaS. So, we had to be able to turn around changes asap if we ran into bugs/feature fixes, etc.

Al-Pragliola:
Thanks for your detailed feedback and for sharing the example, @dhirajsb ! I appreciate you taking the time to explain your perspective on using GORM for schema management.

You've made some good points about GORM's capabilities for portability and handling schema evolution with structs. My initial reasoning for choosing golang-migrate with SQL files stemmed from a few key considerations, and perhaps I can clarify how they relate to the lifecycle of migrations themselves.

One of the primary motivations for using a dedicated tool like golang-migrate was to leverage its robust, out-of-the-box functionalities for managing the entire migration process. This includes critical aspects like:

- Database Locking: golang-migrate drivers (e.g., for MySQL, PostgreSQL) typically implement database-level locking. This ensures that if multiple instances of our application (like multiple pods in Kubernetes) try to run migrations simultaneously at startup, only one proceeds, preventing race conditions and ensuring migrations apply cleanly and sequentially.

- Atomic Version Tracking: It maintains a dedicated schema_migrations table to reliably track which versioned migrations have been applied, and whether the database is in a "dirty" state.

- Explicit Up and Down Migrations: For every schema change (.up.sql), there's a corresponding .down.sql script. This provides clear, explicit, and testable rollback paths for each versioned change, which is invaluable for operational stability.

And also because it's the tool the team approved to use in _RHOAIENG-22102_

The decoupling aspect I mentioned earlier also ties into this: golang-migrate handles the "how and when" of migration execution (locking, versioning, up/down), while the SQL (or even GORM DDL within a Go migration called by such a system) defines the "what." If the "how and when" is tied too closely to the ORM's schema definition tools, it feels like we might lose some of this specialized process machinery.

I'm definitely open to discussing this further and seeing if there's an approach that gives us the best of both worlds! Perhaps there's a way to leverage GORM for defining the schema state within a migration step while still using golang-migrate's core engine for the overall process management.

But in my initial research I didn't find an easy way to integrate the two tools without major rewriting of golang-migrate functionalities.

Thanks again for the discussion!
:wave:

rareddy:
+1 for keeping both DDL and GORM decoupled. For me, 
- It separates concerns effectively.
- The schema is explicit and easy to understand, especially from a management perspective. No need to read the code to understand what the schema is. 
- If we need to make any database-specific changes for performance or maintainability, this approach makes it easier to do so.
- While we're currently using Go, if we ever need to switch to Python or support Python libraries directly (as MLMD did), this design will allow us to adapt quickly.

dhirajsb:
If the team has made the final decision on going with golang migrate and using direct DDL, that's fine by me. I do have some questions:
1. Are we still able to support different DB types like Gorm does, i.e. MySQL, PostgreSQL, SQLite, SQL Server, or is the DDL tied to a DB dialect? 
2. Is there a way we can check whether the Gorm structs that we'll add later are in sync with the _current_ schema from go-migrate?

Al-Pragliola:
Thanks for these excellent follow-up questions, @dhirajsb . They touch upon important aspects of choosing a migration strategy. Let me try to address them:

1.
> Are we still able to support different DB types like Gorm does, i.e. MySQL, PostgreSQL, SQLite, SQL Server, or is the DDL tied to a DB dialect?

You're right to point out GORM's strength in generating DDL that's portable across the database types it supports. With the current approach of using direct DDL in SQL files with `golang-migrate`, the migration scripts are indeed tied to a specific SQL dialect, but the code is ready to expand on different DB types as there's already a layer of abstraction on top of it. Here's the list of DB types supported by `golang-migrate`:

- PostgreSQL
- PGX v4
- PGX v5
- Redshift
- Ql
- Cassandra / ScyllaDB
- SQLite
- SQLite3
- SQLCipher
- MySQL / MariaDB
- Neo4j
- MongoDB
- CrateDB
- Shell
- Google Cloud Spanner
- CockroachDB
- YugabyteDB
- ClickHouse
- Firebird
- MS SQL Server
- rqlite

2. 

> Is there a way we can check whether the Gorm structs that we'll add later are in sync with the current schema from go-migrate?

A key aspect that `golang-migrate` brings to the table here is explicit schema versioning. It establishes a versioned "source of truth" for the database schema through its migration history, which is tracked in the schema_migrations table. Our GORM structs are then intended to reflect this specific, versioned state of the schema (typically the latest version deployed with the application).

We can use https://gorm.io/gen/database_to_structs.html, as we do with OpenAPI, to generate the types and methods of the REST API. We can use the DB with schema version X as a source of truth to generate the GORM structs. If this fails to meet our expectations, we can write the structs by hand and build a solid unit/integration test suite to ensure everything is in sync.

tarilabs:
> With the current approach of using direct DDL in SQL files with golang-migrate, the migration scripts are indeed tied to a specific SQL dialect, but the code is ready to expand on different DB types as there's already a layer of abstraction on top of it. Here's the list of DB types supported by golang-migrate:

impressive list; just noting from a Requirement perspective we're being asked to support Sqlite, Postgresql, MySQL/MariaDB.

rareddy:
> we're being asked to support Sqlite, Postgresql, MySQL/MariaDB.

Yes

>  the migration scripts are indeed tied to a specific SQL dialect, but the code is ready to expand on different DB types as there's already a layer of abstraction on top of it.

I am confused about this abstraction. Can you share any details about it, any ADR doc etc, please, to learn more about it? Or an example. The immediate question I have is whether we maintain separate DDL scripts for each database or a common ANSI and then vendor-specific extensions? On the DML side, I am less concerned, as most can be ANSI.

rareddy:
Thank you for the explanation @Al-Pragliola, that makes sense. so one needs to keep all supported database schemas in sync.  Also is API --> DML also per database?

dhirajsb:
Thanks for the detailed response @Al-Pragliola 

> We can use https://gorm.io/gen/database_to_structs.html, as we do with OpenAPI, to generate the types and methods of the REST API. We can use the DB with schema version X as a source of truth to generate the GORM structs. If this fails to meet our expectations, we can write the structs by hand and build a solid unit/integration test suite to ensure everything is in sync.

Yeah, if we can either generate the orm schema or validate it, we should be good! :+1:

Al-Pragliola:
> Thank you for the explanation @Al-Pragliola, that makes sense. so one needs to keep all supported database schemas in sync. Also is API --> DML also per database?

For API --> DML this is not necessary, as we leverage GORM's capabilities to handle different sql-dialects

rareddy (approved):
I see all questions answered and concensus with design so I am going to approve this

/lgtm
/approve

Issue #1075, opened 2025-05-07
Model Registry Client Cleanup/Hardening


syntaxsdev:
Model Registry python client has a few soon deprecating functions/logic from other libraries that should be cleaned up, along with ensuring other docs issue.

**Purpose:**
Clean up and enhancement MR client for developer experience and eventual deprecations.

Acceptance criteria:

1. Ensure all deprecations are taken care of
2. Verify, change and/or move older logic
3. Standardized pydocs

PR #1111, opened 2025-05-14
refactor: General MR Client Improvements & Hardening

syntaxsdev:
closes https://github.com/kubeflow/model-registry/issues/1075
Reduce tech debt in client and increase developer experience both as library and while developing.

## Description

1. Updated docs
3. Rename and ensure utility functions are correctly named
4. Refactor S3 helper funcs into `utils.py`
5. Add support for large model uploads to S3
6. Tested support for large models uploads regressively
7. Fixed lingering Temp dir from older client release generated when a default path is used
8. Other general fixes and improvements

## How Has This Been Tested?

Tests added:
`tests/test_client.py`
Fixture added to:
`conftest.py`

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -604,17 +604,14 @@ class ModelRegistry:
             access_key_id=access_key_id,
             secret_access_key=secret_access_key,
         )
-        try:
-            return self.__upload_to_s3(
-                path=path,
-                path_prefix=s3_prefix,
-                bucket=bucket_name,
-                s3=s3,
-                endpoint_url=endpoint_url,
-                region=region,
-            )
-        except Exception as e:
-            raise e
+        return self.__upload_to_s3(
+            path=path,
+            path_prefix=s3_prefix,
+            bucket=bucket_name,
+            s3=s3,
+            endpoint_url=endpoint_url,
+            region=region,
+        )
 
     def __upload_to_s3(  # noqa: C901
         self,
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -9,7 +9,7 @@ import os
 from collections.abc import Coroutine, Mapping
 from dataclasses import asdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, TypeVar, Union, get_args
+from typing import Any, Callable, TypeVar, Union, get_args
 from warnings import warn
 
 from .core import ModelRegistryAPIClient
@@ -25,7 +25,10 @@ from .types import (
 from .utils import (
     OCIParams,
     S3Params,
-    get_files_from_path,
+    _connect_to_s3,
+    _get_files_from_path,
+    _s3_creds,
+    _upload_to_s3,
     s3_uri_from,
     save_to_oci_registry,
 )
@@ -56,10 +59,6 @@ logging.basicConfig(
 logger = logging.getLogger("model-registry")
 
 DEFAULT_USER_TOKEN_ENVVAR = "KF_PIPELINES_SA_TOKEN_PATH"  # noqa: S105
-
-# If we want to forward reference
-if TYPE_CHECKING:
-    from botocore.client import BaseClient
 
 
 class ModelRegistry:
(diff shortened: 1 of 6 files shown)


clients/python/src/model_registry/utils.py
```diff
@@ -357,8 +366,10 @@ def save_to_oci_registry(  # noqa: C901 ( complex args >8 )
         msg = f"Backend '{backend}' is selected, but not available on the system. Ensure the dependencies for '{backend}' are installed in your environment."
         raise ValueError(msg)
 
+    dest_dir_cleanup = False
     if dest_dir is None:
         dest_dir = tempfile.mkdtemp()
+        dest_dir_cleanup = True
```
syntaxsdev, line 372: @Crazyglue I know you wrote the initial imp, do you think this is good to delete? I tested and still green.

clients/python/src/model_registry/utils.py
```diff
@@ -375,16 +386,191 @@ def save_to_oci_registry(  # noqa: C901 ( complex args >8 )
...
+    has_env_creds = aws_access_key_id and aws_secret_access_key
+    has_param_creds = access_key_id and secret_access_key
+    if not (has_env_creds or has_param_creds):
+        msg = """Environment variables `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY` were not set.
+            Please either set these environment variables or pass them in as parameters using
+            `access_key_id` or `secret_access_key`.
+            """
+        raise ValueError(msg)
+
+    return endpoint_url, access_key_id, secret_access_key, region
```
syntaxsdev, line 448: moved from `client.py`

clients/python/src/model_registry/utils.py
```diff
@@ -375,16 +386,184 @@ def save_to_oci_registry(  # noqa: C901 ( complex args >8 )
...
+    )
+    return client(
+        "s3",
+        endpoint_url=endpoint_url,
+        aws_access_key_id=access_key_id,
+        aws_secret_access_key=secret_access_key,
+        region_name=region,
+        config=config,
+    ), transfer_config
+
```
syntaxsdev, line 564: moved from `client.py` with SOME changes

clients/python/src/model_registry/utils.py
```diff
@@ -375,16 +386,184 @@ def save_to_oci_registry(  # noqa: C901 ( complex args >8 )
...
+def _upload_to_s3(  # noqa: C901
+    path: str,
+    bucket: str,
+    s3: BaseClient,
+    path_prefix: str,
+    *,
+    endpoint_url: str | None = None,
+    region: str | None = None,
+    transfer_config: TransferConfig | None = None,
```
syntaxsdev, line 459: this is new, we now pass in a TransferConfig object.

this allow S3 to handle multipart uploads of large models

clients/python/tests/conftest.py
```diff
@@ -220,6 +220,31 @@ def get_temp_dir_with_nested_models():
...
+            chunk = os.urandom(min(chunk_size, target_size - bytes_written))
+            f.write(chunk)
+            bytes_written += len(chunk)
+
+    yield temp_dir
+
+    # Cleanup
+    shutil.rmtree(temp_dir)
+
+
```
syntaxsdev, line 247: create a large test file/model fixture

@dbasunag @lugi0 wdyt about this for d/s?

lugi0: This looks ok to me, we could adopt the same in downstream for sure. Just a couple thoughts:
1) Any specific reason to write to file in chunks? I'd guess that `os.urandom(target_size)` would work just as well and simplify the code slightly
2) If we wanted to test real world scenarios with the models going around today we'd probably need this to be in the multiple/hundreds GB range - although not sure those models would be stored in S3 in the first place. Perhaps it'd be worth having the target_size as an input to the fixture?

syntaxsdev: 1. Chunking keeps RAM usage low instead of immediately allocating 400MB at once. The CI env is constrained on memory, so it gets the job done without spiking RAM, and its more realistic to save patterns.

2. Yes true... However if you try to upload 200MB + with previous implementation, it would indeed fail, because there was no TransferConfig set to chunk-ify the data sent to that degree. We can indeed have a `target_size` to show, but we definitely need to be mindful of the CI limits.

lugi0: Ah, right, I was not thinking about the CI env limits at all. It all makes sense, thanks for the reply!

lugi0: Indeed, we have already identified it as a gap that needs to be covered, I'd say upstream is not the place for it for sure though, at the very least not through its CI

syntaxsdev: agreed. I'll leave out the `target_size` expansion for a later larger stress test effort. Is that good?

clients/python/src/model_registry/utils.py
Crazyglue: Would it be possible to do like a `with ...:` block instead? iirc, we would be able to cleanup the temp dir at the end of that block automatically

syntaxsdev:
> @syntaxsdev nice work with the chunk transfer support. Considering the size of the file, would it make sense to split utils.py further for helping new devs/users reading the code?

I agree @dhirajsb. I mostly wanted to get it out of the `_client.py` because it definitely didn't belong there. I am not opposed to splitting it out further, but there are definitely more things that can help make dev onboarding better we can capture in later efforts. I can log a ticket for it and add ideas as they come along for that. wdyt?

clients/python/src/model_registry/_client.py
```diff
@@ -573,6 +570,9 @@ def save_to_s3(
         access_key_id: str | None = None,
         secret_access_key: str | None = None,
         region: str | None = None,
+        multipart_threshold: int = 1024 * 1024,
+        multipart_chunksize: int = 1024 * 1024,
+        max_pool_connections: int = 10,
```
tarilabs, line 575: great extension to the capability thanks @syntaxsdev , which is also covered by e2e test below (and yes we can always expand for stress testing)

clients/python/src/model_registry/utils.py
tarilabs: I think even with a `with ...` block, we'd still need to distinguish between: 
- user has provided oci-layout destination directory (hence, no delete)
- user has NOT provided oci-layout destination directory, set as temp (existing logic) but adding a flag to cleanup later

I see this as a very welcome improvement but not strictly a requirement

tarilabs (commented):
thank you @syntaxsdev , leaving it a bit further open for any last comment/or tracking of follow-ups, then will proceed to merge

syntaxsdev:
| We be expanding such tests in d/s

Is there a required change for this? @dbasunag

clients/python/src/model_registry/utils.py
```diff
@@ -375,16 +386,184 @@ def save_to_oci_registry(  # noqa: C901 ( complex args >8 )
...
+
+    # Check if we have credentials from either source
+    has_env_creds = aws_access_key_id and aws_secret_access_key
+    has_param_creds = access_key_id and secret_access_key
+    if not (has_env_creds or has_param_creds):
+        msg = """Environment variables `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY` were not set.
+            Please either set these environment variables or pass them in as parameters using
+            `access_key_id` or `secret_access_key`.
+            """
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
syntaxsdev: thank you for this, this really clears up some of the tedious logic initially

dbasunag (commented):
looks good from my side.

dbasunag:
> Is there a required change for this? @dbasunag
Nothing as of now. If QE needs something, we will reach out.

tarilabs (approved):
thanks again @syntaxsdev and all for 👀 

- https://github.com/kubeflow/model-registry/pull/1111#pullrequestreview-2855089730
- https://github.com/kubeflow/model-registry/pull/1111#pullrequestreview-2845639933
- https://github.com/kubeflow/model-registry/pull/1111#pullrequestreview-2844447836

/approve

PR #1146, opened 2025-05-22
Add base model fields to registered model

pboyd:
## Description
This extends PR #939 to add fields to `RegisteredModel` that will eventually be part of the catalog. If anyone feels these changes should be in #939 instead, let me know and I'll update that PR and close this one.

## How Has This Been Tested?
Registering a model through the API using the new fields.

## Merge criteria:

- [x] All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)
- [x] The commits have meaningful messages
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).

PR #1193, opened 2025-06-09
fix: allow omitted fields in MLMD types

pboyd:
## Description

The recently added fields to RegisteredModel cause the pod to crash in clusters created without those fields. Allowing omitted fields in MLMD avoids the problem.

The error was:

```
F0609 15:10:31.711149       1 root.go:45] error: error connecting to datastore: error creating MLMD types: error setting up context type kf.RegisteredModel: rpc error: code = AlreadyExists desc = Type already exists with different properties: can_omit_fields is false while stored type has more properties: stored type:
 id: 10
name: "kf.RegisteredModel"
...
```

## How Has This Been Tested?
I upgraded an old cluster with the new code and witnessed the problem. Then I replaced the image after this update and verified that the pod started and model-registry was functional.

## Merge criteria:

### DIFF
--- a/internal/mlmdtypes/mlmdtypes.go
+++ b/internal/mlmdtypes/mlmdtypes.go
@@ -18,6 +18,7 @@ type MLMDTypeNamesConfig struct {
 	InferenceServiceTypeName   string
 	ServeModelTypeName         string
 	CanAddFields               bool
+	CanOmitFields              bool
 }
 
 func NewMLMDTypeNamesConfigFromDefaults() MLMDTypeNamesConfig {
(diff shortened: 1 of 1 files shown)


pboyd:
@tarilabs this is to fix the error I was talking about earlier.

tarilabs (commented):
thanks a lot @pboyd for your proactivity here

there are some test failures, which need be fixing.

I'm not sure I can exactly pin-point to the reason why we need to omit fields... we had used the `CanAddFields` to support adding new properties, but I'm wondering now which fields require the `CanOmitFields`?

pboyd:
Never mind about this. The field `long_description` was added in an early version of #1146 but removed before it was merged. So applying the final list of types to an environment that had the old list caused this error about an omitted field. This was probably isolated to just my dev environments.

I'll follow up with a PR to add a test that migrates from the old list of fields to the new list.

tarilabs:
thanks for being through @pboyd , this is great value to the project 🙏  I appreciated your proactivity even tho we don't have to do code modification at this time.

PR #1205, opened 2025-06-12
feat: implement EmbedMD business logic

### LINKED ISSUE #1020: [MLMD Removal] Write the CRUD operations necessary to the datastore abstraction using GORM
### Description

This task involves implementing the CRUD operations for our datastore abstraction layer using GORM. The focus is on the following domain types:

- registered_model
- model_version
- artifact
- model_artifact
- serving_environment
- inference_service

The objective is to create a clean, interface-driven repository layer in which each domain type has its own fully functional set of CRUD operations. This layer should encapsulate all database interactions using GORM and be consistent with our existing data store configuration, including connection parameters, error handling, and transaction management.

### Acceptance Criteria

- Interface and Implementation:

  - Develop a common repository interface that abstracts these operations for all six types. The actual implementation using GORM should strictly follow idiomatic patterns to ensure that upper layers remain decoupled from the underlying database specifics.

- Domain Modeling:

  - Ensure that each domain type is modeled correctly according to our data schema. The model definitions should handle relationships, keys, and indexes as required.

- Error Handling and Transaction Management:

  - Each operation must include robust error handling and proper transaction controls, where needed, to guarantee atomicity and consistency. Use logging to provide meaningful context in case operations fail.

- Testing & Documentation:

  - Existing unit, integration, e2e tests should pass using the new metadata service instead of MLMD.

Al-Pragliola:
## Description

closes #1020 

In this PR, I have implemented the core logic of the new datastore, all the existing entities are covered:

- registered_model
- model_version
- artifact
- model_artifact
- serving_environment
- inference_service
- serve_model

## How Has This Been Tested?

Two ways:

1) the simplest one is to just run `make test` as several tests have been added to check if the new datastore is doing what MLMD was doing before

2) a bit more complicated
  
    edit the existing manifests to switch to embedmd
    
    **base/model_registry_configmap.yaml**
    ```yaml
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: model-registry-configmap
      labels:
        component: model-registry-server
    data:
      MODEL_REGISTRY_REST_SERVICE_HOST: "model-registry-service"
      MODEL_REGISTRY_REST_SERVICE_PORT: "8080"
      MODEL_REGISTRY_DATA_STORE_TYPE: "embedmd"
      MODEL_REGISTRY_GRPC_SERVICE_HOST: "model-registry-service"
      MODEL_REGISTRY_GRPC_SERVICE_PORT: "9090"
    ```
    **overlays/db/kustomization.yaml**
    ```yaml
    apiVersion: kustomize.config.k8s.io/v1beta1
    kind: Kustomization
    
    resources:
    - model-registry-db-pvc.yaml
    - model-registry-db-deployment.yaml
    - model-registry-db-service.yaml
    - ../../base
    
    
    configMapGenerator:
    - envs:
      - params.env
      name: model-registry-db-parameters
    secretGenerator:
    - envs:
      - secrets.env
      name: model-registry-db-secrets
    generatorOptions:
      disableNameSuffixHash: true
    
    
    images:
    - name: mysql
      newName: mysql
      newTag: "8.3"
    
    patches:
    - path: patches/model-registry-deployment.yaml
    replacements:
    - source:
        fieldPath: data.MYSQL_PORT
        kind: ConfigMap
        name: model-registry-db-parameters
        version: v1
      targets:
      - fieldPaths:
        - spec.template.metadata.annotations.[traffic.sidecar.istio.io/excludeOutboundPorts]
        select:
          group: apps
          kind: Deployment
          name: model-registry-deployment
          version: v1
    ```
    
    **overlays/db/patches/model-registry-deployment.yaml**
    ```yaml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: model-registry-deployment
    spec:
      template:
        metadata:
          annotations:
            # db doesn't use istio
            traffic.sidecar.istio.io/excludeOutboundPorts: MYSQL_PORT_PLACEHOLDER
        spec:
          containers:
            - name: rest-container
              # Remove existing environment variables
              env:
                - $patch: replace
              envFrom:
                - configMapRef:
                    name: model-registry-configmap
                - secretRef:
                    name: model-registry-db-secrets
                - configMapRef:
                    name: model-registry-db-parameters
              args:
                - --hostname=0.0.0.0
                - --port=$(MODEL_REGISTRY_REST_SERVICE_PORT)
                - --datastore-type=$(MODEL_REGISTRY_DATA_STORE_TYPE)
                - --embedmd-database-dsn=$(MYSQL_USER_NAME):$(MYSQL_ROOT_PASSWORD)@tcp(model-registry-db:$(MYSQL_PORT))/$(MYSQL_DATABASE)?charset=utf8mb4&parseTime=True&loc=Local
            - name: grpc-container
              $patch: delete
    ```
    run
    ```shell
    docker build . -f Dockerfile -t docker.io/mr/mr:0.2.31
    cd clients/python
    DEV_BUILD=true IMG=docker.io/mr/mr:0.2.31 make test-e2e
    ```

## Merge criteria:


internal/converter/openapi_embedmd_converter_util.go
```diff
@@ -0,0 +1,752 @@
...
+	idAsString := strconv.FormatInt(int64(*id), 10)
+	return &idAsString
+}
+
+// Int64ToInt32 converts int64 to int32 if numeric, otherwise return error
+func Int64ToInt32(id *int64) (*int32, error) {
+	if id == nil {
+		return nil, nil
+	}
+	idInt32 := int32(*id)
```
pboyd, line 35: Make sure `*id` is less than `math.MaxInt32`.

Makefile
```diff
@@ -137,10 +137,10 @@ stop/mysql:
 
 # generate the gorm structs
 .PHONY: gen/gorm
-gen/gorm: bin/gorm-gen bin/golang-migrate start/mysql
-	@(trap '$(MAKE) stop/mysql' EXIT; \
+gen/gorm: bin/golang-migrate start/mysql
+	@(trap 'cd $(PWD) && $(MAKE) stop/mysql' EXIT; \
```
pboyd: `$(CURDIR)` would help if someone uses `make -C`:

```suggestion
	@(trap 'cd $(CURDIR) && $(MAKE) stop/mysql' EXIT; \
```

internal/converter/embdemd_openapi_converter.go
pboyd, line 1: This can be a problem for another day, but I'm just looking at the file sizes for the new files package. There are ~5000 lines of code in the new manually maintained files (only 1500 excluding tests--but it needs the tests) and only 541 lines of generated code. Is `goverter` worth it? It seems to require exactly the sort of tedious code that it promised to save us from.

internal/converter/openapi_mlmd_converter_util.go
```diff
@@ -114,8 +114,12 @@ func PrefixWhenOwned(ownerId *string, entityName string) string {
 	} else {
 		prefix = uuid.New().String()
 	}
-	prefixedName := fmt.Sprintf("%s:%s", prefix, entityName)
-	return prefixedName
+
+	if entityName == "" {
+		return prefix
+	}
```
tarilabs, line 120: great catch, although I'm now wondering if we should really allow nameless logical model entities.

internal/ptr/utils.go
```diff
@@ -0,0 +1,13 @@
...
+	if e == nil {
+		return *new(E)
+	}
+
+	return *e
+}
+
+func Of[E any](e E) *E {
+	return &e
+}
```
tarilabs: didn't we have this one already here? 🤔 
https://github.com/kubeflow/model-registry/blob/e712d445b076954d8f4c58acaf0eb69235702d40/internal/apiutils/api_utils.go#L50-L53

internal/ptr/utils.go
```diff
@@ -0,0 +1,13 @@
+package ptr
+
+func In[E any](e *E) E {
+	if e == nil {
+		return *new(E)
+	}
+
+	return *e
+}
```
tarilabs: same comment

https://github.com/kubeflow/model-registry/blob/e712d445b076954d8f4c58acaf0eb69235702d40/internal/apiutils/api_utils.go#L43-L48

tarilabs (commented):
extensive work @Al-Pragliola congrats!!

I tried to test locally but with Podman I believe I'm facing newer issues with the TestContainers even when I try to change manually:

https://github.com/kubeflow/model-registry/blob/c945ab5d167be688b295eff6a4ba181af2cd4039/internal/testutils/test_container_utils.go#L22

couple minor comments below, in general
/lgtm
and I would like to see other people testing this out on the their systems.

internal/core/artifact.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	// If modelVersionId is nil, we need to generate a new UUID for the model version ID for the name prefixing,
+	// the id will still be nil when invoking the Save method so that the attribution is created only when needed.
+	if modelVersionId == nil {
+		uuid := uuid.New().String()
+
+		modelVersionId = &uuid
+	} else {
+		convertedId, err := strconv.Atoi(*modelVersionId)
```
pboyd: Be careful with these `int64` to `int32` conversions. I see a few instances of this, and I'm not going to point them all out, but anywhere that the value from `Atoi` is cast to an `int32` should be corrected.

```suggestion
		convertedId, err := strconv.ParseInt(*modelVersionId, 10, 32)
```

pboyd (commented):
I'm still working through this, just some initial observations.

internal/core/test_utils.go
```diff
@@ -0,0 +1,111 @@
...
+	"gorm.io/gorm"
+)
+
+// setupTestDB creates a MySQL testcontainer with migrations for testing
+func setupTestDB(t *testing.T) (*gorm.DB, func()) {
+	ctx := context.Background()
+
+	mysqlContainer, err := cont_mysql.Run(
+		ctx,
+		"mysql:5.7",
```
pboyd, line 24: We use MySQL 8.3 elsewhere. Do we need to support all the way back to 5.7?

internal/db/models/pagination.go
```diff
@@ -0,0 +1,50 @@
...
+	return p.NextPageToken
+}
+
+func (p *Pagination) GetOrderBy() *string {
+	if p.OrderBy == nil {
+		return nil
+	}
+
+	return p.OrderBy
+}
```
pboyd: If you keep the pointers, these don't need a `nil` check:

```suggestion
func (p *Pagination) GetNextPageToken() *string {
	return p.NextPageToken
}

func (p *Pagination) GetOrderBy() *string {
	return p.OrderBy
}
```

internal/db/models/pagination.go
pboyd, line 1: Do the methods in this file really need to return pointers? The logic in `internal/db/scopes/paginate.go` would be simpler with non-pointer types. And IMO you're asking for trouble by returning pointers to `DefaultSortOrder` and `DefaultPageSize` (I don't see any assignments right now, but that's not to say someone won't try it one day).

internal/ptr/utils.go
Al-Pragliola: Thanks I have missed this function :pray:

internal/ptr/utils.go
Al-Pragliola: Thanks I have missed this function 🙏

gorm-gen/main.go
Chanakya-TS, line 1: Would it make more sense to pass the database type as a flag (like we did previously)? Since this custom generator will also support Postgres and potentially other databases in the future. (Maybe also the DSN, although I see that it can be overridden)

gorm-gen/main.go
```diff
@@ -0,0 +1,75 @@
...
+	// Database connection configuration
+	dsn := "root:root@tcp(localhost:3306)/model-registry?charset=utf8mb4&parseTime=True&loc=Local"
+
+	// Allow DSN override via environment variable
+	if envDSN := os.Getenv("GORM_GEN_DSN"); envDSN != "" {
+		dsn = envDSN
+	}
+
+	// Connect to database
+	db, err := gorm.Open(mysql.Open(dsn), &gorm.Config{})
```
Chanakya-TS: Would it make more sense to pass the database type as a flag (like we did previously)? Since this custom generator will also support Postgres and potentially other databases in the future. (Maybe also the DSN, although I see that it can be overridden)

internal/converter/openapi_embedmd_converter_util.go
Al-Pragliola: thanks @pboyd , nice catch

internal/core/artifact.go
Al-Pragliola: Thanks I changed all the Atoi occurrencies

gorm-gen/main.go
Al-Pragliola: Thanks for your review @Chanakya-TS !
I made some changes to gorm-gen. Please check to see if they address your concerns

internal/converter/embdemd_openapi_converter.go
Al-Pragliola: Yeah I agree with you this is a valid concern and we should try to lessen the burden, I also have a plan to refactor the code from this PR (after the epic is done) there's much repetition that can be avoided but I wanted to have everything covered by tests and reviews before touching it 🤔

internal/converter/openapi_mlmd_converter_util.go
Al-Pragliola: yeah that's something that I discovered while running tests, feels a bit strange 🤔

internal/db/service/registered_model_test.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+// Helper functions
+func stringPtr(s string) *string {
+	return &s
+}
+
+func int32Ptr(i int32) *int32 {
+	return &i
+}
```
pboyd: Can you use the functions from your `ptr` package instead?

internal/core/test_utils.go
Al-Pragliola: A question for @rareddy or @tarilabs 

I have followed 

image


from https://docs.redhat.com/en/documentation/red_hat_openshift_ai_cloud_service/1/html/managing_model_registries/creating-a-model-registry_managing-model-registries

gorm-gen/main.go
Chanakya-TS: I just reviewed the latest changes to gorm-gen — everything looks great now!

dhirajsb:
@Al-Pragliola this is a big PR and it's going to take me a while to go through it all. Due to some other tight deadlines, I don't have the bandwidth to do it this week. You are getting good feedback from other team members though.

As long as the present MLMD schema functionality is maintained, we can iron out any implementation tweaks and improvements later. Thanks for the great work on this. :+1:

internal/db/service/registered_model_test.go
Al-Pragliola: yes, addressed in the latest commit (thanks)

internal/core/test_utils.go
tarilabs: I'd say for the beginning is best if we can keep the same version "requirement", we can move the needle past 8.4 once we have at least one LTS out.

tarilabs (approved):
thanks a lot @Al-Pragliola 

- https://github.com/kubeflow/model-registry/pull/1205#pullrequestreview-2931926064
- https://github.com/kubeflow/model-registry/pull/1205#discussion_r2150648080
- https://github.com/kubeflow/model-registry/pull/1205#issuecomment-2978938126
- https://github.com/kubeflow/model-registry/pull/1205#pullrequestreview-2939112294

/approve

PR #1209, opened 2025-06-13
Initial Model Catalog API

pboyd:
## Description
This incorporates @dhirajsb's changes from #725 and extends them to use the updated API spec. 

The list sources endpoint (`/api/model_catalog/v1alpha1/sources`) is implemented (except for pagination). Other endpoints will be implemented in follow up PRs.

## How Has This Been Tested?

Manual tested locally and unit tests.

## Merge criteria:
- [x] All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)


api/openapi/catalog.yaml
```diff
@@ -48,7 +48,7 @@ paths:
           $ref: "#/components/responses/NotFound"
         "500":
           $ref: "#/components/responses/InternalServerError"
-      operationId: getAllCatalogModels
+      operationId: listModels
```
dhirajsb: Let's keep the operation names uniform across registry and catalog.

dhirajsb:
@pboyd Although catalog and registry share common types, do we want to keep the implementation in a common codebase. I mean they don't have much in common beyond that. Do we want to create a sub-directory for the catalog project and keep the code bases isolated? That gives us the option of breaking it out into a separate go project and CI/releases, etc. Thoughts?

api/openapi/catalog.yaml
pboyd: They've been renamed. I think I captured the MR naming convention, but let me know if they need to be adjusted further.

pboyd:
> @pboyd Although catalog and registry share common types, do we want to keep the implementation in a common codebase. I mean they don't have much in common beyond that. Do we want to create a sub-directory for the catalog project and keep the code bases isolated? That gives us the option of breaking it out into a separate go project and CI/releases, etc. Thoughts?

That seems reasonable, I'll get started on it.

catalog/internal/catalog/catalog.go
```diff
@@ -0,0 +1,142 @@
...
+
+// CatalogSourceConfig is a single entry from the catalog sources YAML file.
+type CatalogSourceConfig struct {
+	model.CatalogSource `json:",inline"`
+
+	// Catalog type to use, must match one of the registered types
+	Type string `json:"type"`
+
+	// private properties used for configuring the catalog connection based on catalog implementation
+	PrivateProperties map[string]any `json:"privateProperties,omitempty"`
```
rareddy: Can we call this `properties`?

catalog/internal/catalog/yaml_catalog.go
```diff
@@ -0,0 +1,96 @@
...
+
+	model "github.com/kubeflow/model-registry/catalog/pkg/openapi"
+)
+
+type YamlArtifacts struct {
+	Protocol string `yaml:"protocol"`
+	URI      string `yaml:"uri"`
+}
+
+type YamlModel struct {
```
rareddy, line 18: Why is this "YamlModel"? Should this be elevated to something like "CatalogModel"? In general the comment applies everywhere YAML is mentioned

catalog/internal/catalog/yaml_catalog.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (y yamlCatalogImpl) GetModel(ctx context.Context, name string) (model.CatalogModel, error) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (y yamlCatalogImpl) ListModels(ctx context.Context, params ListModelsParams) (model.CatalogModelList, error) {
+	//TODO implement me
+	panic("implement me")
```
rareddy, line 40: To start with at least provide the whole list, deferring the query portion. or implement some simple name filtering

catalog/internal/catalog/yaml_catalog.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type yamlCatalogImpl struct {
+	contents *YamlCatalog
+	source   *CatalogSourceConfig
+}
+
+func (y yamlCatalogImpl) GetModel(ctx context.Context, name string) (model.CatalogModel, error) {
+	//TODO implement me
+	panic("implement me")
```
rareddy, line 35: Can we provide impl here?

rareddy (changes requested):

catalog/internal/catalog/catalog.go
pboyd: Yes. I've renamed it.

catalog/internal/catalog/yaml_catalog.go
pboyd: This is the implementation for the static YAML catalog source, it wouldn't make sense to call it `CatalogModel`. We could call it `StaticCatalog`, if you like that better.

catalog/internal/catalog/yaml_catalog.go
pboyd: That's part of another ticket, can we do that in another PR?

catalog/internal/catalog/yaml_catalog.go
pboyd: Same as above. That's part of another ticket, can we do that in another PR?

catalog/internal/catalog/yaml_catalog.go
rareddy: I see. This only represents the Yaml Catalog Model. I would suggest not having different version of CatalogModel per source. There is one defined at the API layer, and try to convert the current YAML file into that format. Otherwise, we need to keep writing converters for one another, which seems unnecessary at this stage. The YAML file was hand-built, so we will not be disturbing any automation process in place. The UI already needs to make adjustments for the Catalog API based model.

catalog/templates/go-server/api.mustache
```diff
@@ -0,0 +1,34 @@
+{{>partial_header}}
+package {{packageName}}
+
+import (
```
rareddy: Are these mustache files new? I see in them in the current repo. Did any spacing change? If it is I suggest a commit with formatting and a commit with changes to reflect what is truly changing.

catalog/cmd/catalog.go
```diff
@@ -0,0 +1,44 @@
...
+	RunE:  runCatalogServer,
+}
+
+func init() {
+	CatalogCmd.Flags().StringVarP(&catalogCfg.ListenAddress, "listen", "l", catalogCfg.ListenAddress, "Address to listen on")
+	CatalogCmd.Flags().StringVar(&catalogCfg.ConfigPath, "catalogs-path", catalogCfg.ConfigPath, "Path to catalog source configuration file")
+}
+
+func runCatalogServer(cmd *cobra.Command, args []string) error {
+	sources, err := catalog.LoadCatalogSources(catalogCfg.ConfigPath)
```
rareddy, line 34: In seperate PR we need to look into if can configure such that when the configmap for the sources changes, it triggers a bounce server.

catalog/internal/catalog/yaml_catalog.go
pboyd: We need this struct because that's how the YAML decoder works. In general, we won't be adding a `CatalogModel` for each source unless (like here) we need a structure to hold data specific to that source.

catalog/templates/go-server/api.mustache
pboyd: Yes, per Dhiraj's comment, the catalog is isolated as much as possible from the registry. That required duplicating these files. The only difference is the package in one `import`.

catalog/internal/catalog/yaml_catalog.go
rareddy: I am confused, can you pls explain :( If I reformat `catalog/internal/catalog/testdata/test-yaml-catalog.yaml` file to the same YAML as that supports `[CatalogModel](https://github.com/kubeflow/model-registry/pull/1209/files#diff-bb094295a21224ad92abff10ccf99c352240f9343d0270b2403fe6b3c0451b2d)` would that not suffice, is my question.

catalog/templates/go-server/api.mustache
rareddy: Duplication is not desirable IMO. Why not elevate the templates to higher-level directory? and generate the files in separate directories?

catalog/internal/catalog/yaml_catalog.go
pboyd: I can't remove it completely because we need the `artifacts` in the YAML, but not in the API response (it's a different endpoint in the API). But I can de-duplicate it like [this](https://github.com/kubeflow/model-registry/pull/1209/commits/6426b54efc5960c187f3469c9cc88d08d4c2777b).

rareddy: Sounds good.

catalog/templates/go-server/api.mustache
pboyd: I updated it to remove the duplicated templates.

api/openapi/model-registry.yaml
```diff
@@ -1731,6 +1731,7 @@ components:
         - CREATE_TIME
         - LAST_UPDATE_TIME
         - ID
+        - NAME
```
Al-Pragliola: overall looks good to me, I have only one doubt here, if we add the Name in the order by fields, should we also implement it in all the GetAll functions?

pboyd: Good point, thanks. Unfortunately, MLMD doesn't support sorting by name, so I can't easily add it. But I split up the `orderBy` type so that we can have different fields in registry and catalog and removed this line from the model registry version.

rareddy: Let's capture this as an add-on after MLMD removal, and get this PR merged now.

PR #1211, opened 2025-06-16
Migration error handling

adysenrothman:
This PR adds a readiness check to the new schema_migrations table as part of the MLMD removal. It ensures that the dirty state is false for the current version before allowing traffic to the model registry. If it is in a dirty state, it logs an error in the pod which tells the user to manually fix and force the version for that table. This error requires manual intervention by the user.

### DIFF
--- a/internal/proxy/readiness_test.go
+++ b/internal/proxy/readiness_test.go
@@ -105,8 +105,8 @@ func TestReadinessHandler_EmbedMD_Dirty(t *testing.T) {
 	err = migrator.Migrate()
 	require.NoError(t, err)
 
-	// manually set dirty flag
-	err = db.Exec("INSERT INTO schema_migrations (version, dirty) VALUES (1, 1)").Error
+	// manually set latest migration to dirty
+	err = db.Exec("UPDATE schema_migrations SET dirty = 1").Error
 	require.NoError(t, err)
 
 	ds := datastore.Datastore{


tarilabs (commented):
thanks @adysenrothman !!

in general 
/lgtm
 
If feasible without much additional efforts I would recommend adding at least a couple of tests so we "pin" that readiness probe is only available for the embedmd.
Otherwise, as we will migrate away from MLMD and we will look into E2E testing, we shall include testing this endpoint there.
wdyt?

adysenrothman:
> thanks @adysenrothman !!
> 
> in general /lgtm
> 
> If feasible without much additional efforts I would recommend adding at least a couple of tests so we "pin" that readiness probe is only available for the embedmd. Otherwise, as we will migrate away from MLMD and we will look into E2E testing, we shall include testing this endpoint there. wdyt?

I think that is a good point! I will add some testing for this new logic.

internal/proxy/readiness.go
```diff
@@ -0,0 +1,82 @@
...
+			db, err = connector.Connect()
+		default:
+			http.Error(w, fmt.Sprintf("unsupported database type: %s", dbType), http.StatusServiceUnavailable)
+			return
+		}
+
+		if err != nil {
+			http.Error(w, fmt.Sprintf("database connection error: %v", err), http.StatusServiceUnavailable)
+			return
+		}
```
Al-Pragliola: I think here you can use directly the db.NewConnector(dbType, dsn) and then .Connect()

internal/proxy/readiness_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	req, err := http.NewRequest("GET", "/readyz/isDirty", nil)
+	require.NoError(t, err)
+
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+
+	assert.Equal(t, http.StatusOK, rr.Code)
+	assert.Equal(t, "ok", rr.Body.String())
+}
```
Al-Pragliola, line 67: I think it would be also useful to test a EmbedMD case here

PR #1242, opened 2025-06-24
Add api tests stateless

fege:
## Merge criteria:

If you have UI changes


tarilabs:
(sync'd with Federico offline for a small suggestion)

clients/python/tests/test_api.py
```diff
@@ -0,0 +1,40 @@
...
+    if case.path == "/api/model_registry/v1alpha3/model_versions/{modelversionId}/artifacts":
+        case.path_parameters["modelversionId"] = "test-model-version-1"
+
+        if case.method == "POST":
+            case.body = {
+                "artifactType": "model-artifact",
+                "name": "test-artifact",
+                "uri": "s3://test-bucket/test-artifact",
+                "metadata": {}
+            }
```
lugi0: I would need to double check the internal logic of the library again since it's been some time, but I think that this is not needed. 
From the openAPI schema the library should be able to understand that these fields are required in the request it sends to the endpoint. 
If the fields are not set to required, or if the library somehow does not pick it up automatically, we might need to either revisit the schema or the implementation of the library. 
Could you please double check what happens if these lines are not used?

If instead you are trying to force specific values here, IMHO for the _stateless_ test it is not needed as we are effectively fuzzing the API with random data, and I'd instead suggest implementing this (and additional logic) in the _stateful_ test.

lugi0 (approved):
In general LGTM, however I do have a small comment on how additional data is fixed for the test.

clients/python/tests/test_api.py
```diff
@@ -0,0 +1,30 @@
...
+schema = schemathesis.pytest.from_fixture("generated_schema")
+
+
+@schema.parametrize()
+@settings(
+    deadline=None,
+    suppress_health_check=[
+        HealthCheck.filter_too_much,
+        HealthCheck.too_slow,
+        HealthCheck.filter_too_much,
```
dbasunag: this seems like a duplicate entry (same as line 12).

clients/python/pyproject.toml
```diff
@@ -29,6 +29,7 @@ eval-type-backport = "^0.2.0"
 huggingface-hub = { version = ">=0.20.1,<0.34.0", optional = true }
 olot = { version = "^0.1.6", optional = true }
 boto3 = { version = "^1.37.34", optional = true }
+urllib3 = ">=1.25.4,<1.27"
```
tarilabs: can we do without pinning urllib3 to the user?
note: this is listed as a project dependency.

tarilabs:
Thanks a lot really @fege for the efforts and this work, as well to all collaborations "offline" with @Al-Pragliola and QE folks

/lgtm
/approve

PR #1249, opened 2025-06-26
feat(catalog): implement the getCatalogModel endpoint

pboyd:
## Description
Implement `/api/model_catalog/v1alpha1/sources/{source_id}/models/{model_name}` for static catalogs.

## How Has This Been Tested?
Locally with kind and the included unit tests.

## Merge criteria:


api/openapi/catalog.yaml
```diff
@@ -76,6 +76,78 @@ paths:
           $ref: "#/components/responses/InternalServerError"
       operationId: findSources
       description: Gets a list of all `CatalogSource` entities.
+  /api/model_catalog/v1alpha1/sources/{source_id}/models/{model_name_a}/{model_name_b}:
+    description: >-
+      The REST endpoint/path used to get a `CatalogModel` for sources that have slashes in the model name. See `/api/model_catalog/v1alpha1/sources/{source_id}/models/{model_name}`.
```
rareddy: oh! no.. We should not consider any additions to the API. Slight development modifications it is OK. If you make changes to the API, we need to revise them first!

Coming to this specific one, IMO model names do not have "/" in them. If they have a slash, then it should match to their `provider/grouping/organization/namespace` classification IMO. I also realize the issue you are running into, which we failed to capture earlier.  It seems we put in more emphasis on the source-based identification of the model, but compared to HF API style, this does not seem correct to me.

How about we do not identify based on models on `source` but take `source` as param if there are duplicates? essentially then group and model name become identifiers across Catalogs. i.e. allow duplicated names across different sources. cc: @dhirajsb @kevinhowell

pboyd: I don't consider this to be a change to the API. Not really, anyway. They are new endpoints as far as Open API goes, but it's merely expanding `model_name` to allow a `/`. I'm not aware of another way to express that in Open API (supposedly v4 of the Open API spec will support it: https://github.com/OAI/OpenAPI-Specification/issues/2653).

Perhaps I don't understand your suggestion. But the "source" in the URL is necessary as that identifies which implementation to perform the lookup in. We're limited on what we can do because we don't control the external ID's, and unfortunately several of the planned integrations use `{org}/{model}` style names. If we want to assign IDs to things, then we'll probably need some sort of storage, which sounds like a much bigger redesign than this.

If these endpoints are an absolute non-starter, we can require `/`'s to be URL-escaped (e.g. `mistralai%2fMistral-7B-Instruct-v0.3`), or swap `/` for some other character (`;`, perhaps). But `/` seems correct, it's just hard to handle in Open API.

rareddy: Please ignore my comment about removing `source` and its `id` that could be a separate conversation altogether, not needed now.

What I would like to see is not adding extensions to the current API. What if tomorrow the model name is `/google/gemini/gemini-3b-xxx` and `/google/gemx/gemini-7b-xxx`? There will be no end to variations.

1. Yes, URL encoding is one way to fix this, but it may not be intuitive. However, as long as we control the endpoints from the catalog to all models, this is good. Also I do not see any reason why, user needs to manually URL-encode this, so this seems like a good option.
2. Maybe another option is to change the `router` and intercept any URL `/models/ibm/granite-3b-xxx` grab `/ibm/granite-3b-xxx` as a glob and route appropriately.

pboyd: I found a way to do what we need without new routes.

rareddy: great! thank you for finding the solution.

Al-Pragliola (approved):
/lgtm

PR #1256, opened 2025-06-30
async-job: S3 -> OCI synchornization

Crazyglue:
## Description
This will add much of the functionality to perform the s3 -> OCI model synchornization job. That configuration parsing is there, so this will just add the relevant functions to get the MR client, perform the download and perform the upload.

There are some aspects to this that will require MR Client changes to allow passing down variables to `olot`, so this will not be able to be a complete solution right out of the gate. This will get us quite far though int he mean-time.

## How Has This Been Tested?

Running a local kind cluster and using the `make` targets to deploy a minio instance, a MR instance, and a container registry to the kind cluster. 

Then, logging into the minio-instance's console to add a bucket and initial model (to be synchronized).

Then, using the sample k8s Job that was added, validate that the job runs and perform a synchronization.

## Merge criteria:

If you have UI changes

### DIFF
--- a/jobs/async-upload/tests/test_download.py
+++ b/jobs/async-upload/tests/test_download.py
@@ -1,6 +1,6 @@
 import os
 import pytest
-from unittest.mock import Mock, patch
+from unittest.mock import Mock, call, patch
 from job.download import download_from_s3
 from job.config import get_config
 from job.mr_client import validate_and_get_model_registry_client
@@ -36,9 +36,9 @@ def minimal_env_source_dest_vars():
 
     # Model and registry variables
     model_vars = {
-        "model_name": "my-model",
-        "model_version": "1.0.0",
-        "model_format": "onnx",
+        "model_id": "abc",
+        "model_version_id": "def",
+        "model_artifact_id": "123",
         "registry_server_address": "http://localhost",
         "registry_port": "8080",
         "registry_author": "author",
@@ -56,49 +56,91 @@ def minimal_env_source_dest_vars():
 
 
 def test_download_from_s3(minimal_env_source_dest_vars):
-    """Test download_from_s3 function with proper configuration"""
+    """Test download_from_s3 now that it pages through prefixes."""
 
-    # Get configuration from environment variables
+    # load config from your fixture
     config = get_config([])
 
-    # Verify the configuration is set up correctly
+    # sanity-check config
     assert config["source"]["type"] == "s3"
     assert config["source"]["s3"]["bucket"] == "test-bucket"
     assert config["source"]["s3"]["key"] == "test-key"
-    assert config["source"]["s3"]["access_key_id"] == "test-access-key-id"
-    assert config["source"]["s3"]["secret_access_key"] == "test-secret-access-key"
-    assert config["source"]["s3"]["endpoint"] == "http://localhost:9000"
-    assert config["storage"]["path"] == "/tmp/model-sync"
 
-    # Create mock ModelRegistry client
+    # use whatever path came back in config
+    storage_path = config["storage"]["path"]
+    assert storage_path == "/tmp/model-sync"
+
+    # mock out ModelRegistry so validate_and_get_model_registry_client returns a dummy client
     with patch("job.mr_client.ModelRegistry") as mock_registry_class:
-        mock_client = Mock()
-        mock_registry_class.return_value = mock_client
+        mock_registry_class.return_value = Mock()
         client = validate_and_get_model_registry_client(config)
 
-        # Mock the S3 client and _connect_to_s3 function
-        with patch("job.download._connect_to_s3") as mock_connect:
-            mock_s3_client = Mock()
-            mock_transfer_config = Mock()
-            mock_connect.return_value = (mock_s3_client, mock_transfer_config)
+    # now patch _connect_to_s3 and os.makedirs
+    with patch("job.download._connect_to_s3") as mock_connect, \
+         patch("os.makedirs") as mock_makedirs:
 
-            # Call the function under test
-            download_from_s3(client, config)
+        # prepare our fake s3 client + transfer config
+        mock_s3 = Mock()
+        mock_transfer_cfg = Mock()
+        mock_connect.return_value = (mock_s3, mock_transfer_cfg)
 
-            # Verify _connect_to_s3 was called with correct parameters
-            mock_connect.assert_called_once_with(
-                endpoint_url="http://localhost:9000",
-                access_key_id="test-access-key-id",
-                secret_access_key="test-secret-access-key",
-                region=None,  # Not set in the test fixture
-            )
+        # set up a paginator that yields a single page with two entries
+        fake_page = {
+            "Contents": [
+                {"Key": "test-key/file1.txt"},
+                {"Key": "test-key/dir/"},                # should be skipped
+                {"Key": "test-key/dir/file2.bin"},
+            ]
+        }
+        mock_paginator = Mock()
+        mock_paginator.paginate.return_value = [fake_page]
+        mock_s3.get_paginator.return_value = mock_paginator
 
-            # Verify download_file was called with correct parameters
-            mock_s3_client.download_file.assert_called_once_with(
-                "test-bucket",  # bucket
-                "test-key",  # key
-                "/tmp/model-sync",  # local path
-            )
+        # call under test
+        download_from_s3(client, config)
+
+        # ensure _connect_to_s3 got all args including multipart settings
+        mock_connect.assert_called_once_with(
+            "http://localhost:9000",
+            "test-access-key-id",
+            "test-secret-access-key",
+            None,  # region
+            multipart_threshold=1024 * 1024,
+            multipart_chunksize=1024 * 1024,
+            max_pool_connections=10,
+        )
+
+        # ensure we asked for the right paginator and paginated correctly
+        mock_s3.get_paginator.assert_called_once_with("list_objects_v2")
+        mock_paginator.paginate.assert_called_once_with(
+            Bucket="test-bucket",
+            Prefix="test-key",
+        )
+
+        # build expected download calls using the real storage_path
+        expected = [
+            call(
+                "test-bucket",
+                "test-key/file1.txt",
+                os.path.join(storage_path, "file1.txt"),
+            ),
+            call(
+                "test-bucket",
+                "test-key/dir/file2.bin",
+                os.path.join(storage_path, "dir", "file2.bin"),
+            ),
+        ]
+        mock_s3.download_file.assert_has_calls(expected, any_order=False)
+
+        # directories should be created for each file
+        mock_makedirs.assert_any_call(
+            os.path.dirname(os.path.join(storage_path, "file1.txt")),
+            exist_ok=True
+        )
+        mock_makedirs.assert_any_call(
+            os.path.dirname(os.path.join(storage_path, "dir", "file2.bin")),
+            exist_ok=True
+        )
 
 
 def test_download_from_s3_with_region(minimal_env_source_dest_vars):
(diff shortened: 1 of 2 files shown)


scripts/manifests/minio/deployment.yaml
```diff
@@ -54,3 +58,9 @@ spec:
     - port: 9000
       targetPort: 9000
       protocol: TCP
+      name: api
+    - port: 9001
+      targetPort: 9001
+      protocol: TCP
+      name: console
+      nodePort: 30091
```
Crazyglue, line 66: This allows us to open up the minio console when performing local testing (which is automatically provisioned, anyways)

jobs/async-upload/job/upload.py
```diff
@@ -0,0 +1,107 @@
...
+        model_files_path: Path to the downloaded model files
+
+    Returns:
+        Path to the prepared modelcar directory structure
+    """
+    model_path = Path(model_files_path)
+
+    # Create modelcar structure directory
+    modelcar_base = model_path.parent / "modelcar"
+    models_dir = modelcar_base / "models"
```
tarilabs: I'm not sure I get this.
When we build a ModelCar with olot, the /models prefix is automatically added: https://github.com/containers/olot/blob/main/olot/utils/files.py#L67

That is, when building a ModelCar with olot, you just need to supply the files, analogously to the S3 case. 

If you are planning to keep this as-is, I believe you must use olot programmatically so to specify your own base prefix.

wdyt?

scripts/manifests/minio/deployment.yaml
tarilabs: This is great for testing, thank you!

jobs/async-upload/job/upload.py
```diff
@@ -0,0 +1,107 @@
...
+    upload_params = _get_upload_params(config)
+
+    client.upload_artifact_and_register_model(
+        model_files_path=upload_path,
+        name=config["model"]["name"],
+        version=config["model"]["version"],
+        model_format_name=config["model"]["format"],
+        model_format_version=config["model"]["format_version"],
+        upload_params=upload_params,
+    )
```
tarilabs: iirc, you cannot use this method, since it would complain about the fact that ModelVersion already exists.

reminder from the ADR:
image

do you have a test case which shows the intended behaviour is respected by this function?
If yes, all the better!

By default I was thinking the behaviour here can be respected by looking up the ModelVersion ID and then the ModelArtifact ID and then updating its URI? 🤔  But please correct me if that function actually already behave as intended!

tarilabs (commented):
thanks @Crazyglue for pinging me on this one!

I left a couple of comments.

Additionally, I think starting to provide a K8s Job manifest is indeed required next so we can start E2E testing, wdyt?

jobs/async-upload/job/upload.py
Crazyglue: Oh wow, I completely missed this. Yes, this can all be effectively removed!

jobs/async-upload/Makefile
```diff
@@ -5,6 +5,16 @@ deploy-latest-mr:
...
+.PHONY: deploy-test-minio
+deploy-test-minio:
+	cd ../../ && ./scripts/deploy_minio_on_kind.sh
+	kubectl port-forward -n minio svc/minio 9000:9000 & echo $$! >> .port-forwards.pid
+
+.PHONY: deploy-local-registry
+deploy-local-registry:
+	cd ../../ && ./scripts/deploy_local_kind_registry.sh
+	kubectl port-forward service/distribution-registry-test-service 5001:5001 & echo $$! >> .port-forwards.pid
+
```
syntaxsdev, line 27: why copy the same targets? we can just do 
`cd ../parent_of_make && make deploy-test-minio` and then run the target that way

jobs/async-upload/job/config.py
```diff
@@ -167,12 +174,16 @@ def _validate_s3_config(cfg: Dict[str, Any]) -> None:
         raise ValueError("S3 credentials must be set")
     if cfg["s3"]["bucket"] is None:
         raise ValueError("S3 bucket must be set")
+    if cfg["s3"]["key"] is None:
```
syntaxsdev, line 176: wouldn't this fail if the key isn't set?

I'd prefer `.get(key, None)` unless youre sure a default is provided for `key`

jobs/async-upload/job/config.py
```diff
@@ -167,12 +174,16 @@ def _validate_s3_config(cfg: Dict[str, Any]) -> None:
...
+        raise ValueError("S3 key must be set")
```
syntaxsdev, line 177: I'm not a stickler on this particular lint code, but this does raise the question if ruff lints were applied to this directory as a whole.

If we are aiming for least resistance at first then I digress and it can wait

jobs/async-upload/job/config.py
```diff
@@ -167,12 +174,16 @@ def _validate_s3_config(cfg: Dict[str, Any]) -> None:
...
 
 
 def _validate_model_config(cfg: Dict[str, Any]) -> None:
     """Validates the model config is valid"""
     if cfg["name"] is None or cfg["version"] is None or cfg["format"] is None:
         raise ValueError("Model must be set")
+    if cfg["format_version"] is None:
```
syntaxsdev: similar to what I wrote above, 

if you're positive these keys are set to None by default, then you can also just do
`if not cfg["format_version"]:` just to keep conditionals cleaner

jobs/async-upload/tests/test_download.py
```diff
@@ -0,0 +1,178 @@
...
+from job.config import get_config
+from job.mr_client import validate_and_get_model_registry_client
+
+
+@pytest.fixture
+def minimal_env_source_dest_vars():
+    original_env = dict(os.environ)
+
+    # Destination variables
+    dest_vars = {
```
syntaxsdev, line 14: why not monkeypatch the vars?

syntaxsdev (commented):
looks well, great in-depth testing, I read through!
some comments on core code otherwise looks good

syntaxsdev:
Another thought, in general, for a small job like this, I'd probably aim to use a requirements.txt or `uv` to manage dependencies since its not a package/library. Makes the actual build process a lot smoother, faster, and lighter.

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
```diff
@@ -0,0 +1,54 @@
...
+      restartPolicy: Never
+      containers:
+        - name: async-upload
+          image: kubeflow/model-registry-async-job:local
+          env:
+            # ---- Source ----------
+            - name: MODEL_SYNC_SOURCE_TYPE
+              value: "s3"
+            - name: MODEL_SYNC_SOURCE_AWS_BUCKET
+              value: "default"
```
tarilabs: seen new addition of Manifest; just to be sure, the actual Manifest need to contain mounting of the Secrets, which contains keys as specified in ADR; correct?

it doesn't hurt to have a first sample, but we need a sample which complies with what we defined (ie in parlance "mounting the Connections(Secrets)")

jobs/async-upload/job/entrypoint.py
```diff
@@ -17,16 +25,28 @@ def main() -> None:
...
-            f"Destination: {config.destination.type.upper()} storage at {config.destination.endpoint or 'default endpoint'}"
+            f"Destination: {config['destination']['type'].upper()} storage at {config['destination'].get('endpoint') or 'default endpoint'}"
         )
 
-        # TODO: Implement the main upload logic here
+        # Download the model from the defined source
+        perform_download(client, config)
+
+        # Queue up model registration
+        registered_model = perform_model_registration(client, config)
```
tarilabs: Same comment as https://github.com/kubeflow/model-registry/pull/1256#discussion_r2191713230

reminder from the ADR:
image

the RegisteredModel, ModelVersion, and ModelArtifact are expected to be present, the role of this Job it's to update metadata (which I see some lines below).

Are you planning to guard this method to ensure being called only if the model coordinates are not supplied?

jobs/async-upload/job/entrypoint.py
```diff
@@ -17,16 +25,28 @@ def main() -> None:
...
+
+        # KServe Modelcars compatibility is handled within perform_upload()
+        # For OCI destinations, it automatically creates the required /models directory structure
+        # and prepares the model files in a KServe Modelcars-compatible format
+
+        uri = perform_upload(config)
```
tarilabs: this comment does not reflect the actual `perform_upload()` behaviour at the time of me-looking-and-writing this comment 🤔 

<details>

```py
def perform_upload(config: Dict[str, Any]) -> str:
    """
    Performs the upload of the model to the destination with KServe Modelcars compatibility
    Returns:
        The URI of the uploaded model
    """
    model_files_path = config["storage"]["path"]

    # For OCI uploads, prepare KServe Modelcars-compatible structure
    if config["destination"]["type"] == "oci" and os.path.exists(model_files_path):
        # Prepare modelcar directory structure (/models)
        modelcar_path = _prepare_modelcar_structure(config, model_files_path)
```

https://github.com/kubeflow/model-registry/pull/1256/files#diff-7cffcb80c78e46c037c7bbb1fc74ff717eea9342ed4093a360c9b8b99d70a6b9R80-R92

```py
def _prepare_modelcar_structure(config: Dict[str, Any], model_files_path: str) -> str:
    """
    Prepare model files in KServe Modelcars-compatible structure.
    KServe Modelcars expects models to be in /models directory within the OCI image.
    This function creates the proper directory structure.
    Args:
        config: The configuration dictionary containing model metadata
        model_files_path: Path to the downloaded model files
    Returns:
        Path to the prepared modelcar directory structure
    """
    model_path = Path(model_files_path)

    # Create modelcar structure directory
    modelcar_base = model_path.parent / "modelcar"
    models_dir = modelcar_base / "models"
    # Clean up any existing structure
    if modelcar_base.exists():
        shutil.rmtree(modelcar_base)

    models_dir.mkdir(parents=True, exist_ok=True)
```

https://github.com/kubeflow/model-registry/pull/1256/files#diff-7cffcb80c78e46c037c7bbb1fc74ff717eea9342ed4093a360c9b8b99d70a6b9R9-R33

</details>

Crazyglue: Whoops, forgot to remove it, will do

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
Crazyglue: Yes, this is the Job that I have been using just to test, its a lot easier (shorter cycles) to just code in everything like this. There will be another manifest to demonstrate mounted secrets

jobs/async-upload/job/config.py
Crazyglue: I thought I had ruff applied here, but perhaps I didnt 🤔

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
tarilabs: ack, thanks for the confirmation

jobs/async-upload/job/config.py
Crazyglue: There is a bit of a semantic difference here. This is a bit of a contrived example, but if a user supplied the empty string `""` or the integer `0` and they intended to use those values here (or on any of these checks), then doing `if not my_variable:` would be incorrectly go down that code path.

I'm not sure that that is ever going to be a real use-case or not, so I'm not really sure if this should be updated or not

Crazyglue:
> Another thought, in general, for a small job like this, I'd probably aim to use a requirements.txt or `uv` to manage dependencies since its not a package/library. Makes the actual build process a lot smoother, faster, and lighter.

I had originally set up the repo like this, but later changed it to use poetry similar to the client. I suppose we can take a look again later and adjust things. I would be in favor of simpler dependency management, but poetry so far has been performant enough since there are so few dependencies anyways

jobs/async-upload/job/entrypoint.py
Crazyglue: Take a look at the latest push, this should be more in line with what you are thinking

jobs/async-upload/job/mr_client.py
```diff
@@ -11,26 +11,33 @@ def validate_and_get_model_registry_client(config: Dict[str, Any]) -> ModelRegis
...
 
-def update_model_registration(
-    uri: str, client: ModelRegistry, registered_model: RegisteredModel
+
+def update_model_artifact_uri(
+    uri: str, client: ModelRegistry, config: Dict[str, Any]
 ) -> None:
-    updated_model = registered_model.update(uri=uri)
-    client.update(updated_model)
+    artifact = client.get_model_artifact(
```
tarilabs, line 47: better, but as per previous PR, I'm not comfortable with using RegModel-name and ModelVersion-name, in case we will support renaming in the future.

I'd consider using RegisteredModel.ID, ModelVersion.ID and ModelArtifact.ID

If you want to keep the input name parameters for consistency check, that's fine with me

jobs/async-upload/job/entrypoint.py
```diff
@@ -38,15 +38,12 @@ def main() -> None:
         perform_download(client, config)
 
         # Queue up model registration
-        registered_model = perform_model_registration(client, config)
-
-        # KServe Modelcars compatibility is handled within perform_upload()
-        # For OCI destinations, it automatically creates the required /models directory structure
-        # and prepares the model files in a KServe Modelcars-compatible format
+        set_artifact_pending(client, config)
```
tarilabs: I think the pending state is set by the Dashboard per ADR instruction, but happy to keep this 'to make sure'

tarilabs (commented):
thanks for the updates

jobs/async-upload/job/upload.py
tarilabs: less is more 🚀

jobs/async-upload/job/mr_client.py
Crazyglue: Ah yes, let me switch this over

tarilabs:
please rebase?

tarilabs:
> This will get us quite far though int he mean-time.

thanks a lot @Crazyglue and all contributors, it's not the complete use-case I agree but enough substance to take this in as-in and followup PRs

/lgtm

jonburdo (commented):
/lgtm

syntaxsdev:
/lgtm

great job on this @Crazyglue !

PR #1266, opened 2025-07-02
feat(catalog): reload sources and static catalogs

pboyd:
## Description

Reload sources and static catalogs after changes are made to the files they were loaded from.

## How Has This Been Tested?
Local kind cluster and unit tests.

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/monitor.go
+++ b/catalog/internal/catalog/monitor.go
@@ -1,0 +1,238 @@
+package catalog
+
+import (
+	"fmt"
+	"hash/crc32"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/golang/glog"
+)
+
+// monitor sends events when the contents of a file have changed.
+//
+// Unfortunately, simply watching the file misses events for our primary case
+// of k8s mounted configmaps because the files we're watching are actually
+// symlinks which aren't modified:
+//
+//	drwxrwxrwx    1 root     root           138 Jul  2 15:45 .
+//	drwxr-xr-x    1 root     root           116 Jul  2 15:52 ..
+//	drwxr-xr-x    1 root     root            62 Jul  2 15:45 ..2025_07_02_15_45_09.2837733502
+//	lrwxrwxrwx    1 root     root            32 Jul  2 15:45 ..data -> ..2025_07_02_15_45_09.2837733502
+//	lrwxrwxrwx    1 root     root            26 Jul  2 13:18 sample-catalog.yaml -> ..data/sample-catalog.yaml
+//	lrwxrwxrwx    1 root     root            19 Jul  2 13:18 sources.yaml -> ..data/sources.yaml
+//
+// Updates are written to a new directory and the ..data symlink is updated. No
+// fsnotify events will ever be triggered for the YAML files.
+//
+// The approach taken here is to watch the directory containing the file for
+// any change and then hash the contents of the file to avoid false-positives.
+type monitor struct {
+	watcher *fsnotify.Watcher
+	closed  <-chan struct{}
+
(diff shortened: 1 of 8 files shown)


Al-Pragliola (commented):
Good job, I plan to review and test locally this PR tomorrow!

I have just a small question, have you considered using Viper with the WatchConfig function?

pboyd:
> Good job @pboyd , I plan to review and test locally this PR tomorrow!
> 
> I have just a small question, have you considered using Viper with the WatchConfig function?

Thanks, @Al-Pragliola. I don't think Viper is a good fit for this, though it would certainly have been nice to use `WatchConfig`.

Al-Pragliola (approved):
/lgtm

tested locally as a binary and in a kind cluster everything worked fine

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	go.uber.org/zap v1.27.0
 	google.golang.org/grpc v1.72.1
 	google.golang.org/protobuf v1.36.6
+	gopkg.in/fsnotify.v1 v1.4.7
```
tarilabs: do we want to introduce this `gopkg.in` proxy service, or like other dependencies before we just use from the [tutorial](https://github.com/fsnotify/fsnotify?tab=readme-ov-file#usage) `github.com/fsnotify/fsnotify` ?

otherwise,
/lgtm

catalog/internal/catalog/monitor.go
```diff
@@ -0,0 +1,238 @@
...
+//	drwxr-xr-x    1 root     root            62 Jul  2 15:45 ..2025_07_02_15_45_09.2837733502
+//	lrwxrwxrwx    1 root     root            32 Jul  2 15:45 ..data -> ..2025_07_02_15_45_09.2837733502
+//	lrwxrwxrwx    1 root     root            26 Jul  2 13:18 sample-catalog.yaml -> ..data/sample-catalog.yaml
+//	lrwxrwxrwx    1 root     root            19 Jul  2 13:18 sources.yaml -> ..data/sources.yaml
+//
+// Updates are written to a new directory and the ..data symlink is updated. No
+// fsnotify events will ever be triggered for the YAML files.
+//
+// The approach taken here is to watch the directory containing the file for
+// any change and then hash the contents of the file to avoid false-positives.
```
tarilabs, line 33: I'm a bit surprised that it doesn't exist any other Golang utility for this and avoiding the k8s client.

Maybe something like https://pkg.go.dev/knative.dev/pkg/configmap ?

Al-Pragliola: uhm maybe I missed it but where is the k8s client used?

tarilabs: _and_ avoiding the k8s client (as an alternative). it's not used. I'm not suggesting to use it instead of the current solution; what I'm puzzled is that no utility already exists, for example the Knative link for this use case?

go.mod
pboyd: I overlooked which package was being imported. I switched it to `github.com/fsnotify/fsnotify`.

tarilabs: thanks Paul

catalog/internal/catalog/monitor.go
pboyd: Viper's `WatchConfig` is an option, but we weren't using it before and I don't see much reason to use it now, except for this one feature.

`knative.dev/pkg/configmap` has two implementations for a "watcher". One uses the k8s API, and the other is a "manual" watcher which we'd have to have trigger (presumably by handling `fsnotify` events).

I'm not aware of any another package that handles this case.

Al-Pragliola:
- https://github.com/kubeflow/model-registry/pull/1266#pullrequestreview-2984513453
- https://github.com/kubeflow/model-registry/pull/1266#pullrequestreview-2994035658

/approve

PR #1294, opened 2025-07-11
chore(mr-client): match up the hand-rolled artifact data model to match the api

Crazyglue:
## Description
There are usecases where a ModelArtifact will not have a `uri` associated to it, such as when kicking off an model synchronization job (the uri will be pending as the synchronization happens). The data model for a ModelArtifact does not require the uri to be populated, so having these asserts in the client will cause errors to be thrown even though the returned (or supplied) data is valid.

This PR does present a breakage risk for consumers of the MR Client.

They might have logic which wraps a call to fetch a ModelArtifact with a try/catch block, and handle that case accordingly. With this change, that logic would need to be changed to check for existence (`is None`) instead.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/types/artifacts.py
+++ b/clients/python/src/model_registry/types/artifacts.py
@@ -47,7 +47,7 @@ class Artifact(BaseResourceModel, ABC):
     """
 
     name: str | None = None
-    uri: str
+    uri: str | None = None
     state: ArtifactState = ArtifactState.UNKNOWN
 
     @classmethod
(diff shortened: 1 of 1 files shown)


Crazyglue:
> thanks @Crazyglue , I'd keep changes to the minimal we need to unblock the async-upload Job work: removing the assert on uri
> 
> AFAIK even if state is not declared as required: in openapi, we set a default internally
> 
> ie I would keep changes focused on the minimum common need, wdyt?

Totally fair, I will adjust this accordingly and push a commit

syntaxsdev (approved):
this looks fine because this will support future work of experiment and runs as metrics build on the artifact type and dont use uri's.

/lgtm

tarilabs (approved):
great catch @Crazyglue 

thanks @syntaxsdev for 👀 

/approve

PR #1310, opened 2025-07-14
[async-job

Crazyglue:
## Merge criteria:

If you have UI changes

### DIFF
--- a/jobs/async-upload/README.md
+++ b/jobs/async-upload/README.md
@@ -38,45 +38,47 @@ Below is a table of the configuration that can be passed into the job
 
 See asterisks below table for details
 
-| Environment Variable                         | Arg                                 | Default Value     | Required | Description                                                      |
-| -------------------------------------------- | ----------------------------------- | ----------------- | -------- | ---------------------------------------------------------------- |
-| MODEL_SYNC_SOURCE_TYPE                       | --source-type                       | s3                | ✅       |                                                                  |
-| MODEL_SYNC_SOURCE_S3_CREDENTIALS_PATH        | --source-s3-credentials-path        |                   |          |                                                                  |
-| MODEL_SYNC_SOURCE_OCI_CREDENTIALS_PATH       | --source-oci-credentials-path       |                   |          |                                                                  |
-| MODEL_SYNC_SOURCE_AWS_BUCKET                 | --source-aws-bucket                 |                   | ✅\*     | When --source-type is "s3"                                       |
-| MODEL_SYNC_SOURCE_AWS_KEY                    | --source-aws-key                    |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_SOURCE_AWS_REGION                 | --source-aws-region                 |                   |          | "                                                                |
-| MODEL_SYNC_SOURCE_AWS_ACCESS_KEY_ID          | --source-aws-access-key-id          |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_SOURCE_AWS_SECRET_ACCESS_KEY      | --source-aws-secret-access-key      |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_SOURCE_AWS_ENDPOINT               | --source-aws-endpoint               |                   |          | "                                                                |
-| MODEL_SYNC_SOURCE_OCI_URI                    | --source-oci-uri                    |                   | ✅\+     | When --source-type is "oci"                                      |
-| MODEL_SYNC_SOURCE_OCI_USERNAME               | --source-oci-username               |                   | ✅\+     | "                                                                |
-| MODEL_SYNC_SOURCE_OCI_PASSWORD               | --source-oci-password               |                   | ✅\+     | "                                                                |
-| MODEL_SYNC_DESTINATION_TYPE                  | --destination-type                  | oci               | ✅       |                                                                  |
-| MODEL_SYNC_DESTINATION_S3_CREDENTIALS_PATH   | --destination-s3-credentials-path   |                   |          |                                                                  |
-| MODEL_SYNC_DESTINATION_OCI_CREDENTIALS_PATH  | --destination-oci-credentials-path  |                   |          |                                                                  |
-| MODEL_SYNC_DESTINATION_AWS_BUCKET            | --destination-aws-bucket            |                   | ✅\*     | When --destination-type is "s3"                                  |
-| MODEL_SYNC_DESTINATION_AWS_KEY               | --destination-aws-key               |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_DESTINATION_AWS_REGION            | --destination-aws-region            |                   |          | "                                                                |
-| MODEL_SYNC_DESTINATION_AWS_ACCESS_KEY_ID     | --destination-aws-access-key-id     |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_DESTINATION_AWS_SECRET_ACCESS_KEY | --destination-aws-secret-access-key |                   | ✅\*     | "                                                                |
-| MODEL_SYNC_DESTINATION_AWS_ENDPOINT          | --destination-aws-endpoint          |                   |          | "                                                                |
-| MODEL_SYNC_DESTINATION_OCI_URI               | --destination-oci-uri               |                   | ✅\+     | When --destination-type is "oci"                                 |
-| MODEL_SYNC_DESTINATION_OCI_USERNAME          | --destination-oci-username          |                   | ✅\+     | "                                                                |
-| MODEL_SYNC_DESTINATION_OCI_PASSWORD          | --destination-oci-password          |                   | ✅\+     | "                                                                |
-| MODEL_SYNC_MODEL_ID                          | --model-id                          |                   | ✅       | The `RegisteredModel.id`                                         |
-| MODEL_SYNC_VERSION_ID                        | --model-version-id                  |                   | ✅       | The `ModelVersion.id`                                            |
-| MODEL_SYNC_ARTIFACT_ID                       | --model-artifact-id                 |                   | ✅       | The `ModelArtifact.id`                                           |
-| MODEL_SYNC_STORAGE_PATH                      | --storage-path                      | `/tmp/model-sync` | ✅       | Temporary storage, must be large enough to hold the entire model |
-| MODEL_SYNC_REGISTRY_SERVER_ADDRESS           | --registry-server-address           |                   | ✅       | Server address for the model-registry client to connect to       |
-| MODEL_SYNC_REGISTRY_PORT                     | --registry-port                     | 443               |          |                                                                  |
-| MODEL_SYNC_REGISTRY_IS_SECURE                | --registry-is-secure                | True              |          |                                                                  |
-| MODEL_SYNC_REGISTRY_AUTHOR                   | --registry-author                   |                   |          |                                                                  |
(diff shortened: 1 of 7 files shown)


jobs/async-upload/README.md
Crazyglue, line 1: Not sure why the diff is crazy here, 2 entries were added for the new config garibales

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
 type: kubernetes.io/dockerconfigjson
 stringData:
   .dockerconfigjson: '{"auths": {"distribution-registry-test-service.default.svc.cluster.local:5001": {"auth": "base64(username:password)","email": "user@example.com"}}}'
-  ACCESS_TYPE: ["Push", "Pull"]
+  ACCESS_TYPE: "Push,Pull"
```
tarilabs:
```suggestion
  ACCESS_TYPE: '["Push", "Pull"]'
```

I think the OCI Connection type expect a String representation of a JSON array

jobs/async-upload/job/config.py
```diff
@@ -27,6 +28,7 @@ def _parser() -> cap.ArgumentParser:
     p.add("--source-aws-endpoint")
     # OCI registry
     p.add("--source-oci-uri")
+    p.add("--source-oci-registry")
```
Crazyglue: This (along with the destination) are added because a `.dockerconfigjson` file may contain multiple credentials for each registry. So its better for the consumer to declare which registry they intend to upload to, to make credential parsing more straight forward

jobs/async-upload/job/config.py
```diff
@@ -149,11 +154,13 @@ def _load_oci_credentials(
...
     # Load the credentials from the docker config, the URI passed in via config is used as a key to find the correct credentials
-    auth = docker_config["auths"][store["uri"]]["auth"]
-    username, password = base64.b64decode(auth).decode("utf-8").split(":")
-    store["username"] = username
-    store["password"] = password
-    store["email"] = docker_config["auths"][store["uri"]]["email"]
+    registry = store["oci"]["registry"]
+    auth = docker_config["auths"][registry]["auth"]
+    # TODO: This might not be the correct way to parse this
+    username, password = auth.split(":")
```
Crazyglue: Ive seen the credentials be uplaoded as base-64 encoded, but in other places ive seen it passed in as plain-text. I'm not sure which is the "correct" way, but for now i will assume plain-text

jobs/async-upload/job/config.py
```diff
@@ -236,10 +244,11 @@ def get_config(argv: list[str] | None = None) -> Dict[str, Any]:
                 "region": None,
                 "access_key_id": None,
                 "secret_access_key": None,
-                "endpoint": None,
+                "endpoint_url": None,
```
Crazyglue, line 262: This is a correction on the official S3 env vars.

https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html

And so, `AWS_ENDPOINT_URL` -> `endpoint_url`

jobs/async-upload/job/config.py
```diff
@@ -345,4 +355,41 @@ def get_config(argv: list[str] | None = None) -> Dict[str, Any]:
...
     _validate_config(cfg)
 
+    # Log the configuration (with sensitive values sanitized)
+    sanitized_cfg = _sanitize_config_for_logging(cfg)
+    logger.debug("Configuration loaded: %s", json.dumps(sanitized_cfg, indent=2))
+
     return cfg
+
+
+def _sanitize_config_for_logging(cfg: Dict[str, Any]) -> Dict[str, Any]:
```
Crazyglue, line 384: This is just to help with debugging. Can be removed later on if need-be, but this helped with running the job in the cluster and seeing what was being parsed

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
Crazyglue: I'm not 100% sure if this will be a comma-separated list or a stringified array. in either case, an array-type is not valid for a secret's value

tarilabs: https://github.com/kubeflow/model-registry/pull/1310#pullrequestreview-3016839082

Crazyglue: ah, we had commented aroudn the same time haha

jobs/async-upload/job/upload.py
```diff
@@ -15,18 +15,23 @@ def _get_upload_params(config: Dict[str, Any]) -> S3Params | OCIParams:
...
+            base_image=destination_config["oci"]["base_image"],
             oci_ref=destination_config["oci"]["uri"],
             dest_dir=config["storage"]["path"],
             oci_username=destination_config["oci"]["username"],
             oci_password=destination_config["oci"]["password"],
+            # Same as the default backend, but with additional args included
+            custom_oci_backend=utils._get_skopeo_backend(
+                push_args=["--dest-tls-verify=false"]
+            ),
+
```
tarilabs: I think this is _awesome_ as demonstrating the extensibility we had placed in the MR py client orchestration code, very good job!

Unfortunately I don't think we can leave this "here" in running code, it would imply all job skips tls verifcation.

We need a way to parametrize/configure this, so that it's added when running say in E2E testing, but by default it's _not present_.

I would suggest this is just another configuration parameter, wdyt?

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
```diff
@@ -64,9 +67,13 @@ spec:
             - name: MODEL_SYNC_DESTINATION_TYPE
               value: "oci"
             - name: MODEL_SYNC_DESTINATION_OCI_URI
-              value: "oci://distribution-registry-test-service.default.svc.cluster.local:5001/my-model:latest"
+              value: "distribution-registry-test-service.default.svc.cluster.local:5001/my-model:latest"
+            - name: MODEL_SYNC_DESTINATION_OCI_REGISTRY
+              value: "distribution-registry-test-service.default.svc.cluster.local:5001"
             - name: MODEL_SYNC_DESTINATION_OCI_CREDENTIALS_PATH
-              value: "/opt/creds/destination"
+              value: "/opt/creds/destination/.dockerconfigjson"
```
tarilabs, line 74: I think we can avoid parsing the content for user/psw and just point skopeo to this.

wdyt of trying

```
skopeo copy --authfile /opt/creds/destination/.dockerconfigjson ...rest of the command
```

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
```diff
@@ -64,9 +67,13 @@ spec:
...
+            - name: MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE
+              value: "alpine:latest"
```
tarilabs: I don't think we ever tested `alpine:latest`, I would recommend `busybox` which is the default also in KServe documentation: https://kserve.github.io/website/latest/modelserving/storage/oci/#prepare-an-oci-image-with-model-data:~:text=by%20the%20runtime%3A-,FROM%20busybox,-RUN%20mkdir%20/models

jobs/async-upload/job/upload.py
Crazyglue: I agree, this is a use case that will need to be configurable at the job level. I will save it for a follow-up pr, as im sure there are even more parameters we would want to expose!

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
Crazyglue: That would be excellent, however i believe the OCIParams object does require a "username" and "password". I will investigate but I prefer your solution more

Crazyglue: Ok! I just checked and this is feasible. However, the fix will require a bit more rework of the job configuration it self, since that path is not stored in the config itself.

What do you think of a follow up PR to adjust this?

Crazyglue: Auth and TLS added

jobs/async-upload/job/config.py
```diff
@@ -18,66 +19,72 @@ def _parser() -> cap.ArgumentParser:
...
-    p.add("--source-aws-access-key-id")
-    p.add("--source-aws-secret-access-key")
-    p.add("--source-aws-endpoint")
+    p.add_argument("--source-type", choices=["s3", "oci"], default="s3")
+    p.add_argument("--source-aws-bucket")
+    p.add_argument("--source-aws-key")
+    p.add_argument("--source-aws-region")
+    p.add_argument("--source-aws-access-key-id")
+    p.add_argument("--source-aws-secret-access-key")
+    p.add_argument("--source-aws-endpoint")
```
Crazyglue, line 28: I ended up switching out the alias `add` for the actual method `add_argument`. IDE was able to pick up the typings better

jobs/async-upload/samples/sample_job_s3_to_oci.yaml
tarilabs: > What do you think of a follow up PR to adjust this?

yes 
> you can move those in a follow-up PRs if preferred

([ref](https://github.com/kubeflow/model-registry/pull/1310#pullrequestreview-3017384942))

and thank you for the work on this

jobs/async-upload/job/config.py
```diff
@@ -18,66 +19,72 @@ def _parser() -> cap.ArgumentParser:
...
-    p.add("--destination-oci-username")
-    p.add("--destination-oci-password")
+    p.add_argument("--destination-oci-uri")
+    p.add_argument("--destination-oci-registry")
+    p.add_argument("--destination-oci-username")
+    p.add_argument("--destination-oci-password")
+    p.add_argument("--destination-oci-base-image", default="busybox:latest")
+    # The `type` converter is needed here to support env-based booleans
+    # See: https://github.com/bw2/ConfigArgParse/tree/master?tab=readme-ov-file#special-values
+    p.add_argument("--destination-oci-enable-tls-verify", default=True, type=str2bool)
```
tarilabs, line 53: thank you

jobs/async-upload/job/upload.py
```diff
@@ -15,18 +15,30 @@ def _get_upload_params(config: Dict[str, Any]) -> S3Params | OCIParams:
...
             region=destination_config["s3"]["region"],
         )
     elif destination_config["type"] == "oci":
+        push_args = []
+        # Note: These are all skopeo args, see: https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md
+        if not destination_config["oci"]["enable_tls_verify"]:
+            push_args.append("--dest-tls-verify=false")
+        if destination_config["credentials_path"]:
+            push_args.append("--authfile")
+            push_args.append(destination_config["credentials_path"])
```
tarilabs, line 30: fantastic, thank you for this!

tarilabs (commented):
thanks a lot @Crazyglue 

/lgtm

jonburdo:
LGTM! Thanks for adding the test steps and result. I was able to get this working as well with similar output :)

tarilabs:
- https://github.com/kubeflow/model-registry/pull/1310#pullrequestreview-3019192458
- https://github.com/kubeflow/model-registry/pull/1310#issuecomment-3073868307

/approve

Issue #1225, opened 2025-06-20
Add MLflow SDK support for Model Registry as a store


dhirajsb:
**Is your feature request related to a problem? Please describe.**
Model Registry has it's own REST API, but a domain model very similar to MLflow. Model Registry is also going to add support for Experiment Tracking in another enhancement request #1224 . 
MLflow SDK supports loading [plugins](https://mlflow.org/docs/latest/ml/plugins) for integrating with other tracking and registry stores. 

Model Registry MUST implement an MLflow plugin to allow millions of existing MLflow users to seamlessly use Model Registry for experiment tracking to begin with. Support for Registry API/store can be added in the future. 

**Describe the solution you'd like**
Implement an MLFlow tracking store for tracking experiments and runs. 

**Describe alternatives you've considered**
Implementing a gRPC server in the Model Registry service compatible with MLflow gRPC API was also considered. But, it complicates the existing Model Registry service implementation. It's easier to use MLFlow plugin architecture and directly make calls to existing Model Registry REST API. 

**Additional context**

dhirajsb:
POC work for this feature has already started and the proposed usage can be seen in the [README.md](https://github.com/dhirajsb/model-registry-kfp/blob/feat/mlflow-plugin/clients/modelregistry_plugin/README.md).

dhirajsb:
# Model Registry MLflow Plugin

This is a detailed overview of the MLflow plugin for Model Registry, covering its purpose, design, implementation, testing, and other relevant details for reviewers.

## Purpose

The primary purpose of this plugin is to enable users to use the MLflow SDK for experiment tracking with Kubeflow Model Registry as the backend. This allows users to leverage their existing MLflow-based tools and workflows while storing experiment data in the Model Registry. This addresses the requirements of the original feature request by providing a standards-based and familiar interface for interacting with the Model Registry's upcoming experiment tracking features.

## Design and Implementation

The plugin is designed as a Python package that implements the `mlflow.tracking.AbstractStore` interface. This allows MLflow to recognize and use the plugin as a valid tracking store when the `modelregistry://` or `modelregistry+https://` scheme is used in the tracking URI.

### Architecture

The plugin's architecture is modular, with a clear separation of concerns between the main store implementation, API client, and various operations. This design enhances maintainability and testability.

Here is a class diagram illustrating the main components of the plugin:

```mermaid
classDiagram
    class ModelRegistryStore {
        +create_experiment(name, artifact_location)
[...]

PR #1318, opened 2025-07-15
Add support for Experiment tracking in Model Registry, fixes #1224

### LINKED ISSUE #1224: Add support for Experiment tracking in Model Registry
Model Registry currently only handles models and versions. ML/AI development starts with experiments and runs. Model Registry API MUST add support for creating and tracking the following:
* Experiments
* Runs within Experiments
* DataSets that are part of input data set
* Metrics and metric history associated with runs
* Parameters associated with runs

The above artifact data should also be accessible in Model Versions. 

Add schema types for the Contexts (Experiment and Run), and Artifacts(DataSet, Metric, Parameter). All of this is possible to be supported and implemented in the current MLMD schema based architecture and REST API. 

Adding an mlflow compliant gRPC/RESTish API. This alternative was rejective due to the cognitive complexity introduced by such a bipolar API scheme with the existing model registry REST API and a partial gRPC API. 
MLflow compatibility is going to be addressed with another feature enhancement next. 

**Additional context**

dhirajsb:
## Description

Add support for Experiment Tracking APIs similare to MLflow and other registries like wandb. 
This is required for experiment tracking using model registry SDK as well as forms the basis of the MLflow SDK integration feature #1225. 

See the feature request #1224 for detailed description of enhancements in this PR. 

## How Has This Been Tested?

Added extensive unit and integration tests for every entity and endpoint for Experiment tracking. 
Detailed unit tests for the filterQuery search enhancement. 

## Merge criteria:

If you have UI changes

### DIFF
--- a/internal/core/artifact_test.go
+++ b/internal/core/artifact_test.go
@@ -172,6 +172,139 @@ func TestUpsertArtifact(t *testing.T) {
 		assert.Contains(t, err.Error(), "metric value is required")
 	})
 
+	// Tests for null name handling - should generate UUID for all artifact types
+	t.Run("create model artifact with null name generates UUID", func(t *testing.T) {
+		modelArtifact := &openapi.ModelArtifact{
+			// Name is intentionally nil/not set
+			Uri: apiutils.Of("s3://bucket/model-no-name.pkl"),
+		}
+
+		artifact := &openapi.Artifact{
+			ModelArtifact: modelArtifact,
+		}
+
+		result, err := _service.UpsertArtifact(artifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		require.NotNil(t, result.ModelArtifact)
+		assert.NotNil(t, result.ModelArtifact.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.ModelArtifact.Name, "Generated name should not be empty")
+		// Check if it looks like a UUID (basic check for format)
+		assert.Len(t, *result.ModelArtifact.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.ModelArtifact.Name, "-", "Generated name should have UUID format")
+	})
+
+	t.Run("create doc artifact with null name generates UUID", func(t *testing.T) {
+		docArtifact := &openapi.DocArtifact{
+			// Name is intentionally nil/not set
+			Uri: apiutils.Of("s3://bucket/doc-no-name.pdf"),
+		}
+
+		artifact := &openapi.Artifact{
+			DocArtifact: docArtifact,
+		}
+
+		result, err := _service.UpsertArtifact(artifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		require.NotNil(t, result.DocArtifact)
+		assert.NotNil(t, result.DocArtifact.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.DocArtifact.Name, "Generated name should not be empty")
+		assert.Len(t, *result.DocArtifact.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.DocArtifact.Name, "-", "Generated name should have UUID format")
+	})
+
+	t.Run("create dataset with null name generates UUID", func(t *testing.T) {
+		dataSet := &openapi.DataSet{
+			// Name is intentionally nil/not set
+			Uri: apiutils.Of("s3://bucket/dataset-no-name.csv"),
+		}
+
+		artifact := &openapi.Artifact{
+			DataSet: dataSet,
+		}
+
+		result, err := _service.UpsertArtifact(artifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		require.NotNil(t, result.DataSet)
+		assert.NotNil(t, result.DataSet.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.DataSet.Name, "Generated name should not be empty")
+		assert.Len(t, *result.DataSet.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.DataSet.Name, "-", "Generated name should have UUID format")
+	})
+
+	t.Run("create metric with null name generates UUID", func(t *testing.T) {
+		metric := &openapi.Metric{
+			// Name is intentionally nil/not set
+			Value: apiutils.Of(0.99),
+		}
+
+		artifact := &openapi.Artifact{
+			Metric: metric,
+		}
+
+		result, err := _service.UpsertArtifact(artifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		require.NotNil(t, result.Metric)
+		assert.NotNil(t, result.Metric.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.Metric.Name, "Generated name should not be empty")
+		assert.Len(t, *result.Metric.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.Metric.Name, "-", "Generated name should have UUID format")
+	})
+
+	t.Run("create parameter with null name generates UUID", func(t *testing.T) {
+		parameter := &openapi.Parameter{
+			// Name is intentionally nil/not set
+			Value: apiutils.Of("param-value"),
+		}
+
+		artifact := &openapi.Artifact{
+			Parameter: parameter,
+		}
+
+		result, err := _service.UpsertArtifact(artifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		require.NotNil(t, result.Parameter)
+		assert.NotNil(t, result.Parameter.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.Parameter.Name, "Generated name should not be empty")
+		assert.Len(t, *result.Parameter.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.Parameter.Name, "-", "Generated name should have UUID format")
+	})
+
+	t.Run("update artifact with null name preserves existing name", func(t *testing.T) {
+		// First create an artifact with a specific name
+		originalName := "original-artifact-name"
+		modelArtifact := &openapi.ModelArtifact{
+			Name: apiutils.Of(originalName),
+			Uri:  apiutils.Of("s3://bucket/original.pkl"),
+		}
+
+		created, err := _service.UpsertModelArtifact(modelArtifact)
+		require.NoError(t, err)
+		require.NotNil(t, created.Id)
+
+		// Update with nil name - should preserve existing name
+		updateArtifact := &openapi.ModelArtifact{
+			Id: created.Id,
+			// Name is intentionally nil
+			Uri: apiutils.Of("s3://bucket/updated.pkl"),
+		}
+
+		updated, err := _service.UpsertModelArtifact(updateArtifact)
+		require.NoError(t, err)
+		require.NotNil(t, updated)
+		assert.Equal(t, originalName, *updated.Name, "Name should be preserved during update")
+		assert.Equal(t, "s3://bucket/updated.pkl", *updated.Uri, "Uri should be updated")
+	})
+
 	t.Run("unicode characters in model artifact name", func(t *testing.T) {
 		// Test with unicode characters: Chinese, Russian, Japanese, and emoji
 		unicodeName := "模型工件-тест-モデルアーティファクト-🚀"
@@ -1493,6 +1626,25 @@ func TestUpsertModelArtifact(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, mixedName, *retrieved.Name)
 		assert.Equal(t, "tensorflow@#$%-测试", *retrieved.ModelFormatName)
+	})
+
+	t.Run("create with null name generates UUID", func(t *testing.T) {
+		modelArtifact := &openapi.ModelArtifact{
+			// Name is intentionally nil
+			Uri:             apiutils.Of("s3://bucket/direct-no-name.pkl"),
+			ModelFormatName: apiutils.Of("tensorflow"),
+		}
+
+		result, err := _service.UpsertModelArtifact(modelArtifact)
+
+		require.NoError(t, err)
+		require.NotNil(t, result)
+		assert.NotNil(t, result.Name, "Name should be auto-generated")
+		assert.NotEmpty(t, *result.Name, "Generated name should not be empty")
+		assert.Len(t, *result.Name, 36, "Generated name should be UUID length")
+		assert.Contains(t, *result.Name, "-", "Generated name should have UUID format")
+		assert.Equal(t, "s3://bucket/direct-no-name.pkl", *result.Uri)
+		assert.Equal(t, "tensorflow", *result.ModelFormatName)
 	})
 
 	t.Run("pagination test", func(t *testing.T) {
--- a/internal/core/artifact.go
+++ b/internal/core/artifact.go
@@ -12,6 +12,34 @@ import (
 	"gorm.io/gorm"
 )
 
+// ensureArtifactName ensures that an artifact has a name during creation.
+// If the artifact has no ID (creation) and no name, it generates a UUID.
+func ensureArtifactName(artifact *openapi.Artifact) {
+	if artifact == nil {
+		return
+	}
+
+	// Helper function to generate name if needed
+	generateNameIfNeeded := func(id *string, name **string) {
+		if id == nil && *name == nil {
+			// This is a create operation with no name, generate UUID
+			*name = converter.GenerateNewName()
+		}
+	}
+
+	if artifact.ModelArtifact != nil {
+		generateNameIfNeeded(artifact.ModelArtifact.Id, &artifact.ModelArtifact.Name)
+	} else if artifact.DocArtifact != nil {
+		generateNameIfNeeded(artifact.DocArtifact.Id, &artifact.DocArtifact.Name)
+	} else if artifact.DataSet != nil {
+		generateNameIfNeeded(artifact.DataSet.Id, &artifact.DataSet.Name)
+	} else if artifact.Metric != nil {
+		generateNameIfNeeded(artifact.Metric.Id, &artifact.Metric.Name)
+	} else if artifact.Parameter != nil {
+		generateNameIfNeeded(artifact.Parameter.Id, &artifact.Parameter.Name)
+	}
+}
+
 func (b *ModelRegistryService) upsertArtifact(artifact *openapi.Artifact, parentResourceId *string) (*openapi.Artifact, error) {
 	var parentResourceIDPtr *int32
 
@@ -20,6 +48,9 @@ func (b *ModelRegistryService) upsertArtifact(artifact *openapi.Artifact, parent
 	if artifact == nil {
 		return nil, fmt.Errorf("invalid artifact pointer, cannot be nil: %w", api.ErrBadRequest)
 	}
+
+	// Ensure artifact has a name if it's being created
+	ensureArtifactName(artifact)
 
 	// Only convert parentResourceId to int32 if it's provided
 	if parentResourceId != nil {


dhirajsb:
I'm working on a couple of more fixes for minor issues in testing against the MLflow plugin POC.

dhirajsb:
Fixed the last few issues with metric history. All tests now pass with the mlflow plugin coming in another PR.

dhirajsb:
Looks like the embedmd DB testing classes were modified on `main`. So, I need to rebase this PR again, hopefully tomorrow.

api/openapi/model-registry.yaml
```diff
@@ -1731,7 +2434,55 @@ components:
         - CREATE_TIME
         - LAST_UPDATE_TIME
         - ID
+        - NAME
```
pboyd: The `OrderByField` was separate between model catalog and model registry before because we couldn't support sorting by name in model registry (MLMD, really). Is that addressed in this PR?

api/openapi/catalog.yaml
pboyd, line 1: Are the changes in this file necessary? It appears to just add model registry types that aren't referenced.

pboyd (commented):
This is bigger than I can effectively review in one sitting. Is it possible to split it up?

syntaxsdev:
Would it be worth it to remove the `mr_openapi/` python src files and do a separate PR?

dhirajsb:
@syntaxsdev I added the generated client files to pass a check that was failing. But they are not part of the MR client SDK and we'd still need the client SDK work you did in another PR after merging this PR. 
Or, we'll have to disable that check if we want to merge this PR without the generated python client code.

api/openapi/catalog.yaml
dhirajsb: I think that's coming from the scripts that merge files in src.

api/openapi/model-registry.yaml
dhirajsb: Good catch, I'll need to check to make sure we can now do that with embedmd (needs a better name, metadb perhaps).

api/openapi/catalog.yaml
dhirajsb: Both the filterQuery and artifactType query params should be useful for Catalog service too. It's not on the radar right now, but it should be coming soon. Catalog models should also support multiple artifact types like models, datasets, doc-artifacts, metrics and parameters like Model Registry.

Makefile
```diff
@@ -174,7 +174,7 @@ endif
 
 .PHONY: vet
 vet:
-	${GO} vet ./...
+	${GO} vet $$(${GO} list ./... | grep -v github.com/kubeflow/model-registry/internal/db/filter)
```
jonburdo:
```suggestion
	${GO} vet $$(${GO} list ./... | grep -vF github.com/kubeflow/model-registry/internal/db/filter)
```
not critical, but `-F / --fixed-strings` with `grep` is useful if the pattern is an exact string, not a pattern.

adysenrothman (approved):
/lgtm with consideration to the size of the PR, nothing stands out as needing to be changed

Makefile
tarilabs: I'm not entirely sure why these are excluded from the analysis... are you going to log a ticket to followup on later PRs?

pkg/api/api.go
```diff
@@ -127,4 +128,23 @@ type ModelRegistryApi interface {
...
+	GetExperimentRunById(id string) (*openapi.ExperimentRun, error)
+	GetExperimentRunByParams(name *string, experimentId *string, externalId *string) (*openapi.ExperimentRun, error)
+	GetExperimentRuns(listOptions ListOptions, experimentId *string) (*openapi.ExperimentRunList, error)
+
+	// EXPERIMENT RUN ARTIFACTS
+	UpsertExperimentRunArtifact(artifact *openapi.Artifact, experimentRunId string) (*openapi.Artifact, error)
+	GetExperimentRunArtifacts(artifactType openapi.ArtifactTypeQueryParam, listOptions ListOptions, experimentRunId *string) (*openapi.ArtifactList, error)
+
+	// EXPERIMENT RUN METRIC HISTORY
+	GetExperimentRunMetricHistory(name *string, stepIds *string, listOptions ListOptions, experimentRunId *string) (*openapi.ArtifactList, error)
```
tarilabs: should we add go docs like per the other API entries?

api/openapi/src/model-registry.yaml
```diff
@@ -1115,28 +1127,469 @@ paths:
...
+        - $ref: "#/components/parameters/filterQuery"
+        - $ref: "#/components/parameters/name"
+        - $ref: "#/components/parameters/stepIds"
+        - $ref: "#/components/parameters/pageSize"
+        - $ref: "#/components/parameters/orderBy"
+        - $ref: "#/components/parameters/sortOrder"
+        - $ref: "#/components/parameters/nextPageToken"
+      responses:
+        "200":
+          $ref: "#/components/responses/ArtifactListResponse"
```
tarilabs: tools suggest we should document the expectation is that the server will always return Artifact of type metric here. wdyt?

internal/converter/embdemd_openapi_converter_util.go
```diff
@@ -604,3 +604,337 @@ func MapEmbedMDPropertyModelVersionIdServeModel(source *[]models.Properties) (st
...
+	return nil
+}
+
+// DataSet mapping functions
+func MapEmbedMDURIDataSet(source *models.DataSetAttributes) *string {
+	return source.URI
+}
+
+func MapEmbedMDArtifactTypeDataSet(source *models.DataSetAttributes) *string {
+	return of("dataset")
```
syntaxsdev: The OpenAPI schema is expecting either "DataSet" or "dataset-artifact" type and here is it wrong.

```suggestion
	return of("dataset-artifact")
```

This might also need some changes for tests?

internal/core/artifact_test.go
```diff
@@ -1676,3 +1689,638 @@ func TestDocArtifactNilFieldsPreservation(t *testing.T) {
...
+	// Test all artifact types for GetArtifacts (general endpoint)
+	t.Run("GetArtifacts endpoint filtering", func(t *testing.T) {
+		testCases := []struct {
+			name         string
+			artifactType openapi.ArtifactTypeQueryParam
+			expectField  string
+		}{
+			{"model-artifact filter", openapi.ARTIFACTTYPEQUERYPARAM_MODEL_ARTIFACT, "ModelArtifact"},
+			{"doc-artifact filter", openapi.ARTIFACTTYPEQUERYPARAM_DOC_ARTIFACT, "DocArtifact"},
+			{"dataset-artifact filter", openapi.ARTIFACTTYPEQUERYPARAM_DATASET_ARTIFACT, "DataSet"},
```
syntaxsdev, line 2406: do we want this discrepancy everywhere moving forward?

The param or type adds a `_artifact`/`-artifact` suffix  but models as a DataSet for everything?

syntaxsdev (changes requested):

api/openapi/model-registry.yaml
```diff
@@ -1259,7 +1727,7 @@ components:
               type: string
             externalId:
               description: |-
-                The external id that come from the clients' system. This field is optional.
+                The external id that come from the clients’ system. This field is optional.
```
jonburdo, line 1730: This is probably okay, but odd...

jonburdo (commented):
lgtm from looking over it, though I wasn't able to read it all very thoroughly

api/openapi/model-registry.yaml
syntaxsdev: its the machine... 😉

internal/datastore/embedmd/service.go
```diff
@@ -90,14 +90,52 @@ func (s *EmbedMDService) Connect() (api.ModelRegistryApi, error) {
...
 	servingEnvironmentRepository := service.NewServingEnvironmentRepository(connectedDB, typesMap[defaults.ServingEnvironmentTypeName])
 	inferenceServiceRepository := service.NewInferenceServiceRepository(connectedDB, typesMap[defaults.InferenceServiceTypeName])
 	serveModelRepository := service.NewServeModelRepository(connectedDB, typesMap[defaults.ServeModelTypeName])
+	experimentRepository := service.NewExperimentRepository(connectedDB, typesMap[defaults.ExperimentTypeName])
+	experimentRunRepository := service.NewExperimentRunRepository(connectedDB, typesMap[defaults.ExperimentRunTypeName])
+
+	dataSetRepository := service.NewDataSetRepository(connectedDB, typesMap[defaults.DataSetTypeName])
+	metricRepository := service.NewMetricRepository(connectedDB, typesMap[defaults.MetricTypeName])
+	parameterRepository := service.NewParameterRepository(connectedDB, typesMap[defaults.ParameterTypeName])
+	metricHistoryRepository := service.NewMetricHistoryRepository(connectedDB, typesMap[defaults.MetricHistoryTypeName])
```
tarilabs, line 138: in comparison to what highlighted in https://github.com/kubeflow/model-registry/issues/1224#issuecomment-3068005968, 

> Metric (new Artifact type)

my understanding is that this PR actually defines 2 metric artifact types for the backend RDBMS: one is a generic Metric type of Artifact (which we can bind to any Contexts), while the "MetricHistory" Artifact is only affected to the ExperimentRun history Context.

Seems a departure from the original intent from: (https://github.com/kubeflow/model-registry/issues/1224#issuecomment-3068005968)
image

Is there any future plan to limit/avoid this cardinality multiplication in the backend?

tarilabs (commented):
Impressive work! I've sampled ~100 files, with some comments below for my understanding but also hopeful of help/interest to others.

I'm aware this is work related to PoC done with KFP and we are now under some unfortunate pressure to merge it, to which I'm supportive. For the future we shall do a better service to all other contributors by:
1. make sure to track any outstanding comment at merge time to follow-up on
2. define as one of the PoC exit-criteria the drafting of a pragmatic plan to transition from experimentation to industrialization, and avoid PRs of this size (since reviewing PRs of this size is a challenge on several dimensions).

api/openapi/model-registry.yaml
jonburdo: I know, I just wish the machine didn't do this :)

internal/apiutils/query_translator.go
```diff
@@ -0,0 +1,421 @@
...
+
+// FieldMapping represents a mapping from REST API field to MLMD field
+type FieldMapping struct {
+	MLMDField    string
+	FieldType    FieldType
+	PropertyType string // For properties: "string_value", "int_value", "double_value", "bool_value"
+}
+
+// getFieldMappings returns the field mappings for each entity type
+func getFieldMappings() map[EntityType]map[string]FieldMapping {
```
pboyd, line 44: Can this be a package variable instead?

internal/apiutils/query_translator.go
```diff
@@ -0,0 +1,421 @@
...
+
+// isValidValueType checks if the provided value type is valid for MLMD filter queries
+func isValidValueType(valueType string) bool {
+	validTypes := map[string]bool{
+		"string_value": true,
+		"int_value":    true,
+		"double_value": true,
+		"bool_value":   true,
+	}
+	return validTypes[valueType]
```
pboyd: Nit: a switch  would seem a little more standard.
 
```suggestion
	switch valueType {
		case "string_value", "int_value", "double_value", "bool_value":
			return true
		default:
			return false
	}
```

internal/apiutils/query_translator.go
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// unescapeDots converts escaped dots (\.) back to regular dots (.) in property names
+func unescapeDots(s string) string {
+	return strings.ReplaceAll(s, "\\.", ".")
```
pboyd: Do you need to check for an odd number of `\`? I wouldn't normally suggest it, but `findLastUnescapedDot` makes an effort to exclude `\\.`.

If you really don't need it here, can `findLastUnescapedDot` be simplified as well?

internal/apiutils/query_translator.go
```diff
@@ -0,0 +1,421 @@
...
+func parseCustomPropertyField(fieldName string) (string, string, bool) {
+	// Find the last unescaped dot
+	dotPos := findLastUnescapedDot(fieldName)
+	if dotPos == -1 {
+		// No unescaped dot found - this is not a custom property with value type
+		return "", "", false
+	}
+
+	propertyName := fieldName[:dotPos]
+	valueType := fieldName[dotPos+1:]
```
pboyd, line 177: Will this panic on strings that end in `.`?

internal/db/filter/parser.go
```diff
@@ -0,0 +1,286 @@
...
+	{Name: "Punct", Pattern: `[().,]`},
+})
+
+// Global parser instance - built once, reused everywhere (thread-safe)
+var (
+	globalParser *participle.Parser[WhereClause]
+	parserOnce   sync.Once
+)
+
+// initParser builds the parser once using sync.Once for thread safety
```
pboyd: This comment is a little misleading. I see that this function is called from `sync.Once`, but it doesn't use it itself.

internal/db/filter/parser.go
```diff
@@ -0,0 +1,286 @@
...
+
+// getParser returns the singleton parser instance (thread-safe)
+func getParser() *participle.Parser[WhereClause] {
+	parserOnce.Do(initParser)
+	return globalParser
+}
+
+// Grammar structures for SQL WHERE clauses
+
+//nolint:govet
```
pboyd, line 57: I'm guessing that `go vet` complains about participle's tags?

internal/db/filter/parser.go
```diff
@@ -0,0 +1,286 @@
...
+
+	return convertToFilterExpression(whereClause.Expression), nil
+}
+
+// Deprecated: NewParser is deprecated, use Parse() function directly
+// This is kept for backward compatibility but will be removed in future versions
+func NewParser() *Parser {
+	return &Parser{}
+}
+
```
pboyd: Do you need to add this function?

```suggestion
```

internal/db/filter/parser.go
```diff
@@ -0,0 +1,286 @@
...
+	} else {
+		// Use the new property mapping system - but we need entity type context
+		// For now, use a fallback approach and let the query builder handle it properly
+		isCustom = true // Will be properly determined later in query builder
+		valueType = inferValueType(value)
+	}
+
+	return &PropertyReference{
+		Name:      name,
+		IsCustom:  isCustom,
```
pboyd, line 238: If `IsCustom` is always `true`, why not set it here and remove the variable:

```suggestion
		IsCustom:  true,
```

Does it need to be set at all?

internal/db/filter/query_builder.go
```diff
@@ -0,0 +1,360 @@
...
+
+	default:
+		return db
+	}
+}
+
+// conditionResult holds a condition string and its arguments
+type conditionResult struct {
+	condition string
+	args      []interface{}
```
pboyd: I'd recommend `any` for new code:

```suggestion
	args      []any
```

internal/db/filter/query_builder.go
```diff
@@ -0,0 +1,360 @@
...
+		return "string_value"
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+		return "int_value"
+	case float32, float64:
+		return "double_value"
+	case bool:
+		return "bool_value"
+	default:
+		return "string_value" // fallback
+	}
```
pboyd: I've seen `"string_value"`, `"int_value"`, etc. a number of times now. Is it worth introducing a constant?

internal/db/filter/query_builder.go
```diff
@@ -0,0 +1,360 @@
...
+type EntityType string
+
+const (
+	EntityTypeContext   EntityType = "context"
+	EntityTypeArtifact  EntityType = "artifact"
+	EntityTypeExecution EntityType = "execution"
+)
+
+// QueryBuilder builds GORM queries from filter expressions
+type QueryBuilder struct {
```
pboyd, line 22: What's the advantage of creating a query builder instead of using what GORM provides? Or an open-source query builder (https://pkg.go.dev/github.com/Masterminds/squirrel, for instance)?

Makefile
dhirajsb: The package had too many false positives in vet output and it doesn't support comments in code to disable certain checks. I'll revisit this again tomorrow to see if there is an alternative. I agree that this is certainly sub-optimal.

pkg/api/api.go
dhirajsb: Good catch, added comments.

dhirajsb:
> I tried to test the filterQuery param in both existing and new types but I didn't manage to make it work (using curl, not with unit testing)
> 
> Also I would add validation on the "endTimeSinceEpoch" field so that it cannot be less than the create/update time

Got examples of filterQuery params that didn't work? I found that it required http escaping on the command line. I'll try to add integration tests to verify queries. 

Good idea to add validation for `endTimeSinceEpoch`.

api/openapi/src/model-registry.yaml
dhirajsb: Tools got that right. Instead of clarifying it in documentation, I replaced the return value with a MetricListResponseType instead. That way there is no ambiguity and we won't force clients to check artifacts in the list response.

internal/datastore/embedmd/service.go
dhirajsb: Good question. Since metric history is an internal type we store to keep track of each metric value update for ExperimentRuns, the API never exposes that. So, from a user and logical API entity point of view the above diagram is correct. 
I'll add another diagram for DB entities with MetricHistory to the feature request  and how it maps to the API entities.

internal/core/artifact_test.go
dhirajsb: Model and Doc were too overloaded, so we added Artifact suffix for those initially. The `-artifact` suffix kinda shows these are external artifacts we point to, as opposed to metric and parameter, which are just metadata values stored directly in the registry. 
Do we think we need to rename `DataSet` type to `DataSetArtifact` to follow the pattern, or does it make the name too verbose? 
Any opinion one way or another @tarilabs @Al-Pragliola ? 
If the consensus is to change it I can do that tomorrow.

internal/converter/embdemd_openapi_converter_util.go
dhirajsb: Good catch, it also looks like unit tests missed it. Fixed it in both the implementation and tests.

internal/datastore/embedmd/service.go
tarilabs: thanks, but this does not really address the question:

> Is there any future plan to limit/avoid this cardinality multiplication in the backend?

Can you expand on why this internal type differences, and reasons why it diverges from existing pattern please.

dhirajsb: Right, it's a unique situation where in experiment tracking, we need to be able to do the following:

1. Record and return the last updated value of a Metric
2. Record and return ALL updates to a Metric value (essentially, log all changes to a Metric)

For both generative and predictive AI Experiments, there is a requirement to be able to see how a metric value changed overtime as the experiment run progressed. 

I considered alternatives like using the same entity type and filtering by a discriminator property, but searching and filtering (including use of DB indices) becomes much harder with that approach. Another alternative was to explicitly introduce a version context like `ModelVersion` for Metrics. But, that ends up becoming very heavy handed and burdensome to use.

internal/db/filter/parser.go
dhirajsb: Yup, that was it and there is no way to turn it off at the line/file level, afaik. Maybe we can move that one file into a dedicated package to ignore.

internal/db/filter/parser.go
dhirajsb: Good catch, will remove the deprecated function and struct.

internal/db/filter/parser.go
dhirajsb: I was doing some experimentation :smile: with advanced queries, and this property will come in handy here in the future. Since it doesn't hurt anything at this point, I'm inclined to leave it here to use in the future.

internal/db/filter/query_builder.go
dhirajsb: Standard SQL parsers couldn't handle type suffixes, non-SQL compliant custom property names, etc. QueryBuilder maps custom filterQuery to Gorm where clauses. It also maps REST properties and custom property names to internal metadata table schema. 
There are also advanced query features we'll need soon that can't be supported directly with standard SQL.

internal/db/filter/query_builder.go
dhirajsb: Fixed.

internal/db/filter/query_builder.go
dhirajsb: Replaced all interface{} with any in query_builder.

internal/core/artifact_test.go
dhirajsb: Since there weren't any comments here, keeping it as it is for now.

Makefile
dhirajsb: Good suggestion, made the change in a couple of places in the makefile.

dhirajsb: The particle annotations caused vet to fail. Added extra logic to Makefile to still run vet on that package separately while ignoring the annotation errors.

api/openapi/model-registry.yaml
dhirajsb: Copied orderby type and parameters back to registry and catalog to have different values.

internal/apiutils/query_translator.go
dhirajsb: Having it as a function ensures immutability and encapsulation. It's called once per query parse, so it shouldn't be a big performance burden.

internal/apiutils/query_translator.go
dhirajsb: Fixed it by adding the same counting logic as findLastUnescapedDot. But, since mlmd is being removed, it's kinda redundant.

internal/db/filter/parser.go
dhirajsb: Fixed to reference caller of the function uses sync.Once.

internal/apiutils/query_translator.go
dhirajsb: It doesn't panic, added a test to verify.

internal/core/artifact_test.go
syntaxsdev: Yeah, I think its fine now imo.

syntaxsdev (approved):
Looks good and changes were fixed.

No further comments on it

dhirajsb:
@Al-Pragliola I added a comprehensive set of integration tests to verify that filterQuery now works with Service endpoints. Let me know if you think we need to do anything more in that area.

syntaxsdev:
looking into failing tests

dhirajsb:
@syntaxsdev the version should be getting prefixed in [MapModelVersionAttributesEmbedMD](https://github.com/kubeflow/model-registry/blob/9e2ee3c220d749dcda2f153c6f1110ad5e8524c5/internal/converter/openapi_embedmd_converter_util.go#L325)
And there are unit and integration tests in the server that verify it.

dhirajsb:
I'm working on a fix and test for this on the server side.

dhirajsb:
Yay, all tests are now passing. :partying_face:

pboyd (commented):
/lgtm

internal/datastore/embedmd/service.go
Al-Pragliola: Just thinking out loud if we use the same entity isn't that:

1. Take the latest metric (bound to an experiment) considering the create timestamp
2. Take all the metrics (bound to an experiment)

to avoid name conflicts we can create (as you already do for metric history) both for CREATE and UPDATE metrics new metrics with the timestamp suffix, wdyt? @dhirajsb

Al-Pragliola:
Two nits @dhirajsb 

1) Before this PR, artifacts without a name followed the pattern name = `uuid`. In this PR, this becomes `uuid:uuid`. I checked if this caused an error with the DB having mixed strategies, but it seems to support both. So, my question is whether we want to keep this new pattern or revert to the old one. It was also like this with MLMD.

2) Is the value required for a metric? I can create one value-less but does it make sense?

dhirajsb:
> * Before this PR, artifacts without a name followed the pattern name = `uuid`. In this PR, this becomes `uuid:uuid`. I checked if this caused an error with the DB having mixed strategies, but it seems to support both. So, my question is whether we want to keep this new pattern or revert to the old one. It was also like this with MLMD.

I'm thinking we need to refactor the whole prefixed name mechanism eventually. So, in the short term this is ok for standalone artifacts (which is an edge use case anyway). 
We should replace the unique name constraint with a uniqie name+parentResourceId constraint in the schema. That way names only need to be unique within a parent, and this implementation with a uuid: prefix can also be migrated at that point. 

> * Is the value required for a metric? I can create one value-less but does it make sense?

Good catch, added a check and test for that. We can't make it a required property in the schema because it breaks the generic artifact discriminiator hierarchy. Apparently json schema derived types MUST have a common set of required properties, not per derived type. :shrug:

internal/datastore/embedmd/service.go
dhirajsb: Because it breaks listing artifacts for parents like runs and versions. We'd end up grabbing the entire metric history for each metric along with other artifacts. It also breaks paging.

dhirajsb:
I created an issue for future name handling fix #1417

clients/python/tests/test_api.py
```diff
@@ -56,19 +59,146 @@ def test_post_model_version_artifacts(auth_headers: dict, artifact_type: str, ur
...
+        payload["value"] = 0.95  # Example accuracy value
+        # Use current time in milliseconds plus random offset to ensure uniqueness
+        import time
+        current_ms = int(time.time() * 1000)
+        random_offset = secrets.randbelow(1000000)  # Random offset up to 1 million ms (1000 seconds)
+        payload["timestamp"] = str(current_ms + random_offset)
+    elif artifact_type == "parameter":
+        payload["value"] = "0.001"  # Example learning rate
+        payload["parameterType"] = "string"
+
```
fege: can this be moved to an utils/helper? same code is added to the previous test, lines 60-8

clients/python/tests/test_api.py
```diff
@@ -85,13 +215,25 @@ def test_patch_artifact(auth_headers: dict, artifact_resource: Callable, artifac
...
+        create_payload["uri"] = "https://docs.example.com/initial-doc.pdf"
+    elif artifact_type == "dataset-artifact":
+        create_payload["uri"] = "s3://my-test-bucket/datasets/initial-dataset.parquet"
+        create_payload["sourceType"] = "s3"
+    elif artifact_type == "metric":
+        create_payload["value"] = 0.85
+        create_payload["timestamp"] = "1000000000"
+    elif artifact_type == "parameter":
+        create_payload["value"] = "0.01"
+        create_payload["parameterType"] = "string"
```
fege, line 308: also this is very similar

clients/python/tests/test_api.py
```diff
@@ -56,19 +59,146 @@ def test_post_model_version_artifacts(auth_headers: dict, artifact_type: str, ur
...
+        payload["value"] = 0.95  # Example accuracy value
+        # Use current time in milliseconds plus random offset to ensure uniqueness
+        import time
+        current_ms = int(time.time() * 1000)
+        random_offset = secrets.randbelow(1000000)  # Random offset up to 1 million ms (1000 seconds)
+        payload["timestamp"] = str(current_ms + random_offset)
+    elif artifact_type == "parameter":
+        payload["value"] = "0.001"  # Example learning rate
+        payload["parameterType"] = "string"
+
```
fege: maybe a support function can be used here

fege (approved):
/lgtm

clients/python/tests/test_api.py
dhirajsb: Added helper functions

clients/python/tests/test_api.py
dhirajsb: Moved into a common function.

clients/python/tests/test_api.py
dhirajsb: Done

dhirajsb:
Identified one more thing that needs to be fixed and detected by fuzzy tests. Working on a fix now. That should be the last one for all fuzzy tests.

dhirajsb:
All fuzzer tests are now passing. The PR is ready to merge.

rareddy:
Any concerns @Al-Pragliola @tarilabs in moving forward and merging this?

tarilabs:
> Any concerns @Al-Pragliola @tarilabs in moving forward and merging this?

1.
we should account for unresolved comment and ensure the "edge cases" in the backend do not spawn further 

2.
same goes for general comment that doing feature this way is not efficient

that said we already discussed 1. Release, 2. merge this, 3. merge MLMD codepath removal PR, did you had another plan of step in mind @rareddy ? 🤔

syntaxsdev:
/lgtm

reviewed delta changes since last approval

tarilabs (commented):
- https://github.com/kubeflow/model-registry/pull/1318#issuecomment-3175032821
- https://github.com/kubeflow/model-registry/pull/1318#issuecomment-3174245963
- https://github.com/kubeflow/model-registry/pull/1318#issuecomment-3167293677
- https://github.com/kubeflow/model-registry/pull/1318#pullrequestreview-3092830371
- https://github.com/kubeflow/model-registry/pull/1318#discussion_r2226524112
- https://github.com/kubeflow/model-registry/pull/1318#pullrequestreview-3030698211
- https://github.com/kubeflow/model-registry/pull/1318#pullrequestreview-3047791578

again many thanks to @dhirajsb for leading this impressive work and the exceptional review efforts by all involved 

per [discussions](https://github.com/kubeflow/model-registry/pull/1318#issuecomment-3174270785)
/approve

rareddy:
Great work by all the folks in seeing this through!

dhirajsb:
Thanks for the detailed reviews on a big PR everyone. Great team work for a huge feature improvement in the next phase of this project.

PR #1326, opened 2025-07-16
[async-job] E2E Test with Sample Job

Crazyglue:
## Description

This adds a somewhat manually run e2e test case for the sample job provided. It will (in-order):

1. Start a KinD cluster and load all dependent services (MR, Minio, Container Registry, etc)
2. Create some sample data in the MR Service (RegisteredModel, ModelVersion, and ModelArtifact)
3. Download a _tiny_ `.onnx` file (mnist) to local file system
4. Upload that `.onnx` file to the minio instance within the KinD cluster
5. Using the results from above, create a Kustomize patch-file to apply to the sample job manifest
6. Apply those manifests to kick off the Async Job
7. Awaits its completion (`10m` by default)

## How Has This Been Tested?
Locally by running the script and make commands against brand new KinD clusters

## Merge criteria:

If you have UI changes


manifests/kustomize/base/model-registry-deployment.yaml
```diff
@@ -67,7 +67,7 @@ spec:
             timeoutSeconds: 2
           readinessProbe:
             initialDelaySeconds: 10
-            periodSeconds: 60
+            periodSeconds: 20
```
Crazyglue, line 70: Running a E2E case can take a long time since the readiness probe was set to 60s, despite the server being ready much sooner than that. This is a slight adjustment just to get cycle times lower

jobs/async-upload/samples/setup-and-apply-job.sh
Crazyglue, line 1: This file essentially outlines what was mentioned in the PR Description -- Ie, creating/uploading dummy data, and apply the job manifest

jobs/async-upload/samples/patches/job-values.yaml
Crazyglue, line 1: Its not ideal, but in order to pass in the all of these variables, we are now dependent on the _order_ of the env variables. We could adjust the sample to map a ConfigMap as an ENV var to help clean this up, but that would not be representative of the "typical" use case (where the job and only the job, in its entirety, is created and applied to the cluster at a single point in time)

jonburdo (commented):
lgtm - works for me with:
```
IMG_VERSION=e2e make test-e2e-run-sample-job 
```
Eventually this script could probably be a test written in python.

jobs/async-upload/samples/setup-and-apply-job.sh
```diff
@@ -0,0 +1,105 @@
...
+
+# Apply the patched kustomization
+echo "Applying job..."
+kubectl apply -k $TEMP_DIR
+
+# Wait for job completion and show logs
+echo "Waiting for job completion..."
+kubectl wait --for=condition=complete job/$JOB_NAME --timeout=10m
+
+echo "Job completed successfully!"
```
tarilabs: I think we should check somewhere that:
- ModelArtifact.uri is valorized, and status changed
- on the OCI registry, there is the modelcar image (checking for the presence of the Manifest is fine)

jobs/async-upload/samples/setup-and-apply-job.sh
tarilabs: up to you, but for long term maintenance, I believe doing it in python allow us to leverage the MR py client, make more complete assertions, etc.

Crazyglue: I was concerned about having to do some things in one language (bash, applying manifests via kubectl and `envsubst`-ing them) and doing other things in another language (python to leverage the client for creating records, and then later validations of the job) so I tried to stick to one since `kustomize` is a pretty standard way of modifying k8s manifest files, i wanted to stay true-er to that side of things.

wdyt?

Crazyglue:
> looks to me `./scripts/deploy_on_kind.sh` is trying to load in kind another MR server image, likely some env variable/default settings?

afaik, `deploy_on_kind.sh` will check to see if the kind cluster exists, create it if need-be, and ensure the MR service is running in that cluster. Its used both in the mr-client e2e tests and on the async-job's e2e tests. i guess it has some levers to change whether it pulls an existing image vs builds the image from source before loading it into the kind cluster.

but yeah its doing a bunch of defaulting (like database deployment, db secrets, etc) just to get the service up and running and usable.

jobs/async-upload/samples/setup-and-apply-job.sh
tarilabs: indeed, but even at the current stage, looks to me the most of the lines are dedicated to REST invocations, which can leverage existing clients etc. that's why I want to at least consider invoking kustomize/kubectl from python, or use py binding libraries for it. wdyt?
this also would align with QE strategy which has moved tests to python (also as "high level scripting") but if you think it's more consistent in bash, we will maintain bash until refactoring will be needed to extend testing 👍

jobs/async-upload/pyproject.toml
```diff
@@ -10,11 +10,16 @@ packages = [{include = "job"}]
 python = ">=3.9,<4.0"
 configargparse = ">=1.7.1,<2.0.0"
 model-registry = { version = ">=0.2.19,<0.3.0", extras = ["boto3", "olot"] }
+requests = "^2.31.0"
+pyyaml = "^6.0"
```
tarilabs: shoudln't these 2 be dev/integraton dependencies too? 🤔

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -0,0 +1,378 @@
...
+        print("Preparing job patches...")
+        base_job_path = Path(__file__).parent.parent.parent / "samples" / "sample_job_s3_to_oci.yaml"
+        
+        # Define patches to apply to the job
+        patches = {
+            "/metadata/name": job_name,
+            "/spec/template/spec/containers/0/image": container_image_uri,
+            "/spec/template/spec/containers/0/env/12/value": rm.id,
+            "/spec/template/spec/containers/0/env/13/value": mv.id,
+            "/spec/template/spec/containers/0/env/14/value": ma.id,
```
tarilabs: can we patch for a json-path, possibly using some library, instead of hardcoding 12, 13, 14?
otherwise, please add a comment of the Env name for 12, 13, 14 for the next dev who will be looking into this :)

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -0,0 +1,378 @@
...
+                # Job doesn't exist yet
+                time.sleep(5)
+                continue
+            else:
+                raise
+    
+    return False
+
+
+def upload_to_minio(file_path: str, bucket: str, key: str) -> None:
```
tarilabs, line 202: this works for me, but since boto3 is already part of our dev/test dependencies, wouldn't be easier to just use boto3 here?

tarilabs (commented):
thanks a lot for the update @Crazyglue !

some comments for your consideration, also in follow\up PR if preferred!

jobs/async-upload/pyproject.toml
Crazyglue: this was moved to the correct section, whoops!

jobs/async-upload/tests/integration/test_integration_async_upload.py
Crazyglue: I converted over to a `patches`-style kustomize, this did require me to create another temp-file with the actual patches but it should be much cleaner and maintainable now

jobs/async-upload/tests/integration/test_integration_async_upload.py
Crazyglue: This was a huge oversight. Fixed.

.github/workflows/async-upload-test.yml
```diff
@@ -17,11 +17,7 @@ on:
       - ".github/workflows/**"
 
 env:
-  IMG_REGISTRY: ghcr.io
-  IMG_ORG: kubeflow
```
tarilabs, line 21: To ease maintenance in m/s i would really prefer to have these at the top, this way we just adjust OCI reference in a single place when porting. Can you restore these Envs here, please?

jobs/async-upload/Makefile
```diff
@@ -1,15 +1,22 @@
...
+IMG ?= ghcr.io/kubeflow/model-registry/server:$(IMG_VERSION)
 
 .PHONY: deploy-latest-mr
 deploy-latest-mr:
 	cd ../../ && \
 	$(if $(filter true,$(BUILD_IMAGE)),\
-		IMG_VERSION=${IMG_VERSION} make image/build ARGS="--load$(if ${DEV_BUILD}, --target dev-build)" && \
+		IMG_VERSION=${IMG_VERSION} IMG=${IMG} make image/build ARGS="--load$(if ${DEV_BUILD}, --target dev-build)" && \
+	,\
+		docker pull $(IMG) && \
```
tarilabs: this is a bit suspicious; we're entering this code path when BUILD_IMAGE is true, which is also the default with `BUILD_IMAGE ?= true` definition; we don't have to docker pull which may pull from remote container registry?

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -0,0 +1,365 @@
...
+        model_file = Path(temp_dir) / "mnist-8.onnx"
+        
+        # Download the model
+        response = requests.get(
+            "https://github.com/onnx/models/raw/refs/heads/main/validated/vision/classification/mnist/model/mnist-8.onnx"
+        )
+        response.raise_for_status()
+        
+        with open(model_file, "wb") as f:
+            f.write(response.content)
```
tarilabs, line 292: in a follow-up PR, I would place this file in `tests/data` to avoid the network call, we will take onnx from one of our examples.

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -0,0 +1,365 @@
...
+        # Wait a bit for the artifact to be updated
+        time.sleep(2)
+        
+        # Fetch the updated artifact
+        updated_ma = model_registry_client.get_model_artifact(model_name, "v1.0.0")
+        assert updated_ma
+        
+        # Validate the artifact was updated correctly
+        assert updated_ma.uri != "PLACEHOLDER", f"URI was not updated: {updated_ma.uri}"
+        assert updated_ma.state == ArtifactState.LIVE, f"State was not updated to LIVE: {updated_ma.state}"
```
tarilabs, line 349: awesome, thank you

jobs/async-upload/Makefile
Crazyglue: This second block (the `docker pull ...`) should be executed when `BUILD_IMAGE=false` (technically, when `BUILD_IMAGE` is anything _but_ `true`).

I added this because during the later step to load the image into the kind registry, it would fail since the image was not present locally. I suppose this block should not be needed anymore since I've changed the ci/cd to always build the image anyways. I will remove them

.github/workflows/async-upload-test.yml
Crazyglue: Yeah, there are some issues with how the ENV vars are being passed around to downstream make commands, etc. So I think a more comprehensive refactor of all the make files will be needed. This was really done to get all the tests to actually use the variables it should be using. For example, even providing an `IMG` to the make file for the mr-server image will not always take the correct `IMG_VERSION`. There are some hard-coded instances in some of these makefiles.

Let me take a look and see if I can do a minimal refactor with the env vars here restored

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -0,0 +1,365 @@
...
+    assert mv and mv.id
+    print(f"  Created ModelVersion with ID: {mv.id}")
+    
+    ma = model_registry_client.get_model_artifact(model_name, "v1.0.0")
+    assert ma and ma.id
+    print(f"  Created ModelArtifact with ID: {ma.id}")
+    
+    # Verify initial state
+    assert ma.uri == "PLACEHOLDER"
+    assert ma.state == ArtifactState.UNKNOWN
```
tarilabs, line 278: this makes me realize in the MR python client we should probably by default create Artifact State live ... (in a followup PR/ticket!) or check which alignment with the pure REST API call

Makefile
```diff
@@ -31,8 +31,10 @@ IMG_REPO ?= model-registry/server
 # container image build path
 BUILD_PATH ?= .
 # container image
-ifdef IMG_REGISTRY
-    IMG := ${IMG_REGISTRY}/${IMG_ORG}/${IMG_REPO}
+ifdef IMG
+	IMG := ${IMG}
+else ifdef IMG_REGISTRY
+    IMG := ${IMG_REGISTRY}/${IMG_ORG}/${IMG_REPO}:${IMG_VERSION}
```
tarilabs, line 37: to me this is a great catch, if IMG was defined the makefile would have ignored previously

cc @Al-Pragliola  (if we want to do differently I'm also open to followups)

jobs/async-upload/Makefile
```diff
@@ -1,17 +1,22 @@
...
+JOB_IMG_REGISTRY ?= ghcr.io
+JOB_IMG_ORG ?= kubeflow
+JOB_IMG_NAME ?= model-registry/job/async-upload
+JOB_IMG ?= $(JOB_IMG_REGISTRY)/$(JOB_IMG_ORG)/$(JOB_IMG_NAME):$(JOB_IMG_VERSION)
 BUILD_IMAGE ?= true # whether to build the MR server image
+CLUSTER_NAME ?= mr-e2e
+
+# MR Server Params
+IMG_VERSION ?= latest
+IMG ?= ghcr.io/kubeflow/model-registry/server:$(IMG_VERSION)
```
tarilabs, line 11: I'm really tempted to 
```suggestion
IMG ?= $(JOB_IMG_REGISTRY)/$(JOB_IMG_ORG)/model-registry/server:$(IMG_VERSION)
```

but I don't want to diverge on the scope of the PR too much :)

tarilabs (approved):
Thanks a lot for the great work @Crazyglue !

Minor comment below I'd say in followup PRs/ticket because I believe this is good to take this in and unblock further work

/lgtm
/approve

PR #1337, opened 2025-07-19
[Draft] Add MLflow SDK support for Model Registry as a Tracking Store, fixes #1225

### LINKED ISSUE #1225: Add MLflow SDK support for Model Registry as a store

dhirajsb:
## Description

MLflow SDK supports integrations with non-mlflow Tracking stores by implementing a tracking.AbstractStore API. 
This PR implements that API to allow MLflow SDK to directly call Model Registry APIs for storing Experiment and Run metadata. 
See the feature request issue #1225 for motivation, design and implementation details. 

This PR is based on the Experiment Tracking PR #1318 and must be merged after that PR. 
Fixes #1225.

## How Has This Been Tested?

Extensive unit and e2e tests have been provided. A local e2e test is provide to test against the go REST server, as well as an e2e test to run against a deployed model registry to verify RBAC support. 

## Merge criteria:

If you have UI changes


dhirajsb:
The long commit history should become much shorter once #1318 is merged in main.

dhirajsb:
To review this PR, focus on the `clients/modelregistry_plugin` plugin project.

dhirajsb:
Re-based branch on main on top of mlmd removal. Ready for review.

clients/modelregistry_plugin/.gitignore
pboyd, line 1: Should this be its own repo? Or, at least, should it be in `clients/mlflow/modelregistry_plugin` instead?

clients/modelregistry_plugin/README.md
```diff
@@ -0,0 +1,384 @@
...
+The plugin supports various authentication methods:
+
+```bash
+# Token-based auth
+export MODEL_REGISTRY_TOKEN="your-token"
+
+# Kubernetes service account token from `/var/run/secrets/kubernetes.io/serviceaccount/token`
+```
+
+### SSL/TLS and CA Certificate Configuration
```
pboyd:
```suggestion
### TLS and CA Certificate Configuration
```

clients/modelregistry_plugin/README.md
```diff
@@ -0,0 +1,384 @@
...
+```bash
+# Option 1: Custom CA certificate via environment variable
+export MODELREGISTRY_CA_CERT_PATH="/path/to/your/ca.crt"
+export MLFLOW_TRACKING_URI="modelregistry+https://registry.example.com:8080"
+
+# Option 2: In Kubernetes - works automatically (no configuration needed)
+export MLFLOW_TRACKING_URI="modelregistry+https://model-registry-service:8080"
+# Plugin automatically uses /run/secrets/kubernetes.io/serviceaccount/ca.crt
+
+# Option 3: HTTP (no SSL) - CA configuration is skipped
```
pboyd:
```suggestion
# Option 3: HTTP (no TLS) - CA configuration is skipped
```

clients/modelregistry_plugin/modelregistry_plugin/api_client.py
```diff
@@ -0,0 +1,176 @@
...
+
+        try:
+            response = self.session.request(method, url, headers=headers, **kwargs)
+            response.raise_for_status()
+            response_json = response.json()
+        except requests.exceptions.SSLError as e:
+            # Handle SSL certificate errors specifically
+            msg = (
+                f"SSL certificate verification failed connecting to Model Registry: {e}"
+            )
```
pboyd:
```suggestion
            # Handle TLS certificate errors specifically
            msg = (
                f"TLS certificate verification failed connecting to Model Registry: {e}"
            )
```

clients/modelregistry_plugin/modelregistry_plugin/auth.py
```diff
@@ -0,0 +1,30 @@
...
+    # Token-based authentication
+    token = os.getenv("MODEL_REGISTRY_TOKEN")
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
+        return headers
+
+    # Kubernetes service account token
+    token_path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+    if os.path.exists(token_path):
+        with open(token_path) as f:
```
pboyd: Should you catch and ignore the exception from `open`?

The behavior seems a little odd to me:
- No token file exists: nothing happens
- Token file exists but is unreadable: exception

clients/modelregistry_plugin/modelregistry_plugin/auth.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+
+
+def get_auth_headers() -> dict[str, str]:
+    """Get authentication headers for Model Registry API requests.
+
+    Returns:
+        Dictionary of headers for authentication
+    """
+    headers = {"Content-Type": "application/json", "Accept": "application/json"}
```
pboyd: These are unrelated to auth. Maybe this function is misnamed?

clients/modelregistry_plugin/.python-version
```diff
@@ -0,0 +1 @@
+3.9
```
jonburdo: Is this just to have us developing on the same version locally? We don't have this in the python client or async-job, but I guess it probably makes sense here since we're using uv

clients/modelregistry_plugin/pyproject.toml
```diff
@@ -0,0 +1,35 @@
+[project]
+name = "modelregistry_plugin"
```
jonburdo:
```suggestion
name = "model-registry-plugin"
```

clients/modelregistry_plugin/pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+]
+
+[project.entry-points."mlflow.tracking_store"]
+modelregistry = "modelregistry_plugin.tracking_store:ModelRegistryTrackingStore"
+"modelregistry+http" = "modelregistry_plugin.tracking_store:ModelRegistryTrackingStore"
+"modelregistry+https" = "modelregistry_plugin.tracking_store:ModelRegistryTrackingStore"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
jonburdo: Do we want to just use uv for builds? Especially since we have `uv build` in the Makefile
```suggestion
requires = ["uv_build>=0.8.8,<0.9.0"]
build-backend = "uv_build"
```

clients/modelregistry_plugin/.gitignore
jonburdo: I'm thinking `clients/package-name` is okay. The package name should use dashes instead of underscores and it would be more consistent to make `model-registry` two words - `model-registry-plugin` or `model-registry-mlflow-plugin`

clients/modelregistry_plugin/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+## Project Structure
+
+```
+clients/modelregistry_plugin/
+├── modelregistry_plugin/          # Main plugin package
+│   ├── __init__.py
+│   ├── auth.py                   # Authentication handling
+│   ├── tracking_store.py         # Main ModelRegistryTrackingStore implementation
+│   └── utils.py                  # Utility functions
```
jonburdo: I'd suggest using a `src` dir. It's common practice and we do it in the python client. `uv init --library <name>` gives a good starting point.

clients/modelregistry_plugin/tests/test_store.py
```diff
@@ -0,0 +1,102 @@
...
+        store = ModelRegistryTrackingStore("modelregistry://localhost:8080")
+
+        assert store.api_client.base_url == store.base_url
+        assert hasattr(store.api_client, "session")
+        assert hasattr(store.api_client, "request")
+
+
+if __name__ == "__main__":
+    # Allow running the tests directly
+    pytest.main([__file__, "-v"])
```
jonburdo: Unless there's a specific need, I don't think we want this. Best to just stick with `pytest ...`

clients/modelregistry_plugin/REFACTORING_SUMMARY.md
```diff
@@ -0,0 +1,223 @@
...
+
+## 🚀 **Usage Examples**
+
+### **Using the New Store**
+```python
+from modelregistry_plugin import ModelRegistryStore
+
+# Initialize store
+store = ModelRegistryStore(
+    store_uri="modelregistry://localhost:8080",
```
jonburdo: Is there any difference between these two:
`http://localhost:8080`
`modelregistry://localhost:8080`

Should we just stick with http?

clients/modelregistry_plugin/pyproject.toml
jonburdo: or
`model-registry-mlflow`
`model-registry-mlflow-plugin`

clients/modelregistry_plugin/.gitignore
dhirajsb: I can change the pkg directory name to `model-registry-plugin`.

clients/modelregistry_plugin/modelregistry_plugin/auth.py
dhirajsb: Good catch. the auth header handling needs cleanup.

clients/modelregistry_plugin/modelregistry_plugin/auth.py
dhirajsb: That particular header can be moved out. Remnants of a refactor from older code.

clients/modelregistry_plugin/REFACTORING_SUMMARY.md
dhirajsb: The prefix is used by mlflow to load the right store plugin.

jonburdo (approved):
/lgtm

clients/model-registry-mlflow/src/model_registry_mlflow/api_client.py
```diff
@@ -97,6 +97,23 @@ def _configure_ca_cert(self, ca_cert_path: Optional[str] = None) -> None:
...
             # Let requests use system CA bundle (default behavior)
 
+    def _get_request_headers(self) -> Dict[str, str]:
+        """Get request headers."""
+
+        from mlflow.exceptions import MlflowException
+
+        headers = {"Content-Type": "application/json", "Accept": "application/json"}
+
+        if self.base_url.startswith("https://"):
```
jonburdo, line 107: I think this is correct - just want to confirm. This is always a place where we use `https://` at the beginning, and not something like `modelregistry+https://something` right?

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
```diff
@@ -1,30 +1,80 @@
...
+        """Get the Kubernetes token with caching."""
+        with self._lock:
+            try:
+                mtime = os.path.getmtime(_TOKEN_PATH)
+                if self._token_mtime == mtime and self._token is not None:
+                    return self._token
+
+                # Need to refresh token - read file
+                with open(_TOKEN_PATH, "r") as f:
+                    self._token = f.read().strip()
```
jonburdo: Is it possible we'd end up writing this file at the same time we're reading it? Seems like this is maybe okay, since we're unlikely to be writing this a lot. If we did want to make it more robust, we could do one of these:
- add file-locking
- make sure mtime is the same before and after we read, and that the token is the expect length

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
```diff
@@ -1,30 +1,80 @@
...
+                self._token = None
+                self._token_mtime = None
+                raise RuntimeError(
+                    f"Error accessing Kubernetes token file {_TOKEN_PATH}: {e}"
+                ) from e
 
-def get_auth_headers() -> dict[str, str]:
+
+# Global token cache instance
+_token_cache = K8sTokenCache()
```
jonburdo: It's okay for now, but here's my suggestion for the future:

Global state comes with caveats, and this is leaning into the module-as-a-class pattern, which is not ideal. Can we either make `_token_cache` and `get_auth_headers` attributes of a new class (`ModelRegistryAuth` or something), and then give `ModelRegistryAPIClient` and `_auth` property?

If we really want to share the same instance of this auth class across multiple `ModelRegistryAPIClient` instances, we can add logic to do that.

If you think this is reasonable, we can add an issue to do it in a followup. Also curious - how much of an problem is the file-read overhead without this? Or if we do a cache per `ModelRegistryAPIClient` instance instead of a single global cache?

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
```diff
@@ -1,30 +1,80 @@
...
+
+
+def _get_k8s_token() -> Optional[str]:
+    """Get Kubernetes service account token with caching."""
+    global _token_cache
```
jonburdo: This line is not needed. You can reference an outer variable from inside a function and even modify it. This would just be needed if you were re-assigning the variable itself from inside the function and wanted the outer variable to be re-assigned:
```python
def f():
    global _token_cache
    _token_cache = K8sTokenCache()
```

clients/model-registry-mlflow/tests/test_api_client.py
```diff
@@ -14,7 +14,7 @@ class TestModelRegistryAPIClient:
     @pytest.fixture
     def api_client(self):
         """Create a ModelRegistryAPIClient instance for testing."""
-        return ModelRegistryAPIClient("http://localhost:8080")
+        return ModelRegistryAPIClient("https://localhost:8080")
```
jonburdo, line 17: I guess the idea is never to use this for actual requests? Feels a little odd to use `https` with `localhost`, but it's fine

clients/model-registry-mlflow/tests/test_auth.py
```diff
@@ -3,44 +3,222 @@
...
+            RuntimeError, match="Error accessing Kubernetes token file.*invalid"
+        ):
+            _get_k8s_token()
+
+        # Verify cache was cleared
+        assert _token_cache._token is None
+        assert _token_cache._token_mtime is None
+
+    @patch("os.path.getmtime")
+    @patch("builtins.open", new_callable=mock_open)
```
jonburdo: I really try to avoid mocking common builtin functions. These can by used by other builtin python functions, by pytest itself, and by other functions we may want to add to the test without the mock applying.

Usually it should be avoidable by adjusting the code slightly and/or the test. In this case you could use the `tmp_path` fixture and patch `_TOKEN_PATH` be `str(tmp_path / "token")`. Alternatively, using one or two more helper functions for the `get_token()` implementation could make testing easier.

We may be okay mocking these builtins in this instance, but a couple other things to note:
- I believe you should be able to narrow the scope of the patch by patching off of the relevant module (i.e. `model_registry_mlflow.auth.open`, `model_registry_mlflow.os.path.getmtime`)
- There are context managers for unittest's `patch` and pytest's `monkeypatch` which can help reduce the scope (not as relevant to this particular test, but just fyi)

clients/modelregistry_plugin/.gitignore
dhirajsb: Changed project dir to `model-registry-mlflow` and package name to `model_registry_mlfow` per python naming conventions.

clients/modelregistry_plugin/modelregistry_plugin/auth.py
dhirajsb: Cleaned up header handling, including a simple K8s auth token cache.

clients/modelregistry_plugin/tests/test_store.py
dhirajsb: Removed.

clients/modelregistry_plugin/CONTRIBUTING.md
dhirajsb: Moved pkg dir to src/ to cleanup the module root dir.

clients/modelregistry_plugin/pyproject.toml
dhirajsb: Switched to uv build from hatchling.

clients/modelregistry_plugin/pyproject.toml
dhirajsb: Changed project name to `model-registry-mlflow`.

clients/modelregistry_plugin/modelregistry_plugin/auth.py
dhirajsb: Moved content-type header out of auth header handling.

clients/modelregistry_plugin/.python-version
dhirajsb: Good catch, pyproject already has `requires-python`, so no need for this file. Removed.

clients/model-registry-mlflow/src/model_registry_mlflow/api_client.py
dhirajsb: There is a util function parse_tracking_uri that helps do this mapping to remove the `modelregistry` prefix.

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
dhirajsb: Good catch, added a second mtime check and extra error handling in a retry loop to ensure that doesn't happen.

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
dhirajsb: I'll leave it module global for now, since token either comes from an env variable or a pod token file. So, it's pretty much a singleton for the process. We can add an issue for future if you think we can cleanup the code a bit.

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
dhirajsb: Removed redundant `global`, remnant from earlier code directly updating shared token.

clients/model-registry-mlflow/src/model_registry_mlflow/auth.py
dhirajsb: Made the class a singleton to avoid global variable.

clients/model-registry-mlflow/tests/test_api_client.py
dhirajsb: It's just for local tests to validate auth token management works.

clients/model-registry-mlflow/tests/test_auth.py
dhirajsb: Good catch, replaced all the io mocking with a mock token file instead.

dhirajsb:
@jonburdo thanks for the detailed review. I have addressed all the comments. On using mtime for token file caching, I wrote a local benchmark script (using cursor) and it found that with a 95%+ hit rate, using mtime to avoid reading the file would be 4-5X faster.

dhirajsb:
@tarilabs @Al-Pragliola please add an approval so we can close this.

rareddy:
@dhirajsb please hold on this with merging; we did not have any conversations in the community on the direction; we need to establish what we are signing up here. Supporting MLflow this way is big commitment, especially given [experiment tracking KEP ](https://github.com/kubeflow/community/pull/892)was closed without resolution in the KFP recently.

PR #1365, opened 2025-07-22
add support to async job for huggingface source

jonburdo:
This enables the model storage async job to have as the source a hf URI in the following forms:

  hf://repo-name/model-name
  hf://repo-name/model-name:hash

## How Has This Been Tested?

config was validated by setting env vars in a file `env`:
```
TERMINATION_MESSAGE_PATH=/dev/null
MODEL_SYNC_SOURCE_TYPE=uri
MODEL_SYNC_SOURCE_URI=hf://repo/model
```
and running from async-job dir:
```
set -a; source ./env; set +a
python -m job.entrypoint
```
This should error on the destination config checks - after the source config is checked.

The `download_from_hf` function was tested by running this python script:
```python
from job.download import perform_download


client = None
config = {
    "source": {
        "type": "uri",
        "uri": "hf://sadhaklal/logistic-regression-iris",
    },
    "storage": {
        "path": "./tmp/model",
    },
}

```
This should download some files in `./tmp/model/`:
```
pytorch_model.bin  README.md
```

I can add some tests, but not sure how much we want at this stage. It will be good to test e2e once the first e2e test is merged.

## Merge criteria:

If you have UI changes

### DIFF
--- a/jobs/async-upload/job/config.py
+++ b/jobs/async-upload/job/config.py
@@ -19,7 +19,7 @@ def _parser() -> cap.ArgumentParser:
     # --- source ---
     # s3
     # TODO: We should be able to infer the type from the credentials provided, therefore no default needed
-    p.add_argument("--source-type", choices=["s3", "oci", "hf"], default="s3")
+    p.add_argument("--source-type", choices=["s3", "oci", "uri"], default="s3")
     p.add_argument("--source-uri")
     p.add_argument("--source-aws-bucket")
     p.add_argument("--source-aws-key")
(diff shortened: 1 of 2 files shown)


jobs/async-upload/job/config.py
```diff
@@ -19,7 +19,8 @@ def _parser() -> cap.ArgumentParser:
     # --- source ---
     # s3
     # TODO: We should be able to infer the type from the credentials provided, therefore no default needed
-    p.add_argument("--source-type", choices=["s3", "oci"], default="s3")
+    p.add_argument("--source-type", choices=["s3", "oci", "hf"], default="s3")
```
tarilabs: thanks @jonburdo , the only minor comment I have is that I would "classify" this as `uri` and have `hf://`, future `http://`, `https://` as supported protocols

jonburdo: Okay. I'm actually a bit confused regarding why we have this source type.  Isn't everything we're supporting a URI? I realize we're currently specifying the bucket and key separately for s3, but idk if we should be. For example, these are all examples of URIs:
```
s3://a/b
oci://a/b
hf://a/b
https://a/b
```
Seems like we could just take a uri and derive the protocol from whatever is before the `://`. Maybe I'm missing something here though.

jonburdo: For the record, it turns out we do this to match existing ODH types where we have: s3, oci, uri

PR #1374, opened 2025-07-24
Refactor async job config to be more pythonic and typesafe

Crazyglue:
## Description
Now that the core functionality of the async job is ready, we can start to go back and clean a few things up. This is one of those things. It will greatly reduce the complexity of the config, as most of the configuration types are shared between source and destination

## How Has This Been Tested?
Updating the tests accordingly and ensuring the `make test`, `make test-e2e` and `make test-integration` all pass. Additionally, executing the sample job locally on a KinD cluster to triple-check that everything is still working as intended

## Merge criteria:

If you have UI changes


jobs/async-upload/.python-version
```diff
@@ -0,0 +1 @@
+3.11.11
```
Crazyglue: For some reason my local development env was using 3.10 for this project. Some of the types used require python 3.11, so placing this here to ensure thats the case going forward

jobs/async-upload/job/models.py
Crazyglue, line 1: This is basically all the new stuff, everything else in the PR is changes to reflect the configuration objects/typings found here

jobs/async-upload/job/upload.py
```diff
@@ -4,58 +4,60 @@
...
-            access_key_id=destination_config["s3"]["access_key_id"],
-            secret_access_key=destination_config["s3"]["secret_access_key"],
-            region=destination_config["s3"]["region"],
+            bucket_name=config.destination.bucket,
+            s3_prefix=config.destination.key,
+            endpoint_url=config.destination.endpoint_url,
+            access_key_id=config.destination.access_key_id,
+            secret_access_key=config.destination.secret_access_key,
+            region=config.destination.region,
         )
```
Crazyglue, line 28: This is representative of most of the changes happening in the repo. Changing a dict-key-lookup for a proper dot-notation lookup

jobs/async-upload/job/models.py
```diff
@@ -0,0 +1,150 @@
...
+    email: str | None = None
+    base_image: str = "busybox:latest"
+    enable_tls_verify: bool = True
+
+    @model_validator(mode='after')
+    def validate_oci_credentials(self) -> 'OCIConfig':
+        """Validate OCI credentials consistency."""
+        if (self.username is None) != (self.password is None):
+            raise ValueError("Both OCI username and password must be set together or both be None")
+        return self
```
tarilabs, line 54: nice setup

jobs/async-upload/job/models.py
```diff
@@ -0,0 +1,150 @@
...
+
+class OCIStorageConfig(BaseStorageConfig, OCIConfig):
+    """OCI storage configuration with validation - can be used for both source and destination."""
+    
+    @model_validator(mode='after')
+    def validate_oci_storage(self) -> 'OCIStorageConfig':
+        """Validate that required OCI fields are present."""
+        if not all([self.registry, self.uri]):
+            raise ValueError("OCI registry and URI must be set")
+        return self
```
tarilabs, line 76: considering these 2 pydantic classes, any particular reasons to keep them separate from the `...Config` variant given they only add additional validation checks? is there a use-case of the `...Config`?

tarilabs (commented):
thanks a lot for this nice refactoring @Crazyglue !

minor question below, otherwise LGTM

jobs/async-upload/job/models.py
Crazyglue: Yes, so for clarity there are 2 differences. 1) the validation functions and 2) the `credentials_path` field.

Keeping them separate is needed in the current implementation because (using S3 as an example) we instantiate the `S3Config(...)` using whatever variables happened to have come from `configargparse`. This can include combinations which are invalid, and so if we performed the validation at this step, we would get a false-negative. 

Then, we parse the file-based credentials and overlay those ontop of the variable. At that point we have all the data needed for the validation to be triggered, which is where the following is called (and the `credentials_path` is set)

```
        source_config = S3StorageConfig(
            **s3_config.model_dump(),
            credentials_path=args.source_s3_credentials_path
        )
```

This is the same case with the `OCIConfig` class (but not the case with the URI-based config, as there are no supported credentials for that source-type, just a URI is needed)

I think its doable to combine the two classes into one, but it seems to me to be equivalent in terms of complexity/maintainability.

I can add a comment to the classes to clarify their purpose and how they are used.

wdyt?

jonburdo (commented):
lgtm - latest changes make sense. A couple overly long lines in the doc-strings, but we can fix this later by adding proper linting/formatting

jobs/async-upload/job/models.py
tarilabs: thanks for having added the comments with https://github.com/kubeflow/model-registry/pull/1374/commits/c0b89238249da842cf4a460d52e050c44eecaec2

tarilabs (approved):
- https://github.com/kubeflow/model-registry/pull/1374#pullrequestreview-3064246821

thanks again @Crazyglue 

/lgtm
/approve

PR #1375, opened 2025-07-25
add support to async job for http source

jonburdo:
This allows individual files to be specified:

  http://example.com/some-file.bin
  https://example.com/some-file.bin

A tar or zip archive will be extracted when they are downloaded:

  https://example.com/some-file.tar.gz
  https://example.com/some-file.zip

## How Has This Been Tested?

I manually tested the download function by creating a script `download_uri.py`:
```python
import sys
from job.download import perform_download


client = None
config = {
    "source": {
        "type": "uri",
        "uri": sys.argv[1],
    },
    "storage": {
        "path": "./tmp/model",
    },
}

```
and testing it on a single model file, a tar, and a zip:
```
python download_uri.py https://huggingface.co/sadhaklal/logistic-regression-iris/resolve/main/pytorch_model.bin
python download_uri.py https://github.com/kubeflow/model-registry/archive/refs/tags/v0.2.21.tar.gz
python download_uri.py https://github.com/kubeflow/model-registry/archive/refs/tags/v0.2.21.zip
```
Then confirming the files were downloaded as expected, and that the tar and zip were extracted.

I tested error cases with bad urls:
```
python download_uri.py https://huggingface.co/sadhaklal/logistic-regression-iris/resolve/main/NOT_A_FILE
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/jobs/async-upload/tests/integration/test_integration_async_upload.py
+++ b/jobs/async-upload/tests/integration/test_integration_async_upload.py
@@ -16,6 +16,8 @@ try:
 except ImportError:
     pytest.skip("model-registry client not available", allow_module_level=True)
 
+HTTP_SOURCE = "https://github.com/onnx/models/raw/refs/heads/main/validated/vision/classification/mnist/model/mnist-8.onnx"
+
 
 @pytest.fixture(scope="session")
 def k8s_client():
@@ -238,9 +240,7 @@ def _setup_s3(tmp_path):
     model_filepath = model_dirpath / "mnist-8.onnx"
 
     # Download the model
-    response = requests.get(
-        "https://github.com/onnx/models/raw/refs/heads/main/validated/vision/classification/mnist/model/mnist-8.onnx"
-    )
+    response = requests.get(HTTP_SOURCE)
     response.raise_for_status()
 
     model_dirpath.mkdir()
@@ -260,7 +260,7 @@ def _setup_s3(tmp_path):
             None,
             {
                 "MODEL_SYNC_SOURCE_TYPE": "uri",
-                "MODEL_SYNC_SOURCE_URI": "https://huggingface.co/sadhaklal/logistic-regression-iris/resolve/main/pytorch_model.bin",
+                "MODEL_SYNC_SOURCE_URI": HTTP_SOURCE,
             },
         ),
         (


jobs/async-upload/job/download.py
```diff
@@ -74,6 +95,79 @@ def download_from_hf(uri: str, dest_dir: str) -> str:
...
+
+        is_archive = True
+        match mimetype:
+            case "application/zip":
+                valid_content_types = ZIP_CONTENT_TYPES
+            case "application/x-tar":
+                valid_content_types = TAR_CONTENT_TYPES
+            case _:
+                valid_content_types = REGULAR_FILE_CONTENT_TYPES
+                is_archive = False
```
jonburdo: This is a simplified version of what we have in the [kserve-storage function](https://github.com/kserve/kserve/blob/4edbb36c520c2e880842229bfc56b7f11d766822/python/storage/kserve_storage/kserve_storage.py#L698-L771). I wonder if we need to be this strict or if we should simply download whatever url we get. Maybe a warning instead of an error would make sense.

jobs/async-upload/job/download.py
```diff
@@ -74,6 +95,79 @@ def download_from_hf(uri: str, dest_dir: str) -> str:
...
+                f"URI {uri} appears to have MIME type {mimetype} but did not respond with any of following for 'Content-Type': {content_types_str}"
+            )
+
+        dest_path = os.path.join(dest_dir, filename)
+        os.makedirs(dest_dir, exist_ok=True)
+        with open(dest_path, "wb") as dest_file:
+            shutil.copyfileobj(response.raw, dest_file)
+
+    if is_archive:
+        dest_dir = _unpack_archive_file(dest_path, mimetype, dest_dir)
```
jonburdo: Would we ever want the option to tell this function to simply download archives without extracting them?

tarilabs: great question, but not in our scope of producing a ModelCar 🚗 
For the time being, let's keep it to always maintain extract/unarchive.

jobs/async-upload/job/download.py
tarilabs: > I wonder if we need to be this strict or if we should simply download whatever url we get. Maybe a warning instead of an error would make sense.

SGTM since it's a superset of the KServe storage fn

tarilabs (commented):
thanks @jonburdo , please add an (e2e possibly?) test, given it should be possible to put under non-regression the `download_from_http()`, wdyt?

jobs/async-upload/job/download.py
```diff
@@ -74,6 +95,79 @@ def download_from_hf(uri: str, dest_dir: str) -> str:
...
+    else:
+        mimetype, encoding = mimetypes.guess_type(url.path)
+
+    if filename == "":
+        raise ValueError(f"No filename contained in URI: {uri}")
+
+    # Get header information from host url
+    host_uri = url.hostname
+    headers_env_name = host_uri + HEADERS_SUFFIX
+    if headers_env_name in os.environ:
```
Crazyglue: is this also from kserve? using the suffix above, I would have to set something like `example.com-headers="content-type=application/gzip"` as an env var, but typical OS env vars are upper-snake-case: `EXAMPLE.COM_HEADERS="content-type=application/gzip"`

rareddy:
qq: Does the HTTP source support the Hugging Face or is this for generic http model source?

tarilabs:
> qq: Does the HTTP source support the Hugging Face or is this for generic http model source?

@rareddy the `hf://` support through this generic "URI Connection" type of source was already done with:
- https://github.com/kubeflow/model-registry/pull/1365

jobs/async-upload/job/download.py
jonburdo: Yeah, it's very odd use of env vars. I think I'll remove this unless there's any objection. If we need support we can add a new option with the `MODEL_SYNC_SOURCE_*` / `MODEL_SYNC_DESTINATION_*` vars. Instead of encoding the uri as part of the env var I think it should be part of the value.

For example instead of:
```
example.com-headers='{"content-type": "application/gzip"}'
other-example.com-headers='{"content-type": "application/x-tar"}'
```
we could have:
```
MODEL_SYNC_SOURCE_HEADERS='{"example.com": {"content-type": "application/gzip"}, "other-example.com": {"content-type": "application/x-tar"}}'
```
Idk if we'd really even need to specify the uri, or support more than one at a time...

But yeah, it's from kserve:
https://github.com/kserve/kserve/pull/1372

tarilabs: > I think I'll remove this unless there's any objection

we don't _have to_ support any possible variation of the KServe specific way to load from URI, for as long as overall we remain semantically equivalent we match the general expectations (load a single file, or load and expand an archive-file). I doubt anyone using the Job would complain we're missing this Env var, so I'm favourable of removing it. ...we can always enhance the code later if somebody ask for it (but I double 🙃 )

jobs/async-upload/pyproject.toml
```diff
@@ -35,3 +35,6 @@ markers = [
 
 [tool.poetry.scripts]
 debug-test = "debugpy:main --listen 5678 --wait-for-client -m pytest jobs/async-upload/tests/test_config.py -v"
+
+[tool.ruff]
+line-length = 119
```
jonburdo, line 40: Copied this value from model-registry python client's `pyproject.toml`

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -60,196 +57,258 @@ def model_registry_client():
...
     2. Downloads and uploads an ONNX model to MinIO
     3. Creates and applies a Kubernetes job using kustomize
     4. Waits for job completion
     5. Validates the final artifact state
     """
-    
+    env = env.copy()
+
+    if "MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE" in os.environ:
+        env["MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE"] = os.environ["MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE"]
```
jonburdo, line 298: Allows to do something like `MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE=quay.io/namespace/busybox:latest make test-integration`

Useful if you hit the limit on pulling `busybox:latest` from docker. Maybe we should have a default in quay

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -60,196 +57,258 @@ def model_registry_client():
...
+
+    # TODO use yield fixtures for setup/teardown
+    # https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
+    if setup is not None:
+        setup(tmp_path)
```
jonburdo, line 303: Not the prettiest but I figure we can improve later.

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -60,196 +57,258 @@ def model_registry_client():
...
+
+
+@pytest.mark.parametrize(
+    "setup,env",
+    [
+        (
+            None,
+            {
+                "MODEL_SYNC_SOURCE_TYPE": "uri",
+                "MODEL_SYNC_SOURCE_URI": "https://huggingface.co/sadhaklal/logistic-regression-iris/resolve/main/pytorch_model.bin",
```
jonburdo: I found a random tiny model file. If anyone has a better test url or wants to have an archive in e2e, happy to change this.

jobs/async-upload/pyproject.toml
tarilabs: Yeah, although I'm not that "integralist" it's possibly an old/initial setup, we can revise and adapt as needed; thank you for the consistency.

jobs/async-upload/tests/integration/test_integration_async_upload.py
tarilabs: I think on ci/GHA we don't have this limitation, but we can have one on ghcr.io as needed 👍

jobs/async-upload/tests/integration/test_integration_async_upload.py
tarilabs: you can use
https://huggingface.co/tarilabs/mnist/resolve/v1.nb20231206162408/mnist.onnx if preferred (I've created the onnx with our samples),
or otherwise why not re-use 
https://github.com/onnx/models/raw/refs/heads/main/validated/vision/classification/mnist/model/mnist-8.onnx
from the other tests? 🤔

tarilabs (commented):
thanks @jonburdo couple of minor comments
/lgtm

lmk if you want to merge now and follow-up in case

jobs/async-upload/tests/integration/test_integration_async_upload.py
jonburdo: Right, this is mainly for local dev. Still, using ghcr.io as the default might be a good idea. Leaving for now, but may change it in a followup

tarilabs:
- https://github.com/kubeflow/model-registry/pull/1375#pullrequestreview-3067541548

thanks again @jonburdo 

/lgtm
/approve

PR #1390, opened 2025-07-30
feat: added a generic readyz endpoint

Al-Pragliola:
## Description

In this PR I have added a new endpoint to check for the model registry status `/readyz/health` 
here we do a combined check about the database state (not dirty) and that the datastore correctly responds (internally) to requests. Additionally in order to have less restarts I introduced a startup probe to make sure that the db migrations are done and the state is not "dirty"

## How Has This Been Tested?

in a local kind cluster

## Merge criteria:

### DIFF
--- a/cmd/proxy.go
+++ b/cmd/proxy.go
@@ -74,8 +74,8 @@ func (c *ConditionalModelRegistryHealthChecker) Check() proxy.HealthCheck {
 	service := c.holder.Get()
 	if service == nil {
 		return proxy.HealthCheck{
-			Name:    "model-registry",
-			Status:  "fail",
+			Name:    proxy.HealthCheckModelRegistry,
+			Status:  proxy.StatusFail,
 			Message: "model registry service not yet initialized",
 			Details: map[string]interface{}{
 				"service_ready": false,
--- a/internal/proxy/readiness.go
+++ b/internal/proxy/readiness.go
@@ -11,6 +11,42 @@ import (
 	"github.com/kubeflow/model-registry/pkg/api"
 )
 
+const (
+	// Health check names
+	HealthCheckDatabase      = "database"
+	HealthCheckModelRegistry = "model-registry"
+	HealthCheckMeta          = "meta"
+
+	// Health check statuses
+	StatusPass = "pass"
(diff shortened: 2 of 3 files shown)


cmd/proxy.go
```diff
@@ -46,29 +47,78 @@ hostname and port where it listens.`,
 	}
 )
 
+// ModelRegistryServiceHolder safely holds the model registry service
+type ModelRegistryServiceHolder struct {
```
dhirajsb, line 51: Sorry I' missing the context here. Why is the service holder required?

internal/proxy/readiness.go
```diff
@@ -1,63 +1,265 @@
...
+
+func NewDatabaseHealthChecker(datastore datastore.Datastore) *DatabaseHealthChecker {
+	return &DatabaseHealthChecker{
+		datastore: datastore,
+	}
+}
+
+func (d *DatabaseHealthChecker) Check() HealthCheck {
+	check := HealthCheck{
+		Name:    "database",
```
dhirajsb: Could this be a list of consts? Same thing for other string literals like Status.

dhirajsb (commented):
Haven't looked too deep into the implementation, but lgtm in general. 
Can we also test recovery from a bad state like a DB reconnect back to healthy state?

Al-Pragliola:
as requested from @dhirajsb 

Tested with this scenario:

1) db up and running mr up and running
2) db goes offline (replicas set to 0)
3) mr liveness probe fails and restart the pod
4) db back online (replicas set to 1)
5) mr up and running

(3)

image


(4)-(5)

image

cmd/proxy.go
Al-Pragliola: service at the start is nil, in a goroutine that initialize the datastore is set to the correct service so to avoid nil dereference or race conditions I wrapped it in the Holder as this process is async

dhirajsb (approved):
/lgtm

Al-Pragliola:
- https://github.com/kubeflow/model-registry/pull/1390#pullrequestreview-3079873886


/approve

PR #1453, opened 2025-08-12
feat: Local dev setup for MR, catalog, and UI via containers

chambridge:
## Description
* Use docker/podman-compose to start the model registry (with postgresql database), model catalog, and ui frontend (uses mock-bff)
* Adds a mock-bff layer which decouples basic setup from Kubernetes deployment

## How Has This Been Tested?

Follow directions in [local deployment doc](https://github.com/chambridge/model-registry/blob/feat/local-dev-via-compose/clients/ui/docs/local-deployment-guide-mock.md)

## Merge criteria:

If you have UI changes


docker-compose-dev.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - AUTH_METHOD=internal
+      - PROXY_PROTOCOL=http
+      - PROXY_HOST=mock-bff
+      - PROXY_PORT=4000
+      - HOST=0.0.0.0
+      - PORT=3000
+      - PUBLIC_PATH=/
+      - BASE_PATH=/
+    depends_on:
+      - mock-bff
```
tarilabs, line 86: can you also please add these to:

https://github.com/kubeflow/model-registry/blob/main/docker-compose-mysql.yaml

wdyt?

docker-compose-dev.yaml
```diff
@@ -0,0 +1,89 @@
+# Development docker-compose file with PostgreSQL, local builds, and full stack
+# Composes:
+# - PostgreSQL database
+# - model-registry server built locally with direct PostgreSQL connection (embedmd)
+# - model-catalog service built locally using same image as model-registry
+# - UI frontend built locally using webpack dev server with proxy
+version: '3.8'
```
tarilabs, line 7: I would especially ad a docker-compose-postgresql like it exists for
https://github.com/kubeflow/model-registry/blob/main/docker-compose-mysql.yaml

docker-compose-dev.yaml
chambridge: sure. I can add to that setup.

docker-compose-dev.yaml
chambridge: would you prefer that I just rename this file `docker-compose-postgres.yaml` if I'm going to add this setup to the mysql version too?

tarilabs: we can keep this docker compose, it's building the referred containerfile (see the "`-local`" one)
I believe is helpful to have a compose file which just take `:latest` from ghcr.io where we publish periodically our container, so that folks who want to try it out without building/"compiling" this, can just Pull latest.

docker-compose-dev.yaml
chambridge: I think I've added what you were looking for. Let me know if there are alternative changes needed.

docker-compose-dev.yaml
chambridge: ok. I've added a postgres docker-compose that mirrors the mysql with the new components added, but using the registry image.

clients/ui/frontend/config/webpack.common.js
```diff
@@ -18,7 +18,7 @@ const FAVICON = process.env.FAVICON;
 const PRODUCT_NAME = process.env.PRODUCT_NAME;
 const COVERAGE = process.env.COVERAGE;
 const DEPLOYMENT_MODE = process.env._DEPLOYMENT_MODE;
-const BASE_PATH = DEPLOYMENT_MODE === 'kubeflow' ? '/model-registry/' : PUBLIC_PATH;
```
lucferbux, line 21: Be careful with this change, have you tried it against kubeflow? this is intended this way since there are some set of prefix that needs to follow.

clients/ui/mock-bff/mock-bff.js
```diff
@@ -0,0 +1,162 @@
+const express = require('express');
```
lucferbux, line 1: This goes against the core principle of modular architecture, BFF is mocked running it in mock mode, we intend to ALWAYS ship both things, never use the frontend as standalone

Dockerfile
```diff
@@ -9,6 +9,7 @@ COPY ["go.mod", "go.sum", "go.work", "./"]
 COPY ["pkg/openapi/go.mod", "pkg/openapi/"]
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
+USER root
 RUN go mod download
```
lucferbux, line 13: Any particular reason to add root here?

clients/ui/docs/local-deployment-guide-mock.md
```diff
@@ -0,0 +1,226 @@
+# Local Deployment Guide (Mock BFF)
+
+## Overview
+
+This guide walks you through setting up a complete local development environment for the Model Registry UI using Docker/Podman Compose. This setup includes:
+
+* PostgreSQL database for model registry storage
+* Model Registry server with direct database connection
+* Model Catalog service for browsing community models
+* Mock BFF service for UI backend functionality
```
lucferbux, line 10: Modular architecture has some opinionated workflows, apologies for the lack of docs, we are working on it, but this would go against them. We should always ship BFF, we should never create a dev setup that gets an standalone UI development.

We already have a defined local dev, any feedback is welcome and we might have not considered some use cases so I'm happy to discuss the reason behind this.

If the main goal is to have a local deployment, we already support running the UI in local mode (which should be the default) and then make sure the bff can hit different services. 

If the goal here would be to have the bff in mock mode for all k8s calls but wanting to hit both the MR service and the Catalog API service, I would suggest adapt the BFF to change that logic, because right now we are forcing users to go with local dev with BFF in mock mode.

@ederign and I are really happy to discuss this, cause it's not a bad idea having docker support (we did have it but remove it at some point due to some refactors) and we can even include it as an standard for modular architecture, but this kind of changes should be run by the dashboard paltform team.

Thanks a lot!

lucferbux (changes requested):
Just left my concerns in the first comment, I'm really happy to discuss this if you want alongside @ederign to find a better fit for this workflow.

Dockerfile
chambridge: `go mod download` will fail for me if this is not in place.

chambridge:
Thanks for the background. I'll look more at the BFF and try and digest the current documentation more./

lucferbux:
Thanks a lot, docs are high priority for use, we are working to make sure it's easy for people to onboard in the architecture.

PR #1487, opened 2025-08-18
fix filterQuery IN operator

jonburdo:
This adds handling of array values so we can do something like `id IN (6,7,8)` to filterQuery:

  curl 'http://localhost:8080/api/model_registry/v1alpha3/experiment_runs?filterQuery=id+IN+(6,7,8)'

AI-assisted: Cursor (claude-4-sonnet)

## Merge criteria:

If you have UI changes

### DIFF
--- a/internal/db/filter/parser.go
+++ b/internal/db/filter/parser.go
@@ -107,7 +107,7 @@ type Value struct {
 
 //nolint:govet
 type ValueList struct {
-	Values []*SingleValue `"(" @@ ("," @@)* ")"`
+	Values []*SingleValue `"(" (@@  ("," @@)*)? ")"`
 }
 
 //nolint:govet
--- a/internal/db/filter/parser.go
+++ b/internal/db/filter/parser.go
@@ -257,7 +257,7 @@ func convertPropertyRef(prop *PropertyRef, value *Value) *PropertyReference {
 
 func convertValue(val *Value) interface{} {
 	if val.String != nil {
-		return processStringValue(*val.String)
+		return unquoteStringValue(*val.String)
 	}
 
 	if val.Integer != nil {
@@ -274,7 +274,7 @@ func convertValue(val *Value) interface{} {
 
 	if val.ValueList != nil {
 		// Convert list of values to slice
-		var values []interface{}
+		var values []any
 		for _, singleVal := range val.ValueList.Values {
 			values = append(values, convertSingleValue(singleVal))
 		}
@@ -286,7 +286,7 @@ func convertValue(val *Value) interface{} {
 
 func convertSingleValue(val *SingleValue) interface{} {
 	if val.String != nil {
-		return processStringValue(*val.String)
+		return unquoteStringValue(*val.String)
 	}
 
 	if val.Integer != nil {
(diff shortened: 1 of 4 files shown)


internal/db/filter/parser.go
```diff
@@ -265,11 +272,72 @@ func convertValue(val *Value) interface{} {
 		return strings.ToLower(*val.Boolean) == "true"
 	}
 
+	if val.ValueList != nil {
+		// Convert list of values to slice
+		var values []interface{}
```
pboyd: We should use `any` for new code:

```suggestion
		var values []any
```

internal/db/filter/parser.go
```diff
@@ -265,11 +272,72 @@ func convertValue(val *Value) interface{} {
...
+
+	if val.Boolean != nil {
+		return strings.ToLower(*val.Boolean) == "true"
+	}
+
+	return nil
+}
+
+// processStringValue removes quotes and handles escape sequences
+func processStringValue(str string) string {
```
pboyd: Maybe call this `unquoteStringValue`?

internal/db/filter/parser.go
```diff
@@ -265,11 +272,72 @@ func convertValue(val *Value) interface{} {
...
+		return DoubleValueType
+	}
+	if val.Boolean != nil {
+		return BoolValueType
+	}
+	if val.ValueList != nil && len(val.ValueList.Values) > 0 {
+		// For lists, infer type from the first value
+		return inferSingleValueType(val.ValueList.Values[0])
+	}
+	return StringValueType // default to string
```
pboyd: Can this code be simplified to:

```suggestion
	if val.ValueList != nil && len(val.ValueList.Values) > 0 {
		// For lists, infer type from the first value
		return inferSingleValueType(val.ValueList.Values[0])
	}
	return inferSingleValueType(val)
```

internal/db/filter/query_builder.go
```diff
@@ -386,7 +386,21 @@ func (qb *QueryBuilder) buildOperatorCondition(column string, operator string, v
...
+			if len(valueSlice) == 0 {
+				// Empty list should return false condition
+				return conditionResult{condition: "1 = 0", args: []any{}}
+			}
+			// Create placeholders for each value
+			placeholders := make([]string, len(valueSlice))
+			for i := range placeholders {
+				placeholders[i] = "?"
+			}
+			condition := fmt.Sprintf("%s IN (%s)", column, strings.Join(placeholders, ","))
```
pboyd: Do what you want, of course, but here's another option:

```suggestion
			// Create placeholders for each value
			condition := fmt.Sprintf("%s IN (?%s)", column, strings.Repeat(",?", len(valueSlice)-1))
```

pboyd (approved):
I left a few comments, but it's nothing critical.

/lgtm

dhirajsb:
@jonburdo let's also add a service level test in `TestGetExperimentsWithFilterQuery()` in `internal/core/experiment_test.go`

dhirajsb (approved):
/lgtm

internal/db/filter/parser.go
jonburdo: Yeah, although we do have `Value` and `SingleValue` now so we have to do:
```go
singleVal := &SingleValue{
        String:  val.String,
        Integer: val.Integer,
        Float:   val.Float,
        Boolean: val.Boolean,
}
```
This is still more succinct though. I guess the other option would be to define `Value` like this:
```go
type Value struct {
    SingleValue      // Embed SingleValue
    ValueList *ValueList `| @@`
}
```
but I won't worry about it unless anyone has opinions on this

Issue #323, opened 2024-08-30
Introduce dependency license monitoring


tarilabs:
This project onboarded as Apache-2.0, [see original thread](https://github.com/kubeflow/community/issues/685#issuecomment-1889559126). 

We have received questions about the dependencies' license compatibility, for which we referenced: https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist

In this repo, we have:
- main REST Model Registry server: Go
- MR Python client: Python
- UI: React, Node
- BFF: Go
- CSI: Go

(we could have more in the future more)

To help with automating on the dependency checks,
👉 we propose to introduce FOSSA scanning in this repo, from the list of tools CNCF can support with: https://contribute.cncf.io/resources/project-services/hosted-tools/#tools

---

If you are a Kubeflow member reading this issue and you need to be added in the FOSSA tool:

- general access to CNCF's FOSSA which is a prerequisite is [to be requested via CNCF service desk](https://contribute.cncf.io/resources/project-services/hosted-tools/#tools:~:text=Most%20services%20can%20be%20requested%20through%20a%20Service%20Desk%20ticket): ask a KFSC representative to manage the service desk request for you/your KF project.
- then, you can reach out to myself, @rimolive and @terrytangyuan to be added in the existing Fossa team "Kubeflow" in the CNCF's FOSSA, or ask a KFSC representative to manage the service desk request for you/your KF project 

Suggested steps for new projects/repos:

1. ensure the Maintainer is added to the correct Team (in our case, "Kubeflow" Team)
![Screenshot 2024-09-10 at 09 54 36](https://github.com/user-attachments/assets/037d555e-af06-451f-89dd-6c412dc89c67)

2. the Maintainer is to create a "push only" token api key, [see example](https://github.com/kubeflow/model-registry/pull/324/files#diff-472dd216a27f2e558659b83ec1d8874189994adca4da6eefb3b00f63dab93fd2R16).

3. the Maintainer should first use the fossa cli with the `--team` flag for the desired repo, and the appropriate configurations keys from within a [.fossa.yml](https://github.com/fossas/fossa-cli/blob/master/docs/references/files/fossa-yml.md) file.
That would spawn a new Fossa project in the CNCF Org _and_ assign it to the Fossa Team "Kubeflow", at once.

> [!TIP]
[...]

PR #1528, opened 2025-08-26
Add support for setting verify ssl, user token and request headers in tests

dbasunag:
In order to support running the tests against live clusters, I need ability to set verify=False, headers to post/patch calls and user_token to modelRegistry client

## Merge criteria:

If you have UI changes


tarilabs (commented):
thanks @dbasunag , this is helpful to parametrize Test Suite to leverage parametric user-token and verify-ssl

/lgtm

clients/python/tests/conftest.py
```diff
@@ -78,16 +78,41 @@ def root(request) -> Path:
     return (request.config.rootpath / "../..").resolve()  # resolves to absolute path
 
 
-def poll_for_ready():
+@pytest.fixture(scope="session")
+def user_token() -> str:
+    return os.environ.get("AUTH_TOKEN") if os.environ.get("AUTH_TOKEN") else None
```
syntaxsdev:
```suggestion
    return os.getenv("AUTH_TOKEN", None)
```

clients/python/tests/conftest.py
```diff
@@ -78,16 +78,41 @@ def root(request) -> Path:
...
+    if user_token:
+        headers["Authorization"] = f"Bearer {user_token}"
+    return headers
+
+
+@pytest.fixture(scope="session")
+def verify_ssl() -> bool:
+    verify_ssl_env = os.environ.get("VERIFY_SSL")
+    if verify_ssl_env is None:
+        return None
```
syntaxsdev, line 98:
```suggestion
    return os.getenv("VERIFY_SSL", None)
```

clients/python/tests/conftest.py
```diff
@@ -78,16 +78,41 @@ def root(request) -> Path:
...
+    return verify_ssl_env.lower() == "true"
```
syntaxsdev, line 99: the change I suggested above will evaluate back to true when compared against.
```suggestion

```

clients/python/tests/conftest.py
```diff
@@ -78,16 +78,41 @@ def root(request) -> Path:
...
         elapsed_time = time.time() - start_time
         if elapsed_time >= MAX_POLL_TIME:
             print("Polling timed out.")
             break
-
-        print("Attempt to connect")
+        print(f"Attempt to connect to server {REGISTRY_URL}")
         try:
-            response = requests.get(REGISTRY_URL, timeout=MAX_POLL_TIME)
+            response = requests.get(**params, timeout=MAX_POLL_TIME, )
```
syntaxsdev: linting maybe?
```suggestion
            response = requests.get(**params, timeout=MAX_POLL_TIME)
```

clients/python/tests/REST_bindings_test.py
```diff
@@ -20,8 +20,13 @@
...
 @pytest.fixture
 @cleanup
-async def client() -> AsyncIterator[ModelRegistryServiceApi]:
-    config = mr_openapi.Configuration(REGISTRY_URL)
+async def client(user_token: str, verify_ssl: bool) -> AsyncIterator[ModelRegistryServiceApi]:
+    params = {}
+    if user_token:
+        params["access_token"] = user_token
+    if verify_ssl is not None:
+        params["verify_ssl"] = verify_ssl
```
syntaxsdev: not nitpicky on this one, but you can always set these params directly in the dict even if its None, because they are still going to default to `None` in the function param since they are optional?

```suggestion
params = {"verify_ssl": verify_ssl, "access_token": user_token}
```

syntaxsdev (changes requested):
Overall looks good, thank you @dbasunag.

Some general changes to make more readable

clients/python/tests/conftest.py
dbasunag: I need to send boolean True/False or None. returing `os.getenv("VERIFY_SSL", None)` would send back the string "true", "false" and None.

clients/python/tests/conftest.py
dbasunag: I believe this is needed. Without that it would return the string set to environment variable.

clients/python/tests/conftest.py
syntaxsdev: Ah I see. Ignore then :)

clients/python/tests/REST_bindings_test.py
```diff
@@ -20,8 +20,9 @@
 
 @pytest.fixture
 @cleanup
-async def client() -> AsyncIterator[ModelRegistryServiceApi]:
-    config = mr_openapi.Configuration(REGISTRY_URL)
+async def client(user_token: str, verify_ssl: bool) -> AsyncIterator[ModelRegistryServiceApi]:
+params = {"verify_ssl": verify_ssl, "access_token": user_token}
```
syntaxsdev:
```suggestion
    params = {"verify_ssl": verify_ssl, "access_token": user_token}
```

syntaxsdev:
/lgtm

thanks for adopting recommendations!

tarilabs (approved):
thanks @dbasunag and all for 👀 

- https://github.com/kubeflow/model-registry/pull/1528#pullrequestreview-3160044803
- https://github.com/kubeflow/model-registry/pull/1528#pullrequestreview-3160044803

/approve

tarilabs:
we resolved the false-positive on FOSSA

<details>

Logged with: https://github.com/kubeflow/model-registry/issues/323#issuecomment-2343738915 In the fossa project CNCF/Kubeflow/Model Registry, the "github.com/docker/docker" which for us is a transitive dependency appears to be showing multiple licenses, namely Apache-2.0 and GPL-2.0-or-later. This can be seen by scrolling in the same page to the beginning of
vendor/github.com/spdx/tools-golang/LICENSE.code to see: The tools-golang source code is provided and may be used, at your option,
Package
under either: * Apache License, version 2.0 (Apache-2.0), OR * GNU General Public License, version 2.0 or later (GPL-2.0-or-later). This is actually the license for "github.com/spdx/tools-golang", which is vendored by "github.com/docker/docker" (for example, here). The
"github.com/docker/docker" is distributed under Apache-2.0 as can also be seen in this page:
https://pkg.go.dev/github.com/docker/docker@v27.2.1+incompatible#:
~:text=Sep%206%2C%202024-,License%3A%20Apache%2D2.0,-
Details Currently, FOSSA does not support the automatic creation of dual licenses on the backend. Without FOSSA Org-wise permission which is our case since we're a subset, the "Kubeflow" FOSSA Team inside CNCF's FOSSA Org, the advised acton is to: ignoring the issue with a note that the dependency is dual-licensed may be the easiest solution Recorded as part of TKT-11490.

</details>

PR #1535, opened 2025-08-28
feat: add metric history bulk fetch endpoint

jonburdo:
This adds a new endpoint to fetch metric history items for multiple experiment runs at once:

  GET /experiment_runs/metric_history

This supports filtering with params like filterQuery.

AI-assisted: Claude Code 1.0.63 (claude-sonnet-4@20250514)

## Merge criteria:

If you have UI changes


api/openapi/src/model-registry.yaml
```diff
@@ -1520,6 +1520,35 @@ paths:
           type: string
         in: path
         required: true
+  "/api/model_registry/v1alpha3/experiment_runs/metric_history":
```
tarilabs, line 1523: adding in /src/ ...

api/openapi/model-registry.yaml
```diff
@@ -253,6 +253,35 @@ paths:
       operationId: createExperimentRun
       summary: Create an ExperimentRun
       description: Creates a new instance of an `ExperimentRun`.
+  "/api/model_registry/v1alpha3/experiment_runs/metric_history":
```
tarilabs, line 256: ... and reflected in api/openapi/model-registry.yaml as expected.

api/openapi/model-registry.yaml
```diff
@@ -253,6 +253,35 @@ paths:
...
+    summary: Path used to get metric history for multiple experiment runs.
+    description: >-
+      The REST endpoint/path used to get metric history for multiple `ExperimentRun` entities. This path contains a `GET` operation to retrieve metrics with optional filtering.
+    get:
+      tags:
+        - ModelRegistryService
+      parameters:
+        - $ref: "#/components/parameters/filterQuery"
+        - $ref: "#/components/parameters/name"
+        - $ref: "#/components/parameters/stepIds"
```
tarilabs, line 266: added parameters per descriptions.

tarilabs (commented):
in general LGTM and letting other people time for comment; while the scope of the PR is clear, I'm only curious whether we should emit a warning on attempting to "dump" the entire metric metadata available when called as:

```
service.GetExperimentRunMetricHistory(nil, nil, api.ListOptions{}, nil)
```

I understand pagination support already avoids potential issues, 
and that filter queries can be applied,
yet I'm curious of the need for the "plain dump" use-case?

If you could expand on the need of it please.

jonburdo:
> yet I'm curious of the need for the "plain dump" use-case?
> 
> If you could expand on the need of it please.

I'm not sure of a specific request for a full dump. One reason, would be for consistency with our other endpoints.

The main scenario I can think of would be if the user (especially in a relatively new model registry instance) simply wanted to export all their runs and related data. It would be useful to simply allow this on each endpoint:
```
GET /experiment_runs
GET /experiment_runs/metric_history
```
As a user, I think I would also find it surprising if I could do:
```
GET /experiment_runs/metric_history?filterQuery=experimentRunId+IN+(1,2,3,4,5,...)
```
but not
```
GET /experiment_runs/metric_history
```

Because it's paginated (and the user can choose the page size), it's also useful if we were to want the ability to begin a iterating lazily through records.

That being said, the most important use case is with `filterQuery`, particularly `filterQuery=experimentRunId+IN+(...)`.

api/openapi/src/model-registry.yaml
```diff
@@ -1520,6 +1520,35 @@ paths:
...
+          $ref: "#/components/responses/Unauthorized"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+        "503":
+          $ref: "#/components/responses/ServiceUnavailable"
+      operationId: getExperimentRunsMetricHistory
+      summary: Get metric history for multiple ExperimentRuns
+      description: Gets the metric history for multiple `ExperimentRun` entities with optional filtering by metric name, step IDs, and experiment run IDs.
```
dhirajsb, line 1551: Is the experiment run ID reference in description about being able to use `filterQuery` to achieve it or do we want to also support a standalone query param like stepIds?

internal/core/experiment_run.go
```diff
@@ -239,22 +239,24 @@ func (b *ModelRegistryService) GetExperimentRunArtifacts(artifactType openapi.Ar
 
 func (b *ModelRegistryService) GetExperimentRunMetricHistory(name *string, stepIds *string, listOptions api.ListOptions, experimentRunId *string) (*openapi.MetricList, error) {
 
-	// Validate experiment run exists
-	if experimentRunId == nil {
-		return nil, fmt.Errorf("experiment run ID is required: %w", api.ErrBadRequest)
-	}
+	var experimentRunIdInt32Ptr *int32
 
-	// Validate experiment run exists
```
dhirajsb, line 247: Since this validation is removed, would we not get an error message if user gets metric history for a non-existent experiment run id? If so, it would cause failure in fuzzer tests.

dhirajsb (commented):
I added a couple of questions in comments.

api/openapi/src/model-registry.yaml
jonburdo: Just using `filterQuery`. Just my opinion, but I don't think we want too many ways to do the same thing. It could be confusing if people were coming across both of these examples:
```
/experiment_runs/metric_history?filterQuery=experimentRunIds+IN+(1,2,3)
/experiment_runs/metric_history?runIds=1,2,3
```
I think we can and should provide useful examples for common scenarios though in our readme/docs

internal/core/experiment_run.go
jonburdo: We still have this check when an individual experiment run id is passed, so nothing should change for:
```
/experiment_runs/32/metric_history
```
but just like with a sql where-clause, we don't make sure experiment ids exist for filterQuery. For example, if experiment run 3 does not exist then these are equivalent:
```
/experiment_runs/metric_history?filterQuery=experimentRunIds+IN+(1,2)
/experiment_runs/metric_history?filterQuery=experimentRunIds+IN+(1,2,3)
```
I'm not sure we need validation in this case. The typical scenario in my mind would be:

1. get experiment runs
  ```
  /experiment_runs?filterQuery=...
  ```
2. get metric history for returned experiment run ids:
  ```
  /experiment_runs/metric_history?filterQuery=experimentRunIds+IN+(...)
  ```

I think any additional validation can be done with code, unless I'm missing some additional need.

dhirajsb: I was just wondering about the `/experiment_runs/32/metric_history` scenario. Whether there would be any url representation that would drop the id or set it to null somehow. But, I don't think there is a way, so this should be ok.

PR #1548, opened 2025-09-03
Added Register Model Capability from Model Catalog

YuliaKrimerman:
## Description

Added the capability to Register a model From Model Catalog, and added Registered From field in Model Registry Version Details Page 
Screenshot 2025-09-03 at 4 13 19 PM
Screenshot 2025-09-03 at 4 13 10 PM

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/concepts/modelRegistry/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    registeredFrom.string_value !== 'Model catalog'
+  ) {
+    return null;
+  }
+
+  // Extract catalog model information
+  const sourceModel = customProperties['Source model'];
+  const sourceModelVersion = customProperties['Source model version'];
+  const sourceModelId = customProperties['Source model id'];
+  const provider = customProperties.Provider;
```
mturley: We need to use the `modelSource*` properties on the artifact to reference the catalog, not customProperties. Just like we have today in the original ODH version. `modelSourceKind`, `modelSourceClass`, `modelSourceGroup`, `modelSourceId` and `modelSourceName` are top level general purpose properties intended for this use.

Also (cc @ppadti ) once the changes for integrating UI with BFF land, we'll need to make sure this linking still works - I think for example the model version is not something we'll uniquely track anymore. I think it's ok for that to not block this PR as long as the reference matches what's necessary to link back to the catalog with its mock data, but we'll need to follow up on any changes necessary here.

mturley: @YuliaKrimerman hmm -- I also just noticed there is already a `clients/ui/frontend/src/concepts/modelRegistry/utils.ts` (without the `/app/`) that already has its own `modelSourcePropertiesToCatalogParams` util added by @rsun19 that already uses the `modelSource*` properties correctly. We should not create this new file, and we should use that existing util (maybe change it if necessary? It might be fine as-is.)

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+      if (
+        tags.some((tag) => tag.toLowerCase().includes(pattern)) ||
+        description.toLowerCase().includes(pattern)
+      ) {
+        return pattern.charAt(0).toUpperCase() + pattern.slice(1);
+      }
+    }
+
+    return 'PyTorch'; // Default fallback
+  };
```
mturley: I don't think this is a good idea. Did this come from a requirement I missed? We shouldn't hard code expected metadata and have special behavior for it, and I'm not sure if these framework tags are on the real data. If there's no definitive property in the catalog data for the framework and framework version, we should not try to prefill those values. Falling back to PyTorch feels especially wrong, we don't want to prefill wrong info if we're not sure.

I apologize if I missed a conversation about this from the MR team or something. Maybe we take this out and don't prefill `sourceModelFormat` or `sourceModelFormatVersion` and open a bug for adding it back.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+      for (const tag of tags) {
+        const match = tag.match(versionPattern);
+        if (match) {
+          return match[1];
+        }
+      }
+    }
+    return '1.0.0'; // Default fallback
+  };
```
mturley: Same here, we probably don't want to assume the version will look like this or fall back to a value.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+      setSubmitError(new Error('Source model format is required'));
+      setIsSubmitting(false);
+      return;
+    }
+
+    if (!formData.sourceModelFormatVersion || formData.sourceModelFormatVersion.trim() === '') {
+      setSubmitError(new Error('Source model format version is required'));
+      setIsSubmitting(false);
+      return;
+    }
```
mturley: Source model format and version should not be required, they aren't required in the original form.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
```diff
@@ -37,7 +38,7 @@ const RegisterModelDetailsFormSection = <D extends RegisterModelFormData>({
   );
 
   const modelDescriptionInput = (
-    <TextArea
+    <TextInput
```
mturley: Why change this from a TextArea to a TextInput? This will also affect the regular registration form, and I'm not sure where the change is coming from.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
```diff
@@ -69,7 +70,11 @@ const RegisterModelDetailsFormSection = <D extends RegisterModelFormData>({
         <FormFieldset component={modelDescriptionInput} field="Model Description" />
         <FormHelperText>
           <HelperText>
-            <HelperTextItem>Enter a brief summary of the model&apos;s key details.</HelperTextItem>
+            <HelperTextItem>
+              {isCatalogModel
+                ? 'Once your model is registered, you can create a model card to include more details'
```
mturley: Where is this microcopy coming from? I'm not sure it's accurate, you can't create a model card after you register, what does that mean?

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
```diff
@@ -17,12 +16,14 @@ type RegisterModelDetailsFormSectionProp<D extends RegisterModelFormData> = {
...
   hasModelNameError: boolean;
   isModelNameDuplicate?: boolean;
+  isCatalogModel?: boolean;
 };
 const RegisterModelDetailsFormSection = <D extends RegisterModelFormData>({
   formData,
   setData,
   hasModelNameError,
   isModelNameDuplicate,
+  isCatalogModel = false,
```
mturley: There are a few other things this `isCatalogModel` prop should do here - compare with the original model catalog in ODH. If you're registering a catalog model, the model location section should disable the radio buttons to force you to choose "URI", and it should change the URI input field to be static text. The user can't change the model location that's coming from the catalog.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+      setSubmitError(new Error('Model registry API is not available'));
+      setIsSubmitting(false);
+      return;
+    }
+
+    try {
+      const {
+        data: { registeredModel, modelVersion, modelArtifact },
+        errors,
+      } = await registerModel(apiState, formData, 'user'); // TODO: Get actual user
```
mturley: You can get this value from the AppContext just like the regular RegisterModel form does.
```ts
  const { user } = React.useContext(AppContext);
  const author = user.userId || '';
```

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+      if (registeredModel && modelVersion && modelArtifact) {
+        const navigationPath = `/model-registry/${encodeURIComponent(preferredModelRegistry.name)}/registeredModels/${registeredModel.id}`;
```
mturley: You can use the `registeredModelUrl` function from `~/app/pages/modelRegistry/screens/routeUtils` instead of hard coding the URL here.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+  // Get the first available model registry from the context
+  const preferredModelRegistry = modelRegistries.length > 0 ? modelRegistries[0] : null;
+
+  // Check to see if data is loaded
+  const isDataReady = !loading && !error && model !== undefined;
+
+  return (
+    <ApplicationsPage
+      title={`Register ${model?.name || ''} model`}
+      description="Create a new model and register the first version of your new model."
```
mturley: This doesn't match the description of the regular registration page "Create and register the first version of a new model." We should change it to be consistent.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+      breadcrumb={
+        <Breadcrumb>
+          <BreadcrumbItem render={() => <Link to="/model-catalog">Model catalog</Link>} />
+          <BreadcrumbItem
+            data-testid="breadcrumb-model-name"
+            render={() =>
+              !model?.name ? (
+                'Loading...'
+              ) : (
+                <Link to={`/model-catalog/${modelId}`}>{model.name}</Link>
```
mturley: ~~These hard coded URLs -- we forgot to create a model catalog `routeUtils` file. We should open a followup issue for that.~~

Edit: looks like you did create a util for this URL, let's use it here.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { modelRegistries } = React.useContext(ModelRegistrySelectorContext);
+
+  const model: ModelCatalogItem | undefined = React.useMemo(() => {
+    for (const source of sources) {
+      const found = source.models?.find((m) => m.id === modelId);
+      if (found) {
+        return found;
+      }
+    }
+    return undefined;
```
mturley, line 31: It shouldn't block this PR, but we'll need to match the model in the source data more specifically by passing more parameters to the route. cc @ppadti

mturley (changes requested):
Hey @YuliaKrimerman - there are a few things to address here, sorry for the delay with my full review

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
YuliaKrimerman: Yep, it's coming from here : https://redhat-internal.slack.com/archives/C07C0447EAV/p1757319129592229 this is an additional requirement we just added as part of this tory

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
YuliaKrimerman: because when it's textArea for some reason it looks like this 
Screenshot 2025-09-10 at 12 09 55 PM

mturley: I think that's a bug with the Material theme and we should leave it as a TextArea. It looks fine in the PF theme, we should raise that separately

rsun19: Should we tag Jenny on this? Or start a thread?

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+      },
+      'Source model version': {
+        // eslint-disable-next-line camelcase
+        string_value: versionTag || '',
+        metadataType: ModelRegistryMetadataType.STRING,
+      },
+      'Source model id': {
+        // eslint-disable-next-line camelcase
+        string_value: model.id || '',
+        metadataType: ModelRegistryMetadataType.STRING,
```
mturley: I don't think we need these since they're redundant with the modelSource* artifact properties we'll use for the "Registered from" link. They were a holdover from before we had that.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+    setData('modelLocationEndpoint', initialFormData.modelLocationEndpoint);
+    setData('modelLocationBucket', initialFormData.modelLocationBucket);
+    setData('modelLocationRegion', initialFormData.modelLocationRegion);
+    setData('modelLocationPath', initialFormData.modelLocationPath);
+    setData('modelLocationURI', initialFormData.modelLocationURI);
+    setData('modelRegistry', initialFormData.modelRegistry);
+    setData('modelCustomProperties', initialFormData.modelCustomProperties);
+    setData('versionCustomProperties', initialFormData.versionCustomProperties);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [model.id, preferredModelRegistry.name, versionTag]);
```
mturley: We shouldn't need this effect if we just pass `initialFormData` into `useRegisterCatalogModelData` as a default, like we did in [the ODH version](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx#L98). This is why we don't render RegisterCatalogModelForm until the data is ready, so that we can use it as the state default instead of updating it with an effect.

I think the original port of `useRegisterCatalogModelData` was outdated and we must have added that default value later, see t[he ODH version of that hook](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts#L82-L87).

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
mturley: @jenny-s51 are you already aware of this issue in the Material theme? (TextArea labels occluding the content)? Should we open an issue / discuss elsewhere?

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const versionTag = extractVersionTag(model.tags);
+
+  const catalogParams = {
+    modelName: model.name,
+    tag: versionTag || '',
+    sourceName: model.provider || '',
+    repositoryName: model.id || '',
+  };
+  const sourceProperties = catalogParamsToModelSourceProperties(catalogParams);
+  const sourceCustomProperties = modelSourcePropertiesToCustomProperties(sourceProperties);
```
mturley: We don't need `sourceCustomProperties` -- we are only using the `sourceProperties`.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,243 @@
...
+        string_value: model.framework || '',
+        metadataType: ModelRegistryMetadataType.STRING,
+      },
+      Task: {
+        // eslint-disable-next-line camelcase
+        string_value: model.task || '',
+        metadataType: ModelRegistryMetadataType.STRING,
+      },
+    },
+    additionalArtifactProperties: sourceCustomProperties,
```
mturley: The `additionalArtifactProperties` are the `sourceProperties`, not custom properties.

clients/ui/frontend/src/concepts/modelRegistry/utils.ts
```diff
@@ -26,3 +27,61 @@ export const modelSourcePropertiesToCatalogParams = (
...
+      customProperties['Source model id'] = {
+        // eslint-disable-next-line camelcase
+        string_value: properties.modelSourceGroup,
+        metadataType: ModelRegistryMetadataType.STRING,
+      };
+    }
+  }
+
+  return customProperties;
+};
```
mturley: This isn't necessary. We do not want to use custom properties AT ALL for this feature. We can get rid of the `modelSourcePropertiesToCustomProperties` function.

mturley (changes requested):

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,243 @@
...
+    setData('modelLocationBucket', initialFormData.modelLocationBucket);
+    setData('modelLocationRegion', initialFormData.modelLocationRegion);
+    setData('modelLocationPath', initialFormData.modelLocationPath);
+    setData('modelLocationURI', initialFormData.modelLocationURI);
+    setData('modelRegistry', initialFormData.modelRegistry);
+    setData('modelCustomProperties', initialFormData.modelCustomProperties);
+    setData('versionCustomProperties', initialFormData.versionCustomProperties);
+    setData('additionalArtifactProperties', initialFormData.additionalArtifactProperties);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [model.id, preferredModelRegistry.name, versionTag]);
```
mturley: Repeating a comment from an earlier review because it disappeared from the diff --

We shouldn't need this effect if we just pass `initialFormData` into `useRegisterCatalogModelData` as a default, like we did in [the ODH version](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx#L98). This is why we don't render RegisterCatalogModelForm until the data is ready, so that we can use it as the state default instead of updating it with an effect.

I think the original port of `useRegisterCatalogModelData` was outdated and we must have added that default value later, see t[he ODH version of that hook](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts#L82-L87).

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts
```diff
@@ -20,6 +20,7 @@ export type RegistrationCommonFormData = {
   modelLocationURI: string;
   versionCustomProperties?: ModelRegistryCustomProperties;
   modelCustomProperties?: ModelRegistryCustomProperties;
+  additionalArtifactProperties?: ModelRegistryCustomProperties;
```
mturley: This type should be `Partial<ModelArtifact>` like it is [in the ODH version here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelRegistry/screens/RegisterModel/useRegisterModelData.ts#L22). We are allowing passing additional top level properties on the artifact, not customProperties. The word "properties" here is confusing us 😓 

This usage of top level artifact properties instead of custom properties is the same mechanism you added to ODH in this PR: https://github.com/opendatahub-io/odh-dashboard/pull/4072

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -103,7 +103,9 @@ export const registerVersion = async (
     modelArtifact = await apiState.api.createModelArtifactForModelVersion({}, modelVersion.id, {
       name: `${formData.versionName}`,
       description: formData.versionDescription,
-      customProperties: {},
+      customProperties: {
+        ...formData.additionalArtifactProperties,
+      },
```
mturley: We need to put these directly on the artifact, not wrapped in `customProperties`.

clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+    sourceModelFormat: '',
+    sourceModelFormatVersion: '',
+    modelLocationType: ModelLocationType.URI,
+    modelLocationEndpoint: '',
+    modelLocationBucket: '',
+    modelLocationRegion: '',
+    modelLocationPath: '',
+    modelLocationURI: model.url || '',
+    modelRegistry: preferredModelRegistry.name,
+    modelCustomProperties: {},
```
mturley: Ok, I missed this before... comparing with the ODH version of RegisterCatalogModelForm (we really need to closely look at it and make sure we're matching the behavior), we do have `modelCustomProperties` that get passed in, [see here](https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx#L55) where we have `const labels = createCustomPropertiesFromModel(model);` which is a util we also need to move over.

Note that util has a call to `removeIlabLabels` which is not something we're moving to kubeflow, so it should just keep all the model labels and tasks. Although it looks like now it's a single `task` and we're putting it in as a Task property instead of a label for the version... let's do that for the model too (so the `modelCustomProperties should have labels from the catalog labels and a Task property for the catalog task.

clients/ui/frontend/src/app/routes/modelCatalog/catalogModelDetails.ts
```diff
@@ -0,0 +1,11 @@
+export interface CatalogModelDetailsParams {
+  modelName: string;
+  tag: string;
+  sourceName?: string;
+  repositoryName?: string;
+}
```
mturley: You don't need to add this here, it's already in `clients/ui/frontend/src/app/pages/modelRegistry/screens/types.ts` and you can `import { CatalogModelDetailsParams } from '~/app/pages/modelRegistry/screens/types';`

PR #1549, opened 2025-09-04
fix(python): replace assertions with proper StopIteration handling

tarilabs:
## Description
when calling something ~like:

```py
    for registered_model in client.get_registered_models():
        print(registered_model)
```

from [the docs](https://model-registry.readthedocs.io/en/latest/#listing-models), on an empty Model Registry (for example, just created MR instance) it raised AssertionError here:

https://github.com/kubeflow/model-registry/blob/7251769867eef1635efecfc0e01e80dfc7ead71f/clients/python/src/model_registry/types/pager.py#L133

which was an unforeseen consequence of asserting on an empty list.

- Replace `assert self._current_page` with explicit None checks in both sync and async _next_item methods
- Fix async version to raise StopAsyncIteration instead of StopIteration for consistency as part of THIS PR
- Reproducer tests for zero and single model pagination scenarios in commit ee656f95c561b31856eb811487e2d93bdc6e439f

This prevents AssertionError when _next_page() returns empty lists and ensures
proper iterator protocol compliance with StopIteration/StopAsyncIteration.

## How Has This Been Tested?
see ee656f95c561b31856eb811487e2d93bdc6e439f

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/types/pager.py
+++ b/clients/python/src/model_registry/types/pager.py
@@ -130,7 +130,8 @@ class Pager(Generic[T], Iterator[T], AsyncIterator[T]):
         if self._needs_fetch():
             self._current_page = self._next_page()
             self._i = 0
-        assert self._current_page
+        if self._current_page is None: # for example when the MR server is empty
+            raise StopIteration
         if self._i >= len(self._current_page):
             raise StopIteration
 
@@ -149,9 +150,10 @@ class Pager(Generic[T], Iterator[T], AsyncIterator[T]):
         if self._needs_fetch():
             self._current_page = await self._anext_page()
             self._i = 0
-        assert self._current_page
+        if self._current_page is None: # for example when the MR server is empty
+            raise StopAsyncIteration
         if self._i >= len(self._current_page):
-            raise StopIteration
+            raise StopAsyncIteration
 
         item = self._current_page[self._i]
         self._i += 1


tarilabs:
Reproducer demonstrates failure, with:

```
=================================== FAILURES ===================================
________________________ test_page_through_zero_models _________________________

client = <model_registry._client.ModelRegistry object at 0x7fabb90bf7d0>

    @pytest.mark.e2e
    def test_page_through_zero_models(client: ModelRegistry):
        """Test that we can page through zero models (i.e. a Model Registry just created)"""
        # leave with no models in the MR server
>       for _ in client.get_registered_models():

tests/test_client.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.nox/e2e-3-12/lib/python3.12/site-packages/model_registry/types/pager.py:163: in __next__
    item = self._next_item()
           ^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Pager(page_fn=<function ModelRegistry.get_registered_models.<locals>.rm_list at 0x7fabb8d7fec0>, options=ListOptions(limit=None, order_by=None, is_asc=True, next_page_token='', artifact_type_param=None))

    def _next_item(self) -> T:
        """Get the next item in the pager.
    
        This variant won't check for looping, so it's useful for manual iteration/scripting.
    
        NOTE: This won't check for looping, so use with caution.
        If you want to check for looping, use the pythonic `next()`.
        """
        if self._needs_fetch():
            self._current_page = self._next_page()
            self._i = 0
>       assert self._current_page
               ^^^^^^^^^^^^^^^^^^
E       AssertionError

.nox/e2e-3-12/lib/python3.12/site-packages/model_registry/types/pager.py:133: AssertionError
```

ref: https://github.com/kubeflow/model-registry/actions/runs/17459187537/job/49579683178?pr=1549

from reproducer commit https://github.com/kubeflow/model-registry/pull/1549/commits/ee656f95c561b31856eb811487e2d93bdc6e439f

fege (approved):
/lgtm

clients/python/tests/test_client.py
```diff
@@ -71,6 +71,22 @@ async def test_register_new_using_s3_uri_builder(client: ModelRegistry):
     assert ma.uri == uri
 
 
+@pytest.mark.e2e
```
jonburdo, line 74: I suggest we leave these tests as is for now. I just want to remark for future work - It could be nice to test client logic like this with a mock server or something similar, so it doesn't have to be e2e. Seems like we have a pretty loose definition of "end-to-end" in general. There are caveats to my suggestion too, and this isn't really a huge problem right now, but I wanted to mention it.

jonburdo (commented):
lgtm!

clients/python/tests/test_client.py
syntaxsdev: We can always convert tests we tag to later, but I don't think we should waste cycles for mock servers. The E2E tests already pretty fast in CI.

tarilabs: agree, but I would like to keep mocking to the bare, minimal of the minimal.

for example another PRs I've mocked the connection as I wanted to test just a very specific behaviour:
- https://github.com/kubeflow/model-registry/pull/1550
- https://github.com/kubeflow/model-registry/pull/1553

so I've mocked the get_registered_models() since that is used by the constructor...

I would avoid mocking for testing more complex logic and flow simply because you end up testing the mock, more than the actual code (server) instead. (the latest time I've re-encountered this principle of keeping mocks to a bare minimal was from Bill Kennedy statements, and I cannot agree more)

tarilabs:
- https://github.com/kubeflow/model-registry/pull/1549#pullrequestreview-3184326560
- https://github.com/kubeflow/model-registry/pull/1549#pullrequestreview-3189415317
- https://github.com/kubeflow/model-registry/pull/1549#issuecomment-3258463168

thanks again all for 👀 

/approve

PR #1553, opened 2025-09-05
feat(client): add K8s SA token fallback for user authn

tarilabs:
## Description

- When user do not supply explicitly user_token=, add fallback to read from K8s SA token file when envvar `KF_PIPELINES_SA_TOKEN_PATH` is not set
- Change token from envvar log level from warning to info for consistency
- Add test coverage for token source priority and fallback logic
- Ensure envvar `KF_PIPELINES_SA_TOKEN_PATH` token takes precedence over K8s file token when both are available

### Details

In a simple Notebook, typically we rely on envvar `KF_PIPELINES_SA_TOKEN_PATH` to point to the path containing a SA token, since in most environment we faced this envvar is set into the Notebook with the KFP SA token (and we refer to that, not necessarily the SA token of the Jupyter Notebook itself).

Instead, in a simple example of K8s workload (eg KFP, or simply a Pod), I typically resort to something like the following:

```py
@dsl.component(base_image="registry.redhat.io/ubi9/python-311", packages_to_install=["model-registry"])
def my_component(run_id: str = None, run_name: str = None) -> str:
    import os
    try:
        with open("/var/run/secrets/kubernetes.io/serviceaccount/token", "r") as f:
            user_token = f.read().strip()
        print("Token loaded successfully")
    except FileNotFoundError:
        raise ValueError("Service account token file not found")
    except Exception as e:
        raise ValueError("Couldn't read the token", e)
    
    registry = ModelRegistry(
        server_address="https://my-registry-rest.apps.rosa.mmortari-rosa2.otuf.p3.openshiftapps.com",
        author="pipeline-author",
        user_token=user_token
    )
    print("Connected to ModelRegistry")
```

that is, just source the standard K8s SA token and feed it into the client initialization, per documented API.

However, I've received the following comment:

> I think one of your examples had some magic around service account tokens

This PR fallback into attempting with a lower priority to source the K8s SA token when no user_token is provided and when the Notebook envvar is not set (ie inside a KFP or inside a simple Pod).

## How Has This Been Tested?

Pytest suite are added.

In addition to pytests, I've manually tested inside a barebone Pod, by loading the locally build .whl and:

Screenshot 2025-09-05 at 14 39 23

to confirm the envvar `KF_PIPELINES_SA_TOKEN_PATH` is not present on this barebone pod:

Screenshot 2025-09-05 at 14 41 24

overriding the default log_level show the intended messages:

Screenshot 2025-09-05 at 14 44 35

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/tests/test_client.py
+++ b/clients/python/tests/test_client.py
@@ -1,5 +1,7 @@
 import os
+import tempfile
 from itertools import islice
+from unittest.mock import patch
 
 import pytest
 import requests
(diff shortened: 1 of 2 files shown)


clients/python/tests/test_client.py
```diff
@@ -1206,3 +1208,129 @@ async def test_as_mlops_engineer_i_would_like_to_store_a_longer_documentation_fo
...
+            os.unlink(token_file_path)
+
+    @patch.object(ModelRegistry, "get_registered_models")
+    def test_user_token_from_k8s_file(self, mock_get_registered_models):
+        """Test for user not providing explicitly user_token,
+        reading user token from Kubernetes service account token file."""
+        test_token = "test-token-from-k8s-file" # noqa: S105
+        k8s_token_path = "/var/run/secrets/kubernetes.io/serviceaccount/token" # noqa: S105
+
+        with patch("model_registry._client.Path") as mock_path_class, \
```
jonburdo: This seems a little fragile. Maybe it's okay given it's **only** for a call to the `ModelRegistry` constructor (and you've mocked only for the `model_registry._client` module), but it could cause a small headache later if we add anything else that uses `Path` (or even some builtin utility that uses `Path` internally) in the constructor. I also hesitate to set the precedent.

Instead of mocking `Path` could we maybe set a top-level variable for the path value? (i.e. at the same place we're setting `DEFAULT_USER_TOKEN_ENVVAR` we could set `USER_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"`). Then you could simply patch this variable to be the path of a temp file that you write to.

clients/python/tests/test_client.py
```diff
@@ -1206,3 +1208,129 @@ async def test_as_mlops_engineer_i_would_like_to_store_a_longer_documentation_fo
...
+        """Test for user not providing explicitly user_token,
+        reading user token from environment variable."""
+        test_token = "test-token-from-envvar" # noqa: S105
+
+        with tempfile.NamedTemporaryFile(mode="w", delete=False) as token_file:
+            token_file.write(test_token)
+            token_file_path = token_file.name
+
+        try:
+            with patch.dict(os.environ, {"KF_PIPELINES_SA_TOKEN_PATH": token_file_path}):
```
jonburdo: nit - I'm thinking we should mostly stick with our test framework's machinery, and use pytest's `monkeypatch.setenv`. I would even use pytest-mock's `mocker` fixture instead of directly using unittest's mocks when possible. But I realize we already have a bit of variation in how we do things. Also fine to keep as is

jonburdo (commented):
A couple minor suggestions. Otherwise looks good!

clients/python/tests/test_client.py
tarilabs: thanks, let me try that!

clients/python/tests/test_client.py
tarilabs: > I'm thinking we should mostly stick with our test framework's machinery, and use pytest's monkeypatch.setenv

I've done this with https://github.com/kubeflow/model-registry/pull/1553/commits/81f87352e0d90d4fc868cbd2dab8bdab3a63573a

> I would even use pytest-mock's mocker fixture instead of directly using unittest's mocks when possible

it doesn't seem to me we import pytest-mock at this time, so we could do that with a follow-up PR to keep this focused on the feat change.

clients/python/tests/test_client.py
tarilabs: done with https://github.com/kubeflow/model-registry/pull/1553/commits/60db060d4c78ebf0c89a36157d2262453d2f30df
and look overall simpler, do you agree?

clients/python/tests/test_client.py
```diff
@@ -1206,3 +1208,122 @@ async def test_as_mlops_engineer_i_would_like_to_store_a_longer_documentation_fo
...
+        return self.path_arg in self.file_contents
+
+    def read_text(self):
+        return self.file_contents.get(self.path_arg, "<wrong, returning arbitrary string>")
+
+    def __str__(self):
+        return self.path_arg
+
+    def __repr__(self):
+        return f"MockPath('{self.path_arg}', file_contents={self.file_contents})"
```
jonburdo: Should we get rid of this class? If you think it's worth keeping, I'm not opposed. Just hesitant to encourage mocking things like Path. Usually it can be avoided with slightly improved code design. Either way, this isn't a blocker for merging this.

tarilabs: thank you for calling it out; I simply forgot while I refactored according to your feedback with : https://github.com/kubeflow/model-registry/pull/1553#discussion_r2329522938

tarilabs: done with https://github.com/kubeflow/model-registry/pull/1553/commits/bb5835720f1f7afbb374d9860d3d84afc02f9613

tarilabs:
thanks @jonburdo for the 👀 

- https://github.com/kubeflow/model-registry/pull/1553#pullrequestreview-3198714409

/lgtm
/approve

tarilabs:
adding label per https://github.com/kubeflow/model-registry/pull/1553#pullrequestreview-3198714409

tarilabs:
label has been added per https://github.com/kubeflow/model-registry/pull/1553#pullrequestreview-3198714409

PR #1583, opened 2025-09-11
fix: quote table names to support postgresql queries

Al-Pragliola:
## Description

unquoted table names are auto set to lowercase by PostgreSQL, so that are those not found when doing queries because PostgreSQL is also case-sensitive 

e.g.

on `/api/model_registry/v1alpha3/registered_models/1/versions`

```
{
  "code":"Internal Server Error",
  "message":"error listing model versions: ERROR: relation \"parentcontext\" does not exist (SQLSTATE 42P01)"
}
```

on `/api/model_registry/v1alpha3/experiment_runs?filterQuery=name="exp-run8`

```
{
  "code": "Internal Server Error",
  "message": "error listing experiment runs: ERROR: missing FROM-clause entry for table \"context\" (SQLSTATE 42P01)"
}
```

as a follow up I strongly suggest to add PostgreSQL to end to end tests

## How Has This Been Tested?

manually run tests on a local cluster by using affected endpoints

## Merge criteria:

### DIFF
--- a/internal/db/service/inference_service.go
+++ b/internal/db/service/inference_service.go
@@ -66,8 +66,8 @@ func applyInferenceServiceListFilters(query *gorm.DB, listOptions *models.Infere
 	}
 
 	if listOptions.ParentResourceID != nil {
-		query = query.Joins("JOIN ParentContext ON ParentContext.context_id = Context.id").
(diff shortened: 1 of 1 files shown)


tarilabs:
⏲️ considering this fixes for PostgreSQL without regressing for MySQL I'm going to 
/approve
it and we can revisit this later if needed by others.

PR #1615, opened 2025-09-16
Update FE routes and model catalog image

ppadti:
## Description
This PR aims to fix some route in FE. while constructing the route I assumed we always have repositoryName, but looks like few might not have it. So this PR updates the routes to have modelName, if the model has repository name it will encode it with modelName and will consider as single param, otherwise it will be just modelName.
Also update the model catalog image. fixed breaking tests.

## How Has This Been Tested?
make sure the model catalog pages works if the model does not have repository name and model with no artifacts.

## Merge criteria:

If you have UI changes


ppadti:
@mturley one question that I have is - if the model does not have uri in it - what should be the flow for register from catalog.
Currently we are pre-filling the uri value and its a read-only field. if the model does not have uri, should we just make it text field?

mturley:
@ppadti I think we can assume there will always be a uri, otherwise there's nothing to register. If the uri is missing that's an error and we maybe should block the register button

ppadti:
> @ppadti I think we can assume there will always be a uri, otherwise there's nothing to register. If the uri is missing that's an error and we maybe should block the register button

yeah, but when I run the application with catalog server, it has a model with no artifacts - just wanted to make sure we cover that case as well.
So if we don't have uri - we should disable the register button with a tooltip right?

mturley:
@ppadti I think that's correct, we can just use a tooltip with "Model location is unavailable". Is it just some of the data that has no artifacts?

mturley (approved):
LGTM, tested and working, thanks @ppadti !

PR #1638, opened 2025-09-20
fix user token handling

jonburdo:
Previously the user_token_envvar kwarg for ModelRegistry.__init__ was given an explicit default string value, so passing in None or another falsy value would override the default. With this change, passing a falsy value results in the default being tried. This also errors when the user passes a non-empty string value but the env var is not set.

This allows the async-job to automatically use the default value as expected. Previously, we had to explicitly specify a value like this:
```
- name: MODEL_SYNC_REGISTRY_USER_TOKEN_ENVVAR
  value: KF_PIPELINES_SA_TOKEN_PATH
```
because failing to do so sets user_token_envvar to None.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -86,7 +86,7 @@ class ModelRegistry:
         author: str,
         is_secure: bool = True,
         user_token: str | None = None,
-        user_token_envvar: str = DEFAULT_USER_TOKEN_ENVVAR,
+        user_token_envvar: str | None = None,
         custom_ca: str | None = None,
         custom_ca_envvar: str | None = None,
         log_level: int = logging.WARNING,
@@ -124,19 +124,9 @@ class ModelRegistry:
             logger.debug("Setting up reentrant async event loop")
             nest_asyncio.apply()
 
-        if not user_token and user_token_envvar:
-            logger.info("Reading user token from %s", user_token_envvar)
-            # /var/run/secrets/kubernetes.io/serviceaccount/token
-            if sa_token := os.environ.get(user_token_envvar):
-                if user_token_envvar == DEFAULT_USER_TOKEN_ENVVAR:
-                    logger.info(
-                        f"Sourcing user token from default envvar: {DEFAULT_USER_TOKEN_ENVVAR}"
-                    )
-                user_token = Path(sa_token).read_text()
-            elif Path(DEFAULT_K8S_SA_TOKEN_PATH).exists():
-                user_token = Path(DEFAULT_K8S_SA_TOKEN_PATH).read_text()
-                logger.info("Sourced user token from K8s default path: %s.", DEFAULT_K8S_SA_TOKEN_PATH)
-            else:
+        if not user_token:
+            user_token = self._get_user_token(user_token_envvar)
+            if not user_token:
                 warn("User access token is missing", stacklevel=2)
 
         self.hint_server_address_port(server_address, port)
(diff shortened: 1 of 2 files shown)


clients/python/src/model_registry/_client.py
```diff
@@ -72,7 +72,7 @@
 logger = logging.getLogger("model-registry")
 
 DEFAULT_USER_TOKEN_ENVVAR = "KF_PIPELINES_SA_TOKEN_PATH"  # noqa: S105
-DEFAULT_K8S_SA_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token" # noqa: S105
+DEFAULT_K8S_SA_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"  # noqa: S105
```
tarilabs:
```suggestion
DEFAULT_K8S_SA_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token" # noqa: S105
```

not sure I get this modification? 🙃

tarilabs (commented):
thanks @jonburdo , a minor comment please

Al-Pragliola:
it would be useful to have few lines in the description of this PR of what is changing in the user token handling 
e.g. before this PR we did X, this PR aims to do Y instead and why

clients/python/src/model_registry/_client.py
```diff
@@ -86,7 +86,7 @@ def __init__(  # noqa: C901
         author: str,
         is_secure: bool = True,
         user_token: str | None = None,
-        user_token_envvar: str = DEFAULT_USER_TOKEN_ENVVAR,
+        user_token_envvar: str | None = None,
```
jonburdo, line 89: We want to distinguish between:
- falling back to the default value of `DEFAULT_USER_TOKEN_ENVVAR` because `None` is provided (or `user_token_envvar` kwarg is not used)
- and the user explicitly passing a non-empty string value, even if that value happens to be `DEFAULT_USER_TOKEN_ENVVAR`. In this case we error if the env var is not set, and log messages are slightly different.

We also want passing `user_token_envvar=None` to be equivalent to not passing the kwarg at all.

clients/python/tests/test_client.py
```diff
@@ -13,9 +13,9 @@
 from model_registry.types.artifacts import DocArtifact
 
 
-def test_secure_client():
-    os.environ["CERT"] = ""
-    os.environ["KF_PIPELINES_SA_TOKEN_PATH"] = ""
+def test_secure_client(monkeypatch):
+    monkeypatch.delenv("CERT", raising=False)
+    monkeypatch.delenv("KF_PIPELINES_SA_TOKEN_PATH", raising=False)
```
jonburdo, line 18: We have a module-scoped fixture that temporarily sets `KF_PIPELINES_SA_TOKEN_PATH`, used by multiple tests. I was getting test failures due to `KF_PIPELINES_SA_TOKEN_PATH` being modified:

https://github.com/kubeflow/model-registry/blob/198a8576dd7821149d0b29ffbc9cb8921ac73132/clients/python/tests/conftest.py#L180-L193

clients/python/src/model_registry/_client.py
```diff
@@ -167,6 +157,43 @@ def __init__(  # noqa: C901
...
+    @staticmethod
+    def _get_user_token(user_token_envvar: str | None = None) -> str | None:
+        sa_token_path: str
+        user_provided: bool = True
+        if user_token_envvar:
+            try:
+                sa_token_path = os.environ[user_token_envvar]
+            except KeyError:
+                msg = f"user_token_envvar is {user_token_envvar!r} but no such env var is set"
+                raise ValueError(msg) from None
```
jonburdo, line 169: We error if the user explicitly gives a value

clients/python/src/model_registry/_client.py
jonburdo: I ran `ruff` on this file which mostly formatted my code changes, but affected a couple other lines. I updated the PR to revert this and other formatting outside of my code, since we don't have make commands and CI maintaining this currently.

jonburdo:
Just updated this to:
- add PR description and more descriptive commit message
- revert unrelated formatting changes

clients/python/src/model_registry/_client.py
tarilabs: ah, that's weird of Ruff 🤔

PR #1652, opened 2025-09-22
replace black formatter with ruff

jonburdo:
This improves consistency since ruff is already used for linting.

## Description

This replaces `black` with `ruff format` in `make tidy` to be consistent. (We already have `ruff check` in `make tidy`.)

Most of the changes are extraneous newlines that are removed by ruff, and change to poetry.lock. To focus on other changes, use:
```
git diff main --ignore-blank-lines -- . ':!poetry.lock'
```

Note:
```
$ git diff main --shortstat --ignore-blank-lines -- . ':!poetry.lock'
 4 files changed, 6 insertions(+), 9 deletions(-)
```

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [ ] The commits have meaningful messages
- [ ] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work.
- [ ] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).
- [ ] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

- [ ] The developer has added tests or explained why testing cannot be added.
- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

PR #1661, opened 2025-09-24
format client code

jonburdo:
This changes `make tidy` to format everything under src so the actual client code in src/model_registry is included. Previously only src/mr_openapi was being formatted.

## Description

Follow-up to: https://github.com/kubeflow/model-registry/pull/1652

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/Makefile
+++ b/clients/python/Makefile
@@ -91,8 +91,8 @@ lint:
 
 .PHONY: tidy
 tidy:
-	poetry run ruff check --fix --unsafe-fixes --fix-only || true
-	poetry run ruff format src/mr_openapi
+	poetry run ruff check --fix-only --unsafe-fixes
+	poetry run ruff format src
 
 .PHONY: build
 build: install tidy
--- a/clients/python/src/model_registry/utils.py
+++ b/clients/python/src/model_registry/utils.py
@@ -149,9 +149,7 @@ def _kwargs_to_params(kwargs: dict[str, str]) -> list[str]:
     return args
 
 
-def _get_skopeo_backend(
-    pull_args: list[str] | None = None, push_args: list[str] | None = None
-) -> BackendDefinition:
+def _get_skopeo_backend(pull_args: list[str] | None = None, push_args: list[str] | None = None) -> BackendDefinition:
     try:
         from olot.backend.skopeo import is_skopeo, skopeo_pull, skopeo_push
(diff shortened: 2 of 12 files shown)


clients/python/Makefile
```diff
@@ -91,8 +91,8 @@ lint:
 
 .PHONY: tidy
 tidy:
-	poetry run ruff check --fix --unsafe-fixes --fix-only || true
-	poetry run ruff format src/mr_openapi
+	poetry run ruff check --fix-only --unsafe-fixes
+	poetry run ruff format src
```
jonburdo, line 95: The main change is replacing `src/mr_openapi` with `src` so that `src/model_registry` gets checked. If we were to add another dir under `src`, it should be checked too.

Also:
- Removing redundant `--fix` option because `--fix-only` implies `--fix`
- Removing `|| true` because `--fix-only` prevents non-zero exit status due to lint violations. If we get non-zero exit status for another reason such as missing `ruff` command or invalid option, we should fail fast rather than ignoring it.

A general note on makefile commands: We can basically do the equivalent of `... || true` by prefixing with a `-`. For instance, if we **really** wanted to ignore non-zero exit status in **every** scenario, we could:
```diff
- poetry run ruff check --fix-only --unsafe-fixes
+ -poetry run ruff check --fix-only --unsafe-fixes
```

From `ruff check -h`:
```
      --fix-only
          Apply fixes to resolve lint violations, but don't report on, or exit non-zero for,
          leftover violations. Implies `--fix`. Use `--no-fix-only` to disable or
          `--unsafe-fixes` to include unsafe fixes
```

clients/python/src/model_registry/utils.py
```diff
@@ -724,7 +716,7 @@ def get_auth_reference(image_path: str) -> str:
     return repo_path
 
 
-def create_auth_object(oci_ref: str, username: str, password: str) -> dict[str: dict[str, dict[str, str]]]:
+def create_auth_object(oci_ref: str, username: str, password: str) -> dict[str, dict[str, dict[str, str]]]:
```
jonburdo, line 719: The one manual change outside of the `Makefile` - fixing this typo.

Everything else is from running `make tidy` with the updated `Makefile`.

tarilabs: nice catch!

PR #1664, opened 2025-09-24
MC Hardware configuration table with mock data

YuliaKrimerman:
## Description

Implemented a comprehensive Hardware Configuration table within the Performance Insights tab of the Model Catalog details page, replacing the previous "Coming Soon" placeholder with a fully functional data table.
HardwareConfigurationTable: Main table component with sorting, and loading states
 In addition includes a fork from @mturley  that includes the scrolling of the table, as well as the left sticky columns and the use of mod-arch Table

## How Has This Been Tested?

https://github.com/user-attachments/assets/02dce588-ed25-48f9-98b0-58b8155fe8f9

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelCatalog/mocks/hardwareConfigurationMock.ts
```diff
@@ -0,0 +1,184 @@
+import { HardwareConfiguration } from '~/app/pages/modelCatalog/types/hardwareConfiguration';
+
+export const mockHardwareConfigurations: HardwareConfiguration[] = [
+  {
+    id: '1',
+    hardwareConfiguration: '1 x A100-80',
```
ppadti: @YuliaKrimerman @mturley I just wanted to understand the format of these metric data. I think these will be present in the customProperties of artifacts endpoint response right, or am  I missing something? 
for BFF mock data, I added these to the customProperties. so...

YuliaKrimerman: For my PR I just used a random mock for those fields. the true data should come once the API will be implemented

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+                return column.label;
+              };
+
+              return (
+                <Th
+                  key={column.field}
+                  className={
+                    index < 2
+                      ? 'pf-v6-c-table__sticky-column pf-v6-c-table__sticky-column--left'
+                      : ''
```
ppadti: when I run this in patternfly theme - I am not seeing the sticky column. 
I think we can add this `isStickyColumn`property to `hardwareConfigColumns` directly instead making this condition here.
`SortableData<HardwareConfiguration>[]` can have is  `isStickyColumn` along with field, label and sortable.

ppadti (changes requested):

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+            max-width: 100%;
+            overflow-x: auto;
+          }
+          .hardware-table-container table {
+            width: max-content;
+            min-width: 100%;
+            table-layout: auto;
+          }
+        `}
+      </style>
```
mturley: Let's avoid inline styles. If we really need custom CSS we should import a separate `.scss` file, but even that should be avoided if possible - the container sizing and scrolling behavior seems like something we could achieve with PatternFly props and/or [utility classes](https://www.patternfly.org/utility-classes). What happens without these styles?

Edit: ok, maybe this is tricky without custom styles, but we should still put them in a .scss file. I'll mess around to see if I can help solve this.

We do still get a double scrollbar effect here which we really need to avoid. If we can't figure it out as part of this PR and it's blocking things we could follow up in another PR but let's leave the issue open until we solve that I think.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
mturley: Hmm... I was going to suggest we use the `isStickyColumn` prop directly on the `Th` like in [this PF sticky columns example](https://www.patternfly.org/components/table#sticky-column), but looking closer at that `SortableData` type it seems to extend `ThProps`. I don't think that property will end up on the Th unless you're spreading `{...column}` to it here right? ~~I imagine we need to do that...~~

Ok looking closer, we are mixing component structures here. You're importing the PF `Table` directly but using the `SortableData` type for your columns, which is intended to be used with our own `Table` wrapper imported from `mod-arch-shared`. It should really have its own name...

Let's shift to using that mod-arch-shared Table component @YuliaKrimerman. Use our other tables from MR as a reference like the `RegisteredModelTable`. That way you pass your `hardwareConfigColumns` array directly into its `columns` prop and properties in there like `isStickyColumn` should work that way (and you won't need to render your own `Th`'s. Instead of your own `Tbody` with a .map you'd use a `rowRenderer` callback.

Or if there is some other reason you can't use that abstraction let's talk through it.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+                      : ''
+                  }
+                  style={{
+                    width: `${column.width}ch`,
+                    minWidth: `${column.width}ch`,
+                    verticalAlign: 'middle',
+                    height: 'auto',
+                    lineHeight: '1.2',
+                    textAlign: 'left',
+                  }}
```
mturley: We really need to avoid inline styles. Things like this break when we upgrade PF. If you can't achieve what you need with existing component props or utility classes let's talk through what you're trying to solve here.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTableRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+        return value;
+      };
+
+      return (
+        <Td
+          key={column.field}
+          dataLabel={column.label.replace('\n', ' ')}
+          className={
+            index < 2 ? 'pf-v6-c-table__sticky-column pf-v6-c-table__sticky-column--left' : ''
+          }
```
mturley: You should also be able to avoid passing in special sticky classes like this by rendering this as a `Th` instead of a `Td` and passing it the `isStickyColumn` prop (and other props depending on whether it's the first one or second one, [see the description of the PF example](https://www.patternfly.org/components/table/#multiple-left-aligned-sticky-columns)). You'll have to conditionally use Th for only those first two columns I think. If conditionals are getting too crazy here, maybe just write out those two `Th`s and then only do your .map() for the rest of the array.

clients/ui/frontend/src/app/pages/modelCatalog/mocks/hardwareConfigurationMock.ts
mturley: @ppadti my understanding is that you're right that we'll have customProperties with these metrics on them. @YuliaKrimerman we should probably use the real object types here - [we already have the `CatalogModelArtifact` type](https://github.com/kubeflow/model-registry/blob/main/clients/ui/frontend/src/app/modelCatalogTypes.ts) and we need to add `artifactType` and `metricsType` properties to it. The artifacts we'll use for this table will specifically have `artifactType: 'metrics-artifact'` and `metricsType: 'performance-metrics'`, and then their customProperties contain things like `hardware: { string_value:'H100' }`, `hardware_count: { int_value: 2 }`, `ttft_mean: { double_value: 35.48818160947744 }`. See the ["Exposing Jounce Metrics in the Model Catalog" section of this ADR doc](https://docs.google.com/document/d/12wGInEpLsPS4P8XiczgQwYhJavn5fAUkt4XQTyAVLns/edit?tab=t.0).

Let's create some discriminated types for these different kinds of artifacts. Can you take the existing `CatalogModelArtifact` type, rename it to `CatalogModelUriArtifact`, move its `createTimeSinceEpoch`, `lastUpdateTimeSinceEpoch` and `customProperties` into a new `CatalogModelArtifactBase`, and make a new `CatalogModelPerformanceMetricsArtifact` type that also extends that base? That last type can have those literal artifactType and metricsType values and its own `customProperties` which contain the literal expected properties from the ADR instead of the generic `ModelRegistryCustomProperties`. Finally, we can make a union type `CatalogModelArtifact = CatalogModelUriArtifact | CatalogModelPerformanceMetricsArtifact`. Later we will also need to add an additional member to this union for the `artifactType: 'accuracy-metrics'`, that will be needed for part of what we show on the updated landing page cards, but it's out of scope for this PR.

The existing code that fetches artifacts can still work with `CatalogModelArtifact`, and then in places where we need specific kinds of artifacts we can do something like `if (artifact.uri)` or `if (artifact.metricsType === 'performance-metrics)` to narrow it, then it will typecheck the use of the specific custom properties.

This may require looking at where we currently pull the URI out of an artifact on the details and register pages and having those places narrow the artifact. In fact, I think currently those places just grab the first artifact and assume it's the URI one, we'll need to fix that so they find an artifact that has a URI. We may have to make that fix as part of this PR if you're changing the types here:

https://github.com/kubeflow/model-registry/blob/main/clients/ui/frontend/src/app/pages/modelCatalog/screens/RegisterCatalogModelForm.tsx#L63

https://github.com/kubeflow/model-registry/blob/main/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsView.tsx#L94

I believe the updated API will have an `artifactType: 'uri-artifact'` but it doesn't currently so to keep things from breaking we can just look for the uri property.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -1,17 +1,47 @@
...
+    <PageSection style={{ overflow: 'hidden', maxWidth: '100vw', width: '100%' }}>
+      <style>
+        {`
+          .performance-insights-container {
+            width: 100%;
+            max-width: 100%;
+            overflow-x: hidden;
+          }
+        `}
+      </style>
```
mturley: Yeah let's kill all the inline styles and start fresh, try using the sticky columns pattern with the mod-arch-shared version of the Table and then see what minimal tweaks it needs.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -1,17 +1,47 @@
...
+              </FlexItem>
+              <FlexItem>
+                <p>
+                  Compare the performance metrics of hardware configuration to determine the most
+                  suitable option for deployment.
+                </p>
+              </FlexItem>
+            </Flex>
+          </FlexItem>
+          <FlexItem style={{ width: '100%', overflow: 'hidden' }}>
```
mturley: Same here with the inline styles

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
ppadti: oh, I didn't see we are importing Table from patternfly - I assumed we use it from shared-library, so suggested using `isStickyColumn` property directly on `hardwareConfigColumns`, so that everything else will be handled there, sorry.

clients/ui/frontend/src/app/pages/modelCatalog/mocks/hardwareConfigurationMock.ts
ppadti: @mturley if we can get [this PR](https://github.com/kubeflow/model-registry/pull/1666) in, so we can avoid re-work on this maybe?

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
ppadti: maybe the double scrollbar is only happening with MUI and not with patternfly, I didn't see that happening with patternfly...

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTableRow.tsx
```diff
@@ -0,0 +1,85 @@
...
+      case 'mean_output_tokens':
+        return formatTokenValue(getDoubleValue(customProperties, field));
+      case 'framework_version':
+        return getStringValue(customProperties, field);
+      default:
+        return '-';
+    }
+  };
+
+  // TODO sticky isn't quite working with both columns and the scroll container is weird. double check PF docs
```
mturley, line 64: I forgot to remove this comment - I want to unblock your rebase, can you remove in the next PR @YuliaKrimerman ?

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTableRow.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <Tr>
+      {hardwareConfigColumns.map((column) => (
+        <Td
+          key={column.field}
+          dataLabel={column.label.replace('\n', ' ')}
```
mturley, line 71: @YuliaKrimerman in the next PR can you also just switch this to `dataLabel={column.label}`? I forgot to remove the replace here which is no longer necessary

PR #1698, opened 2025-10-01
Basic filters rendering

rsun19:
## Description

Added filters and types
Screenshot 2025-10-01 at 4 18 53 PM

## How Has This Been Tested?

Go to the model catalog page and check the functionality. Added cypress tests as well

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogPage.tsx
```diff
@@ -27,6 +28,9 @@ const ModelCatalogPage: React.FC<ModelCatalogPageProps> = ({ searchTerm }) => {
     searchTerm,
   );
 
+  const { filterData } = React.useContext(ModelCatalogContext);
+  const filteredModels = filterModelCatalogModels(catalogModels.items, filterData);
```
mturley: You're filtering on the client side, which means it's only filtering the current page of results since we also have the lazy-loading "load more models" button. We'll need to instead apply the filter to the API request params using @andrewballantyne's code, but that's out of scope for this issue. We should either remove this and the `filterModelCatalogModels` function (just have the filters not actually do anything for this PR) or add comments in both places saying it needs to be replaced with server-side filtering. I think I'd lean towards just removing it now tbh, sorry 😦

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -41,6 +55,14 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
   const [selectedSource, setSelectedSource] =
     React.useState<ModelCatalogContextType['selectedSource']>(undefined);
+  const [filterData, setFilterData] = useGenericObjectState<ModelCatalogFilterDataType>({});
+
+  // const setTypedFilterData = React.useCallback(
+  //   <K extends keyof ModelCatalogFilterStatesByKey>(key: K, value: ModelCatalogFilterState<K>) => {
+  //     setFilterData(key, value);
+  //   },
+  //   [setFilterData],
+  // );
```
mturley: Looks like some commented out code to remove here

clients/ui/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -74,20 +81,17 @@ export type CatalogFilterNumberOption = {
   };
 };
 
-export type CatalogFilterStringOption = {
+export type CatalogFilterStringOption<T> = {
```
mturley: Can you change `T` to `T extends string` here? Might not matter yet but will prevent us accidentally using a number type as the type param in the future etc.

clients/ui/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -74,20 +81,17 @@ export type CatalogFilterNumberOption = {
...
   type: 'string';
-  values: string[];
+  values: T[];
 };
 
-export type CatalogFilterOption = CatalogFilterNumberOption | CatalogFilterStringOption;
+export type CatalogFilterOptions =
+  | Partial<ModelCatalogFilterTypesByKey>
+  | CatalogFilterNumberOption;
```
mturley: I don't think this is right, wouldn't this just be `Partial<ModelCatalogFilterTypesByKey>`? When we add numeric filters later on they would just become part of ModelCatalogFilterTypesByKey, this has one numeric filter optionally replacing the whole filters object.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+export enum ModelCatalogTasks {
+  AUDIO_TO_TEXT = 'audio-to-text',
+  IMAGE_TEXT_TO_TEXT = 'image-text-to-text',
+  IMAGE_TO_TEXT = 'image-to-text',
+  TEXT_GENERATION = 'text-generation',
+  TEXT_TO_TEXT = 'text-to-text',
+  VIDEO_TO_TEXT = 'video-to-text',
+}
```
mturley: Nit, but can we rename the enums to be singular like `ModelCatalogFilterKey`, `ModelCatalogTask`? That makes it easier to read if we use an array of them, `ModelCatalogTask[]` is valid for an array of those values. Same for ModelCatalogProvider, ModelCatalogLicense

clients/ui/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -130,3 +134,34 @@ export type CatalogModelDetailsParams = {
...
+  ModelCatalogFilterState<ModelCatalogFilterKeys.LANGUAGE>;
+
+export type ModelCatalogFilterStatesByKey = {
+  [K in ModelCatalogFilterKeys]: ModelCatalogFilterState<K>;
+};
+
+export type ModelCatalogFilterDataType = Partial<ModelCatalogFilterStatesByKey>;
+
+export type FilterValue<K extends ModelCatalogFilterKeys> =
+  ModelCatalogFilterTypesByKey[K]['values'][number];
```
mturley: I think you can maybe simplify these types. `ModelCatalogFilterState` doesn't need to be generic, it could be explicit just like `ModelCatalogFilterTypesByKey` and rely on narrowing by property access instead of generics. I think otherwise we code ourselves into a corner where this only works for checkbox filters, and in the future we'll have numeric filters.

Can we try making this type explicit, and instead of using nested objects with booleans we can use arrays of strings to represent the selected checkboxes? I think that will be easier to traverse, and allow us to add different value types more easily later.

```ts
export type ModelCatalogFilterStatesByKey = {
  [ModelCatalogFilterKeys.TASK]: ModelCatalogTask[];
  [ModelCatalogFilterKeys.PROVIDER]: ModelCatalogProvider[];
  [ModelCatalogFilterKeys.LICENSE]: ModelCatalogLicense[];
  [ModelCatalogFilterKeys.LANGUAGE]: ModelCatalogLanguage[];
}
```

Then your `ModelCatalogProvidersFilterStateType` is just `ModelCatalogFilterStatesByKey[ModelCatalogFilterKeys.PROVIDER]`, etc. and I don't think you end up needing `ModelCatalogFilterState` or `FilterValue`. (NIt but I would also call the entire `ModelCatalogFilterStatesByKey` object `ModelCatalogFilterState` since it's the single state object we'll pass around, but that's not important).

Edit: Ok I see why you need `FilterValue<K>` later for your `filterToNameMapping`, but I think it is specific to the string filters (numeric filters won't have a `values` array). So I think you can maybe call it `StringFilterValue<K extends ModelCatalogFilterKeys>` and we can deal with how to narrow ModelCatalogFilterKeys better when we add numeric filters perhaps. We might need a `ModelCatalogStringFilterKeys` type that is a subset of ModelCatalogFilterKeys`, but for now they would be the same type.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogFilters.tsx
```diff
@@ -1,7 +1,44 @@
...
+import ProviderFilter from './globalFilters/ProviderFilter';
+import LicenseFilter from './globalFilters/LicenseFilter';
+import LanguageFilter from './globalFilters/LanguageFilter';
 
-const ModelCatalogFilters: React.FC = () => (
-  <div data-testid="model-catalog-filters">{/* // TODO: add filters */}</div>
-);
+const ModelCatalogFilters: React.FC = () => {
+  const { apiState } = React.useContext(ModelCatalogContext);
+  const [filterOptions] = useCatalogFilterOptionList(apiState);
```
mturley: We need to handle loading and error states here and render either a spinner or an alert. Actually I think we probably should load the filter options list in the ModelCatalogContext since we'll need it on the details page as well, those filters will be client side but they'll need access to the same value options. We can make the loading and error state part of the entire page's spinner/alert.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogStringFilter.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const ModelCatalogStringFilter = <K extends ModelCatalogFilterKeys>({
+  title,
+  filterKey,
+  filterToNameMapping,
+  filters,
+}: ModelCatalogStringFilterProps<K>): JSX.Element => {
+  const [showMore, setShowMore] = React.useState(false);
+  const [searchValue, setSearchValue] = React.useState('');
+  const [filteredValues, setFilteredValues] = React.useState<FilterValue<K>[]>(filters.values);
```
mturley: The filter/search naming here is a bit confusing, I initially thought this was duplicated filter selection state. Can we call this something like `valuesMatchingSearch`?

Also, actually, it does not need to be state. It is derived entirely from `filters.values` and `searchValue`. You can instead have `const valuesMatchingSearch = useMemo(...)` to memoize your `filters.values.filter()` call.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogStringFilter.tsx
```diff
@@ -0,0 +1,99 @@
...
+        return value.toLowerCase().includes(lowerCaseValue) || label.includes(lowerCaseValue);
+      }),
+    );
+  };
+
+  const onToggle = (checkbox: FilterValue<K>, checked: boolean) => {
+    const nextState: Partial<Record<FilterValue<K>, boolean>> = {
+      ...(filterData[filterKey] ?? {}),
+      [checkbox]: checked,
+    };
```
mturley: Ok. This is where we'd need to adjust to match the type changes I suggested. To be compatible with the util Andrew wrote to help us with converting to the API query and to simplify types a bit, I think we should use `FilterValue<K>[]` instead of `Partial<Record<FilterValue<K>, boolean>>`. The update logic gets slightly more complicated - you'll need to take the prior state's array and either filter out the value or append the value depending on whether you're checking or unchecking the box. And then in places where you are accessing the boolean, you can just use `.includes()`. I get what you're going for here with mapping it nicely to booleans but I think it makes types a little harder to work with.

clients/ui/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -130,3 +134,34 @@ export type CatalogModelDetailsParams = {
...
+export type ModelCatalogTasksFilterStateType = ModelCatalogFilterState<ModelCatalogFilterKeys.TASK>;
+
+export type ModelCatalogProvidersFilterStateType =
+  ModelCatalogFilterState<ModelCatalogFilterKeys.PROVIDER>;
+
+export type ModelCatalogLicensesFilterStateType =
+  ModelCatalogFilterState<ModelCatalogFilterKeys.LICENSE>;
+
+export type ModelCatalogLanguagesFilterStateType =
+  ModelCatalogFilterState<ModelCatalogFilterKeys.LANGUAGE>;
```
mturley: I notice you aren't using these 4 types anywhere, do we need them?

mturley (changes requested):
Getting really close here, this is great work. Just some thoughts on making the types simpler and more flexible and a couple other minor nits.

clients/ui/frontend/src/app/modelCatalogTypes.ts
rsun19: i meant to do something like Record<string, CatalogFilterNumberOption> in case the api returned something like that, but maybe that isn't an issue

rareddy:
@mturley qq: are you looking for static `filter options` on client side? or are you expecting the this served through the enpoint?

mturley:
@rareddy our expectation is that the filter options are served by the endpoint as stated in the ADR, but based on conversations with @andrewballantyne my understanding is we will not get the nice microcopy versions of filter names and filter value names from the API (just underlying keys) so we still need a static list of those in the UI. But we still rely on the API telling us what is valid, and if the API returns more filter values that we don't expect (for a filter we do expect) we will render them "raw" and they'll still work.

We did however decide to have the UI only render the filters that it knows about (as opposed to any string filter the API gives us) in order to let us do things like combine multiple filters to render the latency filter controls etc. we can explore making it more dynamic in the future.

andrewballantyne:
This is one of many conversations we should have in the growth space of this API. We got a few more topics of conversation too 🙂

PR #1701, opened 2025-10-03
Validated model cards in Model catalog

YuliaKrimerman:
Here's the filled PR form for your validated model cards implementation:

## Description

Created validated model cards for the Model Catalog to display Red Hat AI validated models with performance and accuracy metrics. This PR introduces a new `ValidatedModelCard` component that shows validated models with special styling, performance metrics (hardware configuration, RPS, TTFT), accuracy scores, and benchmark navigation.

**UI Changes:**
- Validated models now display as special cards with performance metrics instead of standard model cards
- Cards include tooltips for metrics explanation (currently with placeholder)
Screenshot 2025-10-01 at 2 45 19 PM

## How Has This Been Tested?

- Validated model cards integrate seamlessly with existing ModelCatalogPage
- Conditional rendering works correctly (ValidatedModelCard for validated models, ModelCatalogCard for others)
- Mock data integration functions properly with utility functions

## Merge criteria:

- [x] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogLabels.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogLabels.tsx
@@ -14,12 +14,13 @@ const ModelCatalogLabels: React.FC<ModelCatalogLabelsProps> = ({
   provider,
   labels = [],
 }) => (
-  <LabelGroup numLabels={5} isCompact>
+  <LabelGroup numLabels={2} isCompact>
     {tasks.map((task) => (
       <Label data-testid="model-catalog-label" key={task} variant="outline">
         {task}
       </Label>
     ))}
+    {provider && <Label isCompact>{provider}</Label>}
     {labels.map((label) => (
       <Label data-testid="model-catalog-label" key={label} variant="outline">
         {label}
@@ -30,7 +31,6 @@ const ModelCatalogLabels: React.FC<ModelCatalogLabelsProps> = ({
         {license}
       </Label>
     )}
-    {provider && <Label isCompact>{provider}</Label>}
   </LabelGroup>
 );
 
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
@@ -1,0 +1,247 @@
+import React from 'react';
+import {
+  Badge,
+  Button,
+  Card,
+  CardBody,
+  CardFooter,
+  CardHeader,
+  CardTitle,
+  Flex,
+  FlexItem,
+  Tooltip,
+  Content,
+  ContentVariants,
+  Skeleton,
+} from '@patternfly/react-core';
+import text from '@patternfly/react-styles/css/utilities/Text/text';
+import { ChartBarIcon, HelpIcon, ChevronLeftIcon, ChevronRightIcon } from '@patternfly/react-icons';
+import { Link } from 'react-router-dom';
+import {
+  CatalogModel,
+  CatalogSource,
+  CatalogPerformanceMetricsArtifact,
+  CatalogAccuracyMetricsArtifact,
+} from '~/app/modelCatalogTypes';
+import { getModelName } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+import { catalogModelDetailsFromModel } from '~/app/routes/modelCatalog/catalogModel';
+import { getLabels } from '~/app/pages/modelRegistry/screens/utils';
+import { extractValidatedModelMetrics } from '~/app/pages/modelCatalog/utils/validatedModelUtils';
+import ModelCatalogLabels from '~/app/pages/modelCatalog/components/ModelCatalogLabels';
+
+type ValidatedModelCardProps = {
+  model: CatalogModel;
+  source: CatalogSource | undefined;
+  performanceMetrics?: CatalogPerformanceMetricsArtifact;
+  accuracyMetrics?: CatalogAccuracyMetricsArtifact;
+  truncate?: boolean;
+};
+
+const ValidatedModelCard: React.FC<ValidatedModelCardProps> = ({
+  model,
+  source,
+  performanceMetrics,
+  accuracyMetrics,
+  truncate = false,
+}) => {
+  const allLabels = model.customProperties ? getLabels(model.customProperties) : [];
+  const validatedLabels = allLabels.includes('validated') ? ['validated'] : [];
+
+  const metrics = extractValidatedModelMetrics(performanceMetrics, accuracyMetrics);
+
+  return (
+    <Card
+      isFullHeight
+      data-testid="validated-model-catalog-card"
+      key={`${model.name}/${model.source_id}`}
+    >
+      <CardHeader>
+        <CardTitle>
+          <Flex alignItems={{ default: 'alignItemsCenter' }} style={{ paddingLeft: '8px' }}>
+            {model.logo ? (
+              <img
+                src={model.logo}
+                alt="model logo"
+                style={{ height: '56px', width: '56px', marginRight: '8px' }}
+              />
+            ) : (
+              <Skeleton
+                shape="square"
+                width="56px"
+                height="56px"
+                screenreaderText="Brand image loading"
+                style={{ marginRight: '8px' }}
+              />
+            )}
+            <FlexItem align={{ default: 'alignRight' }}>
+              <Badge className="pf-m-purple">Validated</Badge>
+            </FlexItem>
+          </Flex>
+          <div style={{ marginTop: '8px' }}>
+            <Flex alignItems={{ default: 'alignItemsCenter' }} gap={{ default: 'gapSm' }}>
+              <div
+                data-testid="validated-model-status-indicator"
+                className={text.textColorStatusSuccess}
+                style={{
+                  width: '8px',
+                  height: '8px',
+                  backgroundColor: 'var(--pf-t--global--palette--green-400)',
+                  borderRadius: '50%',
+                }}
+              />
+              <Content component={ContentVariants.small} className={text.textColorStatusSuccess}>
+                Success
+              </Content>
+            </Flex>
+            <Link to={catalogModelDetailsFromModel(model.name, source?.id)}>
+              <Button
+                data-testid="model-catalog-detail-link"
+                variant="link"
+                tabIndex={-1}
+                isInline
+                style={{ fontSize: '1.25rem', fontWeight: 'bold' }}
+              >
+                {truncate ? (
+                  <span>{getModelName(model.name)}</span>
+                ) : (
+                  <span>{getModelName(model.name)}</span>
+                )}
+              </Button>
+            </Link>
+          </div>
+        </CardTitle>
+      </CardHeader>
+      <CardBody>
+        <div>
+          <div style={{ paddingLeft: '8px', marginBottom: '16px' }}>
+            <Flex
+              alignItems={{ default: 'alignItemsCenter' }}
+              justifyContent={{ default: 'justifyContentSpaceBetween' }}
+            >
+              <Flex
+                alignItems={{ default: 'alignItemsCenter' }}
+                gap={{ default: 'gapSm' }}
+                wrap="nowrap"
+              >
+                <ChartBarIcon />
+                <span
+                  data-testid="validated-model-accuracy"
+                  style={{ fontSize: '1.25rem', fontWeight: 'bold' }}
+                >
+                  {metrics.accuracy}%
+                </span>
+              </Flex>
+              <Flex
+                alignItems={{ default: 'alignItemsCenter' }}
+                gap={{ default: 'gapSm' }}
+                wrap="nowrap"
+              >
+                <span className="pf-v5-c-content pf-m-small">Average accuracy</span>
+                <Tooltip
+                  content="Average accuracy
+
+The weighted average of normalized scores from all benchmarks. Each benchmark is normalized to a 0-100 scale. All normalized benchmarks are then averaged together."
+                  position="top"
+                  className="pf-v6-c-tooltip"
+                >
+                  <HelpIcon style={{ cursor: 'help' }} />
+                </Tooltip>
+              </Flex>
+            </Flex>
+          </div>
+
+          {/* Hardware Configuration and Latency */}
+          <div style={{ paddingLeft: '8px', marginBottom: '16px' }}>
+            <Flex justifyContent={{ default: 'justifyContentSpaceBetween' }}>
+              <div>
+                <Content component={ContentVariants.small} style={{ color: '#6a6e73' }}>
+                  Hardware
+                </Content>
+                <Content component={ContentVariants.p} data-testid="validated-model-hardware">
+                  {metrics.hardwareCount}x{metrics.hardware}
+                </Content>
+              </div>
+              <div>
+                <Content component={ContentVariants.small} style={{ color: '#6a6e73' }}>
+                  RPS/rep.
+                </Content>
+                <Content component={ContentVariants.p} data-testid="validated-model-rps">
+                  {metrics.rpsPerReplica}
+                </Content>
+              </div>
+              <div>
+                <Content component={ContentVariants.small} style={{ color: '#6a6e73' }}>
+                  TTFT
+                </Content>
+                <Flex
+                  alignItems={{ default: 'alignItemsCenter' }}
+                  gap={{ default: 'gapSm' }}
+                  wrap="nowrap"
+                >
+                  <span data-testid="validated-model-ttft">{metrics.ttftMean} ms</span>
+                  <Tooltip
+                    content="Latency
+
+The delay (in milliseconds) between sending a request and receiving the first response.
+
+*TTFT (Time to First Token)
+The time between when a request is sent to a model and when the model begins streaming its first token in the response.
+
+*ITL (Inter-Token Latency)
+The average time between successive output tokens after the model has started generating.
+
+*E2E (End-to-End latency)
+The total time from when the request is sent until the last token is received."
+                    position="top"
+                    className="pf-v6-c-tooltip"
+                  >
+                    <HelpIcon />
+                  </Tooltip>
+                </Flex>
+              </div>
+            </Flex>
+          </div>
+
+          {/* Benchmarks */}
+          <div style={{ paddingLeft: '8px' }}>
+            <Flex
+              alignItems={{ default: 'alignItemsCenter' }}
+              justifyContent={{ default: 'justifyContentSpaceBetween' }}
+            >
+              <Content component={ContentVariants.p} data-testid="validated-model-benchmarks">
+                1 of 3{' '}
+                <Button variant="link" isInline style={{ padding: 0, fontSize: 'inherit' }}>
+                  benchmarks
+                </Button>
+              </Content>
+              <Flex gap={{ default: 'gapSm' }}>
+                <Button
+                  variant="plain"
+                  icon={<ChevronLeftIcon />}
+                  aria-label="Previous benchmark"
+                  data-testid="validated-model-benchmark-prev"
+                />
+                <Button
+                  variant="plain"
+                  icon={<ChevronRightIcon />}
+                  aria-label="Next benchmark"
+                  data-testid="validated-model-benchmark-next"
+                />
+              </Flex>
+            </Flex>
+          </div>
+        </div>
+      </CardBody>
+      <CardFooter>
+        <ModelCatalogLabels
+          tasks={model.tasks ?? []}
+          license={model.license}
+          provider={model.provider}
+          labels={validatedLabels}
+        />
+      </CardFooter>
+    </Card>
+  );
+};
+
+export default ValidatedModelCard;
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogPage.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogPage.tsx
@@ -13,6 +13,10 @@ import { ModelCatalogContext } from '~/app/context/modelCatalog/ModelCatalogCont
 import { useCatalogModelsBySources } from '~/app/hooks/modelCatalog/useCatalogModelsBySource';
 import { CatalogModel } from '~/app/modelCatalogTypes';
 import ModelCatalogCard from '~/app/pages/modelCatalog/components/ModelCatalogCard';
+import ValidatedModelCard from '~/app/pages/modelCatalog/components/ValidatedModelCard';
+import { isModelValidated } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+import { mockPerformanceMetricsArtifacts } from '~/app/pages/modelCatalog/mocks/hardwareConfigurationMock';
+import { mockAccuracyMetricsArtifacts } from '~/app/pages/modelCatalog/mocks/accuracyMetricsMock';
 import EmptyModelCatalogState from '~/app/pages/modelCatalog/EmptyModelCatalogState';
 
 type ModelCatalogPageProps = {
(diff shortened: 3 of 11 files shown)


mturley:
/hold
this reverts changes from https://github.com/YuliaKrimerman/model-registry/pull/2, will follow up with @YuliaKrimerman

mturley:
This PR should cherrypick the reverted commits back onto this branch and fix the diff here: https://github.com/YuliaKrimerman/model-registry/pull/3

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+type ValidatedModelCardProps = {
+  model: CatalogModel;
+  source: CatalogSource | undefined;
+  performanceMetrics?: CatalogPerformanceMetricsArtifact;
+  accuracyMetrics?: CatalogAccuracyMetricsArtifact;
+  truncate?: boolean;
+};
+
+const ValidatedModelCard: React.FC<ValidatedModelCardProps> = ({
```
mturley: This component repeats a decent amount of code from `ModelCatalogCard`. Can we refactor a bit so we're deduplicating that code and we just have conditions for the parts that are different? That'll be easier to maintain and prevent inconsistencies like the font size being different on the two cards.

I think what we could do is remove this new `ValidatedModelCard` and always render `ModelCatalogCard` with a few changes:
* If the model has the `validated` label, render the purple "Validated" badge in the top right instead of the source badge
* ~~If the model has the `validated` label, render the little checkmark indicator~~ (edit: actually let's exclude this, see my next comment)
* Move the model name link up from CardBody to CardTitle like you did here (edit: but leave its implementation as-is, see later comments)
* Move the `CardBody` and the rest of its contents to a new `ModelCatalogCardBody` component
  * In that component, we'll later want to fetch the artifacts for models that have the `validated` label to determine if there are accuracy and performance metrics to render. That fetch is out of scope for this PR, but setting up this body component gives it a good place to live where we can later just return a spinner during that loading state.
  * In that component, if we determine there are metrics to render (which for this PR can just be whether the validated label is present, maybe leave a comment explaining that later it will be based on fetching artifacts), render your new version of the CardBody children, otherwise render the old contents (just the truncated description).
    * If you want to you could further break this down into ValidatedModelCardBody and StandardModelCardBody components or something, but mostly I think it's just important to wrap the "which body should I render" logic.

This should make things a bit more consistent - generally, font sizes and such should look like they already did and these models just show more data. As it is the two cards look pretty out of place with one another.

Screenshot 2025-10-07 at 5 51 09 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                  width: '8px',
+                  height: '8px',
+                  backgroundColor: 'var(--pf-t--global--palette--green-400)',
+                  borderRadius: '50%',
+                }}
+              />
+              <Content component={ContentVariants.small} className={text.textColorStatusSuccess}>
+                Success
+              </Content>
+            </Flex>
```
mturley: So this little success indicator needs to either not exist (and be baked into the logo image we get from the backend) or we need to render it as an actual checkbox icon that floats over the corner of the logo as shown in the mockups, not the Success text we get here.

Right now in the logos we get from the backend they have the green checkmarks in them already, so let's just leave this out. I'll follow up on whether we want to open a separate issue for making that driven by the UI, in which case we'd want it to be an icon using absolute positioning CSS.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                tabIndex={-1}
+                isInline
+                style={{ fontSize: '1.25rem', fontWeight: 'bold' }}
+              >
+                {truncate ? (
+                  <span>{getModelName(model.name)}</span>
+                ) : (
+                  <span>{getModelName(model.name)}</span>
+                )}
+              </Button>
```
mturley: I don't think we needed this model name link to change. With this version the font size is bigger than the rest of the cards, and the `truncate` condition isn't doing anything. The existing code for this in ModelCatalogCard can stay as-is (using a Button with styles driven by CSS variables and using the Truncate component), but we can just move it unchanged up into the CardTitle to help with factoring out the CardBody to a child component as described above.

If moving that code into CardTitle breaks its styles, let's use care to fix that without explicit inline styles. We can rely on existing PF component props or CSS variables.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+            </Link>
+          </div>
+        </CardTitle>
+      </CardHeader>
+      <CardBody>
+        <div>
+          <div style={{ paddingLeft: '8px', marginBottom: '16px' }}>
```
mturley: Ok, the rest of my comments in this file apply to the code you'd be moving to the new ModelCatalogCardBody component and conditionally rendering only for validated models.

These nested divs aren't necessary (the outer one doesn't do anything), and we should avoid inline styles with arbitrary numbers. If you need to apply some extra margin/padding to the Flex child here, you can do it by applying [PatternFly utility classes](https://www.patternfly.org/utility-classes/spacing) to the `className` prop of the Flex component. Reference the [PF design foundations section on spacers](https://www.patternfly.org/design-foundations/spacers) to see how big each spacer size is (`sm` is 8px and `md` is 16px), and then look for the corresponding utility classes for adding those spacer sizes as padding-left (`pl`) and and margin-bottom (`mb`): `className="pf-v6-u-pl-sm pf-v6-u-mb-md"`

Actually -- scratch that, I'll leave it here for future reference, but instead of a <div>

Also, instead of using divs with explicit spacing to separate the accuracy area from the performance area, you can probably just use a [Stack](https://www.patternfly.org/layouts/stack) and pass it the `hasGutter` prop. You may also need some margin/padding using utility classes but start there, maybe you don't. If `hasGutter` introduces too much space you can still use a stack with utility classes.

You're also missing a horizontal divider between the accuracy and performance metrics areas. You can probably just [render a `Divider`](https://www.patternfly.org/components/divider) as one of the children of the Stack.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <Flex
+              alignItems={{ default: 'alignItemsCenter' }}
+              justifyContent={{ default: 'justifyContentSpaceBetween' }}
+            >
+              <Flex
+                alignItems={{ default: 'alignItemsCenter' }}
+                gap={{ default: 'gapSm' }}
+                wrap="nowrap"
+              >
+                <ChartBarIcon />
```
mturley: This is the wrong icon, use the `MonitoringIcon`. [Icon docs](https://www.patternfly.org/design-foundations/icons#all-icons)

Screenshot 2025-10-07 at 5 02 06 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                <span
+                  data-testid="validated-model-accuracy"
+                  style={{ fontSize: '1.25rem', fontWeight: 'bold' }}
```
mturley: The fontSize is unnecessary and makes the text too big, it seems to me to be the same size as other text on the card but just bold. You can make the text bold with `className="pf-v6-u-font-weight-bold"` instead of an inline style. [PF Text utility classes](https://www.patternfly.org/utility-classes/text)

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                shape="square"
+                width="56px"
+                height="56px"
+                screenreaderText="Brand image loading"
+                style={{ marginRight: '8px' }}
+              />
+            )}
+            <FlexItem align={{ default: 'alignRight' }}>
+              <Badge className="pf-m-purple">Validated</Badge>
+            </FlexItem>
```
mturley: Let's still use the Label component as the existing card does so they have the same shape/size, the only difference for validated models is we'll have it say Validated instead of the source name and be purple (Label has a `color="purple"` prop as well).

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                >
+                  {metrics.accuracy}%
+                </span>
+              </Flex>
+              <Flex
+                alignItems={{ default: 'alignItemsCenter' }}
+                gap={{ default: 'gapSm' }}
+                wrap="nowrap"
+              >
+                <span className="pf-v5-c-content pf-m-small">Average accuracy</span>
```
mturley: You can use `<Content component="small">` here instead of a span with classes.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                <Tooltip
+                  content="Average accuracy
+
+The weighted average of normalized scores from all benchmarks. Each benchmark is normalized to a 0-100 scale. All normalized benchmarks are then averaged together."
```
mturley: The whitespace in this content gets collapsed to a space. ~~You can pass JSX to the tooltip's content prop and use that to render this as a fragment `<></>` with some line breaks in it `<br /><br />` so "Average accuracy" appears separately.~~

Actually, looking closer at the mockups this should be a Popover (appears on click) and not a Tooltip, and "Average accuracy" should be used as the popover title. Same with the Latency help text.

Screenshot 2025-10-07 at 5 17 19 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                  position="top"
+                  className="pf-v6-c-tooltip"
+                >
+                  <HelpIcon style={{ cursor: 'help' }} />
```
mturley: Our other help popovers don't override the cursor, it should use the default.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                </Tooltip>
+              </Flex>
+            </Flex>
+          </div>
+
+          {/* Hardware Configuration and Latency */}
+          <div style={{ paddingLeft: '8px', marginBottom: '16px' }}>
```
mturley: As above, see what happens if you just don't use custom styles here and use Stack/StackItem to separate these vertical areas.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <Flex justifyContent={{ default: 'justifyContentSpaceBetween' }}>
+              <div>
+                <Content component={ContentVariants.small} style={{ color: '#6a6e73' }}>
+                  Hardware
+                </Content>
+                <Content component={ContentVariants.p} data-testid="validated-model-hardware">
+                  {metrics.hardwareCount}x{metrics.hardware}
+                </Content>
+              </div>
```
mturley: A couple things here, comparing this to the mockups.
- The subtitles "Hardware", "RPS/rep." and "TTFT" should appear below the metrics, not above.
- I think "p" is the wrong component to use for the metric, it puts too much whitespace between the caption and the metric. Maybe just don't put that part in a Content component, it can go in a span.
- The metric numbers should be bold, using `className="pf-v6-u-font-weight-bold"`.
- Avoid inline styles, especially hard coded colors. The captions here can just be the default color they get from ContentVariants.small, it's already slightly lighter than regular text.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+*ITL (Inter-Token Latency)
+The average time between successive output tokens after the model has started generating.
+
+*E2E (End-to-End latency)
+The total time from when the request is sent until the last token is received."
+                    position="top"
+                    className="pf-v6-c-tooltip"
+                  >
+                    <HelpIcon />
+                  </Tooltip>
```
mturley: Again here we need a Popover instead of a Tooltip, with "Latency" as the popover title and a real rendered bulleted list inside it with bold text where present on terms in the mockups.

Screenshot 2025-10-07 at 5 33 28 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                </Flex>
+              </div>
+            </Flex>
+          </div>
+
+          {/* Benchmarks */}
+          <div style={{ paddingLeft: '8px' }}>
```
mturley: Same here with the inline styles - see what Stack gets you and if needed use spacing utility classes.

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                  aria-label="Previous benchmark"
+                  data-testid="validated-model-benchmark-prev"
+                />
+                <Button
+                  variant="plain"
+                  icon={<ChevronRightIcon />}
+                  aria-label="Next benchmark"
+                  data-testid="validated-model-benchmark-next"
+                />
+              </Flex>
```
mturley: A couple things on the benchmarks switcher thing here

- You're hard-coding "1 of 3" and your arrow buttons don't do anything. I see now that you're passing in a single performance artifact. You need to use an array of performance artifacts, and you need to set up state for using these arrow buttons to toggle through which artifact is being used to show the hardware/RPS/TTFT details and show X of Y where X is the current artifact and Y is the total number of artifacts. I should be able to click through and get "2 of 3" and see different numbers from the second artifact, etc.
- Your text and buttons aren't vertically aligned (testing using the PF theme):

Screenshot 2025-10-07 at 5 43 28 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ValidatedModelCard.tsx
```diff
@@ -0,0 +1,247 @@
...
+                alignItems={{ default: 'alignItemsCenter' }}
+                gap={{ default: 'gapSm' }}
+                wrap="nowrap"
+              >
+                <ChartBarIcon />
+                <span
+                  data-testid="validated-model-accuracy"
+                  style={{ fontSize: '1.25rem', fontWeight: 'bold' }}
+                >
+                  {metrics.accuracy}%
```
mturley: Coming back to accuracy here (this comment will be out of order if you're reading the diff, sorry) - similarly to my comment on the performance artifacts, you need to use an array of accuracy artifacts - this is not the value from a single artifact, you need to average the accuracy value across those artifacts to get this average accuracy value.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogPage.tsx
```diff
@@ -91,7 +108,7 @@ const ModelCatalogPage: React.FC<ModelCatalogPageProps> = ({ searchTerm }) => {
           </Bullseye>
         </div>
       )}
-    </>
+    </Gallery>
```
mturley: You moved the "Load more models" button inside the Gallery, which makes it appear in the grid of cards instead of centered below them as it was before.

clients/ui/frontend/src/app/pages/modelCatalog/utils/validatedModelUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+export const extractValidatedModelMetrics = (
+  performanceMetrics?: CatalogPerformanceMetricsArtifact,
+  accuracyMetrics?: CatalogAccuracyMetricsArtifact,
+): ValidatedModelMetrics => ({
+  accuracy: accuracyMetrics?.customProperties.overall_average?.double_value || 53.9,
+  hardware: performanceMetrics?.customProperties.hardware?.string_value || '8xH100-80',
+  hardwareCount: performanceMetrics?.customProperties.hardware_count?.int_value || '8',
+  rpsPerReplica: performanceMetrics?.customProperties.requests_per_second?.double_value || 1,
+  ttftMean: performanceMetrics?.customProperties.ttft_mean?.double_value || 1428,
+});
```
mturley: This will need to be refactored a bit since you need to use multiple accuracy metrics artifacts (to average their accuracy) and multiple performance metrics artifacts (to show details from one at a time). I'd maybe split this utility in two (one for each artifact type) so you can use them independently. We should also add some basic unit tests for them.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCard.tsx
```diff
@@ -48,57 +62,48 @@ const ModelCatalogCard: React.FC<ModelCatalogCardProps> = ({ model, source, trun
...
+              {truncate ? (
+                <Truncate
+                  data-testid="model-catalog-card-name"
+                  content={model.name}
+                  position="middle"
+                  tooltipPosition="top"
+                  style={{ textDecoration: 'underline' }}
+                />
+              ) : (
+                <span data-testid="model-catalog-card-name">{model.name}</span>
```
mturley: You still need to use `getModelName(model.name)` as previously used here, which strips away the repository (e.g. `repo1/`) at the front of these names.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCard.tsx
```diff
@@ -10,33 +10,47 @@ import {
...
   const allLabels = model.customProperties ? getLabels(model.customProperties) : [];
   const validatedLabels = allLabels.includes('validated') ? ['validated'] : [];
+  const isValidated = isModelValidated(model);
 
   return (
     <Card isFullHeight data-testid="model-catalog-card" key={`${model.name}/${model.source_id}`}>
       <CardHeader>
         <CardTitle>
-          <Flex alignItems={{ default: 'alignItemsCenter' }}>
+          <Flex alignItems={{ default: 'alignItemsCenter' }} gap={{ default: 'gapSm' }}>
```
mturley: I didn't notice this before, but `alignItemsCenter` isn't correct here - the label should be top-aligned, so we want `alignItemsFlexStart`.

Also, I don't think the `gapSm` is doing anything here (it would put a gap between the children of the flex, which already aren't near each other because they're split across the header). The model name needs some space between it and the logo, which you can do by just adding className="pf-v6-u-mb-md" on this Flex.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+                  type="button"
+                  aria-label="More info for average accuracy"
+                  style={{
+                    all: 'unset',
+                    cursor: 'pointer',
+                    display: 'inline-flex',
+                  }}
+                >
+                  <HelpIcon />
+                </button>
```
mturley: To get the correct styles here you want to use a PF Button with the `icon` (instead of children) and `hasNoPadding` props, and `variant="plain"`.

```tsx
                  <Button
                    icon={<HelpIcon />}
                    hasNoPadding
                    aria-label="More info for average accuracy"
                    variant="plain"
                  />
```
This does what your inline styles here are doing plus a color change on hover.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+                    type="button"
+                    aria-label="More info for latency"
+                    style={{
+                      all: 'unset',
+                      cursor: 'pointer',
+                      display: 'inline-flex',
+                    }}
+                  >
+                    <HelpIcon />
+                  </button>
```
mturley: Same here, switch to the PF icon button.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+          <Flex
+            alignItems={{ default: 'alignItemsCenter' }}
+            justifyContent={{ default: 'justifyContentSpaceBetween' }}
+          >
+            <Content component={ContentVariants.p} data-testid="validated-model-benchmarks">
+              {currentPerformanceIndex + 1} of {performanceMetrics.length}{' '}
+              <Button variant="link" isInline style={{ padding: 0, fontSize: 'inherit' }}>
+                benchmarks
+              </Button>
+            </Content>
```
mturley: The "X of Y benchmarks" text is still not vertically aligned with the buttons. I think it might be because you're using ContentVariants.p here and paragraphs get some extra bottom margin on them. If you just replace `Content` here with a `span` it should render correctly. We don't really need the Content component unless we want non-default text styles (like your `small` text above) since the CardBody implies some text styles for its children.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+                variant="plain"
+                icon={<ChevronLeftIcon />}
+                aria-label="Previous benchmark"
+                data-testid="validated-model-benchmark-prev"
+                onClick={handlePreviousBenchmark}
+                isDisabled={performanceMetrics.length <= 1}
+              />
+              <Button
+                variant="plain"
+                icon={<ChevronRightIcon />}
```
mturley: I didn't notice this before but these are the wrong icons, they render larger than in the mockups. I think you want `AngleLeftIcon` and `AngleRightIcon`.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+                content={model.name}
+                position="middle"
+                tooltipPosition="top"
+                style={{ textDecoration: 'underline' }}
+              />
+            ) : (
+              <span>{getModelName(model.name)}</span>
+            )}
+          </Button>
+        </Link>
```
mturley: This model name link isn't necessary because you moved it up to the CardTitle. With this you get a duplicate. You can remove it, and that also means you don't need the Stack at all since you'll just be left with one StackItem which won't be necessary - you can just render the children of that remaining StackItem here instead.

Screenshot 2025-10-08 at 3 38 22 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -0,0 +1,262 @@
...
+      <StackItem isFilled data-testid="model-catalog-card-description">
+        {truncate ? (
+          <div
+            style={{
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+              WebkitLineClamp: 4,
+              WebkitBoxOrient: 'vertical',
+              display: '-webkit-box',
+            }}
```
mturley: Hmmm... I also notice that we're never actually passing in a true value for `truncate` - According to the designs I think we should always be truncating. Can we remove the truncate prop from this component and ModelCatalogCard and just always render this truncated?

As for the inline styles here, I can't remember if I already tried to find an alternative... looking now I don't see a good way to replace them so they can stay for now, but we may want to follow up with PF to ask for a way to achieve this with PF components.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogLabels.tsx
```diff
@@ -14,12 +14,13 @@ const ModelCatalogLabels: React.FC<ModelCatalogLabelsProps> = ({
   provider,
   labels = [],
 }) => (
-  <LabelGroup numLabels={5} isCompact>
+  <LabelGroup numLabels={2} isCompact>
```
mturley: Ok so (sorry) I was wrong when I said numLabels should be 2 for all models - for non-validated models it should apparently be 3. Can we lift out a numLabels prop from here and pass 2 for validated models and 3 for other models?

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogPage.tsx
```diff
@@ -67,29 +71,37 @@ const ModelCatalogPage: React.FC<ModelCatalogPageProps> = ({ searchTerm }) => {
...
-              </>
+              </Flex>
             ) : (
               <Button variant="tertiary" onClick={catalogModels.loadMore} size="lg">
                 Load more models
               </Button>
             )}
           </Bullseye>
-        </div>
+        </Flex>
```
mturley: There's now no gap between the button and the cards. Can we restore this to how it was before (just a bullseye with no flex inside or out) but just replace the old `<div style={{ marginTop: '2rem' }}>` with a `className="pf-v6-m-mt-lg"` on the Bullseye?

Screenshot 2025-10-09 at 12 15 09 PM

PR #1702, opened 2025-10-03
fix: align MR and MC route paths 

manaswinidas:
## Description

fix: align MR and MC route paths with internal structure improvements

- Update internal route paths to use kebab-case convention:
  * registeredModels -> registered-models  
  * registerModel -> register/model
  * registerVersion -> register/version
- Fix route path mismatches between URL generation functions and route definitions
- Update all Cypress test files to use new internal route structure

## How Has This Been Tested?

Go to different model registry and model catalog routes and check whether the pages load fine

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelDetailsCard.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelDetailsCard.cy.ts
@@ -104,7 +104,7 @@ describe('Model Details Card', () => {
   });
 
   it('displays model details correctly', () => {
-    cy.visit('/ai-hub/registry/modelregistry-sample/registered-models/1/overview');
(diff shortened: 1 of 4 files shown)


clients/ui/frontend/src/app/AppRoutes.tsx
```diff
@@ -32,15 +33,15 @@ export const useNavData = (): NavDataItem[] => {
   const baseNavItems = [
     {
       label: 'Model Registry',
-      path: '/model-registry',
+      path: '/ai-hub/registry',
```
lucferbux: This is fine for now but we should take a look at the istio rules cause I think this is gonna break kubeflow. Maybe we can add some TODOs

mturley: @manaswinidas I don't think we should need to change AppRoutes, in ODH these routes are not used and the MR+MC relative routes are directly imported. Maybe I was wrong about that? I don't think we should need to make the "ai-hub" naming move upstream.

PR #1719, opened 2025-10-07
All models view

ppadti:
## Description
This PR aims to add all models view with categories sections and also will have show all models link to view the indivisual section

https://github.com/user-attachments/assets/e8101973-7290-427b-a22d-c2c4792e65bb

## How Has This Been Tested?
- go to model catalog, you will be in all models view by default
- you will have each section with 4 cards and each section will have show all models link to view individual section and also can view it by clicking on each of the toggle
- make sure the search will work as expected

## Merge criteria:

If you have UI changes


ppadti:
As I removed the sorting from the categories, the cypress tests are failing now - i will fix that.

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
```diff
@@ -0,0 +1,117 @@
...
+              <Button
+                variant="link"
+                size="sm"
+                isInline
+                icon={<ArrowRightIcon />}
+                iconPosition="right"
+                data-testid={`show-more-button-${label.toLowerCase().replace(/\s+/g, '-')}`}
+                onClick={() => handleShowMoreCategory(label)}
+              >
+                Show all {displayName ?? label} models
```
ppadti, line 78: when we click on this link - there is still a scroll issue that, the page won't scroll to top - will look into this

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
```diff
@@ -0,0 +1,117 @@
...
+    pageSize,
+    searchTerm,
+  );
+
+  const handleShowMoreCategory = (categoryLabel: string) => {
+    onShowMore(categoryLabel, true);
+  };
+
+  return (
+    <>
```
rsun19, line 54: `<StackItem>` but no `<Stack>`? We can use `<Stack hasGutter>` for the spacing too

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
rsun19: Hmmm, it seems to work on my machine

rsun19 (commented):
At first glance it looks good, I haven't taken a deep look into the functions yet though

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
ppadti: we have `<Stack hasGutter>` in parent component.

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
```diff
@@ -0,0 +1,117 @@
...
+                data-testid={`show-more-button-${label.toLowerCase().replace(/\s+/g, '-')}`}
+                onClick={() => handleShowMoreCategory(label)}
+              >
+                Show all {displayName ?? label} models
+              </Button>
+            </FlexItem>
+          )}
+        </Flex>
+      </StackItem>
+      <StackItem style={{ paddingBottom: '20px' }}>
```
rsun19:
```suggestion
      <StackItem className="pf-v6-u-pb-md">
```
Can we use patternfly styling here

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -32,14 +40,17 @@ const ModelCatalog: React.FC = () => {
...
-            <ModelCatalogSourceSelectorNavigator
-              getRedirectPath={(sourceId: string) => modelCatalogUrl(sourceId)}
+            <ModelCatalogSourceLabelSelectorNavigator
               searchTerm={searchTerm}
               onSearch={handleSearch}
               onClearSearch={handleClearSearch}
             />
-            <PageSection isFilled>
-              <ModelCatalogPage searchTerm={searchTerm} />
+            <PageSection isFilled style={{ paddingLeft: '0px', paddingTop: '25px' }}>
```
rsun19: Same here, wondering if patternfly classes can replace this

mturley: +1, you should be able to use [PF spacing utility classes](https://www.patternfly.org/utility-classes/spacing) here. [The `lg` spacer is 24px](https://www.patternfly.org/design-foundations/spacers)

clients/ui/frontend/src/app/pages/modelCatalog/ModelCatalogCoreLoader.tsx
```diff
@@ -9,54 +9,47 @@ import {
...
+    if (
+      catalogSourcesLoaded &&
+      catalogSources?.items &&
+      catalogSources.items.length > 0 &&
+      !selectedSourceLabel
+    ) {
+      updateSelectedSourceLabel('All models');
     }
-  }, [modelCatalogFromRoute, updateSelectedSource, selectedSource]);
+  }, [catalogSourcesLoaded, catalogSources, selectedSourceLabel, updateSelectedSourceLabel]);
```
mturley: Maybe we don't need this useEffect if we just have the state default to "All models"? Also instead of hard-coding the "All models" string everywhere can we create an enum of our aggregate source label strings? Will probably just be this one and then `"null"` (which should replace `"Other"`, see my later comment)

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
mturley: I am seeing this, but only if I make my window height very small. I suspect @rsun19 isn't seeing it because navigating to one of the label pages shows only the first 10 models, which on a big enough screen doesn't take the full height of the browser, so it seems to scroll up.

We could put a `<ScrollViewOnMount shouldScroll />` above the Gallery on `ModelCatalogPage` since it isn't mounted while you're on the All models tab now. Speaking of which, since `ModelCatalogPage` isn't the top level view anymore maybe we should rename it to ModelCatalogGalleryView or something?

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelBlocks.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const labelBlocks: SourceLabelBlock[] = uniqueLabels.map((label) => ({
+      id: `label-${label}`,
+      label,
+      displayName: label,
+    }));
+
+    const noLabelsBlock: SourceLabelBlock = {
+      id: 'no-labels',
+      label: 'Other',
```
mturley: So if I'm reading this correctly we'll end up passing `sourceLabel="Other"` to the API when this is selected. I think we don't want that - if there are sources with that label it will have already appeared as one of the labelBlocks above, so this tab would end up as a duplicate of that.

I think what we want [based on this slack thread](https://redhat-internal.slack.com/archives/C08G31WCV16/p1760043597708769?thread_ts=1759329274.975099&cid=C08G31WCV16) is to use `"null"` as the sourceLabel value for this tab.

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <Button
+                variant="link"
+                size="sm"
+                isInline
+                icon={<ArrowRightIcon />}
+                iconPosition="right"
+                data-testid={`show-more-button ${label.toLowerCase().replace(/\s+/g, '-')}`}
+                onClick={() => handleShowMoreCategory(label)}
+              >
+                Show all {displayName ?? label} models
```
mturley, line 78: On larger screens, this "Show all ___ models" link is appearing way off to the side and the model cards are not taking the full width. Can we maybe combine these two StackItems into one and have its width fit its content, so that the link is always right-aligned with the right side of the cards? Either that or we could make sure the cards always stretch to fill the width of the screen but that might not be desirable.

If either is difficult don't worry about it and we can follow up.

Screenshot 2025-10-09 at 5 44 45 PM

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -17,6 +24,18 @@ const ModelCatalog: React.FC = () => {
     setSearchTerm('');
   }, []);
 
+  const resetAllFilters = React.useCallback(() => {
+    setFilterData(ModelCatalogStringFilterKey.TASK, []);
+    setFilterData(ModelCatalogStringFilterKey.PROVIDER, []);
+    setFilterData(ModelCatalogStringFilterKey.LICENSE, []);
+    setFilterData(ModelCatalogStringFilterKey.LANGUAGE, []);
```
mturley: Can we loop over the `Object.values(ModelCatalogStringFilterKey)` to set all these so we don't have to maintain the list here?

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -195,3 +196,37 @@ export const filtersToFilterQuery = (
...
+): CatalogSource | undefined => {
+  if (!catalogSources || !sourceId) {
+    return undefined;
+  }
+
+  return catalogSources.items.find((source) => source.id === sourceId);
+};
+
+export const hasFiltersApplied = (filterData: ModelCatalogFilterStates): boolean =>
+  Object.values(filterData).some((value) => Array.isArray(value) && value.length > 0);
```
mturley: Eventually we'll have values here that aren't arrays, we can have number filters supported by our types already (even though we don't have any here yet). So I think this should still return true for a filter value if it's defined but not an array. (not just truthy though - a number value of 0 is still an applied filter)

clients/ui/frontend/src/app/routes/modelCatalog/catalogModel.ts
```diff
@@ -1,7 +1,6 @@
 import { getCatalogModelDetailsRoute } from '~/app/routes/modelCatalog/catalogModelDetails';
 
-export const modelCatalogUrl = (sourceId?: string): string =>
-  `/model-catalog${sourceId ? `/${sourceId}` : ''}`;
```
mturley, line 5: Hmm, I think we still need this as it was - if you need just the root `/model-catalog` you can just not pass in a sourceId.

clients/ui/frontend/src/app/routes/modelCatalog/catalogModelDetails.ts
```diff
@@ -1,11 +1,10 @@
...
 import { encodeParams } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
-import { modelCatalogUrl } from './catalogModel';
 
 export const getCatalogModelDetailsRoute = (params: CatalogModelDetailsParams): string => {
   const { sourceId = '', modelName = '' } = encodeParams({
     sourceId: params.sourceId,
     modelName: params.modelName,
   });
-  return `${modelCatalogUrl(sourceId)}/${modelName}` || '#';
+  return `/model-catalog/${sourceId}/${modelName}` || '#';
```
mturley: We can't hard-code these paths, we need to still use the util so we can override it downstream.

clients/ui/frontend/src/app/routes/modelCatalog/catalogModelRegister.ts
```diff
@@ -1,10 +1,9 @@
 import { encodeParams } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
-import { modelCatalogUrl } from './catalogModel';
 
 export const getRegisterCatalogModelRoute = (id = '', name = ''): string => {
   const { sourceId = '', modelName = '' } = encodeParams({
     sourceId: id,
     modelName: name,
   });
-  return `${modelCatalogUrl(sourceId)}/${modelName}/register` || '#';
+  return `/model-catalog/${sourceId}/${modelName}/register` || '#';
```
mturley: We can't hard-code these paths, we need to still use the util so we can override it downstream.

ppadti:
> We could put a <ScrollViewOnMount shouldScroll /> above the Gallery on ModelCatalogPage since it isn't mounted while you're on the All models tab now.

If we add that to ModelCatalogPage - it will only scroll upto the first card and search bar and toggle for sourceLabel will still be hidden from the view, like in  the screenshot. I tried this before, got this result. I will look into it.
Screenshot 2025-10-10 at 3 29 13 PM

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
ppadti: +1, it doesn't work with smaller screen - adding <ScrollViewOnMount shouldScroll /> to ModelCatalogPage will lead to this https://github.com/kubeflow/model-registry/pull/1719#issuecomment-3389157486

PR #1725, opened 2025-10-09
Add filter toolbar for hardware configuration table

manaswinidas:
## Description

Added Workload type, Latency, RPS and Hardware type filters for hardware configuration table.
Note: Hardware type and RPS filters are working - investigating the other filters(Workload type and Max latency are hardcoded as of now and the filters don't work)
Chip toolbar hasn't been added but the Clear all filters functionality has been added
Need to add unit tests

https://github.com/user-attachments/assets/217e58ff-39c6-4c6e-a871-c11b910c1409

## How Has This Been Tested?

Go to Performance insights of first model of Model catalog and play around with the filters

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -6,7 +6,10 @@ export enum ModelCatalogStringFilterKey {
 }
 
 export enum ModelCatalogNumberFilterKey {
-  // TODO add number filters here
+  MIN_RPS = 'rps_mean',
+  MAX_LATENCY = 'ttft_mean',
+  WORKLOAD_TYPE = 'workload_type',
+  HARDWARE_TYPE = 'hardware',
```
mturley: According to [the ADR (see "Performance Metric Properties"](https://docs.google.com/document/d/12wGInEpLsPS4P8XiczgQwYhJavn5fAUkt4XQTyAVLns/edit?tab=t.0#heading=h.40y985s08lwd) these two aren't real filters - well, workload_type isn't and hardware is a string filter.

I think what we need here is keys based on the properties we're filtering by which isn't going to match exactly 1:1 with the way these ones are presented in the UI. So for the workload type filter, really we are hard-coding those 4 options and filtering on the numbers in them (maxInputTokens and maxOutputTokens), and we'll need to translate the selected dropdown option into logic that filters by both of those. The max latency filter is a little more complex - we'll need to use the combination of the user's selected Metric and Percentile to determine which number we're filtering by (e.g. if they selected E2E and P90, we're filtering by `e2e_p90`, ~~and we'll need to use that key to pull out the min/max numbers for the slider from the filter options from the API~~ (edit: see my comment below this review - we can't get min/max from the API for these yet). When the user changes that selection we need to clear the old filter and move its new value to the new filter key - and in the process make sure it's always between the new min and max - if the min/max of the new filter key are different and the current value is out of range change it to the new min or max (nearest). So we'll need our types and mocks etc to have all those permutations of metric and percentile added as separate keys 😅 

One thing I want to avoid is having to sync state between "what dropdown item is selected" and "what filters are present in the filter values lifted into context". I wonder if we can entirely use the filter values in context as the source of truth. i.e. instead of state for which dropdown item is selected, we have a function that determines which one is selected based on what we know about the filter values that will match - if the filter state has maxInputTokens: 2048 and maxOutputTokens: 256, we know the UI should show the "Summarization" workload type as selected. Conversely, when the user clicks to select a different workload type we can just update those two numbers and that'll cause the dropdown to update. So for each dropdown we'd need utilities to convert a partial filter state object to a dropdown item and vice versa (or in the case of max latency, converting a partial filter state object to both the metric and percentile dropdown items, and converting back from the pair of those to the filter key we want to use). If this is too much of a pain we could sync the two with a useEffect but that feels like asking for trouble. We'll need fallback behavior for weird states (what happens if there are filters on two of the max latency filter properties? That's invalid and shouldn't be possible to encounter from our dropdowns, so maybe we log a warning and just default to the first one present or something. Good unit tests for those will be good.

Also it looks like when filtering, we'll have `rpsMean` and `ttftMean` rather than the snake_case versions... seems inconsistent with the properties on the objects, I'll start a thread to ask the MR team...

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+ * Clears all active filters
+ */
+export const clearAllFilters = (
+  setFilterData: <K extends keyof ModelCatalogFilterStates>(
+    key: K,
+    value: ModelCatalogFilterStates[K],
+  ) => void,
+): void => {
+  // Clear hardware filters
+  setFilterData(ModelCatalogStringFilterKey.PROVIDER, []);
```
mturley: Provider isn't one of the hardware filters, that's one of the model filters e.g. Meta or IBM.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  // Clear number filters
+  setFilterData(ModelCatalogNumberFilterKey.MIN_RPS, undefined);
+  setFilterData(ModelCatalogNumberFilterKey.MAX_LATENCY, undefined);
+  setFilterData(ModelCatalogNumberFilterKey.WORKLOAD_TYPE, undefined);
```
mturley: I think maybe we can just loop over the `Object.values(ModelCatalogNumberFilterKey)` to reset them all.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareTypeFilterState.ts
```diff
@@ -0,0 +1,16 @@
...
+  setAppliedHardwareTypes: React.Dispatch<React.SetStateAction<string[]>>;
+  clearHardwareFilters: () => void;
+} => {
+  const [appliedHardwareTypes, setAppliedHardwareTypes] = React.useState<string[]>([]);
+
+  return {
+    appliedHardwareTypes,
+    setAppliedHardwareTypes,
+    clearHardwareFilters: () => setAppliedHardwareTypes([]),
+  };
```
mturley, line 28: Ideally we can also drive this from the central filter state object. You could just use `useCatalogStringFilterState(ModelCatalogStringFilterKey.hardware)`.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Toolbar clearAllFilters={handleClearAllFilters} clearFiltersButtonText="Reset all filters">
+      <ToolbarContent>
+        <ToolbarGroup>
+          <ToolbarItem>
+            <WorkloadTypeFilter filterOptions={filters} />
+          </ToolbarItem>
+          <ToolbarItem
+            style={{ borderLeft: '1px solid #d2d2d2', paddingLeft: '16px', marginLeft: '16px' }}
+          >
```
mturley: Can we try to avoid inline styles? We should be able to use `<ToolbarItem variant="separator" />` here. [Toolbar docs](https://www.patternfly.org/components/toolbar)

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <MaxLatencyFilter filterOptions={filterOptions} />
+          </ToolbarItem>
+          <ToolbarItem>
+            <MinRpsFilter filterOptions={filterOptions} />
+          </ToolbarItem>
+          <ToolbarItem
+            style={{ borderLeft: '1px solid #d2d2d2', paddingLeft: '16px', marginLeft: '16px' }}
+          >
```
mturley: Same here

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    // Hardware Type Filter (using dedicated hardware filter state)
+    if (appliedHardwareTypes && appliedHardwareTypes.length > 0) {
+      const hardwareType = getStringValue(artifact.customProperties, 'hardware');
+      if (!hardwareType || !appliedHardwareTypes.includes(hardwareType)) {
+        return false;
+      }
+    }
+
+    // Min RPS Filter
+    const minRpsFilter = filterState[ModelCatalogNumberFilterKey.MIN_RPS];
```
mturley, line 90: Ugh ok this one is weird... `ModelCatalogNumberFilterKey` should match the filter options we get from the API, but I don't think there is one for "min RPS" or "total RPS"... we may have to adjust this after we know what we're getting exactly. I'll check with the MR team. but this is a weird case where we indeed need one number in our state that we filter with a combined value from the artifact... maybe it's fine for now for this to be a filter key in our enum that isn't present in the API filter options... later when we want to go and apply this as a filter on the models list we'll need to ask for it to be a single key anyway.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (minRpsFilter !== undefined) {
+      const totalRps = getTotalRps(artifact.customProperties);
+      if (totalRps < minRpsFilter) {
+        return false;
+      }
+    }
+
+    // Max Latency Filter (enhanced with metric and percentile support)
+    const maxLatencyFilter = filterState[ModelCatalogNumberFilterKey.MAX_LATENCY];
```
mturley, line 101: Yeah here's a case where we won't have a single max_latency filter key in state but instead need to look at all the possible filter keys - when applying the filter here actually it'll be pretty simple, because you can just filter by whichever one is present in the filterState (or all of them that are present there really since this function doesn't need to care that our UI will only ever have one of them in state).

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+        value: maxLatencyFilter,
+      };
+
+      if (!applyMaxLatencyFilter(artifact, defaultConfig)) {
+        return false;
+      }
+    }
+
+    // Workload Type Filter (Task-based, would need to be enhanced with real workload mapping)
+    const workloadTypeFilter = filterState[ModelCatalogNumberFilterKey.WORKLOAD_TYPE];
```
mturley: As mentioned before this one will need to be based on a combination of the maxInputTokens and maxOutputTokens. Also, I think actually it should treat the options as minimums when applying the filter - if you selected RAG (4096 input, 512 output tokens), you're saying "show me artifacts whose maxInputTokens and maxOutputTokens are at least 4096 and 512". we're looking for a minimum threshold of that "max value the hardware is capable of".

We'll want to adjust our mock artifacts so Max output tokens are higher numbers, right now they're in the single digits which isn't realistic and won't ever match these filters.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+      // This could be enhanced when workload type is available in the artifact data
+    }
+
+    return true;
+  });
+
+/**
+ * Gets the count of active filters
+ */
+export const getActiveFilterCount = (filterState: ModelCatalogFilterStates): number => {
```
mturley: It looks like we never use this? If we do need it, we could maybe do it by looping over enum values (splitting our existing enum values so we have a subset of them that are hardware configuration filter keys (which would probably need to be further split to string and number ones so we can still include them in ModelCatalogStringFilterKey and ModelCatalogNumberFilterKey).

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/HardwareTypeFilter.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }, [performanceArtifacts]);
+
+  const selectedCount = localSelectedHardwareTypes.length;
+
+  // Initialize local state from applied state when dropdown opens
+  React.useEffect(() => {
+    if (isOpen) {
+      setLocalSelectedHardwareTypes([...appliedHardwareTypes]);
+    }
+  }, [isOpen, appliedHardwareTypes]);
```
mturley: Hmm, looking at the mockups it doesn't have an Apply/Reset button on this one, I think we can just update the real applied state when you toggle the checkboxes and get rid of the local state.

Screenshot 2025-10-09 at 1 57 05 PM

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
```diff
@@ -0,0 +1,309 @@
...
+                    <span>Metric: {localFilter.metric}</span>
+                  </MenuToggle>
+                )}
+              >
+                <SelectList>
+                  {METRIC_OPTIONS.map((option) => (
+                    <SelectOption key={option.value} value={option.value}>
+                      {option.label}
+                    </SelectOption>
+                  ))}
```
mturley: I think we need to filter the available options by which properties are present in the filterOptions from the API. If they suddenly stop giving us all the `ttft_*` options for example we shouldn't show TTFT as an option. We can maybe somehow drive this from the same mapping we use to convert between the filter keys and the pairs of dropdown options?

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
```diff
@@ -0,0 +1,309 @@
...
+        <Flex spaceItems={{ default: 'spaceItemsSm' }}>
+          <FlexItem>
+            <Button variant="primary" onClick={handleApplyFilter}>
+              Apply filter
+            </Button>
+          </FlexItem>
+          <FlexItem>
+            <Button variant="secondary" onClick={handleReset}>
+              Reset
+            </Button>
```
mturley: The reset buttons need to be `variant="link"` so they don't have a border, based on the mockups

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
```diff
@@ -0,0 +1,309 @@
...
+    <Dropdown
+      isOpen={isOpen}
+      onOpenChange={setIsOpen}
+      toggle={toggle}
+      shouldFocusToggleOnSelect={false}
+    >
+      <div
+        onMouseDown={(e) => e.stopPropagation()}
+        onClick={(e) => e.stopPropagation()}
+        onKeyDown={(e) => e.stopPropagation()}
```
mturley: heh, dropdowns in dropdowns...

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MinRpsFilter.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const handleReset = () => {
+    setSavedFilterValue(undefined);
+    setLocalValue(1);
+    setIsOpen(false);
+  };
+
+  // Get min/max values from filter options or use defaults
+  // TODO: Use real min/max values when available from API
+  const minValue = 1; // Default minimum value
+  const maxValue = 300; // Default maximum value
```
mturley: Since there currently isn't a filter from the API for this and we're rendering our max RPS by doing math, I think we need to do the same math here to get the min and max. we can map over all the artifacts and use the existing `getTotalRps` util to convert them to an array of those numbers and then `Math.min(...array)` and `Math.max(...array)`, probably memoized.

I do think we should leave the TODO here though.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MinRpsFilter.tsx
```diff
@@ -0,0 +1,156 @@
...
+} from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import { ModelCatalogNumberFilterKey } from '~/concepts/modelCatalog/const';
+import { CatalogFilterOptionsList } from '~/app/modelCatalogTypes';
+import { useCatalogNumberFilterState } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+
+const filterKey = ModelCatalogNumberFilterKey.MIN_RPS;
+
+type MinRpsFilterProps = {
+  filterOptions?: CatalogFilterOptionsList | null;
```
mturley: Looks like you don't need this prop since it isn't used

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MinRpsFilter.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <Flex spaceItems={{ default: 'spaceItemsSm' }}>
+          <FlexItem>
+            <Button variant="primary" onClick={handleApplyFilter}>
+              Apply filter
+            </Button>
+          </FlexItem>
+          <FlexItem>
+            <Button variant="secondary" onClick={handleReset}>
+              Reset
+            </Button>
```
mturley: `variant="link" on this one too`

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+  Select,
+  SelectList,
+  SelectOption,
+} from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import { ModelCatalogNumberFilterKey } from '~/concepts/modelCatalog/const';
+import { useCatalogNumberFilterState } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+import { CatalogFilterOptions } from '~/app/modelCatalogTypes';
+
+const filterKey = ModelCatalogNumberFilterKey.WORKLOAD_TYPE;
```
mturley: Ok yeah so there isn't a workload type filter key in the API's filterOptions, so I do think we want to drive this from a combination of the maxInputTokens and maxOutputTokens (or `max_input_tokens` and `max_input_tokens` is probably what we really want since it matches the artifact data.... sorry for being inconsistent in these comments, let's use snake_case for all of these and I'll follow up with the team)

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+type WorkloadTypeFilterProps = {
+  filterOptions?: CatalogFilterOptions | null;
```
mturley: Looks like you're not using this prop and it can go. Honestly I don't really think these filter options can really come from the API even in the future, unless we ask them for a true top level workload type property/filter.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+    value: 'code_generation',
+    label: 'Code Generation (1024 input | 512 output tokens)',
+    description: 'Code generation and completion tasks',
+  },
+];
+
+const WorkloadTypeFilter: React.FC<WorkloadTypeFilterProps> = () => {
+  const { value: appliedValue, setValue } = useCatalogNumberFilterState(filterKey);
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [localSelectedValue, setLocalSelectedValue] = React.useState<string | undefined>(undefined);
```
mturley: Let's not hold local state here, and instead use two `useCatalogNumberFilterState` calls for the max input and output tokens numbers, and then derive the selected option from what those numbers are. we'll need a mapping of the option values to each pair of those numbers, maybe we lift out an enum for WorkloadTypeOptionValue.

Utils like `maxInputOutputTokensToWorkloadType` and `workloadTypeToMaxInputOutputTokens` might be worthwhile. I think if the values in state don't exactly match one of our dropdown items we can just fall back to showing nothing selected, and that should be an unreachable state.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const workloadOptions = WORKLOAD_TYPE_OPTIONS;
+
+  // Initialize local state from applied state when dropdown opens
+  React.useEffect(() => {
+    if (isOpen) {
+      const applied = workloadOptions.find((opt) => opt.value === appliedValue?.toString());
+      setLocalSelectedValue(applied?.value);
+    }
+  }, [isOpen, appliedValue, workloadOptions]);
```
mturley: Yeah, no local state will mean we need no useEffect here.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const appliedOption = workloadOptions.find((option) => option.value === appliedValue?.toString());
+  const displayText = appliedOption ? appliedOption.label : 'Select workload type';
```
mturley: Interesting... looks like we won't have a default here, but there's no way to reset this filter aside from the "reset all filters" that only appears if there's no result. I think we may need to revisit that with UX, but the workaround is to reload the page so I'm not worried about for now.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const toggle = (toggleRef: React.Ref<MenuToggleElement>) => (
+    <MenuToggle
+      ref={toggleRef}
+      onClick={() => setIsOpen(!isOpen)}
+      isExpanded={isOpen}
+      style={{ minWidth: '300px', width: 'fit-content' }}
+    >
+      <span style={{ marginRight: '8px' }}>Workload type:</span>
```
mturley: Hm, I don't see a good alternative to the minWidth and width styles here but can we use a utility class for the marginRight instead of inline styles? `className="pf-v6-u-mr-sm"` should give 8px

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+          // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+          setLocalSelectedValue(selectedVal as string);
+          // Apply immediately when selecting from dropdown
+          const selectedOption = workloadOptions.find((opt) => opt.value === selectedVal);
+          setValue(
+            selectedOption
+              ? // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+                (selectedVal as unknown as number)
+              : undefined,
+          );
```
mturley: Yeah if we drive this from the pair of number states as the only source of truth we don't have to duplicate this

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,132 @@
...
+        bodyContent="Select a workload type to view performance under specific input and output token lengths."
+        appendTo={() => document.body}
+      >
+        <Button
+          variant="plain"
+          aria-label="More info for workload type"
+          onClick={(e) => e.stopPropagation()}
+        >
+          <HelpIcon />
+        </Button>
```
mturley: I think we want to use the `icon` prop instead of passing the icon as children. That should give a color change hover effect on the icon. Same for the other two help popovers in the latency/RPS dropdowns too.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -115,22 +121,34 @@ describe('filtersToFilterQuery', () => {
...
+          max: 10,
+        },
+      },
+      [ModelCatalogNumberFilterKey.HARDWARE_TYPE]: {
+        type: 'number',
+        range: {
+          min: 0,
+          max: 10,
+        },
+      },
```
mturley: yeah it doesn't make sense for these to be number filters

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    'e2e_p99',
+    'tps_mean',
+    'tps_p90',
+    'tps_p95',
+    'tps_p99',
+    'itl_mean',
+    'itl_p90',
+    'itl_p95',
+    'itl_p99',
+  ];
```
mturley: We can derive these values from the ModelCatalogNumberFilterKey enum once we add them there

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/**
+ * Extracts unique hardware types from performance artifacts
+ */
+export const getUniqueHardwareTypes = (
+  artifacts: CatalogPerformanceMetricsArtifact[],
+): string[] => {
+  const hardwareTypes = artifacts
+    .map((artifact) => getStringValue(artifact.customProperties, 'hardware'))
```
mturley: we can put hardware in as a ModelCatalogStringFilterKey and use that instead of hard coding it

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+ */
+export const filterHardwareConfigurationArtifacts = (
+  artifacts: CatalogPerformanceMetricsArtifact[],
+  filterState: ModelCatalogFilterStates,
+  appliedHardwareTypes?: string[],
+): CatalogPerformanceMetricsArtifact[] =>
+  artifacts.filter((artifact) => {
+    // Hardware Type Filter (using dedicated hardware filter state)
+    if (appliedHardwareTypes && appliedHardwareTypes.length > 0) {
+      const hardwareType = getStringValue(artifact.customProperties, 'hardware');
```
mturley: same here for the hardware string

mturley:
@manaswinidas - [from this slack thread](https://redhat-internal.slack.com/archives/C08G31WCV16/p1760037354425149) it looks like:
* these will all be `snake_case`, so please ignore me using camelCase in some of my review comments
* we won't have these numeric filters in the API's filter_options for the initial version. So let's still convert to using filter state keys based on the individual properties we're filtering by (so that this code is future proof since we do eventually want to use it with API filtering), but for this PR let's not try to use `flterOptions` at all. Instead we can calculate the min/max values by looking at the lowest and highest values present in the data.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    // Get all latency values for the currently selected metric/percentile
+    const fieldName = getLatencyFieldName(appliedFilter.metric, appliedFilter.percentile);
+    const latencyValues = performanceArtifacts
+      .map((artifact) =>
+        getDoubleValue(
+          artifact.customProperties,
+          // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+          fieldName as keyof typeof artifact.customProperties,
```
mturley, line 175: We could probably narrow the return type of getLatencyFieldName so that this assertion isn't necessary

Edit: ah, we can just get rid of the getLatencyFieldName util in here and export the one from hardwareConfigurationFilterUtils instead which is more narrow

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -0,0 +1,100 @@
...
+  );
+
+  return (
+    <>
+      <Select
+        isOpen={isOpen}
+        selected={selectedWorkloadType}
+        onSelect={(_, selectedVal) => {
+          // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+          const workloadType = selectedVal as WorkloadTypeOptionValue;
```
mturley, line 58: We can probably use a type guard here instead of an assertion

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,153 @@
...
+  const percentileSuffix = percentile === 'Mean' ? '_mean' : `_${percentile.toLowerCase()}`;
+  const fieldName = `${metricPrefix}${percentileSuffix}`;
+
+  // Validate that the field exists in PerformanceMetricsCustomProperties
+  // Get latency fields from LatencyMetricKey enum
+  const validFields = Object.values(LatencyMetricKey);
+
+  return validFields.some((field) => field === fieldName)
+    ? // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      (fieldName as keyof PerformanceMetricsCustomProperties)
```
mturley, line 38: Ok I see - yeah this may be harder to avoid, will revisit this. some combination of type guards and more narrow param types should help.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (minRpsFilter !== undefined) {
+      const totalRps = getTotalRps(artifact.customProperties);
+      if (totalRps < minRpsFilter) {
+        return false;
+      }
+    }
+
+    // Max Latency Filter (using MAX_LATENCY key for now)
+    // TODO: This currently uses TTFT Mean as default - should be enhanced to use
+    // the specific metric/percentile selected by the user
```
mturley, line 100: @manaswinidas I also missed this - editing my review comment above, seems like we need to fix this?

PR #1740, opened 2025-10-13
Performance artifact bff call

rsun19:
## Description

Added the mock bff call to the model catalog card body so that metrics can be shown on the card.

## How Has This Been Tested?

Go to the model catalog page and observe the metrics on the validated card. You can also test it by adjusting the mocks in `model_catalog_client_mock.go`.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
@@ -242,6 +242,10 @@ describe('ModelCatalogCard Component', () => {
         modelCatalog.findValidatedModelHardware().should('contain.text', '40xA100');
         modelCatalog.findValidatedModelRps().should('contain.text', '15');
         modelCatalog.findValidatedModelTtft().should('contain.text', '42.123791232');
+        modelCatalog.findValidatedModelBenchmarkPrev().click();
+        modelCatalog.findValidatedModelHardware().should('contain.text', '33xRTX 4090');
+        modelCatalog.findValidatedModelRps().should('contain.text', '10');
+        modelCatalog.findValidatedModelTtft().should('contain.text', '67.14892749816');
         modelCatalog.findValidatedModelBenchmarkLink().click();
         cy.url().should('include', 'performance-insights');
       });
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelCatalog.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelCatalog.ts
@@ -119,6 +119,10 @@ class ModelCatalog {
     return cy.findAllByTestId('validated-model-benchmark-next');
   }
 
+  findValidatedModelBenchmarkPrev() {
+    return cy.findAllByTestId('validated-model-benchmark-prev');
+  }
+
   findModelCatalogDescription() {
     return cy.findByTestId('model-catalog-card-description');
   }


clients/ui/bff/internal/mocks/model_catalog_client_mock.go
```diff
@@ -171,7 +171,15 @@ func (m *ModelCatalogClientMock) GetCatalogModelArtifacts(client httpclient.HTTP
 	var allMockModelArtifacts models.CatalogModelArtifactList
 
 	if sourceId == "sample-source" && modelName == "repo1%2Fgranite-8b-code-instruct" {
-		allMockModelArtifacts = GetCatalogPerformanceMetricsArtifactListMock()
+		performanceArtifacts := GetCatalogPerformanceMetricsArtifactListMock()
```
mturley: Can you edit `GetCatalogPerformanceMetricsArtifactListMock` so it returns more than one perf artifact with some different numbers, so that in mock mode we have the switcher behavior in the card?

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -58,6 +61,24 @@ const initIntercepts = ({
...
+    {
+      path: {
+        apiVersion: MODEL_CATALOG_API_VERSION,
+        sourceId: 'sample-source',
+        modelName: 'repo1/model1',
+      },
+    },
+    {
+      items: [
+        mockCatalogPerformanceMetricsArtifact({}),
```
mturley, line 76: Similarly can you have this intercept return more than one perf artifact? we should probably add a test to make sure the switcher works but that's out of scope here, we may just as well fix the mock data while we're adding it.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
```diff
@@ -105,6 +108,24 @@ const initIntercepts = ({
...
+    {
+      path: {
+        apiVersion: MODEL_CATALOG_API_VERSION,
+        sourceId: 'source-2',
+        modelName: 'validated-model',
+      },
+    },
+    {
+      items: [
+        mockCatalogPerformanceMetricsArtifact({}),
```
mturley, line 124: Same here

clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogModelArtifacts.ts
```diff
@@ -20,9 +21,12 @@ export const useCatalogModelArtifacts = (
       if (!modelName) {
         return Promise.reject(new NotReadyError('No model name'));
       }
+      if (isValidated === false) {
+        return Promise.reject(new NotReadyError('Model is not validated'));
+      }
```
mturley: Hmm.. my initial read of this was going to be "this will block us from fetching the model URI on the details page for non-validated models" but it seems it doesn't simply because of the `=== false`. I wonder if there's a way to make it more readable. I know it's not strictly necessary but can we maybe have an additional boolean `onlyFetchIfValidated` that defaults to false?

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
rsun19: Hmmm I did not add it here because we currently do not have any tests that explicitly would test for it, but I added a switcher test to `modelCatalogCard.cy.ts` instead

mturley: Makes sense, sounds good

Issue #1748, opened 2025-10-15
Code gen is broken on MacOS


pboyd:
**Describe the bug**
The code generation is apparently not producing the same output between MacOS and Linux. The output should be identical on every platform. 

**To Reproduce**
1. On a mac, and from a clean working directory, run `make clean gen`
2. `git diff`

**Expected behavior**
There should be no difference in the generated files.

PR #1750, opened 2025-10-15
build: fix SED regex

tarilabs:
## Description
Resolves https://github.com/kubeflow/model-registry/issues/1748

## How Has This Been Tested?
`make clean gen` locally,
BUT,
let's see if it's invariant for linux/CI/GHA

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/scripts/gen_openapi_server.sh
+++ b/catalog/scripts/gen_openapi_server.sh
@@ -32,7 +32,6 @@ sed_inplace 's/"encoding\/json"//' "$PROJECT_ROOT"/internal/server/openapi/api_m
 sed_inplace 's/github.com\/kubeflow\/model-registry\/pkg\/openapi/github.com\/kubeflow\/model-registry\/catalog\/pkg\/openapi/' \
     "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go \
     "$PROJECT_ROOT"/internal/server/openapi/api.go
-sed_inplace 's/\{model_name\+\}/*/g' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
 sed_inplace 's/\{?model_name\+\}?/*/' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
 
 echo "Applying patches to generated code"


tarilabs:
reproduce locally on Mac with

```
rm -rf bin
make build

git restore bin/.gitignore
```

image
image

also

```
rm -rf bin
make clean gen

git restore bin/.gitignore
```

image
image

catalog/scripts/gen_openapi_server.sh
```diff
@@ -32,7 +32,8 @@ sed_inplace 's/"encoding\/json"//' "$PROJECT_ROOT"/internal/server/openapi/api_m
 sed_inplace 's/github.com\/kubeflow\/model-registry\/pkg\/openapi/github.com\/kubeflow\/model-registry\/catalog\/pkg\/openapi/' \
     "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go \
     "$PROJECT_ROOT"/internal/server/openapi/api.go
-sed_inplace 's/{\?model_name+}\?/*/' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
+sed_inplace 's/\{model_name\+\}/*/g' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
+sed_inplace 's/"model_name\+"/"*"/g' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
```
pboyd: Will this work in a single pass? It works on linux with `-E`:

```suggestion
sed_inplace 's/\{?model_name\+\}?/*/' "$PROJECT_ROOT"/internal/server/openapi/api_model_catalog_service.go
```

tarilabs: Let's try it :)

tarilabs: I will have to rework this when I get back to the PC as I believe the suggested change didn't wrap on the 2 lines

jonburdo:
Taking a look at the differences between bsd sed and gnu sed, I'm wondering if we should use something more portable for this like a function that uses python re. I think it can be hard to really achieve identical behavior, so this could end up being a recurring issue.

You can install gnu sed on mac with `brew install gnu-sed` (giving you a `gsed` command) but I don't think we want to make this a requirement for developing.

jonburdo:
Actually looking at our use of `sed`, it's so simple we should stick with this I think. Here's a possible alternative in case we run into problems in the future: https://github.com/kubeflow/model-registry/pull/1754

tarilabs:
after 1093a2a and using @pboyd hints

```
make clean/deps
make clean gen
```

image

image

tarilabs:
@jonburdo sticking with GNU and not BSD sed is definitely an option even tho we tried to keep stuff both BSD and GNU compatible in the past... truly the code generation has been growing in proportion 😅  I like the approach of just using python instead of more and more bash, but that might be my bias 🙃  curious to hear others thought on your PR 👍

tarilabs:
- https://github.com/kubeflow/model-registry/pull/1750#pullrequestreview-3342489015

thanks all
/approve

PR #1763, opened 2025-10-16
Hardware Configuration toolbar Max latency and Workload type fixes

manaswinidas:
## Description

Follow up #1725 

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogStringFilter.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogStringFilter.tsx
@@ -9,14 +9,16 @@ import { useCatalogStringFilterState } from '~/app/pages/modelCatalog/utils/mode
 
 const MAX_VISIBLE_FILTERS = 5;
 
-type ModelCatalogStringFilterProps<K extends ModelCatalogStringFilterKey> = {
+type ArrayFilterKey = Exclude<ModelCatalogStringFilterKey, ModelCatalogStringFilterKey.USE_CASE>;
+
+type ModelCatalogStringFilterProps<K extends ArrayFilterKey> = {
   title: string;
   filterKey: K;
   filterToNameMapping: Partial<Record<ModelCatalogStringFilterValueType[K], string>>;
   filters: ModelCatalogStringFilterOptions[K];
 };
 
-const ModelCatalogStringFilter = <K extends ModelCatalogStringFilterKey>({
+const ModelCatalogStringFilter = <K extends ArrayFilterKey>({
   title,
   filterKey,
   filterToNameMapping,


clients/ui/frontend/src/__mocks__/mockCatalogFilterOptionsList.ts
```diff
@@ -52,6 +54,15 @@ export const mockCatalogFilterOptionsList = (
...
     },
+    [ModelCatalogStringFilterKey.USE_CASE]: {
+      type: 'string',
+      values: [
+        UseCaseOptionValue.CHATBOT,
+        UseCaseOptionValue.CODE_FIXING,
+        UseCaseOptionValue.LONG_RAG,
+        UseCaseOptionValue.RAG,
+      ],
+    },
```
mturley, line 65: I'm not seeing any results when I filter in mock mode based on these, we should follow up to add some of these values to the mock data in the BFF

clients/ui/frontend/src/__mocks__/mockCatalogFilterOptionsList.ts
```diff
@@ -66,19 +77,66 @@ export const mockCatalogFilterOptionsList = (
         max: 893,
       },
     },
-    [ModelCatalogNumberFilterKey.MAX_INPUT_TOKENS]: {
-      type: 'number',
-      range: {
-        min: 0,
-        max: 8192,
-      },
+    // All latency metric combinations for dropdown options (ttft_mean already exists as MAX_LATENCY)
```
mturley, line 80: Hmm, we may want to revisit this to get rid of the MAX_LATENCY enum value and use ttft_mean, but I see the possible problem I've created by not having the latency values be in ModelCatalogNumberFilterKey. We can still get those values where needed with `getLatencyFieldName(LatencyMetric.TTFT, LatencyPercentile.Mean)` and typescript should still be happy with that being used in places where we currently use `ModelCatalogNumberFilterKey.MAX_LATENCY`, might be worth following up with that later

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTableColumns.ts
```diff
@@ -300,6 +276,20 @@ export const hardwareConfigColumns: HardwareConfigColumn[] = [
...
+      a: CatalogPerformanceMetricsArtifact,
+      b: CatalogPerformanceMetricsArtifact,
+    ): number => {
+      const useCaseA = getStringValue(a.customProperties, 'use_case');
+      const useCaseB = getStringValue(b.customProperties, 'use_case');
+      return useCaseA.localeCompare(useCaseB);
+    },
+    width: 20,
+    modifier: 'wrap',
+  },
```
mturley, line 292: Ah we did add a use case column - that's fine, I think we might want to move it elsewhere and maybe have it format the values to look nice but we can consult with UX on that, if it ships like this I'm happy

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTableColumns.ts.backup
```diff
@@ -0,0 +1,317 @@
+import { SortableData } from 'mod-arch-shared';
```
mturley, line 1: Looks like this file shouldn't be committed

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogStringFilter.tsx
```diff
@@ -9,14 +9,16 @@ import { useCatalogStringFilterState } from '~/app/pages/modelCatalog/utils/mode
 
 const MAX_VISIBLE_FILTERS = 5;
 
-type ModelCatalogStringFilterProps<K extends ModelCatalogStringFilterKey> = {
+type ArrayFilterKey = Exclude<ModelCatalogStringFilterKey, ModelCatalogStringFilterKey.USE_CASE>;
```
mturley, line 12: Interesting, I'm curious why this is necessary - in theory we could render a ModelCatalogStringFilter using checkbox options for multiple use_case values

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -11,11 +11,10 @@ import {
...
 } from '~/concepts/modelCatalog/const';
 import { getTotalRps } from './performanceMetricsUtils';
+import { LatencyFilterConfig as SharedLatencyFilterConfig } from './latencyFilterState';
 
-// Type for storing complex latency filter configuration
-export type LatencyFilterConfig = {
-  metric: LatencyMetric;
-  percentile: LatencyPercentile;
+// Type for storing complex latency filter configuration with value
+export type LatencyFilterConfig = SharedLatencyFilterConfig & {
```
mturley, line 17: Hmm, we're exporting this with the same name as the LatencyFilterConfig we're importing, so other places that want to import LatencyFilterConfig have to make sure they're importing the right one. We might want to instead just call this LatencyFilterConfigWithValue or something, or maybe call the shared one LatencyFieldPair or something

clients/ui/frontend/src/app/pages/modelCatalog/utils/latencyFilterState.ts
```diff
@@ -0,0 +1,34 @@
...
+let sharedLatencyConfig: LatencyFilterConfig = {
+  metric: LatencyMetric.TTFT,
+  percentile: LatencyPercentile.Mean,
+};
+
+export const getLatencyFilterConfig = (): LatencyFilterConfig => sharedLatencyConfig;
+
+export const setLatencyFilterConfig = (config: LatencyFilterConfig): void => {
+  sharedLatencyConfig = config;
+};
```
mturley, line 19: It's probably a bad idea to store state in "global" variables like this. If it truly needs to be shared with multiple components it should go in a common ancestor or context, and if it truly needs to be a value mutated outside the render lifecycle you can use useRef.

I see you're using it just in `HardwareConfigurationTable`, so my instinct is to say do neither and ~~just lift the useLatencyFilterConfig hook call out into that component and pass `config` and `updateConfig` down into MaxLatencyFilter.~~

Even that might not be necessary--- in a future refactor we should be able to just use the filterState object as the source of truth - when the user switches the dropdowns we clear the old one (e.g. ttft_mean) and set the new one (e.g. ttft_p90), and then in the filter logic we simply filter by all of them because only one should have a value. But that'll be a little more tricky so we can save it.

Worst case if we ship it as-is that's ok, ~~but I think at least changing this to use a lifted hook call would be good if we can.~~

Edit: we released 3.0 without fixing this, so let's fix it right by doing the full refactor. I'll open an issue.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -118,32 +121,67 @@ export const shouldShowValidatedInsights = (
...
+  [ModelCatalogStringFilterKey.HARDWARE_TYPE]: string;
+};
+
+// Type guard to check if a value is an array of the expected type
+const isArrayOfValues = <T>(value: unknown): value is T[] => Array.isArray(value);
+
+// Type guard to check if filter key is valid for array operations
+const isArrayFilterKey = (filterKey: string): filterKey is ArrayFilterKey =>
+  isEnumMember(filterKey, ModelCatalogStringFilterKey) &&
+  filterKey !== ModelCatalogStringFilterKey.USE_CASE;
```
mturley, line 147: Well, we'll have to come back to this - we're presenting it as single selection with the current designs but it doesn't inherently have to be. When this gets lifted to the models page we'll want it to be multi-select I imagine, and it would match models that have perf artifacts for any of the selected use cases.

I think as far as the types go it's just another string filter like the others. We can likely revert some of these changes and some below, I'll follow up to take a closer look.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -118,32 +121,67 @@ export const shouldShowValidatedInsights = (
...
       : selections.filter((item) => item !== value);
-    if (isStringFilterValid(filterKey, nextState)) {
-      setFilterData(filterKey, nextState);
+
+    if (isArrayFilterKey(filterKey)) {
+      // Type assertion is safe here because we've verified the key is an array filter
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      setFilterData(filterKey, nextState as ModelCatalogFilterStates[K]);
     }
   };
```
mturley, line 183: Yeah, we'll want to revisit this - with type guards we shouldn't need an assertion like this. with the above reverted we probably can just revert this.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -224,6 +262,11 @@ export const filtersToFilterQuery = (
       }
     }
 
+    // Handle single string values (like USE_CASE)
+    if (filterOption?.type === 'string' && typeof data === 'string') {
+      return `${filterId}=${wrapInQuotes(data)}`;
+    }
```
mturley, line 268: Ah I see - it's because we're storing the value in state as a single string instead of an array ---- I think we probably should avoid this and the workload type filter should store its value in state as a single-element array. The fact that you can only select one option is just a presentational concern, the data would work fine if you selected multiple and we may want that in the future.

mturley (approved):
Followups that don't block this PR (and I think don't block the release)
* We don't have `use_case` in our BFF mocks, we should update the corresponding BFF types/mocks and put some values there - currently the workload type filter doesn't match anything when running in mock mode.
* When filtering max latency by a metric/percentile combo where the calculated min and max are the same (data has exactly the same value for all artifacts), the slider renders strangely (a dot you can't drag and a number clipping the side of the container). This will likely only happen with mock data so it's fine, but I imagine in that case the slider should be disabled and somehow forced to render in a way that doesn't look broken (same min/max numbers shown on both ends) - maybe that should be the PF behavior and we can open a PF bug.'
* See inline comments

Approving so we can unblock testing, but let's keep looping back - I think at this point we can ship with it as is if it works though and clean the rest up as tech debt.

mturley:
/approve
/lgtm
See above review comment for caveats - we should follow up again

PR #1768, opened 2025-10-17
Performance metrics fixes

pboyd:
## Description
Import integer values into the `int_value` field, instead of the `double_value` field.

~~Since the data changed, we also need to fix records created before, so also
adding the ability to replace existing artifacts with new data.~~

~~Unfortunately, it loaded every directory after every event. This was never
ideal, but it was quick before the change above. Now it was taking several
minutes to start with a modest data set. So switching it to only try to load
the data for the model we receive an event for.~~

(Dropped the second commit, but it still here: https://github.com/pboyd/model-registry/commit/cf3083325c70611f3a81c8a4a84090bb5f9c04a6)

## How Has This Been Tested?
Imported data locally and checked the resulting database records.

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/performance_metrics.go
+++ b/catalog/internal/catalog/performance_metrics.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


Crazyglue (approved):
/lgtm

PR #1769, opened 2025-10-17
Speed up metrics loading significantly

Crazyglue:
## Description
The initial implementation of the metrics loading left opportunities for performance improvements to how the data was scanned and loaded.

This PR implements a number of the improvements discussed. Namely:

- Directory scanning/caching (don't scan on every model being loaded)
- Add BatchSave method to save all metrics for a given model in 1 operation
- Bulk load all metrics when checking for existence rather than one by one

Total measured load time of my sample data went from ~20s -> ~2s.

Per-model load time went from ~850ms -> ~40ms (depending on number of metrics to be loaded, etc)

## How Has This Been Tested?
Locally using sample data

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/internal/db/service/catalog_metrics_artifact.go
+++ b/catalog/internal/db/service/catalog_metrics_artifact.go
@@ -3,6 +3,7 @@ package service
 import (
 	"errors"
 	"fmt"
+	"math"
 
 	"github.com/kubeflow/model-registry/catalog/internal/db/models"
 	"github.com/kubeflow/model-registry/internal/apiutils"


catalog/internal/db/service/catalog_metrics_artifact.go
```diff
@@ -78,6 +78,98 @@ func (r *CatalogMetricsArtifactRepositoryImpl) Save(ma models.CatalogMetricsArti
...
+	// Execute all batch operations in a single transaction
+	err := config.DB.Transaction(func(tx *gorm.DB) error {
+		// Batch insert artifacts (batch size of 100)
+		if err := tx.CreateInBatches(&schemaArtifacts, 100).Error; err != nil {
+			return fmt.Errorf("failed to batch insert artifacts: %w", err)
+		}
+
+		// Pre-allocate slices for properties and attributions
+		// Estimate ~10 properties per artifact on average
+		allProperties := make([]schema.ArtifactProperty, 0, numArtifacts*10)
```
pboyd: Unless you have benchmarks that say otherwise, I would guess that Go's default behavior is good enough here:


```suggestion
		allProperties := []schema.ArtifactProperty{}
```

catalog/internal/catalog/performance_metrics.go
```diff
@@ -443,43 +458,33 @@ func processPerformanceArtifacts(filePath string, modelID int32, metricsArtifact
...
-		if _, err := metricsArtifactRepo.Save(artifact, &modelID); err != nil {
-			glog.Errorf("Failed to save performance artifact: %v", err)
-			continue
-		}
+	return performanceRecords, nil
+}
 
-		artifactCount++
+// findExistingModel checks if a model with the given metadata already exists in the database
+func findExistingModel(metadata metadataJSON, modelRepo dbmodels.CatalogModelRepository) (dbmodels.CatalogModel, error) {
```
pboyd: This is fine, but you might consider using `modelRepo.GetByName`.

catalog/internal/catalog/performance_metrics.go
```diff
@@ -170,85 +167,27 @@ func NewPerformanceMetricsLoader(path []string, modelRepo dbmodels.CatalogModelR
...
-	if pml == nil {
-		return nil
+		modelDirCache:         make(map[string]string),
 	}
 
-	attrs := record.Model.GetAttributes()
-	if attrs == nil || attrs.Name == nil {
-		return nil
+	// Build the model directory cache once during initialization
+	if err := loader.buildModelDirCache(); err != nil {
```
pboyd, line 174: Should this cache ever be rebuilt? It would only be a problem if files are added or removed after the server is running, so it might not matter very much in practice.

Crazyglue: I suppose we would need to in that case. I'm not sure how realistic it is that that would happen though. If the service restarts, obviously it will be rebuilt, so I guess its a matter of whether or not the file would ever change when the service is not restarted/redployed

pboyd:
@Crazyglue, can you rebase? There was a conflict at some point.

PR #1771, opened 2025-10-20
fix: warn on invalid custom properties

pboyd:
## Description

When a custom property is encountered without a type warn about it, but
don't return an error.

This was an issue because metrics artifacts were being created with an
`updated_at` custom property that had no value set in any column. This
also prevents those properties from being created (they shouldn't have
been custom properties).

## How Has This Been Tested?
On a local dev environment.

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/performance_metrics.go
+++ b/catalog/internal/catalog/performance_metrics.go
@@ -509,6 +509,8 @@ func createAccuracyMetricsArtifact(evalRecords []evaluationRecord, modelID int32
 				updateTime = &updatedAt
 			}
 		}
+		delete(evalRecord.CustomProperties, "updated_at")
+		delete(evalRecord.CustomProperties, "created_at")
 	}
 
 	// Properties can be empty or contain general metadata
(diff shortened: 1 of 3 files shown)


catalog/internal/catalog/performance_metrics.go
```diff
@@ -555,15 +557,25 @@ func createPerformanceArtifact(perfRecord performanceRecord, modelID int32, type
...
-		if createdAtFloat, ok := perfRecord.CustomProperties["created_at"].(float64); ok {
-			createdAt := int64(createdAtFloat)
-			createTime = &createdAt
+		if createdAtNum, ok := perfRecord.CustomProperties["created_at"].(json.Number); ok {
+			createdAt, err := createdAtNum.Int64()
+			if err == nil {
+				createTime = &createdAt
+			} else {
+				glog.Warningf("%s: invalid created_at value: %v", artifactName, err)
+			}
```
lugi0, line 566: Would it make sense to have a "default" value in this case?

catalog/internal/catalog/performance_metrics.go
```diff
@@ -555,15 +557,25 @@ func createPerformanceArtifact(perfRecord performanceRecord, modelID int32, type
...
-	if updatedAtFloat, ok := perfRecord.CustomProperties["updated_at"].(float64); ok {
-		updatedAt := int64(updatedAtFloat)
-		updateTime = &updatedAt
+	if updatedAtNum, ok := perfRecord.CustomProperties["updated_at"].(json.Number); ok {
+		updatedAt, err := updatedAtNum.Int64()
+		if err == nil {
+			updateTime = &updatedAt
+		} else {
+			glog.Warningf("%s: invalid updated_at value: %v", artifactName, err)
+		}
```
lugi0, line 575: Same as above, would it make sense to have a "default" value in this case?

catalog/internal/catalog/performance_metrics.go
pboyd: Yeah, probably. I'll make a follow up PR to set this to the current time.

PR #1754, opened 2025-10-15
feat: replace sed with Python regex in OpenAPI generator script

jonburdo:
Replaced all sed_inplace invocations with py-re-replace function that uses Python's re module. The new function supports:
- count=0: replace all occurrences (equivalent to sed /g flag)
- count=1: replace first occurrence only (equivalent to sed without /g flag)
- count=N: replace first N occurrences
- Multiple file processing in single call
- Cross-platform compatibility with consistent regex behavior

AI-assisted: Claude Code 2.0.15 (claude-sonnet-4@20250514)

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [ ] The commits have meaningful messages
- [ ] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work.
- [ ] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).
- [ ] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

- [ ] The developer has added tests or explained why testing cannot be added.
- [ ] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

PR #1792, opened 2025-10-23
Revert "feat: replace sed with Python regex in OpenAPI generator script"

jonburdo:
Reverts kubeflow/model-registry#1754

This caused failures on jobs where python is not available:
- https://github.com/kubeflow/model-registry/actions/runs/18751313576/job/53491428154#step:8:1018
- https://github.com/kubeflow/model-registry/actions/runs/18751313578/job/53491428069#step:6:986

When using an image like `registry.access.redhat.com/ubi9/ubi-minimal:latest` we don't have python:
```
./scripts/gen_openapi_server.sh: line 24: python: command not found
```


jonburdo:
Closing in favor of https://github.com/kubeflow/model-registry/pull/1793

PR #1817, opened 2025-10-29
Added chip and "Reset all filters" to model catalog page 

YuliaKrimerman:
## Description

Adds a "Reset all filters" button and active filter chips to the model catalog page.

- Active filter chips: grouped by category (Provider, License, Task, Language, Hardware type) and shown in bordered containers with the category name inside each container.
- Reset all filters: button clears the search input, global filter chips

## How Has This Been Tested?
- Manual testing: verified filter chips display correctly, individual chip removal works, category clear buttons work, "Reset all filters" clears all filters without affecting category toggles

https://github.com/user-attachments/assets/0ae36819-6b9a-4722-ae8b-d5a29cf899a9

in PF mode 
Screenshot 2025-10-29 at 1 08 48 PM

## Merge criteria:
- All the commits have been signed-off (To pass the `DCO` check)

**For UI changes:**

**Note:** Screenshots should be added showing:
1. Active filter chips displayed with grey backgrounds in bordered containers
2. "Reset all filters" button visible when filters are active
3. Verification that category toggles remain unchanged after reset

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
@@ -1,14 +1,6 @@
 import React from 'react';
-import {
-  Button,
-  Flex,
-  FlexItem,
-  Label,
-  LabelGroup,
-  Stack,
-  StackItem,
-} from '@patternfly/react-core';
-import { TimesIcon } from '@patternfly/react-icons';
+import { ToolbarFilter, ToolbarLabelGroup, ToolbarLabel } from '@patternfly/react-core';
+import { isEnumMember } from 'mod-arch-core';
 import { ModelCatalogContext } from '~/app/context/modelCatalog/ModelCatalogContext';
 import {
   ModelCatalogStringFilterKey,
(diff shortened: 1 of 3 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,220 @@
...
+};
+
+// Filter category display names
+const FILTER_CATEGORY_NAMES: Record<ModelCatalogStringFilterKey, string> = {
+  [ModelCatalogStringFilterKey.PROVIDER]: 'Provider',
+  [ModelCatalogStringFilterKey.LICENSE]: 'License',
+  [ModelCatalogStringFilterKey.TASK]: 'Task',
+  [ModelCatalogStringFilterKey.LANGUAGE]: 'Language',
+  [ModelCatalogStringFilterKey.HARDWARE_TYPE]: 'Hardware type',
+  [ModelCatalogStringFilterKey.USE_CASE]: 'Use case',
```
mturley: A couple things here:

1. The microcopy we present for this column/filter is "Workload type" even though it's `use_case` behind the scenes.
2. We're starting to duplicate a lot of these names in a few places now (e.g. "Provider" is here, in HardwareConfigurationTableColumns.ts, and in LicenseFilter.tsx), I wonder if we should consolidate them somewhere. I don't think it would be exactly a 1:1 mapping of the filter keys to strings since the columns don't map to them perfectly, so maybe an enum like `ModelCatalogPropertyName` that's just a simple set of strings and is used in all 3 places.
3. Eventually this will need to be a mapping of all the `ModelCatalogFilterKey`s (see modelCatalogTypes.ts, it's a union of `ModelCatalogStringFilterKey | ModelCatalogNumberFilterKey`), but I don't think that should be done in this PR since @manaswinidas 's refactoring will change the ModelCatalogNumberFilterKeys. I.e. later on there will need to be a separate filter category name for "Max latency (TTFT Mean)" and "Max latency (TTFT P99)" etc, but the types aren't ready for that and they aren't needed for this PR. So can you add a TODO comment here to switch this to be a `Record<ModelCatalogFilterKey, string>` when performance filters need to be included?

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,220 @@
...
+    categoryName: string;
+    values: string[];
+  }> = [];
+
+  // Process each filter category
+  Object.values(ModelCatalogStringFilterKey).forEach((filterKey) => {
+    if (filterKey === ModelCatalogStringFilterKey.USE_CASE) {
+      // Skip USE_CASE as it's a single value, not an array
+      return;
+    }
```
mturley: ...

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,220 @@
...
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          return AllLanguageCodesMap[languageValue] || value;
+        }
+        default:
+          return value;
+      }
+    } catch {
+      return value;
+    }
+  };
```
mturley, line 65: We always want to avoid disabling the consistent-type-assertions rule, we can usually solve it by using better types or narrowing the type with a type guard function. In this case we can import the `isEnumMember` type guard from `mod-arch-core` and use it here. Also, with these guards in place you shouldn't need the try/catch.
```suggestion
  const getFilterLabel = (filterKey: ModelCatalogStringFilterKey, value: string): string => {
    switch (filterKey) {
      case ModelCatalogStringFilterKey.PROVIDER: {
        return isEnumMember(value, ModelCatalogProvider)
          ? MODEL_CATALOG_PROVIDER_NAME_MAPPING[value]
          : value;
      }
      case ModelCatalogStringFilterKey.LICENSE: {
        return isEnumMember(value, ModelCatalogLicense)
          ? MODEL_CATALOG_LICENSE_NAME_MAPPING[value]
          : value;
      }
      case ModelCatalogStringFilterKey.TASK: {
        return isEnumMember(value, ModelCatalogTask)
          ? MODEL_CATALOG_TASK_NAME_MAPPING[value]
          : value;
      }
      case ModelCatalogStringFilterKey.LANGUAGE: {
        return isEnumMember(value, AllLanguageCode) ? AllLanguageCodesMap[value] : value;
      }
      default:
        return value;
    }
  };
```

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
mturley: So, 2 things here and the second one makes the first one not matter but FYI.

1. We should avoid introducing new USE_CASE-specific logic before we close @manaswinidas 's USE_CASE refactor issue since it will eliminate this problem.
2. We actually don't want to show all filters here anyway. Right now you'll use this ModelCatalogActiveFilters component only on the landing page where you only want it to show filters that appear there (currently just Provider, License, Task and Language), and in the future we'll also want to use it on the details page and only show the filters that appear there (so none of those 4 basic filters). Because these two pages share the same filter state object (intentionally - when we add perf filters to the landing page we'll have overlap and we'll want those filter values to persist between pages), with your changes if you go to the details page and apply the hardware type filter, then return to the landing page, you'll see a chip for that filter.

What you'll need to do is expose a prop like `filtersToShow: ModelCatalogFilterKey[]` and use that here to filter out what shows up in your `activeFiltersByCategory` similar to what you're doing to filter out USE_CASE here. Then where you render it pass in the 4 basic filter keys, and you won't have to worry about USE_CASE showing up anyway.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    padding: 'var(--pf-v6-global--spacer--xs) var(--pf-v6-global--spacer--sm)',
+                    border: '1px solid var(--pf-v6-global--BorderColor--200)',
+                    borderRadius: 'var(--pf-v6-global--BorderRadius--sm)',
+                    backgroundColor: 'var(--pf-v6-global--BackgroundColor--100)',
+                    minHeight: '32px',
+                    boxSizing: 'border-box',
+                    width: 'fit-content',
+                  }}
+                  data-testid={`${category}-filter-container`}
+                >
```
mturley: We really need to stop introducing inline styles. If you're ever not sure how to avoid them, let's start a thread with the team first to ask for help on how you might implement a layout without them. They override the theming system and lead to tricky bugs when PF changes later.

Let's talk it through - what are you trying to achieve here? It looks like this is a toolbar group, can you not use the PF ToolbarGroup component? In fact, looking at the PF Toolbar docs, here's an example that includes the same filter labels we're trying to render here: https://www.patternfly.org/components/toolbar#with-filters

Ok... I think I have the answer here, let me think out loud:

From looking at the code in that example it looks like the `ToolbarFilter` component is designed to render filter controls you pass as children and then magically handles labels for those filters through a portal to the Toolbar parent. We need to not render our filter controls in the toolbar, but I still think we can use this. I'm editing the code of that example to remove the children from all the ToolbarFilter components. Check this out, from simply rendering this anywhere as a child under a Toolbar:
```tsx
        <ToolbarFilter
          labels={['foo', 'bar']}
          deleteLabel={(category, label) => onDelete(category as string, label as string)}
          deleteLabelGroup={(category) => onDeleteGroup(category as string)}
          categoryName="Status"
        />
        <ToolbarFilter
          labels={['hello', 'world']}
          deleteLabel={(category, label) => onDelete(category as string, label as string)}
          deleteLabelGroup={(category) => onDeleteGroup(category as string)}
          categoryName="Risk"
        />
```
we get this:

Image

So that gets us 90% of the way there - the only problem remaining is that the `categoryName` we pass in there is used both for the human-readable rendered name and as the `category` passed to the `deleteLabel` and `deleteLabelGroup` callbacks. Looking closer at the type of `categoryName` it looks like it can take `string | ToolbarLabelGroup`. [Looking at the PF source, it seems like ToolbarLabelGroup is simply a key/name pair](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx#L8-L13):

```ts
export interface ToolbarLabelGroup {
  /** A unique key to identify this label group category */
  key: string;
  /** The category name to display for the label group */
  name: string;
}
```

Similarly, we have a problem with the `labels` we pass in - if we pass nice looking strings to render, those are what we'll get out as the `label` param of `deleteLabel` which we can't match back to our state. Looking at the type for `labels`, it has a similar mechanism: `(string | ToolbarLabel)[]`. [Let's look at that ToolbarLabel type](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Toolbar/ToolbarFilter.tsx#L15-L20):

```ts
export interface ToolbarLabel {
  /** A unique key to identify this label */
  key: string;
  /** The ReactNode to display in the label */
  node: React.ReactNode;
}
```

Bingo, we can use those. In both cases our `key` is what we need for manipulating state in the callbacks, and the `name` or `node` is how it will be rendered.

So I think that's our answer. No need to render anything you have here at all, from the Stack all the way down. We can simply map over the filters in our state and render a `ToolbarFilter` for each one without any children, map over our values to pass in the `labels` as { key, node } objects, and pass { key, name } objects for the `categoryLabels`.

The only remaining thing here - by default this means it will render the "Clear all filters" link at the Toolbar level (in this example they pass a `clearAllFilters` function to the `Toolbar`), and that's outside your ModelCatalogActiveFilters. You probably just need to not pass in a `onResetAllFilters` prop at all and instead pass in a `clearAllFilters` up where the `<Toolbar>` is rendered.

If any part of this doesn't work the way I'm expecting please let me know, we can figure it out. No need to do a bunch of custom rendering where PF has this handled for us, even if their solution is a bit confusing 😅

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,120 @@
...
+      {filtersToShow.map((filterKey) => {
+        // Only process string filter keys that are arrays
+        if (!isEnumMember(filterKey, ModelCatalogStringFilterKey)) {
+          return null;
+        }
+
+        const filterValues = filterData[filterKey];
+        if (!Array.isArray(filterValues) || filterValues.length === 0) {
+          return null;
+        }
```
mturley, line 78: FYI this will need to be removed when we refactor this to also work for numeric filters, which will be required for the chip bar on the details page (and should probably wait for the ongoing refactors of the filter state)

mturley:
@YuliaKrimerman needs rebase to fix the lint issue

PR #1835, opened 2025-11-03
Slider Issues for Hardware Configuration

claudialphonse78:
## Description

This PR fixes the slider filters for identical values and manual input

**Issue 1**: Slider breaks when all values are identical
Disable slider with tooltip when min === max detected
Set maxValue to minValue + 1 to prevent rendering errors

**Issue 2**: Manual input not working
Separated Slider from TextInput with native HTML5 min/max attributes
Manual value entry now works correctly

https://github.com/user-attachments/assets/2f5f082d-6b1a-4d38-bc18-88f44e6510e3

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/SliderWithInput.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/SliderWithInput.tsx
@@ -1,13 +1,5 @@
 import * as React from 'react';
-import {
-  Flex,
-  FlexItem,
-  InputGroup,
-  InputGroupItem,
-  InputGroupText,
-  Slider,
-  TextInput,
-} from '@patternfly/react-core';
+import { Slider, SliderOnChangeEvent } from '@patternfly/react-core';
 
 type SliderWithInputProps = {
   value: number;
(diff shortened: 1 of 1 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MinRpsFilter.tsx
```diff
@@ -57,22 +60,31 @@ const MinRpsFilter: React.FC<MinRpsFilterProps> = ({ performanceArtifacts }) =>
...
     const rpsValues = performanceArtifacts
       .map((artifact) => getDoubleValue(artifact.customProperties, 'requests_per_second'))
       .filter((rps) => rps > 0); // Filter out invalid values
 
     if (rpsValues.length === 0) {
-      return { minValue: 1, maxValue: 300 }; // Default values when no valid RPS values
+      return { minValue: 1, maxValue: 300, isSliderDisabled: false }; // Default values when no valid RPS values
     }
 
+    const calculatedMin = Math.min(...rpsValues);
```
manaswinidas: we are duplicating this .. can this be a shared util ?

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MinRpsFilter.tsx
```diff
@@ -124,19 +136,38 @@ const MinRpsFilter: React.FC<MinRpsFilterProps> = ({ performanceArtifacts }) =>
...
+                    const numValue = Number(value);
+                    setLocalValue(numValue);
+                  }}
+                  style={{ width: '80px' }}
+                  aria-label="RPS value input"
+                />
+              </InputGroupItem>
+              <InputGroupText isDisabled={isSliderDisabled}>RPS</InputGroupText>
+            </InputGroup>
+          </FlexItem>
```
manaswinidas: Does it make sense for us to make this into a shared component ?

claudialphonse78: Created a common slider component which we can use in the minrps and the maxlatency filters

ppadti:
@claudialphonse78 just wanted to know - the slider and number input doesn't seems to show up correctly in mui theme, but it does work with patternfly theme. Is that known issues already, do we have any bug created to track those?
Screenshot 2025-11-06 at 9 11 29 PM

ppadti:
Also, I am not seeing the empty state for mui theme, when i update the filters(empty state show up for patternfly theme) - not related to this PR though, we can log separate issue to track this. 
Screenshot 2025-11-06 at 9 11 39 PM

claudialphonse78:
> Also, I am not seeing the empty state for mui theme, when i update the filters(empty state show up for patternfly theme) - not related to this PR though, we can log separate issue to track this. Screenshot 2025-11-06 at 9 11 39 PM

@ppadti yes, looks like the sticky header here is conflicting with the empty state when we run in the mui-theme

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/SliderWithInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+              max={max}
+              isDisabled={isDisabled}
+              onChange={handleInputChange}
+              onBlur={handleBlur}
+              style={{ width: '80px' }}
+              aria-label={ariaLabel}
+            />
+          </InputGroupItem>
+          <InputGroupText isDisabled={isDisabled}>{suffix}</InputGroupText>
+        </InputGroup>
```
mturley: @claudialphonse78 I was curious why the `isInputVisible` prop for Slider didn't work, and I found that the `onChange` prop actually has a few extra arguments. When you change the input box, it doesn't receive the new value as the `value` argument but it _does_ receive it as a third `inputValue` argument. We should be able to go back to using `isInputVisible` and combine our onChange functions as long as we use `inputValue || value`. It is a number type - the built in input rejects invalid values, so we also wouldn't need the Number cast or the isNaN check. We do still need to check that it's in range though.

Image

claudialphonse78: @mturley The input inside the slider component for patternfly at the moment does not support any other event other than onBlur or Enter.
Even if we pass inputValue, the Slider's internal onChangeHandler still doesn't call the parent onChange until Enter/blur. The inputValue prop only syncs the input field's display value, not the slider thumb position. 
So I used a separate input so we could have control over that

mturley: Interesting, ok.. I think the slider not updating until blur is ok, if that's the implemented behavior in PF that's their choice. I think we could implement it the same way [their example here is implemented, see onChangeContinuous in the code](https://www.patternfly.org/components/slider#value-input) - keep our state for value and inputValue, pass an `onChange` handler that takes `value`, `inputValue` and `setLocalInputValue` arguments, and then (1) if `inputValue` is undefined we update just our `value` state , or (2) if `inputValue` is defined, we validate and clamp it and update all 3 of: our `value` state, our `inputValue` state, and their internal input value with `setLocalInputValue`. Honestly I'm not sure why that last part is necessary... but it appears to be.

@claudialphonse78 I still think we can avoid all this custom code, if updating onChange is desired we could open a PF bug, otherwise this is the out of the box behavior.

claudialphonse78: @mturley I will open a patternfly bug for this tomorrow
the slider feels very weird works only on enter/blur plus the max/min has no effect on disabling the stepper only works on the ticks too hacky 
added `setPFInputValue` to update PatternFly's internal input state right away when we clamp a value cause without it, the input might shows the flicker thing of the out of range value

mturley: ah I see, ok thanks. yeah this works well enough for now I think and saves us the maintenance of the custom input, and we'll benefit from the PF fix in a future upgrade. I really appreciate your effort here.

PR #1838, opened 2025-11-03
Include artifacts with `artifactType` query param in model-catalog

Crazyglue:
## Description
When querying for models in the model catalog (or a specific model), a `?artifactType=...` query param can be included to optionally include the artifacts related to that model.

If this query param is not included, the artifacts field will be omitted from the response.

This is a convenience parameter for the dashboard/FE to make fewer calls to fetch _all_ the relevant data of a specific model.

## How Has This Been Tested?
Locally and with unit tests

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/internal/server/openapi/artifact_inclusion_test.go
+++ b/catalog/internal/server/openapi/artifact_inclusion_test.go
@@ -1,0 +1,190 @@
+package openapi
+
+import (
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/kubeflow/model-registry/catalog/internal/catalog"
+	"github.com/kubeflow/model-registry/catalog/pkg/openapi"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// mockCatalogProvider implements catalog.APIProvider for testing
+type mockCatalogProvider struct{}
+
+func (m *mockCatalogProvider) GetModel(ctx context.Context, modelName string, sourceID string) (*openapi.CatalogModel, error) {
+	return nil, nil
+}
+
+func (m *mockCatalogProvider) ListModels(ctx context.Context, params catalog.ListModelsParams) (openapi.CatalogModelList, error) {
+	// Create test models
+	model1ID := "1"
+	model1Name := "test-model-1"
+	model2ID := "2"
+	model2Name := "test-model-2"
+
+	model1 := openapi.CatalogModel{
+		Id:   &model1ID,
+		Name: model1Name,
+	}
+
+	model2 := openapi.CatalogModel{
+		Id:   &model2ID,
+		Name: model2Name,
+	}
+
+	// If artifact types are requested, add artifacts
+	if len(params.ArtifactTypesFilter) > 0 {
+		for _, artifactType := range params.ArtifactTypesFilter {
+			if artifactType == "metrics-artifact" {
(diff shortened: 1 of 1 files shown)


patches/api_model_catalog_service.patch
```diff
@@ -2,7 +2,24 @@ diff --git a/catalog/internal/server/openapi/api_model_catalog_service.go b/cata
...
+ 	nextPageTokenParam := query.Get("nextPageToken")
+-	artifactTypeParam := strings.Split(query.Get("artifactType"), ",")
++
++	// Handle multiple artifactType query parameters
++	artifactTypeParam := make([]string, 0)
++	for _, v := range query["artifactType"] {
++		if v != "" {
++			artifactTypeParam = append(artifactTypeParam, v)
++		}
++	}
```
pboyd, line 17: FWIW, we can use comma-separated values here instead, if you'd prefer (that's what we do for the other parameters). I believe the problem you're working around is fixed in newer versions of openapi-generator-cli.

pboyd (approved):
/lgtm

I tested this and the functionality works as expected.

This does add a bit of time to the response if there are a lot of artifacts. It wasn't significant enough in my testing to be a concern today, but we should keep an eye on it.

mturley:
FYI @pboyd @Crazyglue I think I left a communication gap here - we may not need/want this after all when fetching the list of models since it will slow down the page's responsiveness. It could still be used for a minor optimization on the model details page but it's not a priority from our end. Apologies for confusion there, I know in the past I mentioned this might be helpful, I didn't follow up after discussing further with @andrewballantyne

Crazyglue:
> FYI @pboyd @Crazyglue I think I left a communication gap here - we may not need/want this after all when fetching the list of models since it will slow down the page's responsiveness. It could still be used for a minor optimization on the model details page but it's not a priority from our end. Apologies for confusion there, I know in the past I mentioned this might be helpful, I didn't follow up after discussing further with @andrewballantyne

@mturley Can we confirm that we don't need this at all? I can close this PR if so. I would need to do some performance comparisons between with/without the artifacts to see if there is a meaningful difference, but i'd have thought this wouldnt impact page load speed noticeably. I will do a comparison shortly

pboyd:
@mturley, correct me if I'm wrong, the responsiveness issue isn't so much that the this response time of this request is slow, it's that you can progressively render the page if you make the calls separately? Total time spent fetching is probably faster with this change, but, if I understand the concern, it may not feel that way to a user waiting for the entire pull before seeing anything on the page. If that's the case, @Crazyglue, I don't think comparing the response times will matter.

I guess this was a miscommunication all around, but my impression was that this feature was desirable for the dashboard, but not absolutely necessary. I don't think there's any other use-case for this feature, so if you don't want it, then there's no point maintaining it and maybe the best thing to do is close this PR.

mturley:
@pboyd that's correct. I indicated earlier on that this would be desirable but hadn't considered what you describe - the progressive rendering we do now feels more responsive. I also wonder if this would become a maintenance problem as we add filtering to artifacts. I'm fine with closing this.

Crazyglue:
> @pboyd that's correct. I indicated earlier on that this would be desirable but hadn't considered what you describe - the progressive rendering we do now feels more responsive. I also wonder if this would become a maintenance problem as we add filtering to artifacts. I'm fine with closing this.

I'll close the PR given the circumstances

PR #1844, opened 2025-11-04
Use context filter state for Max Latency filters instead of module-level state

manaswinidas:
## Description

1. Deleted latencyFilterState.ts module-level mutable state and removed MAX_LATENCY enum value.
2. Latency filter stores values directly in ModelCatalogContext using dynamic field names like ttft_mean, e2e_p90 generated by getLatencyFieldName().

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
@@ -14,17 +14,10 @@ import {
   Slider,
 } from '@patternfly/react-core';
 import { HelpIcon } from '@patternfly/react-icons';
-import {
-  LatencyMetric,
-  LatencyPercentile,
-  LATENCY_METRIC_FIELD_NAMES,
-} from '~/concepts/modelCatalog/const';
+import { LatencyMetric, LatencyPercentile } from '~/concepts/modelCatalog/const';
 import { CatalogPerformanceMetricsArtifact } from '~/app/modelCatalogTypes';
 import { getDoubleValue } from '~/app/utils';
-import {
-  getLatencyFieldName,
-  parseLatencyFieldName,
-} from '~/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils';
+import { getLatencyFieldName } from '~/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils';
 import { ModelCatalogContext } from '~/app/context/modelCatalog/ModelCatalogContext';
 
 type LatencyFilterState = {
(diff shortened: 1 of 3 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxLatencyFilter.tsx
```diff
@@ -59,15 +57,34 @@ const MaxLatencyFilter: React.FC<MaxLatencyFilterProps> = ({ performanceArtifact
...
+            fieldName,
+            metric: parsed.metric,
+            percentile: parsed.percentile,
+            value,
+          };
+        }
+      }
+    }
+    return null;
+  }, [filterData]);
```
mturley, line 65: I think we can avoid creating a constant array of all the permutations of metric and percentile and keep those enums as the source of truth. I might have been overabstracting there, but you could just loop over both here:
```suggestion
  // Find the currently active latency filter (if any)
  const currentActiveFilter = React.useMemo(() => {
    for (const metric of Object.values(LatencyMetric)) {
      for (const percentile of Object.values(LatencyPercentile)) {
        const fieldName = getLatencyFieldName(metric, percentile);
        const value = filterData[fieldName];
        if (value !== undefined && typeof value === 'number') {
          return { fieldName, metric, percentile, value };
        }
      }
    }
    return null;
  }, [filterData]);
```

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -110,19 +111,14 @@ export const filterHardwareConfigurationArtifacts = (
...
+    // Max Latency Filter - check for any active latency field
+    for (const fieldName of LATENCY_METRIC_FIELD_NAMES) {
+      const filterValue = filterState[fieldName];
+      if (filterValue !== undefined && typeof filterValue === 'number') {
+        const latencyValue = getDoubleValue(artifact.customProperties, fieldName);
+        if (latencyValue > filterValue) {
+          return false;
+        }
       }
     }
```
mturley, line 125: Same here, you can derive everything from the separate metric and percentile enums
```suggestion
    // Max Latency Filter - check for any active latency field
    for (const metric of Object.values(LatencyMetric)) {
      for (const percentile of Object.values(LatencyPercentile)) {
        const fieldName = getLatencyFieldName(metric, percentile);
        const filterValue = filterState[fieldName];
        if (filterValue !== undefined && typeof filterValue === 'number') {
          const latencyValue = getDoubleValue(artifact.customProperties, fieldName);
          if (latencyValue > filterValue) {
            return false;
          }
        }
      }
    }
```

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -29,6 +28,26 @@ export enum LatencyPercentile {
...
+  'tps_mean',
+  'tps_p90',
+  'tps_p95',
+  'tps_p99',
+  'itl_mean',
+  'itl_p90',
+  'itl_p95',
+  'itl_p99',
+];
+
```
mturley: With the above 2 changes we don't need this array. Alternatively, if we want to have it for ease of use elsewhere, we could derive it using LatencyMetric, LatencyPercentile, getLatencyFieldName, and some map/reduce :)

I do get the sense that my desire to reduce duplication has introduced some complexity here so sorry about that.

mturley (changes requested):
Looking good here, I think we can simplify a little bit though.

PR #1767, opened 2025-10-16
Handle JSON arrays in filterQuery

pboyd:
## Description

Handle `filterQuery` args for JSON array fields more like single-value fields. For instance, `language = 'en'` returns records where at least one entry in the language field is "en".

This only works on PostgreSQL right now, but it's only intended for the catalog.

## How Has This Been Tested?
Tested locally and added unit tests.

```
$ curl -s "http://localhost:8082/api/model_catalog/v1alpha1/models?pageSize=15&filterQuery=$(echo '"license IN (\"mit\",\"apache-2.0\") AND language IN (\"zh\",\"en\")"' | jq -r '@uri')" | jq -r '.items[] | .license + " - " + (.language | join(", "))'
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en, de, es, fr, ja, pt, ar, cs, it, ko, nl, zh
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en
apache-2.0 - en
mit - en
apache-2.0 - en, fr, de, es, it, pt, zh, ja, ru, ko
apache-2.0 - en, fr, de, es, it, pt, zh, ja, ru, ko
apache-2.0 - en, fr, de, es, it, pt, zh, ja, ru, ko
apache-2.0 - en, fr, de, es, it, pt, zh, ja, ru, ko
```

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [x] The commits have meaningful messages
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).

PR #1854, opened 2025-11-10
Refactor use case filter to use multi-select instead of single-select

manaswinidas:
## Description

Screenshot 2025-11-10 at 5 44 32 PM

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
@@ -62,6 +62,7 @@ const WorkloadTypeFilter: React.FC = () => {
                       <Checkbox
                         label={option.label}
                         id={option.value}
+                        data-testid={`workload-type-filter-${option.value}`}
                         isChecked={isUseCaseSelected(option.value)}
                         onChange={(_, checked) => toggleUseCaseSelection(option.value, checked)}
                       />
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelCatalog.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelCatalog.ts
@@ -267,7 +267,9 @@ class ModelCatalog {
   }
 
   findWorkloadTypeOption(label: string) {
-    return this.findWorkloadTypeFilter().findDropdownItem(label);
+    // Workload type uses checkboxes in a panel, not menu items
+    // Find checkbox by its label within the dropdown panel
+    return cy.contains('label', label).parent().find('input[type="checkbox"]');
   }
 
   selectWorkloadType(label: string) {
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
@@ -162,7 +162,7 @@ describe('Model Catalog Page', () => {
     ]).then((interceptions) => {
       const lastInterception = interceptions[interceptions.length - 1];
       expect(lastInterception.request.url).to.include(
-        '%28tasks+LIKE+%27%25%22text-generation%22%25%27+OR+tasks+LIKE+%27%25%22text-to-text%22%25%27%29+AND+provider%3D%27Google%27',
+        'tasks+IN+%28%27text-generation%27%2C%27text-to-text%27%29+AND+provider%3D%27Google%27',
       );
     });
   });
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogTabs.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogTabs.cy.ts
@@ -226,6 +226,7 @@ describe('Model Catalog Details Tabs', () => {
       it('should show workload type options when clicked', () => {
         modelCatalog.findModelCatalogDetailLink().first().click();
         modelCatalog.clickPerformanceInsightsTab();
+        modelCatalog.findWorkloadTypeFilter().click();
         modelCatalog.findWorkloadTypeOption('Chatbot').should('be.visible');
         modelCatalog.findWorkloadTypeOption('Code Fixing').should('be.visible');
         modelCatalog.findWorkloadTypeOption('Long RAG').should('be.visible');
@@ -235,17 +236,19 @@ describe('Model Catalog Details Tabs', () => {
       it('should update toggle text when workload type is selected', () => {
         modelCatalog.findModelCatalogDetailLink().first().click();
         modelCatalog.clickPerformanceInsightsTab();
+        modelCatalog.findWorkloadTypeFilter().click();
         modelCatalog.selectWorkloadType('Code Fixing');
         modelCatalog
           .findWorkloadTypeFilter()
-          .should('contain.text', 'Workload type:')
-          .should('contain.text', 'Code Fixing');
+          .should('contain.text', 'Workload type')
+          .should('contain.text', '1 selected');
       });
 
       it('should filter hardware configuration table by selected workload type', () => {
         modelCatalog.findModelCatalogDetailLink().first().click();
         modelCatalog.clickPerformanceInsightsTab();
         modelCatalog.findHardwareConfigurationTableRows().should('have.length.at.least', 1);
+        modelCatalog.findWorkloadTypeFilter().click();
         modelCatalog.selectWorkloadType('Code Fixing');
         modelCatalog.findHardwareConfigurationTableRows().should('exist');
         modelCatalog.findHardwareConfigurationColumn('Workload type').each(($el) => {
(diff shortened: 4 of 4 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -1,43 +1,48 @@
...
+  Panel,
+  PanelMain,
 } from '@patternfly/react-core';
-import { asEnumMember } from 'mod-arch-core';
 import { ModelCatalogStringFilterKey, UseCaseOptionValue } from '~/concepts/modelCatalog/const';
 import { USE_CASE_OPTIONS } from '~/app/pages/modelCatalog/utils/workloadTypeUtils';
 import { ModelCatalogContext } from '~/app/context/modelCatalog/ModelCatalogContext';
 
-const UseCaseFilter: React.FC = () => {
+type UseCaseOption = {
```
YuliaKrimerman: I think it's worth making sure that the component is now having the same name as the file.

YuliaKrimerman: And we can use import { USE_CASE_OPTIONS, UseCaseOption } from '~/app/pages/modelCatalog/utils/workloadTypeUtils'; in here, instead of creating it again

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
```diff
@@ -46,31 +51,50 @@ const UseCaseFilter: React.FC = () => {
...
+        <Flex direction={{ default: 'column' }} spaceItems={{ default: 'spaceItemsSm' }}>
+          {/* Workload type checkboxes */}
+          <FlexItem>
+            <Flex direction={{ default: 'column' }} spaceItems={{ default: 'spaceItemsXs' }}>
+              {useCaseOptions.map((option) => (
+                <FlexItem key={option.value}>
+                  <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                    <FlexItem flex={{ default: 'flex_1' }}>
+                      <Checkbox
+                        label={option.label}
```
YuliaKrimerman, line 63: Missing data-testid here

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -157,8 +158,8 @@ export const clearAllFilters = (
   setFilterData(ModelCatalogStringFilterKey.LANGUAGE, []);
   setFilterData(ModelCatalogStringFilterKey.HARDWARE_TYPE, []);
 
-  // Clear use case filter (single value)
-  setFilterData(ModelCatalogStringFilterKey.USE_CASE, undefined);
+  // Clear use case filter
+  setFilterData(ModelCatalogStringFilterKey.USE_CASE, []);
```
mturley: Nit, but this is just another array string filter now so it can probably just go above with the others instead of with its own comment. Also we should consider refactoring this clear function to just loop over all the ModelCatalogStringFilterKeys like it does for ModelCatalogNumberFilterKey below, so we don't have to maintain this list of calls.

If we do that refactor, as part of it we should probably also loop over all the metric+percentile filter keys like we did for MaxLatencyFilter and filterHardwareConfigurationArtifacts, to clear all those too (will need that for when we add filter chips to the details page, so it could be part of that PR instead if you prefer)

I'm starting to see a possible need for a `getAllFilterKeys` util that just returns an array of ModelCatalogFilterKey by combining all 3 of these.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -121,13 +121,14 @@ export const shouldShowValidatedInsights = (
...
-// Define array-based filter keys (excluding USE_CASE which is single-selection)
+// Define array-based filter keys (all string filters support multiple selections)
 type ArrayFilterKey =
   | ModelCatalogStringFilterKey.TASK
   | ModelCatalogStringFilterKey.PROVIDER
   | ModelCatalogStringFilterKey.LICENSE
   | ModelCatalogStringFilterKey.LANGUAGE
-  | ModelCatalogStringFilterKey.HARDWARE_TYPE;
+  | ModelCatalogStringFilterKey.HARDWARE_TYPE
+  | ModelCatalogStringFilterKey.USE_CASE;
```
mturley: This type is now the same as `ModelCatalogStringFilterKey` and we don't need it anymore

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -136,16 +137,9 @@ type ArrayFilterValueType = {
   [ModelCatalogStringFilterKey.LICENSE]: ModelCatalogLicense;
   [ModelCatalogStringFilterKey.LANGUAGE]: AllLanguageCode;
   [ModelCatalogStringFilterKey.HARDWARE_TYPE]: string;
+  [ModelCatalogStringFilterKey.USE_CASE]: UseCaseOptionValue;
 };
```
mturley: This is also now the same as `ModelCatalogStringFilterValueType` and we don't need it

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -154,32 +148,19 @@ export const useCatalogStringFilterState = <K extends ArrayFilterKey>(
...
-
-    if (isArrayFilterKey(filterKey)) {
-      // Type assertion is safe here because we've verified the key is an array filter
-      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-      setFilterData(filterKey, nextState as ModelCatalogFilterStates[K]);
-    }
+    // TypeScript limitation: generic K creates intersection type instead of union
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    setFilterData(filterKey, nextState as ModelCatalogFilterStates[K]);
   };
```
mturley: You can avoid these type assertions with a type guard and by making the `selections` array a generic `string[]`. It doesn't protect us against introducing an invalid value in the body of the `setSelected` function, but the param of that function has the correct type `Value` and doesn't reference any other values so it's safe.

```suggestion
export const useCatalogStringFilterState = <K extends ModelCatalogStringFilterKey>(
  filterKey: K,
): {
  isSelected: (value: ModelCatalogStringFilterValueType[K]) => boolean;
  setSelected: (value: ModelCatalogStringFilterValueType[K], selected: boolean) => void;
} => {
  type Value = ModelCatalogStringFilterValueType[K];
  const { filterData, setFilterData } = React.useContext(ModelCatalogContext);
  const selections: string[] = filterData[filterKey];

  const isValidStringState = (state: string[]): state is ModelCatalogFilterStates[K] =>
    Object.values(ModelCatalogStringFilterKey).includes(filterKey);

  const isSelected = React.useCallback((value: Value) => selections.includes(value), [selections]);

  const setSelected = (value: Value, selected: boolean) => {
    const nextState = selected
      ? [...selections, value]
      : selections.filter((item) => item !== value);
    if (isValidStringState(nextState)) {
      setFilterData(filterKey, nextState);
    }
  };
```

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -235,9 +216,8 @@ const isFilterIdInMap = (
   filters: CatalogFilterOptions,
 ): filterId is keyof CatalogFilterOptions => typeof filterId === 'string' && filterId in filters;
 
-// TODO tech debt: different filterQuery syntax is needed depending on whether the API stores an array of values or a single string value.
-//   the current filter_options API response does not indicate the difference between these two types of fields, so for now we hard-code them.
-const KNOWN_ARRAY_FILTER_IDS: (keyof CatalogFilterOptions)[] = [
+// Array-based fields that need LIKE syntax instead of IN syntax for database queries
+const ARRAY_FIELD_FILTER_IDS: (keyof CatalogFilterOptions)[] = [
```
mturley: This should no longer be necessary actually, but just in case we can save that for its separate tech debt ticket. I can't find the backend fix right now 😅

mturley (changes requested):
A couple things here -  also +1 to @YuliaKrimerman's comment, this filter now operates the same as the hardware type filter and we should factor out the inner code between this and HardwareTypeFilter - I'm not sure we can exactly reuse code from ModelCatalogStringFilter though, since that is laid out differently (it's for the sidebar of the main page and includes the title, option search, and "show more/less" behavior.

Maybe we can factor out an alternate ModelCatalogStringDropdownFilter or something. we will need these filters presented differently depending on which page they appear once we add them to the main page (where they should use the existing ModelCatalogStringFilter).

We could perhaps factor out a ModelCatalogFilterCheckboxes component for the contents that are just the checkboxes + search + more/less, and have props to disable those things. and then use it inside both ModelCatalogStringFilter and ModelCatalogStringDropdownFilter, and disable the search + more/less when used in the dropdown. Maybe overkill but we can definitely reuse more code here.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
jonburdo: The fix: https://github.com/kubeflow/model-registry/pull/1767
Just happened to notice this :)

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
manaswinidas: tests haven't been added for this - we have a followup bullet to add this - this section is going to undergo changes as a part of Jounce refactor

YuliaKrimerman:
@manaswinidas tests are failing

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/WorkloadTypeFilter.tsx
manaswinidas: nvm you are right!
This is where the tests were failing - so added it. Thanks.

manaswinidas:
@YuliaKrimerman Just pushed the fix - should pass now. Thanks.

mturley:
@manaswinidas approved, but you have a merge conflict now ☹️ sorry

manaswinidas (commented):
/lgtm

PR #1869, opened 2025-11-11
chore: bump kserve dep to 0.16

Al-Pragliola:
## Description

In this PR I bump kserve dep to 0.16, in this version kserve added a new function to implement the interface kserve.Provider

```go
UploadObject(bucket string, key string, object []byte) error
```

as we can see in the function signature, we are missing info about the URI so we cannot extract the underlying provider to call the provider UploadObject function, for this reason (for now) we return an error when trying to call this function

```go
// UploadObject cannot be implemented for the model-registry provider because:
//  1. Unlike DownloadModel which receives a model-registry URI to resolve, UploadObject
//     receives raw bucket/key parameters with no model-registry context
//  2. There's no way to determine which underlying KServe provider to delegate to
func (p *ModelRegistryProvider) UploadObject(bucket string, key string, object []byte) error {
	return fmt.Errorf("uploading objects is currently not supported when using the model-registry protocol")
}
```

a possible solution for our use case could be proposing to add a new function to the kserve interface, a `GetURI() error`

## How Has This Been Tested?

locally run CSI tests and CI

## Merge criteria:


tarilabs:
do we _have to_ implement the `UploadObject` tho?

I think for all intent and purposes, we're typically interested only in the `DownloadModel`. Did you have a specific new use-case in mind? 🤔

Al-Pragliola:
> do we _have to_ implement the `UploadObject` tho?
> 
> I think for all intent and purposes, we're typically interested only in the `DownloadModel`. Did you have a specific new use-case in mind? 🤔

no, we must implement `UploadObject` because of https://github.com/kserve/kserve/blob/master/pkg/agent/storage/provider.go#L19-L22

without that function we can't have mr csi as a new kserve provider

in kserve 0.15.x the interface didn't include that function: https://github.com/kserve/kserve/blob/v0.15.2/pkg/agent/storage/provider.go#L19-L21

jonburdo (approved):
/lgtm

internal/csi/storage/modelregistry_provider.go
```diff
@@ -197,3 +197,11 @@ func (*ModelRegistryProvider) extractProtocol(storageURI string) (kserve.Protoco
...
 	return "", ErrProtocolNotSupported
 }
+
+// UploadObject cannot be implemented for the model-registry provider because:
+//  1. Unlike DownloadModel which receives a model-registry URI to resolve, UploadObject
+//     receives raw bucket/key parameters with no model-registry context
+//  2. There's no way to determine which underlying KServe provider to delegate to
+func (p *ModelRegistryProvider) UploadObject(bucket string, key string, object []byte) error {
+	return fmt.Errorf("uploading objects is currently not supported when using the model-registry protocol")
+}
```
tarilabs, line 207: to me this solution is correct for the foreseable time being, and beyond the technical feasibility once the remaining details are clarified with KServe, we can eventually consider additional user-flows, thanks a lot again @Al-Pragliola 

/lgtm
/approve

PR #1874, opened 2025-11-13
enforce type-checking with mypy

jonburdo:
This ensures we don't silently ignore mypy errors in CI. It also adds mypy to the `make lint` command and fixes typing issues.

## Merge criteria:

If you have UI changes


clients/python/tests/conftest.py
```diff
@@ -12,7 +12,7 @@
 from unittest.mock import Mock, patch
 
 import pytest
-import requests
+import requests  # type: ignore[import-untyped,unused-ignore]
```
jonburdo, line 15: Ideally we should add `types-requests` to the dev dependencies group just under `types-python-dateutil`, but I hit dependency resolution errors with poetry when trying to add it.

clients/python/pyproject.toml
```diff
@@ -135,3 +136,13 @@ show_column_numbers = true
 show_error_codes = true
 show_error_context = true
 ignore_missing_imports = true
+check_untyped_defs = true
```
jonburdo, line 139: Without this, mypy will not check for type consistency in the bodies of functions that are not typed.

.github/workflows/python-tests.yml
```diff
@@ -61,12 +61,7 @@ jobs:
       - name: Nox lint
         working-directory: clients/python
         run: |
-          if [[ ${{ matrix.session }} == "mypy" ]]; then
-            nox --python=${{ matrix.python }} ||\
-              echo "::error title='mypy failure'::Check the logs for more details"
-          else
-            nox --python=${{ matrix.python }}
-          fi
+          nox --python=${{ matrix.python }}
```
jonburdo, line 64: This `|| echo ...` catches the non-zero exit status, so the step passes even if mypy fails. We don't need this custom error message, so this can be simplified.

clients/python/src/model_registry/_client.py
```diff
@@ -404,14 +413,14 @@ def update(self, model: TModel) -> TModel:
         if not model.id:
             msg = "Model must have an ID"
             raise StoreError(msg)
-        if not isinstance(model, get_args(ModelTypes)):
-            msg = f"Model must be one of {get_args(ModelTypes)}"
+        if not isinstance(model, BaseResourceModel):
+            msg = f"Model must be an instance of {BaseResourceModel.__name__} or a subclass"
```
jonburdo, line 417: One of the few functional changes in this PR. Using `get_args` like this is fragile and not really want we want.

clients/python/pyproject.toml
```diff
@@ -135,3 +136,13 @@ show_column_numbers = true
...
+warn_unused_configs = true
+warn_redundant_casts = true
+warn_unused_ignores = true
```
jonburdo, line 142: We want to clean up unused `type: ignore[*]` comments. In certain cases we have `type: ignore[something,unused-ignore]` only because certain errors are sensitive to the exact libraries installed or python version. For example, mypy may error with `import-untyped` on `import requests` if `types-requests` is not installed.

With an up-to-date environment `make lint` should generally be pretty consistent with the way CI runs mypy, but it's not running in an identical environment.

clients/python/pyproject.toml
tarilabs: sgtm

.github/workflows/python-tests.yml
tarilabs: concur, I believe we had it to start having some harness before opting in completely

PR #1878, opened 2025-11-18
Add OpenAPI spec and STUB endpoints for model catalog settings

ppadti:
## Description
This PR aims to add openapi spec, hanlders, types and mocks for model catalog settings.

## How Has This Been Tested?
run the application in mock mode
`make dev-start`

try these curl commands:

get all source configs:

```
curl -X GET "http://localhost:4000/api/v1/settings/model_catalog/source_configs?namespace=kubeflow" -H "kubeflow-userid: user@example.com"
```

add a catalog source config:

```
curl -X POST "http://localhost:4000/api/v1/settings/model_catalog/source_configs?namespace=kubeflow" -H "kubeflow-userid: user@example.com" -H "Content-Type: application/json" -d '{"data":{"id":"test-catalog","name":"Test Catalog","type":"yaml"}}' 
```

get a catalog source config:
```
curl -s "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com" 
```

Update a catalog source config:

```
curl -X PATCH "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com" -H "Content-Type: application/json" -d '{"data":{"id":"test-catalog","name":"Updated Catalog","type":"yaml"}}'
```

Delete a catalog source config:

```
curl -X DELETE "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com"
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/api/openapi/mod-arch.yaml
+++ b/clients/ui/api/openapi/mod-arch.yaml
@@ -84,7 +84,7 @@ paths:
         - $ref: "#/components/parameters/kubeflowUserId"
       responses:
         "200":
-          $ref: "#/components/responses/ModelRegistryRespone"
+          $ref: "#/components/responses/ModelRegistryResponse"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "500":
@@ -978,62 +978,6 @@ components:
           type: boolean
           example: true
 
-    BaseModel:
-      type: object
-      properties:
-        description:
-          type: string
-          description: Human-readable description of the model.
-        readme:
-          type: string
-          description: Model documentation in Markdown.
-        maturity:
-          type: string
-          description: Maturity level of the model.
-          example: Generally Available
-        language:
-          type: array
-          description: List of supported languages (https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes).
-          items:
-            type: string
-          example:
-            - en
-            - es
-            - cz
-        tasks:
-          type: array
-          description: List of tasks the model is designed for.
-          items:
-            type: string
-          example:
-            - text-generation
-        provider:
-          type: string
-          description: Name of the organization or entity that provides the model.
-          example: IBM
-        logo:
-          type: string
-          format: uri
-          description: |-
-            URL to the model's logo. A [data
-            URL](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data)
-            is recommended.
-        license:
-          type: string
-          description: Short name of the model's license.
-          example: apache-2.0
-        licenseLink:
-          type: string
-          format: uri
-          description: URL to the license text.
-        libraryName:
-          type: string
-          example: transformers
-        customProperties:
-          description: User provided custom properties which are not defined by its type.
-          type: object
-          additionalProperties:
-            $ref: "#/components/schemas/MetadataValue"
     ModelRegistry:
       required:
         - name
@@ -2194,31 +2138,28 @@ components:
         isDefault:
           type: boolean
           example: true
-        properties:
-          type: object
-          properties:
-            yamlCatalogPath:
-              type: string
-              example: sample-catalog.yaml
-            apiKey:
-              type: string
-              description: HuggingFace access token - should be stored in a Kubernetes secret and referenced here, or set via environment variable
-              example: "apiKey"
-            includedModels:
-              type: array
-              items:
-                type: string
-                example: rhelai1/modelcar-granite-7b-starter
-            excludedModels:
-              type: array
-              items:
-                type: string
-                example: model-a:1.0
-                description: list of models name, that will be excluded.
-            allowedOrganization:
-              type: string
-              description: name of the organization that need to ingest from the source.
-              example: org1/
+        yaml:
+          type: string
+          description: Complete YAML catalog content for yaml-type sources.
+        apiKey:
+          type: string
+          description: API key for authenticating with the catalog source.
+          example: "apiKey"
+        includedModels:
+          type: array
+          items:
+            type: string
+            example: rhelai1/modelcar-granite-7b-starter
+        excludedModels:
+          type: array
+          items:
+            type: string
+            example: model-a:1.0
+            description: list of models name, that will be excluded.
+        allowedOrganization:
+          type: string
+          description: name of the organization that need to ingest from the source.
+          example: org1
 
   responses:
     NotFound:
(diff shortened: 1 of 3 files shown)


clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -864,6 +977,63 @@ components:
         clusterAdmin:
           type: boolean
           example: true
+
+    BaseModel:
```
mturley: This BaseModel schema doesn't appear to be used anywhere, and we also already have CatalogBaseModel. But either way I'm not sure we need those for settings - we'll just take the user's YAML and pass it through as a string to be validated on the backend, we don't parse stuff out of it until we get it back out of the catalog API

clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -1972,6 +2142,84 @@ components:
...
+            type: string
+        isDefault:
+          type: boolean
+          example: true
+        properties:
+          type: object
+          properties:
+            yamlCatalogPath:
+              type: string
+              example: sample-catalog.yaml
```
mturley: yamlCatalogPath won't be useful for the UI, I think we should remove it. It is something we'll need to abstract away with how we store catalog YAML in the configmap.

Instead we need a way to pass/retrieve the full YAML contents of a yaml-type source. That could just be a top-level `yaml` or `yamlCatalog` field in our object, but in the actual underlying configmap data that yaml won't be stored in the individual catalog source entry, it will need to be stored as a separate "file" in the configmap and referenced in `yamlCatalogPath`. [See this thread](https://redhat-internal.slack.com/archives/C08G31WCV16/p1763498197266429?thread_ts=1763402216.720219&cid=C08G31WCV16) - the actual configmap might look something like this:

```yaml
sources.yaml:
  catalogs:
  - id: foo
    [...]
    properties:
      yamlCatalogPath: foo.yaml
foo.yaml: <the YAML file>
```

We don't want the UI dealing with this weird configmap structure, so the UI<->BFF communication layer (the CatalogSourceConfig object) should just take the raw yaml itself. When we get to actually implementing the configmap CRUD calls, we'll need to have the BFF manage populating/updating those records like this `foo.yaml` outside the catalogs array and referencing them in the `yamlCatalogPath` within the configmap. That's what I meant when I talked about "converting" back and forth between CatalogSourceConfig and configmap structure, we need some abstraction.

Also, the `yaml`/`yamlCatalog` property you add should be optional - it won't be used for huggingface sources. It would be nice if there was a way for openapi types to have a "discriminated union" like typescript, i.e. when type === yaml it is required, but I'm not sure if that's a thing.

I'm also questioning the nesting of things under `properties` - I know that's matching what the configmap has, so maybe it's simplest to leave it that way, but since we already need this object to look different from the configmap and need to convert it back and forth, we can structure it however is most convenient for the UI.

clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -1972,6 +2142,84 @@ components:
...
+            apiKey:
+              type: string
+              description: HuggingFace access token - should be stored in a Kubernetes secret and referenced here, or set via environment variable
+              example: "apiKey"
```
mturley: The description here isn't quite right for this layer. This apiKey is another example of how we'll need the BFF to provide abstraction - the UI should actually directly pass the value here, and then later when the BFF updates the configmap it will need to deal with creating/updating a k8s secret and referencing the secret in the configmap. Then when the BFF is fetching a CatalogSourceConfig it will need to go fetch that secret and return its value directly in the CatalogSourceConfig.

clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -1972,6 +2142,84 @@ components:
...
+            excludedModels:
+              type: array
+              items:
+                type: string
+                example: model-a:1.0
+                description: list of models name, that will be excluded.
+            allowedOrganization:
+              type: string
+              description: name of the organization that need to ingest from the source.
+              example: org1/
```
mturley: The allowedOrganization won't need a trailing slash, I think we can remove that from the example.

clients/ui/bff/internal/mocks/model_catalog_settings_mock.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if sourceName == "" {
+		return nil, fmt.Errorf("source name is required")
+	}
+	if sourceId == "" {
+		return nil, fmt.Errorf("source ID is required")
+	}
+	if sourceType == "" {
+		return nil, fmt.Errorf("source type is required")
+	}
```
mturley, line 44: It doesn't have to be part of this PR, but I wonder if we want to validate the required fields specific to each type. like for example a yaml-type source will require its yaml contents, and a huggingface-type source will require its allowedOrganization and apiKey. We could also bail if the type isn't one of those two types.

mturley (changes requested):
Looking good, just a couple changes to the types to better set us up for the abstraction we need.

clients/ui/api/openapi/mod-arch.yaml
ppadti: oh, sorry... this was from initial commit long back... I will clean the unused things from this file... Apologies, missed this.

clients/ui/bff/internal/mocks/model_catalog_settings_mock.go
ppadti: I just added a few checks in here for mock, we will need to do the actual validation in `repositories/model_catalog_settings.go`

clients/ui/api/openapi/mod-arch.yaml
ppadti: >I'm also questioning the nesting of things under properties - I know that's matching what the configmap has, so maybe it's simplest to leave it that way, but since we already need this object to look different from the configmap and need to convert it back and forth, we can structure it however is most convenient for the UI.

yeah, that makes sense, as we are going to convert the configMap to CatalogSourceConfig - I will remove the nesting part for properties. While creating this I just thought we will just return the whole configMap entry. :sweat_smile:

mturley: Heh yeah sorry... I thought so too when I wrote up the original story. Thanks for bearing with me with the new info.

clients/ui/bff/internal/mocks/model_catalog_settings_mock.go
mturley: ah right, thanks.

mturley:
/lgtm
/approve

Great stuff, thanks @ppadti ! This will unblock a lot of work.

PR #1880, opened 2025-11-18
Initial Model Catalog Settings Modal

YuliaKrimerman:
## Description

Implemented the complete "Add a source" / "Manage source" form for Model Catalog Settings, allowing users to configure and manage catalog sources.

**Key Features:**
- Full form implementation with all required fields (Name, Source type, Credentials, YAML content, Model visibility filters, Enable source toggle)
- Two source types supported: Hugging Face repository and YAML file
- Dynamic form validation with inline error messages
- Two-column layout with live preview panel
- Three synchronized preview buttons (action group, header, and panel)
- Password reveal/hide toggle for access token field
- Expandable Model visibility section with filter controls
- Dynamic microcopy that updates based on organization input

**UI Screenshots:**
Screenshot 2025-11-18 at 2 40 38 PM
Screenshot 2025-11-18 at 2 40 47 PM

In Patternfly Theme

https://github.com/user-attachments/assets/a7901b9e-2c91-4927-875a-e48d9101df12

## How Has This Been Tested?

- **Manual Testing:** Fully tested all form interactions, validation states, source type switching, and preview panel behavior in local development environment
- **Automated Testing:** Added comprehensive Cypress e2e tests covering:
## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogSettings.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogSettings.cy.ts
@@ -205,16 +205,16 @@ describe('Manage Source Page', () => {
 
     it('should expand and collapse model visibility section', () => {
       manageSourcePage.visitAddSource();
-      manageSourcePage.findAllowedModelsInput().should('not.be.visible');
-      manageSourcePage.findExcludedModelsInput().should('not.be.visible');
+      manageSourcePage.findAllowedModelsInput().should('not.exist');
+      manageSourcePage.findExcludedModelsInput().should('not.exist');
 
       manageSourcePage.toggleModelVisibility();
-      manageSourcePage.findAllowedModelsInput().should('be.visible');
-      manageSourcePage.findExcludedModelsInput().should('be.visible');
+      manageSourcePage.findAllowedModelsInput().should('exist');
+      manageSourcePage.findExcludedModelsInput().should('exist');
 
       manageSourcePage.toggleModelVisibility();
-      manageSourcePage.findAllowedModelsInput().should('not.be.visible');
-      manageSourcePage.findExcludedModelsInput().should('not.be.visible');
+      manageSourcePage.findAllowedModelsInput().should('not.exist');
+      manageSourcePage.findExcludedModelsInput().should('not.exist');
     });
 
     it('should allow entering filter values', () => {


clients/ui/frontend/src/app/pages/modelCatalogSettings/components/YamlSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+      name="yaml-content"
+      data-testid="yaml-content-input"
+      value={formData.yamlContent}
+      onChange={(_event, value) => setData('yamlContent', value)}
+      onBlur={() => setIsYamlTouched(true)}
+      validated={isYamlTouched && !isYamlContentValid ? 'error' : 'default'}
+      rows={10}
+      resizeOrientation="vertical"
+    />
+  );
```
mturley, line 74: This yaml input field should use the PatternFly file upload component. See the "Restricting file size and type" example for how to limit it to only letting you select YAML files. The title of the field should be "Upload a YAML file" instead of just YAML.

https://www.patternfly.org/components/file-upload/simple-file-upload#restricting-file-size-and-type

Image

Also [looking at the final design](https://www.figma.com/design/eHE1KV9bOdKqvgURAIvRH6/Model-Filters-and-Admin-settings-of-MC?node-id=2433-139687&p=f&m=dev), we now have a "view expected file format" button that opens a sidebar, this was not present when I broke down the stories... I need to comb through and look for differences and update the stories 🤦 - we don't have to do that as part of this issue, I'll open a followup story for it that I don't think needs to block the epic.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ModelVisibilitySection.tsx
```diff
@@ -0,0 +1,130 @@
...
+    <ExpandableSection
+      toggleText={FORM_LABELS.MODEL_VISIBILITY}
+      onToggle={onToggle}
+      isExpanded={isExpanded}
+      data-testid="model-visibility-section"
+    >
+      <FormSection>
+        <FormHelperText>
+          <HelperText>
+            <HelperTextItem>{sectionDescription}</HelperTextItem>
```
mturley: In the design, this section description appears even when the expandable section is collapsed, and it is indented to be part of the toggle. I think we can maybe include it in the toggle itself by using `toggleContent` instead of `toggleText`. If that doesn't work out, what we could do is have this be a "detached expandable section" - see [this example](https://www.patternfly.org/components/expandable-section#detached). You can separate the toggle from the content it shows and hides, and we can have the section description appear in between the toggle and the content so it's always visible. Let's play around with these 2 approaches to see if we can get it to look closer to the design.

Image

I think the difference from this design is that the "Model visibility" text should be blue like it is in your version because that's the PF behavior to indicate it's clickable. If you go for the toggleContent approach you may want to override the color of the description text using utility classes so the whole thing isn't blue.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  onToggle={() => setIsFiltersExpanded((prev) => !prev)}
+                  setData={setData}
+                />
+              </StackItem>
+
+              <StackItem>
+                <FormSection>
+                  <FormGroup fieldId="enable-source">
+                    <Checkbox
+                      label={FORM_LABELS.ENABLE_SOURCE}
```
mturley: The label for this checkbox should be bold, you can pass in JSX to `label` for that. (it serves as the field title rather than being an option within a checkbox group).

Image

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/PreviewPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+        configuration, complete all required fields, then click <strong>Preview</strong>.
+      </EmptyStateBody>
+      <EmptyStateFooter>
+        <EmptyStateActions>
+          <PreviewButton
+            onClick={onPreview}
+            isDisabled={!isPreviewEnabled}
+            variant="primary"
+            testId="preview-button-panel"
+          />
```
mturley, line 58: In the empty state here the preview button should be the `"link"` variant. I notice that your `variant` prop in PreviewButton is just `'primary' | 'secondary'`, I think it can be `ButtonProps['variant']` to allow for other variants like this.

Image

The reason it looks like the primary style in a lot of the mocks is because a link button, when disabled, still gets that full grey background.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/PreviewPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const PreviewPanel: React.FC<PreviewPanelProps> = ({ isPreviewEnabled, onPreview }) => (
+  <div
+    data-testid="preview-panel"
+    style={{
+      borderLeft: '1px solid #d2d2d2',
+      paddingLeft: '24px',
+      height: '100%',
+    }}
```
mturley: We need to avoid inline styles.

You will actually get some of these styles for free if you use the Sidebar component instead of using a Grid around this preview panel. The Sidebar has props for `hasBorder`, `hasGutter` and `hasPadding`. Look at the various examples here: https://www.patternfly.org/components/sidebar
Note that some examples have the sidebar on the left, you'll also need `isPanelRight` like this example: https://www.patternfly.org/components/sidebar#panel-right-with-gutter

I'm not sure what that will do to the width of the sidebar, but we can probably manage that with sizing [utility classes](https://www.patternfly.org/utility-classes/about-utility-classes).

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  onToggle={() => setIsFiltersExpanded((prev) => !prev)}
+                  setData={setData}
+                />
+              </StackItem>
+
+              <StackItem>
+                <FormSection>
+                  <FormGroup fieldId="enable-source">
+                    <Checkbox
+                      label={<strong>{FORM_LABELS.ENABLE_SOURCE}</strong>}
```
mturley: `<strong>` is actually too bold 😆 it isn't looking consistent with the other labels - sorry, can you actually replace it with a `<span className="pf-v6-c-form__label-text">`? that should make it look consistent with the other form labels (I guess it's a medium-weight font)

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ModelVisibilitySection.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  return (
+    <>
+      <ExpandableSection
+        toggleText={FORM_LABELS.MODEL_VISIBILITY}
+        onToggle={onToggle}
+        isExpanded={isExpanded}
+        contentId="model-visibility-content"
+        data-testid="model-visibility-section"
+      />
```
mturley: Ok. So. Don't kill me -- the expanded content (and the section description) needs to be indented to match the design, and when I went looking in the docs for how we can do this with form styles I discovered Field Groups, which are expandable sets of form fields. I think that's actually what we need instead of using ExpandableSections at all. Check out this example: https://www.patternfly.org/components/forms/form#form-group-with-additional-label-info

In the code there you can see a `FormFieldGroupExpandable` component with a `FormFieldGroupHeader` that includes a place for the `titleText` and `titleDescription`, and indents its children correctly. This looks exactly like the design, I was steering you the wrong way.

Image

Image

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+                      isChecked={formData.enabled}
+                      onChange={(_event, checked) => setData('enabled', checked)}
+                    />
+                  </FormGroup>
+                </FormSection>
+              </StackItem>
+            </Stack>
+          </Form>
+        </SidebarContent>
+        <SidebarPanel hasPadding className="pf-v6-u-flex-basis-40">
```
mturley: All you need here is the `width` prop on SidebarPanel. [You can set different widths by screen size](https://www.patternfly.org/components/sidebar#responsive-panel-width), but I think we're fine just having it always be 50%.

```suggestion
        <SidebarPanel hasPadding width={{ default: 'width_50' }}>
```

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceFormFooter.tsx
```diff
@@ -0,0 +1,85 @@
...
+  submitLabel,
+  submitError,
+  isSubmitDisabled,
+  isSubmitting,
+  onSubmit,
+  onCancel,
+  isPreviewDisabled,
+  onPreview,
+}) => (
+  <PageSection stickyOnBreakpoint={{ default: 'bottom' }} style={{ boxShadow: 'none' }}>
```
mturley: I think we can remove the style here. I see that `stickyOnBreakpoint` forces a shadow (and I'm actually surprised to see that the `shadowOnTop={false}` prop doesn't override that... that's maybe a PF bug), but  actually letting the shadow be there is consistent with the design for the register form. It looks weird without the shadow if the content overflows into the buttons' padding. I know that's not in the design, but let's opt for consistency.

clients/ui/frontend/src/app/pages/modelCatalogSettings/screens/ManageSourcePage.tsx
mturley: @YuliaKrimerman I realized this is also different from the design - the icon isn't present on this form page, this title can just be a string.

PR #1885, opened 2025-11-19
feat: general include exclude models field in sources file

Al-Pragliola:
## Description

This PR introduces standardized model inclusion/exclusion filtering that works consistently across all catalog source types (YAML, Hugging Face, etc.).

## How Has This Been Tested?

local testing plus unit/integration tests

## Merge criteria:

### DIFF
--- a/api/openapi/catalog.yaml
+++ b/api/openapi/catalog.yaml
@@ -507,6 +507,9 @@ components:
             Optional allow-list of models that are eligible for this source. Entries can be
             exact model names or patterns that use `*` as a wildcard. When provided, only
             models matching at least one pattern are considered.
+
+            Pattern matching is case-insensitive, so `Granite/*` will match `granite/model`,
+            `Granite/model`, and `GRANITE/model`.
           type: array
           items:
             type: string
@@ -514,6 +517,9 @@ components:
           description: |-
             Optional block-list of models that should be removed from the catalog even if
             they match `includedModels`. Patterns support the `*` wildcard.
+
+            Pattern matching is case-insensitive, so `*-beta` will match `Model-Beta`,
+            `model-beta`, and `MODEL-BETA`.
           type: array
           items:
             type: string
--- a/api/openapi/src/catalog.yaml
+++ b/api/openapi/src/catalog.yaml
@@ -390,6 +390,9 @@ components:
             Optional allow-list of models that are eligible for this source. Entries can be
             exact model names or patterns that use `*` as a wildcard. When provided, only
             models matching at least one pattern are considered.
+
+            Pattern matching is case-insensitive, so `Granite/*` will match `granite/model`,
+            `Granite/model`, and `GRANITE/model`.
           type: array
           items:
             type: string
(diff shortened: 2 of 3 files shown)


catalog/internal/catalog/yaml_catalog.go
```diff
@@ -378,7 +364,7 @@ func (p *yamlModelProvider) read() (*yamlCatalog, error) {
 func (p *yamlModelProvider) emit(ctx context.Context, catalog *yamlCatalog, out chan<- ModelProviderRecord) {
 	done := ctx.Done()
 	for _, model := range catalog.Models {
-		if _, excluded := p.excludedModels[model.Name]; excluded {
+		if p.filter != nil && !p.filter.Allows(model.Name) {
```
pboyd: You added a nil check in `Allows`, if you wanted to simplify this:

```suggestion
		if !p.filter.Allows(model.Name) {
```

catalog/internal/catalog/model_filter.go
```diff
@@ -0,0 +1,180 @@
...
+
+func newCompiledPattern(field string, idx int, raw string) (*compiledPattern, error) {
+	value := strings.TrimSpace(raw)
+	if value == "" {
+		return nil, fmt.Errorf("%s[%d]: pattern cannot be empty", field, idx)
+	}
+
+	// Convert a simple glob (only supporting '*') into a regexp.
+	var b strings.Builder
+	b.WriteString("^")
```
pboyd: What about making this case-insensitive? This will likely be user-provided strings, and I don't think they'll think about the difference between `llama/*` and `Llama/*`.

```suggestion
	b.WriteString("(?i)^")
```

pboyd (approved):
The this looks really good. I was only expecting a prefix match, but your regex solution is much better.

/lgtm

api/openapi/catalog.yaml
```diff
@@ -502,6 +502,21 @@ components:
           type: array
           items:
             type: string
+        includedModels:
+          description: |-
+            Optional allow-list of models that are eligible for this source. Entries can be
+            exact model names or patterns that use `*` as a wildcard. When provided, only
+            models matching at least one pattern are considered.
```
jonburdo, line 509: To confirm, we want this to be case-insensitive? Maybe worth mentioning that here. I would generally assume a glob is case-sensitive like it would be in a shell.

jonburdo (approved):
/lgtm as long as we do indeed want this to be case-insensitive

api/openapi/catalog.yaml
Al-Pragliola: yeah making it case insensitive makes totally sense in my opinion in our intended use case, users probably don't even know the exact naming of the models that are needed to be excluded/included or what a glob is

jonburdo: Sounds good. We should just make sure to document that

Al-Pragliola: yes thank you @jonburdo , please check the latest commit https://github.com/kubeflow/model-registry/pull/1885/commits/c049854306a645bde670251e03ae179cd5b780c5

PR #1894, opened 2025-11-20
feat: bump go to 1.25.3

Al-Pragliola:
## Description

Bump golang version to 1.25.3

## How Has This Been Tested?

Local build with docker / binary

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,11 +173,11 @@ bin/envtest:
 
 GOLANGCI_LINT ?= ${PROJECT_BIN}/golangci-lint
 bin/golangci-lint:
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(PROJECT_BIN) v2.0.2
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(PROJECT_BIN) v2.6.2
 
 GOVERTER ?= ${PROJECT_BIN}/goverter
 bin/goverter:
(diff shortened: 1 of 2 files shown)


gorm-gen/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubeflow/model-registry/gorm-gen
 
-go 1.24.4
+go 1.25.3
```
pboyd, line 3: Unless there's a need for a specific minor version, I'd suggest we leave it off:


```suggestion
go 1.25
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubeflow/model-registry
 
-go 1.24.4
+go 1.25.3
```
pboyd, line 3: Here too:

```suggestion
go 1.25
```

gorm-gen/go.mod
lugi0: Do you mean the patch version @pboyd ? i.e. the `.3`? why do you prefer leaving it off? In this case it's just a floor on the compiler version, which is what we want given what moved us to update to 1.25, no?

Al-Pragliola:
- https://github.com/kubeflow/model-registry/pull/1894#pullrequestreview-3489105145

/approve

gorm-gen/go.mod
pboyd: Generally, there's no hard requirement for a specific patch version (not minor--sorry). It is just a floor on the supported version, and since there's nothing but bugfixes in the patch releases there's no reason that we can't build on 1.25.0 just as well as 1.25.3.

PR #1921, opened 2025-11-25
Add submit logic to add/manage source form

ppadti:
## Description
This PR aims to add submit logic to add/manage source form

## How Has This Been Tested?
1. Go to Model catalog settings page, click on Add source button
2. fill the all required fields, click Add button
3. Check that the POST call has been made and its successful and the url will redirect to MC settings list page
4. For update, click on Manage source - it will redirect to manage source form with pre-filled data.
5. update the data, click on Save button
6. Check that the PATCH call is made and its successful. the url will redirect to MC settings page

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/__mocks__/mockCatalogSourceConfigList.ts
```diff
@@ -1,70 +1,69 @@
...
   partial?: Partial<YamlCatalogSourceConfig>,
 ): YamlCatalogSourceConfig => ({
-  id: 'yaml-source-1',
-  name: 'Red Hat AI',
+  id: 'sample-source-1',
+  name: 'Source 1',
   type: CatalogSourceType.YAML,
   enabled: true,
-  labels: ['Red Hat AI'],
+  //   labels: ['category-1'],
```
ppadti: @mturley I just had a question - how are we going to add these labels? we don't have any specific fields to add this.

mturley: @ppadti I think for these custom sources, for now we can just not include any labels. That will make them appear in the "Community and custom models" tab. In the future we may want an enhancement to the settings page for admins to create/edit labels and apply them to their sources, but that's out of scope.

mturley: Let's not merge the commented out code, maybe for this mock function we exclude it and if we need labels we can pass them in - they should probably still be present for default sources, although it doesn't really matter yet on the settings side

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/CredentialsSection.tsx
```diff
@@ -33,6 +39,27 @@ const CredentialsSection: React.FC<CredentialsSectionProps> = ({ formData, setDa
 
   const isOrganizationValid = validateOrganization(formData.organization);
   const isAccessTokenValid = validateAccessToken(formData.accessToken);
+  const [validationError, setValidationError] = React.useState<Error | undefined>(undefined);
+  const [isValidating, setIsValidating] = React.useState(false);
+  const [isValidationSuccess, setIsValidationSuccess] = React.useState(false);
+  const [showAlert, setShowAlert] = React.useState<boolean>(false);
```
mturley: These states are redundant, I think we can just have the alert shown when `isValidationSuccess` is true and set it to false when the alert should hide. We don't block submission on this validation state so it only matters for the alert case.

Also the actual underlying validate logic will be the same as running a Preview once we do that, so we'll need to adapt this code later to make sure we're properly reusing things and the only difference is that in this case we also show the validation success message here on success. Not an issue for this PR.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ModelVisibilitySection.tsx
```diff
@@ -55,8 +53,8 @@ const ModelVisibilitySection: React.FC<ModelVisibilitySectionProps> = ({ formDat
       id="allowed-models"
       name="allowed-models"
       data-testid="allowed-models-input"
-      value={formData.allowedModels}
-      onChange={(_event, value) => setData('allowedModels', value)}
+      value={formData.allowedModels.join(', ')}
+      onChange={(_event, value) => setData('allowedModels', [value])}
```
mturley: Wrapping this value as a single array element `[value]` is probably incorrect behavior - if we want to store the value as an array we need to be splitting on comma to update the state, but that'll be strange because as the user types a comma and space the split behavior will either not let them type a space or automatically insert the space... I wonder if we want the form data to just have this stored as a plain string, and have the join/split behavior only be when we convert to/from the API objects?

clients/ui/frontend/src/app/pages/modelCatalogSettings/utils/modelCatalogSettingsUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+  const commaSeparatedString = arr[0] || '';
+  if (!commaSeparatedString.trim()) {
+    return [];
+  }
+
+  return commaSeparatedString
+    .split(',')
+    .map((item) => item.trim())
+    .filter((item) => item.length > 0);
+};
```
mturley: Yeah this feels weird. We have an array of values but one array element could represent multiple comma-separated values? Seems like it would be simpler to just store the input the way the user types it and perform a simple split or join in your `catalogSourceConfigToFormData` and `transformFormDataToPayload` utils.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/CredentialsSection.tsx
```diff
@@ -85,21 +112,55 @@ const CredentialsSection: React.FC<CredentialsSectionProps> = ({ formData, setDa
...
+                  {VALIDATION_MESSAGES.ACCESS_TOKEN_REQUIRED}
+                </HelperTextItem>
+              </HelperText>
+            </FormHelperText>
+          )}
+          {validationError && (
+            <StackItem>
+              <Alert isInline variant="danger" title="Validation failed">
+                The system cannot establish a connection to the source. Ensure that the organization
+                and access tokena are accurate, then try again.
```
mturley: Typo: `tokena` -> `token`

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/CredentialsSection.tsx
```diff
@@ -85,21 +112,55 @@ const CredentialsSection: React.FC<CredentialsSectionProps> = ({ formData, setDa
...
+          )}
+          {isValidationSuccess && showAlert && (
+            <StackItem>
+              <Alert
+                isInline
+                variant="success"
+                title="Validation successful"
+                actionClose={<AlertActionCloseButton onClose={() => setShowAlert(false)} />}
+              >
+                Tha organisation and accessToken are valid for connection.
```
mturley: Spelling: `organisation` -> `organization`. Technically both are correct in different regions but elsewhere we use the US spelling of "organization".

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/CredentialsSection.tsx
```diff
@@ -85,21 +112,55 @@ const CredentialsSection: React.FC<CredentialsSectionProps> = ({ formData, setDa
...
-        {isAccessTokenTouched && !isAccessTokenValid && (
-          <FormHelperText>
-            <HelperText>
-              <HelperTextItem variant="error" data-testid="access-token-error">
-                {VALIDATION_MESSAGES.ACCESS_TOKEN_REQUIRED}
-              </HelperTextItem>
-            </HelperText>
-          </FormHelperText>
-        )}
+        <Stack hasGutter>
```
mturley: The Stack here means the FormHelperText with the validation error on the access token field has a gap (gutter) between it and the field, which is not true of the Organization field. Do we need a stack here? or can the alert(s) just go below the FormHelperText, perhaps with a margin on top? They could even be outside the FormGroup if necessary.

Image

clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -975,6 +975,7 @@ paths:
       tags:
         - ModelCatalogService
       parameters:
+        - $ref: "#/components/parameters/kubeflowUserId"
```
mturley, line 978: Was this intentional? not sure where it came from or if it's documented somewhere.

clients/ui/frontend/src/app/pages/modelCatalogSettings/utils/modelCatalogSettingsUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+  };
+};
+
+export const generateSourceIdFromName = (name: string): string =>
+  name
+    .trim()
+    .replace(/\s+/g, '_')
+    .replace(/-/g, '_')
+    .replace(/[^a-zA-Z0-9_]/g, '')
+    .toLowerCase();
```
mturley, line 44: This could use a couple unit tests. Perhaps also `catalogSourceConfigToFormData` and `transformFormDataToPayload`, although those are more straightforward.

clients/ui/api/openapi/mod-arch.yaml
ppadti: yeah, we missed adding this here for this endpoint. Yes, its [mentioned here](https://github.com/kubeflow/model-registry/blob/main/clients/ui/bff/README.md#sample-local-calls)

PR #1922, opened 2025-11-25
fix: filterQuery AND condition not working correctly

Al-Pragliola:
## Description

The bug was in the artifact filtering logic for model queries. When multiple artifact property conditions were combined with AND, each condition created its own independent EXISTS subquery, which incorrectly matched models where:

- One artifact matched condition A
- A different artifact matched condition B

## How Has This Been Tested?

Added a regression test, tested locally

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/db_catalog_filterquery_test.go
+++ b/catalog/internal/catalog/db_catalog_filterquery_test.go
@@ -721,7 +721,8 @@ func TestArtifactFilteringCapability(t *testing.T) {
 				".name = $",
 				".double_value <= $",
 			},
-			expectedArgs: []any{"ttft_mean", float64(90), "tpot_mean", float64(50)},
+			// Args order: JOIN property names first, then WHERE value conditions
+			expectedArgs: []any{"ttft_mean", "tpot_mean", float64(90), float64(50)},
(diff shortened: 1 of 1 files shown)


catalog/internal/catalog/db_catalog_filterquery_test.go
```diff
@@ -1120,6 +1120,22 @@ func TestArtifactFilteringEdgeCases(t *testing.T) {
 			},
 			description: "Should handle both exact and case-insensitive matching on artifact properties (ILIKE uses UPPER for cross-DB compatibility)",
 		},
+		{
+			name:        "Bug fix: multiple artifact filters must match SAME artifact",
+			filterQuery: `artifacts.hardware_type LIKE "H200" AND artifacts.ttft_p95 < 50`,
```
jonburdo: Should we test with more than 2 artifacts? And a more complicated expression like `artifacts.hardware_type LIKE "H200" AND (artifacts.ttft_p95 < 50 OR artifacts.ttft_p95 > 90) AND ...`

jonburdo (approved):
/lgtm - Nice work on this!
It would be great to get another opinion on this.

This does highlight to me: Ideally we'd have more rigorous tests for our sql generation in general (i.e. more complex expressions, validating the tree not just checking string fragments). Just something we might want to think about in the future.

PR #1930, opened 2025-11-27
feat: add overall_average custom property when loading perf data

Al-Pragliola:
## Description

In this PR when loading performance data we also add a new customProperty that maps the `overall_accuracy` from the metadata to the `overall_average` field.

This field was already supported when sorting by ACCURACY.

## How Has This Been Tested?

unit and local testing

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/performance_metrics_test.go
+++ b/catalog/internal/catalog/performance_metrics_test.go
@@ -235,6 +235,98 @@ func TestParseMetadataJSON_OnlyIDMatters(t *testing.T) {
 	}
 }
 
+func TestOverallAccuracyToOverallAverage(t *testing.T) {
+	t.Run("parse overall_accuracy from metadata", func(t *testing.T) {
+		tests := []struct {
+			name      string
+			jsonData  string
+			wantNil   bool
+			wantValue float64
+		}{
+			{
+				name:      "overall_accuracy present",
+				jsonData:  `{"id": "model-1", "overall_accuracy": 85.5}`,
+				wantNil:   false,
+				wantValue: 85.5,
+			},
+			{
+				name:      "overall_accuracy is zero",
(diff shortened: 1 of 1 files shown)


catalog/internal/catalog/performance_metrics_test.go
```diff
@@ -235,6 +235,98 @@ func TestParseMetadataJSON_OnlyIDMatters(t *testing.T) {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Error("overall_average custom property not found in artifact")
+		}
+	})
+
+	t.Run("artifact has no overall_average when overall_accuracy is nil", func(t *testing.T) {
```
fege, line 315: is there a test that sort the artifacts by overall_average if it is present and fallback to sort by id if not?

Al-Pragliola: Hey @fege yes you can find those tests in https://github.com/kubeflow/model-registry/blob/main/catalog/internal/db/service/catalog_model_test.go#L343

pboyd:
Why does this field need special handling? I thought it was just another metric field.

Edit: Oh, wait, I see now. I guess it has to be `overall_average` even though the field is called `overall_accuracy`? Can we require the performance data to use `overall_average`?

Al-Pragliola:
> Why does this field need special handling? I thought it was just another metric field.
> 
> Edit: Oh, wait, I see now. I guess it has to be `overall_average` even though the field is called `overall_accuracy`? Can we require the performance data to use `overall_average`?

I think it's not about the field name, but the location. This field is not in the metrics json file but in the metadata one

pboyd:
> I think it's not about the field name, but the location. This field is not in the metrics json file but in the metadata one

I see. Thanks for the explanation.

/lgtm
/approve

PR #1956, opened 2025-12-03
feat: merge sources when an user overrides one

Al-Pragliola:
## Description

**Problem**
Previously, if the same source ID appeared in multiple config files, the server returned an error. This prevented users from layering configurations (e.g., a base config with defaults and a user config with overrides).
Additionally, when users provided sparse config overrides, the entire source definition was replaced instead of merged, requiring users to duplicate all fields.

**Solution**
Introduced a merge operation for catalog sources. Sources with the same ID across multiple config files are now merged field-by-field, with later files taking precedence. Unspecified fields in overrides are inherited from earlier definitions.

**Example**
```yaml
# base-sources.yaml
catalogs:
  - id: enterprise_models
    name: "Enterprise Models"
    enabled: false
    labels: [enterprise]
    type: yaml
    properties:
    -  yamlCatalogPath: ./models.yaml

# user-sources.yaml (sparse override - no longer an error!)
catalogs:  
  - id: enterprise_models
    enabled: true
    # Just enable - inherit everything else
```

**Result**
Source is enabled with all other fields inherited from base.

## How Has This Been Tested?

unit tests and local env manual testing

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/loader.go
+++ b/catalog/internal/catalog/loader.go
@@ -73,7 +73,7 @@ type Loader struct {
 	paths         []string
 	services      service.Services
 	closersMu     sync.Mutex
-	closers       map[string]func()
+	closer        func() // cancels the current model loading goroutines
 	handlers      []LoaderEventHandler
 	loadedSources map[string]bool // tracks which source IDs have been loaded
 }
@@ -96,7 +96,6 @@ func NewLoader(services service.Services, paths []string) *Loader {
 		Labels:        NewLabelCollection(),
 		paths:         paths,
 		services:      services,
-		closers:       map[string]func(){},
 		loadedSources: map[string]bool{},
 	}
 }
@@ -257,11 +256,10 @@ func (l *Loader) updateDatabase(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(ctx)
 
 	l.closersMu.Lock()
-	// Use a single key since we now load all merged sources at once
-	if l.closers["_all_"] != nil {
-		l.closers["_all_"]()
+	if l.closer != nil {
+		l.closer()
 	}
-	l.closers["_all_"] = cancel
+	l.closer = cancel
 	l.closersMu.Unlock()
 
 	// Use merged sources from SourceCollection instead of per-file config.


catalog/internal/catalog/loader.go
```diff
@@ -134,17 +153,45 @@ func (l *Loader) loadOne(ctx context.Context, path string) error {
...
+	return l.updateLabels(path, config)
+}
+
+// loadAllModels loads models from all merged sources.
+func (l *Loader) loadAllModels(ctx context.Context) error {
+	// Clear the loaded sources tracker for a fresh load
+	l.loadedSources = map[string]bool{}
+
+	// Use the first config path's directory for relative path resolution
+	// (all config files should be in the same directory or use absolute paths)
```
pboyd: Does it have to be this way? I think it will be a problem for YAML sources where the configuration is in multiple configmaps. Each set of files will have to be mounted to a different directory in the deployment, and that directory will also need to be used for the `yamlCatalogPath`.

For instance, if we have a default configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-sources
data:
  sources.yaml: |-
    catalogs:
      - name: Default Source
        id: default_source
        enabled: true
        type: yaml
        properties:
          yamlCatalogPath: default_catalog.yaml
  default_catalog.yaml: "catalog contents..."
```

And then another user-configurable one:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: user-sources
data:
  sources.yaml: |-
    catalogs:
      - name: User Source
        id: user_source
        enabled: true
        type: yaml
        properties:
          yamlCatalogPath: user_catalog.yaml
  user_catalog.yaml: "catalog contents..."
```

Sources from the second configmap will have to use an absolute path. Current plans for the admin UI have basically that structure, but it will need to know the path where the user-sources configmap is mounted on the pod in order to configure it correctly.

catalog/internal/catalog/loader.go
```diff
@@ -224,17 +255,21 @@ func (l *Loader) updateLabels(path string, config *sourceConfig) error {
...
 }
 
-func (l *Loader) updateDatabase(ctx context.Context, path string, config *sourceConfig) error {
+func (l *Loader) updateDatabase(ctx context.Context, configDir string, _ *sourceConfig) error {
 	ctx, cancel := context.WithCancel(ctx)
 
 	l.closersMu.Lock()
-	if l.closers[path] != nil {
-		l.closers[path]()
+	// Use a single key since we now load all merged sources at once
```
pboyd: Should `closers` just be a single function now instead of a map?

catalog/internal/catalog/loader.go
```diff
@@ -94,36 +107,42 @@ func (l *Loader) RegisterEventHandler(fn LoaderEventHandler) {
...
 			}
 
 			for range changes {
 				glog.Infof("Reloading sources %s", path)
-
-				err = l.loadOne(ctx, path)
-				if err != nil {
-					glog.Errorf("unable to load sources: %v", err)
-				}
+				l.reloadAll(ctx)
```
pboyd, line 140: One more thing, now that I look at it. Does `reloadAll` stop the goroutines that were launched with the earlier `loadAllModels` call? I don't see anything that does, but I might be overlooking it. The source providers do an initial push with the available and then do subsequent pushes for updates (e.g. the YAML provider re-sends models if the catalog file changes). So if there's nothing to stop the old goroutines, `reloadAll` is going to keep adding goroutines that never end.

Al-Pragliola: The flow should be reloadAll → loadAllModels → updateDatabase

where the previous goroutine get closed

```go
func (l *Loader) updateDatabase(ctx context.Context) error {
	ctx, cancel := context.WithCancel(ctx)

	l.closersMu.Lock()
	if l.closer != nil {
		l.closer()
	}
	l.closer = cancel
	l.closersMu.Unlock()
```

am I mistaken?

catalog/internal/catalog/loader.go
Al-Pragliola: good catch! It should be addressed https://github.com/kubeflow/model-registry/pull/1956/commits/c0f50b008664ad0b46f9213ef0095ac8b4a92bb4 

now by default it's relative to the config location

catalog/internal/catalog/loader.go
Al-Pragliola: yes, you are right thanks 🙏🏼

catalog/internal/catalog/loader.go
pboyd: My bad, you're right. I see it now.

pboyd:
Thanks for the fixes.

/lgtm
/approve

PR #1957, opened 2025-12-03
chore(catalog): allow empty sources list

pboyd:
## Description

Always return items from the sources endpoint, even if it's an empty list.

## How Has This Been Tested?
On a local dev environment.

## Merge criteria:

### DIFF
--- a/api/openapi/catalog.yaml
+++ b/api/openapi/catalog.yaml
@@ -533,6 +533,8 @@ components:
               type: array
               items:
                 $ref: "#/components/schemas/CatalogSource"
+          required:
+            - items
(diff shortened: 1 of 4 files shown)


Al-Pragliola:
I may have missed something, but I tried running model catalog with an empty sources file or a file with all sources disabled from the main branch and I already see an empty items: [] as response

pboyd:
> I may have missed something, but I tried running model catalog with an empty sources file or a file with all sources disabled from the main branch and I already see an empty items: [] as response

I wasn't able to see it return `null` either (or omit it). This is really just updating the API spec to match the actual behavior.

PR #1960, opened 2025-12-04
Preview Filtering Results on Add/Manage Source Form

manaswinidas:
## Description

Populate right-side preview panel for both Add and Manage source forms 
Screenshot 2025-12-04 at 6 18 50 PM

Cypress tests to be added (in progress)

## How Has This Been Tested?

Test the preview functionality while adding a source - you should see included and excluded models 

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
@@ -32,6 +32,15 @@ import ModelVisibilitySection from './ModelVisibilitySection';
 import PreviewPanel from './PreviewPanel';
 import ManageSourceFormFooter from './ManageSourceFormFooter';
 
+type PreviewState = {
+  mode?: 'preview' | 'validate';
+  isLoading: boolean;
+  result?: CatalogSourcePreviewResult;
+  error?: Error;
+  resultDismissed: boolean;
+  lastPreviewedData?: string;
+};
+
 type ManageSourceFormProps = {
   existingData?: Partial<ManageSourceFormData>;
   isEditMode: boolean;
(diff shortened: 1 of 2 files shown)


clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -40,10 +44,108 @@ const ManageSourceForm: React.FC<ManageSourceFormProps> = ({ existingData, isEdi
...
   const isFormComplete = isFormValid(formData);
   const canPreview = isPreviewReady(formData);
 
+  // Track form changes to show outdated preview alert
+  React.useEffect(() => {
+    const currentData = JSON.stringify(formData);
+    if (lastPreviewedData && currentData !== lastPreviewedData) {
+      setHasFormChanged(true);
+    }
+  }, [formData, lastPreviewedData]);
```
mturley: There's some redundant state here. `hasFormChanged` can be derived at render time from `lastPreviewedData && currentData !== lastPreviewedData`, we can memoize it with `useMemo` instead of having it be its own state. I think in order to clear the alert when you hit Preview again you'd just need to change the condition for showing the alert to have it also hidden when isPreviewLoading is true

The only difference I see there would be that if you edit the form and have the alert appear, you could then change your edit back and the alert would disappear. I think that's fine, you'd have returned to a state where the old preview is accurate again anyway.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -40,10 +44,108 @@ const ManageSourceForm: React.FC<ManageSourceFormProps> = ({ existingData, isEdi
...
+  };
+
+  const handlePreview = async () => {
+    if (!apiState.apiAvailable) {
+      setPreviewError(new Error('API is not available'));
+      return;
+    }
+
+    setIsPreviewLoading(true);
+    setPreviewError(undefined);
```
mturley: I think you also need to setValidationError(undefined) here - trying again should clear the error. Also it shouldn't be behind a `isValidateTriggered` condition - if i have a validate error and I click Preview instead of Validate to try again, it should still clear the old validate error.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ManageSourceForm.tsx
```diff
@@ -40,10 +44,108 @@ const ManageSourceForm: React.FC<ManageSourceFormProps> = ({ existingData, isEdi
...
+    undefined,
+  );
+  const [previewError, setPreviewError] = React.useState<Error | undefined>(undefined);
+  const [hasFormChanged, setHasFormChanged] = React.useState(false);
+  const [lastPreviewedData, setLastPreviewedData] = React.useState<string>('');
+  const [isValidateTriggered, setIsValidateTriggered] = React.useState(false);
+
+  // Validation state (for HF credentials)
+  const [validationError, setValidationError] = React.useState<Error | undefined>(undefined);
+  const [isValidationSuccess, setIsValidationSuccess] = React.useState(false);
```
mturley: Thinking more about all this state.. it seems a little fragile to have so many interrelated things we need to make sure to update together. Feel free to push back on this, but what if we had a unified PreviewState object that captured the various combined states here? something like:

```ts
type PreviewState = {
  mode?: 'preview' | 'validate';
  isLoading: boolean;
  result?: CatalogSourcePreviewResult;
  error?: Error;
  resultDismissed: boolean;
  lastPreviewedData?: string;
}
```

Then when any of this state needs to be updated in lockstep you're sure you've updated all of it - triggering a new preview/validate can just be:
```ts
setPreviewState({
  mode: 'preview',
  isLoading: true,
});
```
which will implicitly clear result, error, lastPreviewedData, resultDismissed. When the request resolves you can update it all together.

Then you've deduplicated some things that you can derive from this:
* Instead of needing a separate previewError and validationError, you can use `previewState.mode` to determine where to display the single error property. You also wouldn't need `isValidationTriggered` with different state update behavior based on it.
* `isValidationSuccess` can be derived instead of being its own state: `previewState.mode === 'validate' && !previewState.isLoading && !previewState.isError && !previewState.resultDismissed`
  * Most of the time you'd want to update the whole object so you don't forget to clean up other state, but in the few cases where you want to update just a single property you can do e.g. `setPreviewState({ ...previewState, resultDismissed: true });`
* As I said in an earlier comment you can have `hasFormChanged` just be derived with a useMemo

Let me know if I'm missing some detail that would make this a bad idea - I know elsewhere we have separate state for result/error/isLoading, I think as it gets more complicated here it gets harder to make sure we're updating everything where we should, this is more like a state machine and we can be confident we're not missing stale state corner cases. If you hate this, ignore me 😄

mturley:
Also @manaswinidas as long as we're updating mock data, I wonder if we should have more than 10 preview results so we can test the pagination controls in mock mode.

ppadti:
> A couple small things, plus a refactor suggestion that I'm not 100% sure about - open to your opinion of course.
> 
> Also, I'm noticing that when I try to edit a default source I can't hit Preview and the initial preview-on-mount doesn't happen. Also when I edit one of our mock custom YAML sources the YAML isn't populated...
> 
> I think this is actually a mock data issue. Console.logging the `existingSourceConfig` in ManageSourcePage, I'm seeing that when I go to manage those sources there is no `yaml` value defined. So our `isPreviewReady` util thinks that required field is missing. It also looks like that means we can't save updates to the allowed/disallowed on the default sources... not sure what happened there (cc @ppadti )
> 
> Can you maybe see if you can fix the mock data and make sure we're properly getting the existing YAML into form state for models that have it? I guess we don't want to be sending the YAML when we patch a default source, so we may also want to make sure in `transformFormDataToPayload` we exclude `yaml` from the payload if the source `isDefault`? hmm...
> 
> We could maybe follow up on that issue in another PR, but if so I think let's open a bug and not lose track of it.

yeah, this is because of how we implemented the `/source-configs` endpoint - we are not sending the yaml content in the response, as the payload might get too big. We will get the yaml content on `source_configs/sourceId` endpoint, which will be part of my story. Currently, in the frontend, we are fetching the source_config list and finding the particular source by id, so the yaml value is missing there. This behaviour should be fixed in my issue. Maybe we could ignore the test for the default one here? And test it on my PR. Does that make sense?

mturley:
@ppadti that makes sense. looks like we don't have test coverage for the form validation in the flow to manage a default source, so there's nothing to skip here, we should add that separately after your PR.

/approve
/lgtm

ppadti:
> @ppadti that makes sense. looks like we don't have test coverage for the form validation in the flow to manage a default source, so there's nothing to skip here, we should add that separately after your PR.

oh yeah, we have for yaml and HF, I can add that coverage in my current issue of adding new endpoint for artifact/performance.

PR #1966, opened 2025-12-08
Implementation of CRUD for model catalog settings

ppadti:
## Description
This PR aims to add the implementation of crud for model catalog settings. Updated the readme to include the curl commands.
When we run the application with kind cluster, now we should be able to see that, the CRUD operation works as expected.

## How Has This Been Tested?
**In Mock mode:**

BFF - `make run PORT=4000 MOCK_K8S_CLIENT=true MOCK_MR_CLIENT=true MOCK_MR_CATALOG_CLIENT=true DEV_MODE=true DEPLOYMENT_MODE=standalone`

Frontend - `DEPLOYMENT_MODE=standalone STYLE_THEME=patternfly-theme npm run start:dev`

with Kind cluster:
we need to first create a configMap named `model-catalog-default-sources` and `model-catalog-sources` and then try to run the application with MOCK_K8S_CLIENT=false

BFF - `make run PORT=4000 MOCK_K8S_CLIENT=false MOCK_MR_CLIENT=true MOCK_MR_CATALOG_CLIENT=true DEV_MODE=true DEPLOYMENT_MODE=standalone`

Frontend - `DEPLOYMENT_MODE=standalone STYLE_THEME=patternfly-theme npm run start:dev`

IF you want to try few curl commands:
get all source configs:

```
curl -X GET "http://localhost:4000/api/v1/settings/model_catalog/source_configs?namespace=kubeflow" -H "kubeflow-userid: user@example.com"
```

add a catalog source config:

```
curl -X POST "http://localhost:4000/api/v1/settings/model_catalog/source_configs?namespace=kubeflow" -H "kubeflow-userid: user@example.com" -H "Content-Type: application/json" -d '{"data":{"id":"test-catalog","name":"Test Catalog","type":"yaml"}}' 
```

get a catalog source config:
```
curl -s "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com" 
```

Update a catalog source config:

```
curl -X PATCH "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com" -H "Content-Type: application/json" -d '{"data":{"id":"test-catalog","name":"Updated Catalog","type":"yaml"}}'
```

Delete a catalog source config:

```
curl -X DELETE "http://localhost:4000/api/v1/settings/model_catalog/source_configs/test-catalog?namespace=kubeflow" -H "kubeflow-userid: user@example.com"
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalogSettings/utils/modelCatalogSettingsUtils.ts
+++ b/clients/ui/frontend/src/app/pages/modelCatalogSettings/utils/modelCatalogSettingsUtils.ts
@@ -88,8 +88,19 @@ export const getPayloadForConfig = (
   }
 
   if (isEditMode) {
-    const { id, ...catalogSource } = sourceConfig;
-    return catalogSource;
+    return {
+      name: sourceConfig.name,
+      type: sourceConfig.type,
+      enabled: sourceConfig.enabled,
+      isDefault: sourceConfig.isDefault,
+      includedModels: sourceConfig.includedModels,
+      excludedModels: sourceConfig.excludedModels,
+      ...(sourceConfig.type === CatalogSourceType.YAML && { yaml: sourceConfig.yaml }),
+      ...(sourceConfig.type === CatalogSourceType.HUGGING_FACE && {
+        apiKey: sourceConfig.apiKey,
+        allowedOrganization: sourceConfig.allowedOrganization,
+      }),
+    };
   }
 
   return sourceConfig;


clients/ui/frontend/src/app/hooks/modelCatalogSettings/useCatalogSourceConfigBySourceId.ts
```diff
@@ -0,0 +1,24 @@
+import { FetchState, FetchStateCallbackPromise, NotReadyError, useFetchState } from 'mod-arch-core';
+import React from 'react';
+import { CatalogSourceConfig } from '~/app/modelCatalogTypes';
+import { ModelCatalogSettingsContext } from '~/app/context/modelCatalogSettings/ModelCatalogSettingsContext';
+
+type State = CatalogSourceConfig | null;
```
mturley, line 6: @ppadti  I notice that when we list source configs we aren't fetching the secrets or the yaml contents. This is probably fine since we don't need those details in the table and I see you do make sure to fetch the single source by id on the ManageSourceForm now. However both places use the same type `CatalogSourceConfig` and just have `yaml` and `apiKey` as optional properties. I wonder if we should use a different type, like have `CatalogSourceConfigFull`, where those properties are present and required, but don't include them in the type of the regular `CatalogSourceConfig`, maybe rename it something like `CatalogSourceConfigShallow`? Maybe this is overkill. but it would mean we're not allowed to try accessing `yaml` or `apiKey` when it won't be there.

Alternatively we could have the list endpoint go and fetch those dependent things even though we don't use them on the table today just so we have consistent types. That might be overkill though.

I'm curious if you think this matters or if maybe we just need to add some code comments, like in the type maybe comment on yaml / apiKey to explain they will not be there when fetching these from the list endpoint.

clients/ui/frontend/src/app/pages/modelCatalogSettings/utils/modelCatalogSettingsUtils.ts
```diff
@@ -45,23 +45,36 @@ export const generateSourceIdFromName = (name: string): string =>
...
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    return {
+      enabled: formData.enabled,
+      includedModels: parseModels(formData.allowedModels),
+      excludedModels: parseModels(formData.excludedModels),
+    } as CatalogSourceConfigPayload;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  return {
```
mturley: You can avoid these eslint-disable comments if you change the `CatalogSourceConfigPayload` type to be more permissive - a partial that includes specifically enabled, includedModels and excludedModels is a valid payload.

modelCatalogTypes.ts:265
```ts
export type CatalogSourceConfigPayload =
  | CatalogSourceConfig
  | Pick<CatalogSourceConfig, 'enabled' | 'includedModels' | 'excludedModels'>;
```

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
```diff
@@ -197,3 +197,147 @@ func (kc *SharedClientLogic) GetAllCatalogSourceConfigs(
 
 	return *defaultCM, *userCM, nil
 }
+
+func (kc *SharedClientLogic) GetSecretValue(
```
ederign: @ppadti, I just want to double-check that we really need to expose the secret on the UI after creation. Is this really necessary? For which use cases users needs to get the secret value?

clients/ui/bff/internal/api/model_catalog_settings_handler.go
```diff
@@ -59,14 +58,29 @@ func (app *App) GetCatalogSourceConfigHandler(w http.ResponseWriter, r *http.Req
...
+	client, err := app.kubernetesClientFactory.GetClient(ctx)
+	if err != nil {
+		app.serverErrorResponse(w, r, errors.New("catalog client not found"))
+		return
+	}
+
+	catalogSourceConfig, err := app.repositories.ModelCatalogSettingsRepository.GetCatalogSourceConfig(ctx, client, namespace, catalogSourceId)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
```
ederign: We should not do error detection via string matching. One of the options is to type the error on the repository and check it here, i.e.:
var (
    ErrCatalogSourceNotFound = errors.New("catalog source not found")
)

then you can use errors.Is()

clients/ui/bff/internal/repositories/model_catalog_settings.go
```diff
@@ -152,3 +450,388 @@ func extractStringSlice(value interface{}) []string {
...
+}
+
+func deleteSecretForHuggingFace(ctx context.Context,
+	client k8s.KubernetesClientInterface,
+	namespace string,
+	secretName string) {
+	_ = client.DeleteSecret(ctx, namespace, secretName)
+}
+
+func isNotFoundError(err error) bool {
```
ederign: We should avoid this hardcoded. if I recall correctly, the kuberenes API returns k8serrors.IsNotFound() from k8s.io/apimachinery/pkg/api/errors

Also, this file become too big. Shall we extract it to a helper (i.e. with parsing logic)

clients/ui/bff/internal/repositories/model_catalog_settings.go
```diff
@@ -60,19 +127,202 @@ func (r *ModelCatalogSettingsRepository) CreateCatalogSourceConfig(
...
 	payload models.CatalogSourceConfigPayload,
 ) (*models.CatalogSourceConfig, error) {
-	// TODO ppadti write the real implementation here calling k8s client
-	return nil, fmt.Errorf("not implemented yet")
+	defaultCM, userCM, err := client.GetAllCatalogSourceConfigs(ctx, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch catalog source configmaps: %w", err)
+	}
+
+	existingUserSource := findCatalogSourceById(userCM.Data[k8s.CatalogSourceKey], sourceId, false)
```
ederign: Concurrent updates can overwrite each other. Is that ok for our use case?

clients/ui/bff/internal/repositories/model_catalog_settings.go
```diff
@@ -60,19 +127,202 @@ func (r *ModelCatalogSettingsRepository) CreateCatalogSourceConfig(
...
+		switch catalogType {
+		case "yaml":
+			if payload.Yaml != nil && *payload.Yaml != "" {
+				if yamlFilePath == "" {
+					yamlFilePath = fmt.Sprintf("%s.yaml", sourceId)
+				}
+				userCM.Data[yamlFilePath] = *payload.Yaml
+			}
+
+		case "huggingface":
```
ederign: Magic String we should create const. (also applies for hugging face and apiKEY

clients/ui/bff/internal/repositories/model_catalog_settings.go
```diff
@@ -60,19 +127,202 @@ func (r *ModelCatalogSettingsRepository) CreateCatalogSourceConfig(
...
+		return nil, fmt.Errorf("catalog source '%s' already exists in user managed sources", payload.Id)
+	}
+
+	var secretName string
+	var yamlFileName string
+	yamlContent := make(map[string]string)
+
+	switch payload.Type {
+	case "yaml":
+		yamlFileName = fmt.Sprintf("%s.yaml", payload.Id)
```
ederign, line 171: As this is user input, we need to sanitize this to prevent path traversal or invalid characters.

clients/ui/bff/internal/repositories/model_catalog_settings.go
```diff
@@ -152,3 +450,388 @@ func extractStringSlice(value interface{}) []string {
...
+				"app.kubernetes.io/component": "model-catalog",
+				"catalog-source-id":           catalogId,
+			},
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			"apiKey": apiKey,
+		},
+	}
+	err := client.CreateSecret(ctx, namespace, secret)
```
ederign, line 493: If I recall correctly, secret name has a size limit for it. Can you please double check and validate?

ederign:
@ppadti nice work! I haven't reviewed everything yet (I need to run it on my machine), but I've raised some points there from a initial pass

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
mturley: @ederign we prefill the field with this secret when the user is editing their source, but we could talk to UX about a different approach where the existing secret is hidden and can only be replaced. I think this would require a design change - maybe as simple as just showing some asterisks and when you click the field it empties, but there are corner cases there

clients/ui/bff/internal/repositories/model_catalog_settings.go
mturley: We do prevent the user from submitting concurrent updates from the UI, but that's a poor fix considering we could have multiple users in multiple browsers actually. Is there some kind of locking we could do here to make them atomic?

clients/ui/frontend/src/app/hooks/modelCatalogSettings/useCatalogSourceConfigBySourceId.ts
mturley: Based on Eder's comment about secrets, we certainly don't want to be fetching them on the list request if we're not fetching them for the get request. Maybe the types are fine here and we just need to comment/document the behavior that yaml will only be present on the get request, but if we are getting different partials back in certain contexts it's probably harmless to type them more strictly. If any components need to be able to work with both types we can make them generic and only care about the properties they need.

clients/ui/bff/internal/repositories/model_catalog_settings.go
ppadti: yeah sure.  I will update this and other places to not use hardcoded

>Also, this file become too big. Shall we extract it to a helper (i.e. with parsing logic)

I will move those to helper.

clients/ui/bff/internal/repositories/model_catalog_settings.go
ppadti: Id will be generated from the frontend, we are using this regex to convert the user given name to source id
name
    .trim()
    .replace(/\s+/g, '_')
    .replace(/-/g, '_')
    .replace(/[^a-zA-Z0-9_]/g, '')
    .toLowerCase();

clients/ui/bff/internal/repositories/model_catalog_settings.go
ppadti: yeah sure, i will do that.

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
ederign: If possible, I would try to push UX not to expose this on edit, and just * . It's much safer on our side if we just cover the way you described: " existing secret is hidden and can only be replaced."

clients/ui/bff/internal/repositories/model_catalog_settings.go
ederign: @ppadti, it's always safe and good to double-check those validations on the BFF. We never know in future, if frontend change this without awareness of the issues

clients/ui/bff/internal/repositories/model_catalog_settings.go
ederign: Yeah, something with mu sync.Mutex

func (s *Bla) Update() (err error) {
	s.mu.Lock()
	defer func() {
		if r := recover(); r != nil {
			// make sure we never keep the mutex locked on panic
			err = fmt.Errorf("update panicked: %v", r)
			log.Printf("recovering from panic in Update: %v\n", r)
		}
		s.mu.Unlock()
	}()

	// --- critical section ---
	 
	return nil
}
https://gobyexample.com/mutexes
https://go.dev/tour/concurrency/9

ederign:
@ppadti please let me know when you are ready for another round of review.

ppadti:
> @ppadti please let me know when you are ready for another round of review.

Yeah, I have addressed the feedback provided. It's ready for re-review now.

clients/ui/bff/internal/repositories/model_catalog_settings.go
ppadti: Yeah, okay. I will sanitize this in BFF as well.

clients/ui/bff/internal/integrations/kubernetes/client.go
```diff
@@ -2,13 +2,13 @@ package kubernetes
...
+
 	corev1 "k8s.io/api/core/v1"
 )
 
 const ComponentLabelValue = "model-registry"
 const ComponentLabelValueCatalog = "model-catalog"
 
-// TODO ppadti double check if the config map key is indeed sources.yaml
 const CatalogSourceKey = "sources.yaml"
 const CatalogSourceDefaultConfigMapName = "model-catalog-source-config"
```
mturley: @ppadti the default configmap is called `model-catalog-default-sources`. this is causing a 500 error after the sync to ODH.

mturley:
FYI @ppadti I fixed the name of the default sources configmap in https://github.com/kubeflow/model-registry/pull/1986, shouldn't conflict with this, I didn't want to block that while this wraps up review

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
mturley: @ppadti +1 to @ederign, let's drop the GetSecretValue functionality entirely and just have this support only the create/patch behavior for the secret (write-only). we can update the UI to handle the fact that the apiKey will be missing from the get request. we'll need to do so anyway since it is becoming optional according to some ongoing threads.

We just need to make sure that when the apiKey is excluded from a request to patch a source, we don't do anything with it and leave the secret alone (i.e. we only patch the secret if a replacement apiKey was sent)

ppadti:
@ederign @mturley I had a questions on how we handle listing source config, should we be supporting partial response - 
if it fails to retrieve default configMap - return error
if it fails to retrieve catalog-source configMap - return just the default

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
ppadti: @mturley should I remove the required option from the api token field in this PR? we might need to update the description of that field for Manage source.

ppadti:
> I had a questions on how we handle listing source config, should we be supporting partial response - if it fails to retrieve default configMap - return error if it fails to retrieve catalog-source configMap - return just the default

I have added the code to support this, just need to address this comment https://github.com/kubeflow/model-registry/pull/1966#discussion_r2606500061 - working on it
And @mturley I have added test based on this comment here https://github.com/kubeflow/model-registry/pull/1960#issuecomment-3635892720

clients/ui/bff/internal/repositories/model_catalog_settings.go
ppadti: @ederign  so we should be using Mutex for create, update and delete, right - all of these will be patching the same configMap

mturley:
> if it fails to retrieve default configMap - return error\nif it fails to retrieve catalog-source configMap - return just the default

@ppadti this sounds like a good approach to me

ederign:
@ppadti @mturley  the bff code lgtm! I'll approve the PR, but please just merge it after someone tries to run it locally and also reviews the frontend code.

@ppadti nice to see your work here!

ppadti:
> @ppadti it looks like you have type errors here as a result of my suggestion to use the `| Pick<CatalogSourceConfig, 'enabled' | 'includedModels' | 'excludedModels'>;` approach for the `CatalogSourceConfigPayload` type because we now have code in ManageSourceForm trying to inspect a payload's properties that could be missing. The preview request assumes `transformFormDataToPayload` will always return a full config payload as if we are creating a new source, but since that util now has a `if (formData.isDefault)` condition to return a partial patch payload for default sources, we have a problem. The preview request needs the full payload even for default sources...
> 
> I wonder if we should have `transformFormDataToPayload` instead be `transformFormDataToConfig` and return the full `CatalogSourceConfig`, then use that in places where we always need the full config like in ManageSourceForm's `buildPreviewRequest`. Then we could have a wrapper `getPayloadForConfig` that returns `CatalogSourceConfigPayload`, and that could have the check to see if it's default and if so return only those properties we want for that payload. Then we just need to use the appropriate util in the right places.
> 
> Sorry to send you on one more round here...

yeah, fixed the test failure now...
update: those were passing locally for me - forgot to rebase, sorry... still those type checks are failing - fixing now

mturley:
/approve
/lgtm

@ppadti I tested this in ODH by doing a test sync using [this version of the subtree script](https://github.com/opendatahub-io/odh-dashboard/pull/5736). Everything works great except that I can't save edits to a default source, the save button never enables. We still need to fix the validation there, but I don't want to block this PR further. If you'd like to move on to something else I can open a bug and we can follow up, up to you.

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
mturley: @ppadti I just noticed your last comment here, [we have a new story](https://redhat-internal.slack.com/archives/C08G31WCV16/p1765829872589869?thread_ts=1765388380.988649&cid=C08G31WCV16) to make the token field optional and update microcopy and further discussion about handling the UX for this write-only secret behavior when managing a source [in this thread](https://redhat-internal.slack.com/archives/C08G31WCV16/p1765830523374519)

ppadti: thanks @mturley !! currently the BFF is omitting that field -  we need to handle it in FE.

PR #1978, opened 2025-12-09
Alert to when Performance Insights filters are applied 

YuliaKrimerman:
### Feature Description
Added an alert that appears on the Model Catalog page when users return from the Performance Insights details page after changing performance filters. The alert informs users that the results on the Model Catalog Page list has been updated to match the latest performance criteria.

https://github.com/user-attachments/assets/28137f96-6a0e-4ca1-9fb1-1d2e9c87106d

## How Has This Been Tested?

The alert appears in one case, when all of these conditions are met:
* Performance view toggle is ON
* User is on the catalog page (not the details page)
* Alert hasn't been dismissed
* Performance filters were changed on the details page (performanceFiltersChangedOnDetailsPage is true)

The alert is hidden in these scenarios:
* User dismisses it manually (clicks close)
* Performance toggle is turned OFF
* Filters change on the catalog page
* User navigates to the details page

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
+++ b/clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
@@ -14,12 +14,12 @@ import {
   CategoryName,
   ModelCatalogFilterStates,
 } from '~/app/modelCatalogTypes';
-import { ModelDetailsTab } from '~/app/pages/modelCatalog/screens/ModelDetailsTabs';
-import { BFF_API_VERSION, URL_PREFIX } from '~/app/utilities/const';
 import {
+  ModelDetailsTab,
   ModelCatalogStringFilterKey,
   ModelCatalogNumberFilterKey,
 } from '~/concepts/modelCatalog/const';
+import { BFF_API_VERSION, URL_PREFIX } from '~/app/utilities/const';
 
 export type ModelCatalogContextType = {
   catalogSourcesLoaded: boolean;
--- a/clients/ui/frontend/src/app/pages/modelCatalog/ModelCatalogRoutes.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/ModelCatalogRoutes.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react';
 import { Navigate, Route, Routes } from 'react-router-dom';
 import { ModelCatalogContextProvider } from '~/app/context/modelCatalog/ModelCatalogContext';
+import { ModelDetailsTab } from '~/concepts/modelCatalog/const';
 import ModelCatalogCoreLoader from './ModelCatalogCoreLoader';
 import ModelDetailsPage from './screens/ModelDetailsPage';
 import RegisterCatalogModelPage from './screens/RegisterCatalogModelPage';
 import ModelCatalog from './screens/ModelCatalog';
-import { ModelDetailsTab } from './screens/ModelDetailsTabs';
 
 const ModelCatalogRoutes: React.FC = () => (
   <ModelCatalogContextProvider>
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
@@ -24,7 +24,7 @@ import {
 } from '~/app/modelCatalogTypes';
 import { extractValidatedModelMetrics } from '~/app/pages/modelCatalog/utils/validatedModelUtils';
 import { catalogModelDetailsTabFromModel } from '~/app/routes/modelCatalog/catalogModel';
-import { ModelDetailsTab } from '~/app/pages/modelCatalog/screens/ModelDetailsTabs';
+import { ModelDetailsTab } from '~/concepts/modelCatalog/const';
 import { useCatalogModelArtifacts } from '~/app/hooks/modelCatalog/useCatalogModelArtifacts';
 import { filterArtifactsByType } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
 import { formatLatency } from '~/app/pages/modelCatalog/utils/performanceMetricsUtils';
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsPage.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsPage.tsx
@@ -29,7 +29,8 @@ import { CatalogModelDetailsParams } from '~/app/modelCatalogTypes';
 import { useCatalogModelArtifacts } from '~/app/hooks/modelCatalog/useCatalogModelArtifacts';
 import { modelCatalogUrl } from '~/app/routes/modelCatalog/catalogModel';
 import ScrollViewOnMount from '~/app/shared/components/ScrollViewOnMount';
-import ModelDetailsTabs, { ModelDetailsTab } from './ModelDetailsTabs';
+import { ModelDetailsTab } from '~/concepts/modelCatalog/const';
+import ModelDetailsTabs from './ModelDetailsTabs';
 
 type ModelDetailsPageProps = {
   tab: ModelDetailsTab;
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsTabs.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsTabs.tsx
@@ -12,13 +12,9 @@ import {
   shouldShowValidatedInsights,
   filterArtifactsByType,
 } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+import { ModelDetailsTab } from '~/concepts/modelCatalog/const';
 import ModelDetailsView from './ModelDetailsView';
 import PerformanceInsightsView from './PerformanceInsightsView';
-
-export enum ModelDetailsTab {
-  OVERVIEW = 'overview',
-  PERFORMANCE_INSIGHTS = 'performance-insights',
-}
 
 export enum ModelDetailsTabTitle {
   OVERVIEW = 'Overview',
--- a/clients/ui/frontend/src/app/routes/modelCatalog/catalogModel.ts
+++ b/clients/ui/frontend/src/app/routes/modelCatalog/catalogModel.ts
@@ -1,4 +1,4 @@
-import { ModelDetailsTab } from '~/app/pages/modelCatalog/screens/ModelDetailsTabs';
+import { ModelDetailsTab } from '~/concepts/modelCatalog/const';
 import { getCatalogModelDetailsRoute } from '~/app/routes/modelCatalog/catalogModelDetails';
 
 export const modelCatalogUrl = (sourceId?: string): string =>
--- a/clients/ui/frontend/src/concepts/modelCatalog/const.ts
+++ b/clients/ui/frontend/src/concepts/modelCatalog/const.ts
@@ -333,3 +333,8 @@ export const MODEL_CATALOG_FILTER_CATEGORY_NAMES: Record<ModelCatalogStringFilte
   [ModelCatalogStringFilterKey.HARDWARE_TYPE]: 'Hardware type',
   [ModelCatalogStringFilterKey.USE_CASE]: 'Workload type',
 };
+
+export enum ModelDetailsTab {
+  OVERVIEW = 'overview',
+  PERFORMANCE_INSIGHTS = 'performance-insights',
+}


clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -93,6 +103,26 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+
+    const wasOnDetailsPath = isModelDetailsPage(prevPath);
+    const isNowOnDetailsPath = isModelDetailsPage(currentPath);
+
+    if (!wasOnDetailsPath && isNowOnDetailsPath) {
+      setPerformanceFiltersChangedOnDetailsPage(false);
+    }
+
+    prevLocationRef.current = currentPath;
+  }, [location.pathname]);
```
mturley: I don't think we need this effect or `prevLocationRef`. Instead of looking for the path changing we can just trigger an effect when PerformanceInsightsView mounts. Simply add this useEffect to PerformanceInsightsView (and remove the other useEffects there, see below)

```ts
  React.useEffect(() => {
  }, [setPerformanceFiltersChangedOnDetailsPage]);
```
This should only trigger on its first mount if `setPerformanceFiltersChangedOnDetailsPage` is stable, which it should be.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogFilters.tsx
```diff
@@ -23,6 +36,25 @@ const ModelCatalogFilters: React.FC = () => {
...
+            />
+          </StackItem>
+          <StackItem>
+            <Content component={ContentVariants.small}>
+              Enable performance filters, display model benchmark data, and exclude unvalidated
+              models.
+            </Content>
+          </StackItem>
+        </Stack>
+      </StackItem>
```
mturley, line 63: The mocks have a card border around this, I think you need to use the Card component in here somewhere.

Image

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -34,14 +41,77 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
...
   const filtersApplied = React.useMemo(() => hasFiltersApplied(filterData), [filterData]);
   const hasActiveFilters = React.useMemo(
     () => filtersApplied || (searchTerm && searchTerm.trim().length > 0),
     [filtersApplied, searchTerm],
   );
 
+  const isOnCatalogPage = React.useMemo(
+    () => !isModelDetailsPage(location.pathname),
+    [location.pathname],
+  );
```
mturley: This component is only rendered on the catalog landing page, so you don't need to check which page you're on. When this component mounts, you have reached the catalog landing page.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -34,14 +41,77 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
...
+      performanceFiltersChangedOnDetailsPage,
+    ],
+  );
+
+  React.useEffect(() => {
+    if (!performanceViewEnabled) {
+      setAlertDismissed(true);
+      setPerformanceFiltersChangedOnDetailsPage(false);
+    }
+  }, [performanceViewEnabled, setPerformanceFiltersChangedOnDetailsPage]);
```
mturley: This useEffect can be avoided. Any time you want to update state as a result of other state changing, it is usually better to either (a) eliminate duplicate state or (b) update both states from the same function call directly.

In this case, it looks like you want turning off the performance view toggle to dismiss the alert and reset the "filters changed on details page" state. I think we have both duplicated state and a shared update event here:
* Why not remove `alertDismissed` and use `performanceFiltersChangedOnDetailsPage` as your state for whether to show the alert? To dismiss it, just `setPerformanceFiltersChangedOnDetailsPage(false);`. Once the alert is dismissed, I don't think we need to care about tracking that the filters still haven't been changed since the alert was triggered. If it helps you can rename it to `shouldShowFiltersChangedOnDetailsAlert` or something.
* In ModelCatalogContext, you can rename `setPerformanceViewEnabled` to something else (like `baseSetPerformanceViewEnabled` and pass down your own `setPerformanceViewEnabled` wrapper function which both calls the underlying state setter _and_ does anything you want to happen as a direct result of that state changing. So in that handler you could have `if (!checked) { setPerformanceFiltersChangedOnDetailsPage(false); }`.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -34,14 +41,77 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
...
+    const prevFilters = prevFilterDataRef.current;
+    const filtersChanged = !deepEqual(prevFilters, filterData);
+
+    if (filtersChanged && isOnCatalogPage) {
+      setAlertDismissed(true);
+      setPerformanceFiltersChangedOnDetailsPage(false);
+    }
+
+    prevFilterDataRef.current = filterData;
+  }, [filterData, isOnCatalogPage, setPerformanceFiltersChangedOnDetailsPage]);
```
mturley: We can avoid this effect too. We don't need do deep equal comparison if we piggyback on the state setter.

In ModelCatalogContext, try the same approach: Rename `setFilterData` to some thing like `baseSetFilterData` and pass down your own `setFilterData` wrapper that, in addition to calling the base setter, checks if you're on the details page and calls `setPerformanceFiltersChangedOnDetailsPage` with either true (if you're on the details page) or false (if you're not).

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -34,14 +41,77 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
...
+    if (wasOnDetailsPage && isNowOnCatalogPage) {
+      if (performanceFiltersChangedOnDetailsPage) {
+        setAlertDismissed(false);
+      } else {
+        setAlertDismissed(true);
+      }
+    }
+
+    prevIsOnCatalogPageRef.current = isOnCatalogPage;
+  }, [location.pathname, isOnCatalogPage, performanceFiltersChangedOnDetailsPage]);
```
mturley: You don't need this effect if you eliminate the duplicate state. You'll already have what you need if you just drive whether the alert is shown from `performanceFiltersChangedOnDetailsPage`.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -11,36 +18,98 @@ type PerformanceInsightsViewProps = {
...
+    const initialFilters = initialFilterDataRef.current;
+    const filtersChanged = !deepEqual(prevFilters, filterData);
+    const changedFromInitial = !deepEqual(initialFilters, filterData);
+
+    if (filtersChanged && changedFromInitial && hasPerformanceFiltersApplied(filterData)) {
+      setPerformanceFiltersChangedOnDetailsPage(true);
+    }
+
+    prevFilterDataRef.current = filterData;
+  }, [filterData, location.pathname, setPerformanceFiltersChangedOnDetailsPage]);
```
mturley: I think all of these effects and refs are unnecessary if you follow the other patterns I outlined (wrap the state setters and manage side effects of them being called as part of the call itself). `setPerformanceFiltersChangedOnDetailsPage` only needs to be set `true` in one place: your new wrapper around `setFilterData` in ModelCatalogContext ([as I commented here](https://github.com/kubeflow/model-registry/pull/1978/files#r2604676769)). and it needs only be set to false in 3 places: that same wrapper, a one-line effect when PerformanceInsightsView mounts ([as I commented here](https://github.com/kubeflow/model-registry/pull/1978/files#r2604354422)), and when the alert is dismissed.

If you handle it that way, you no longer have any need for comparing locations with previous ones or comparing filters with previous ones at all.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -19,6 +19,86 @@ import {
...
+/**
+ * Performs a deep equality check between two values.
+ * Handles primitives, arrays, and plain objects.
+ * Note: This is a simple implementation. For complex objects with circular references,
+ * consider using a library like lodash.isEqual.
+ */
+export const deepEqual = <T>(a: T, b: T): boolean => {
+  if (Object.is(a, b)) {
+    return true;
+  }
```
mturley: We don't need this deepEqual anymore, but for future reference if you want a deepEqual we already have `lodash-es` installed and we can import it and use `_.isEqual()`, as ODH does in its `useDeepCompareMemoize`.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -19,6 +19,86 @@ import {
...
+/**
+ * Checks if the given pathname corresponds to a model details page.
+ * Details page pattern: /model-catalog/:sourceId/:modelName or /model-catalog/:modelName
+ */
+export const isModelDetailsPage = (pathname: string | null | undefined): boolean => {
+  if (!pathname || typeof pathname !== 'string') {
+    return false;
+  }
+  return !!pathname.match(/\/model-catalog\/[^/]+\/[^/]+/);
+};
```
mturley: A regex like this shouldn't be necessary. It will also break in ODH where the route changes to `/ai-hub/catalog` In the place where you want to check which page you're on (I think now it'll just only be for your `setFilterData` wrapper in ModelCatalogContext), you can use the `useMatch` hook imported from `react-router-dom`. Check out how it's used in ODH's NavItemHref component. You can pass it the same path spec used by the routes themselves, which you should be able to get by passing placeholders into the `catalogModelDetailsFromModel` function from `clients/ui/frontend/src/app/routes/modelCatalog/catalogModel.ts`:

`catalogModelDetailsFromModel(':sourceId', ':modelName')` (literally that, don't use the source id and model name, but the actual strings `':sourceId'` and `':modelName'`). This should return `"/model-catalog/:sourceId/:modelName"` (or ai-hub, in ODH, which you get for free by reusing this util), and the router's `useMatch` should be able to check if the current page matches that.

If that doesn't work, you could just check that the pathname includes `ModelDetailsTab.PERFORMANCE_INSIGHTS`, since really that's what we care about when determining where the filters were updated from. It's a little less clean but it would work.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -295,3 +375,33 @@ export const hasFiltersApplied = (filterData: ModelCatalogFilterStates): boolean
     }
     return value !== undefined;
   });
+
+/**
+ * Checks if any performance filters are applied
+ * Performance filters include: USE_CASE, MIN_RPS, HARDWARE_TYPE, and latency metric filters
+ */
+export const hasPerformanceFiltersApplied = (filterData: ModelCatalogFilterStates): boolean => {
```
mturley: I don't believe you'll need this anymore. Any change at all to filters while on the details page qualifies for triggering this alert.

mturley (changes requested):
We can do some clever things to simplify and get rid of a lot of these useEffects. In general, we really want to avoid piling on a lot of effects like this to avoid future unpredictable behavior.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -93,6 +102,31 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
   const [selectedSourceLabel, setSelectedSourceLabel] = React.useState<
     ModelCatalogContextType['selectedSourceLabel']
   >(CategoryName.allModels);
+  const [basePerformanceViewEnabled, setBasePerformanceViewEnabled] = React.useState(false);
+  const [performanceFiltersChangedOnDetailsPage, setPerformanceFiltersChangedOnDetailsPage] =
+    React.useState(false);
+
+  const location = useLocation();
+  const isOnDetailsPage = location.pathname.includes('performance-insights');
```
mturley: You can use the enum value for this: `ModelDetailsTab.PERFORMANCE_INSIGHTS`

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -19,6 +19,17 @@ import {
   ModelCatalogNumberFilterKey,
 } from '~/concepts/modelCatalog/const';
 
+/**
+ * Checks if the given pathname corresponds to a model details page.
+ * Details page pattern: /model-catalog/:sourceId/:modelName or /model-catalog/:modelName
+ */
+export const isModelDetailsPage = (pathname: string | null | undefined): boolean => {
```
mturley: This function is no longer used and can be removed.

PR #1993, opened 2025-12-12
Added chip and "Reset all filters" to hardware configuration page

Philip-Carneiro:
## Description
Adds a "Reset all filters" button and active filter chips to the hardware configuration page.

- Active filter chips: grouped by category (Workload type, Max latency, Min RPS, Hardware type) and shown in bordered containers with the category name inside each container.
- Reset all filters: button clears the search input, global filter chips

## How Has This Been Tested?
Manual testing: verified filter chips display correctly, individual chip removal works, category clear buttons work, "Reset all filters" clears all filters

https://github.com/user-attachments/assets/821b2ce5-0c6a-4ebf-bb9f-ef0e159f01e8

in PF mode
image

## Merge criteria:

If you have UI changes

- [X] The developer has added tests or explained why testing cannot be added.
- [X] Verify that UI/UX changes conform the UX guidelines for Kubeflow.


clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationActiveFilters.tsx
```diff
@@ -0,0 +1,236 @@
...
+  type LatencyMetricFieldName,
+} from '~/concepts/modelCatalog/const';
+import {
+  getLatencyFieldName,
+  parseLatencyFieldName,
+  type LatencyFilterConfig,
+} from '~/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils';
+import type { ModelCatalogFilterStates } from '~/app/modelCatalogTypes';
+
+const HardwareConfigurationActiveFilters: React.FC = () => {
```
mturley: So, we intend to implement these same filters on the main catalog landing page soon and they will use the same filterData state (they will be modifiable on both pages). I think rather than creating a new component here, we need to enhance the existing `clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx` so it supports these additional filters. It has a `filtersToShow` prop that was added in anticipation of this, which we can use to (for now) only show the 4 basic filters on the landing page (provider, license, task, language) and only show these performance filters on the details page, but combining this into the same chip bar component will help us when we start using performance filters on both pages.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationActiveFilters.tsx
```diff
@@ -0,0 +1,236 @@
...
+            deleteLabelGroup={(category) => {
+              const categoryKey = typeof category === 'string' ? category : category.key;
+              handleClearCategory(categoryKey);
+            }}
+            data-testid="use-case-filter-container"
+          >
+            {null}
+          </ToolbarFilter>
+        );
+      })()}
```
mturley: When you recombine this with ModelCatalogActiveFilters, I _think_ we won't need a special case for the USE_CASE filter and it should be handled by the existing generic logic there that works for any ModelCatalogStringFilterKey. We'll need to adapt that component so it can also handle additional filters instead of bailing if the key isn't a ModelCatalogStringFilterKey, but let's try to still have these be as generic and reusable as possible.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationActiveFilters.tsx
```diff
@@ -0,0 +1,236 @@
...
+          }}
+          deleteLabelGroup={(category) => {
+            const categoryKey = typeof category === 'string' ? category : category.key;
+            handleClearCategory(categoryKey);
+          }}
+          data-testid="max-latency-filter-container"
+        >
+          {null}
+        </ToolbarFilter>
+      )}
```
mturley: This is an interesting case, because ideally it would be nice to have the ModelCatalogActiveFilters component just happily generically render any filter with a ModelCatalogNumberFilterKey (all driven from the `filtersToShow` list of keys in ModelCatalogActiveFilters) and have our latency field names be part of ModelCatalogNumberFilterKey... I may have steered us wrong by trying to avoid duplication and have those keys all driven from the metric/percentile rather than be part of that enum. But even if we made this generic, we still have the quirks of (a) "max latency" is really several underlying filter keys and (b) it needs special rendering like you have with `formatLatencyLabel`. ~~So yeah I suppose it does make sense for this to be its own case.~~ (Edit: actually I think in addition to the rest of this comment, we can indeed make the ToolbarFilter rendering generic for all filter keys - see my [next comment](https://github.com/kubeflow/model-registry/pull/1993/changes#r2615363591).)

However, I think we can avoid the special `if (categoryKey === 'maxLatency')` logic elsewhere you have. Instead of this single ToolbarFilter being rendered when `activeLatencyFilters.length > 0` and having `labels={activeLatencyFilters.map(...)`, what if we lift the map up to the top and render an entire separate ToolbarFilter for each activeLatencyFilter, each of which has just one label? If there were multiple active latency filters we'd get duplicate category names showing up, but that would actually be a bug elsewhere - if the latency filter dropdown is working correctly we should only ever have one of these active at a time.

If you do that, then instead of `key: 'maxLatency'` you can use the specific `key: fieldName`. Then all the handlers for removing/clearing don't need special logic for max latency filters, they can just clear whatever that filter key is.

One wrinkle there: because the latency fields aren't part of ModelCatalogNumberFilterKey, you'd have to change this (in the remove and clear handlers):
```ts
    } else if (isEnumMember(categoryKey, ModelCatalogNumberFilterKey)) {
      setFilterData(categoryKey, undefined);
    }
```
to a catch-all:
```ts
    } else {
      setFilterData(categoryKey, undefined);
    }
```
I'm pretty sure the way the types work out, if the categoryKey isn't a ModelCatalogStringFilterKey, undefined is a valid value for the other two cases, so an else there should be fine.

One other thing - actually, when you move this into ModelCatalogActiveFilters, to drive it from the `filtersToShow` map I think you'll need to have some kind of `isLatencyFieldName` instead of having an `activeLatencyFilters` to map over at all.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationActiveFilters.tsx
```diff
@@ -0,0 +1,236 @@
...
+            deleteLabelGroup={(category) => {
+              const categoryKey = typeof category === 'string' ? category : category.key;
+              handleClearCategory(categoryKey);
+            }}
+            data-testid="min-rps-filter-container"
+          >
+            {null}
+          </ToolbarFilter>
+        );
+      })()}
```
mturley: Ok, here's a case though where I think we can have generic rendering for any `ModelCatalogNumberFilterKey`. If you look at how we handle string filters generically in ModelCatalogActiveFilters, we pull the category name from `MODEL_CATALOG_FILTER_CATEGORY_NAMES` and we get the label text from a `getFilterLabel` util. Those currently only take `ModelCatalogStringFilterKey` keys, but we should have them also handle `ModelCatalogNumberFilterKey`. `getFilterLabel` can have a special case for MIN_RPS to have it render the requests/sec suffix.

Thinking more here... maybe it's overkill, but honestly couldn't we also have the same generic ToolbarFilter for all the filter keys? rather than special cases at all for each of these types of keys when rendering ToolbarFilter, I think if we make `MODEL_CATALOG_FILTER_CATEGORY_NAMES` and `getFilterLabel` handle not only string and number filters but also the latency field names, there doesn't need to be any other special logic when rendering a ToolbarFilter. In ModelCatalogActiveFilters we'd simply need to remove this:
```ts
        // Only process string filter keys that are arrays
        if (!isEnumMember(filterKey, ModelCatalogStringFilterKey)) {
          return null;
        }

        const filterValues = filterData[filterKey];
        if (!Array.isArray(filterValues) || filterValues.length === 0) {
          return null;
        }
```
and if `filterData[filterKey]` isn't an array we could have `filterValues` fall back to being a one-element array with the single value (if defined).

To make `MODEL_CATALOG_FILTER_CATEGORY_NAMES` and `getFilterLabel` handle number filter keys will be simple, but to get it to handle latency fields I think we'd need to:
1. Have `MODEL_CATALOG_FILTER_CATEGORY_NAMES` also spread in a reduced object with all the latency field names as keys which all have the value "Max latency". We may want a util `getAllLatencyFieldNames` to make this easier, to go along with `getLatencyFieldName` and `parseLatencyFieldName` utils. (Edit: maybe just a computed constant instead of a function, see my [later comment](https://github.com/kubeflow/model-registry/pull/1993/changes#r2615373476))
    * There's actually already a TODO comment above the definition of this constant indicating we should expand it, but I think that was written when we assumed ModelCatalogNumberFilterKey would contain the latency fields.
3. Give `getFilterLabel` a case for `isLatencyFieldName` (I guess we need that util too) and have it use your `formatLatencyLabel` logic.

The more stuff we can change from specifically handling `ModelCatalogStringFilterKey` to handling the wider `ModelCatalogFilterKey` type, the better.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationActiveFilters.tsx
```diff
@@ -0,0 +1,236 @@
...
+            }}
+            deleteLabelGroup={(category) => {
+              const categoryKey = typeof category === 'string' ? category : category.key;
+              handleClearCategory(categoryKey);
+            }}
+            data-testid="hardware-type-filter-container"
+          >
+            {null}
+          </ToolbarFilter>
+        );
```
mturley: I'm pretty sure instead of having special rendering for hardware type, in this case we can just rely on the existing generic behavior ModelCatalogActiveFilters has for string filter keys. Simply including `ModelCatalogStringFilterKey.HARDWARE_TYPE` in the `filtersToShow` prop when rendering that component on the details page should do the trick.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -173,7 +173,11 @@ export const clearAllFilters = (
...
-  // TODO: Clear all latency metric filters (e.g., ttft_mean, ttft_p90, etc.)
-  // This will be needed when we add filter chips to the details page.
-  // Can be part of that PR - will need to loop over all LatencyMetricFieldName combinations.
+  // Clear all latency metric filters (e.g., ttft_mean, ttft_p90, etc.)
+  for (const metric of Object.values(LatencyMetric)) {
+    for (const percentile of Object.values(LatencyPercentile)) {
+      const fieldName = getLatencyFieldName(metric, percentile);
+      setFilterData(fieldName, undefined);
+    }
+  }
```
mturley: If we are factoring out a `getAllLatencyFieldNames` util we can just use it here too. In fact it doesn't really need to be a function, it could be a constant array computed once since it's based only on enums.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -23,9 +25,11 @@ type HardwareConfigurationFilterToolbarProps = {
 const HardwareConfigurationFilterToolbar: React.FC<HardwareConfigurationFilterToolbarProps> = ({
   performanceArtifacts,
 }) => {
-  const { filterOptions, filterOptionsLoaded, filterOptionsLoadError, setFilterData } =
+  const { filterOptions, filterOptionsLoaded, filterOptionsLoadError, filterData, setFilterData } =
     React.useContext(ModelCatalogContext);
 
+  const hasActiveFilters = React.useMemo(() => hasFiltersApplied(filterData), [filterData]);
```
mturley: So here's an interesting corner case... because we are not showing the "basic filters" on this page, only the performance filters, but we're sharing one big filter state object, I think we need to change `hasFiltersApplied` and `clearAllFilters` so they can optionally take a list of filter keys. On this page we only want to recognize and clear the performance filters we are showing controls for - clicking "reset all filters" shouldn't clear the basic filters you had applied on the previous page. You could use the same array of filter keys you'll need to pass into `filtersToShow` when rendering ModelCatalogActiveFilters on this page.

I'm starting to see the pain we've caused by having all this filter state centralized - it still makes sense since we are driving everything from the filter_options the API gives us and we'll be using it to generaste one big filterQuery (which is all centralized), but we could maybe have converted things to a structure that works better for us... at any rate, I think it's too late to pivot on that for now and let's make it work for our needs here.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -288,10 +289,21 @@ export const getSourceFromSourceId = (
   return catalogSources.items.find((source) => source.id === sourceId);
 };
 
-export const hasFiltersApplied = (filterData: ModelCatalogFilterStates): boolean =>
-  Object.values(filterData).some((value) => {
+/**
+ * Checks if any filters are applied. If filterKeys is provided, only checks those specific filters.
+ * Otherwise checks all filters.
+ */
+export const hasFiltersApplied = (
```
manaswinidas, line 306: Worth adding a unit test in `modelCatalogUtils.spec.ts` file for this util.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -28,6 +28,23 @@ export enum LatencyPercentile {
...
+      (percentile) =>
+        `${metric.toLowerCase()}_${percentile.toLowerCase()}` as LatencyMetricFieldName,
+    ),
+);
+
+/**
+ * Type guard to check if a string is a valid LatencyMetricFieldName
+ */
+export const isLatencyFieldName = (key: string): key is LatencyMetricFieldName =>
+  ALL_LATENCY_FIELD_NAMES.includes(key as LatencyMetricFieldName);
```
mturley: In addition to the lint errors here, we should avoid `as` assertions where we can. In this case you should be able to just reuse utils we already have in `clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts`:
* For `ALL_LATENCY_FIELD_NAMES`, you can use `getLatencyFieldName` to get the lowercased combined value from metric and percentile values.
* In `isLatencyFieldName` you can just use `parseLatencyFieldName`. It will return null if it's given an invalid latency field. We should also move it into that utils file. The parse util does expect a `LatencyMetricFieldName` instead of a string, so you do probably need the ` as LatencyMetricFieldName` here, we can add an eslint-ignore-line for that. Not sure if there's a better way there.
  * Edit: actually I don't think you need this util at all, see other comments

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -9,33 +9,81 @@ import {
...
+  ModelCatalogStringFilterKey.USE_CASE,
+  ModelCatalogStringFilterKey.HARDWARE_TYPE,
+  ModelCatalogNumberFilterKey.MIN_RPS,
+];
+
+/**
+ * Gets the active filter keys including any active latency filters from filterData
+ */
+const getActivePerformanceFilterKeys = (
+  filterData: Record<string, unknown>,
```
mturley: We should use the existing type for filterData (`ModelCatalogFilterStates`).

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -9,33 +9,81 @@ import {
...
+): ModelCatalogFilterKey[] => {
+  const activeLatencyKeys = Object.keys(filterData).filter(
+    (key) => isLatencyFieldName(key) && filterData[key] !== undefined,
+  ) as ModelCatalogFilterKey[];
+
+  return [...PERFORMANCE_FILTER_KEYS, ...activeLatencyKeys];
+};
```
mturley, line 53: We should use the existing type for filterData (`ModelCatalogFilterStates`). Also we don't need isLatencyFieldName here if we invert the logic:

```suggestion
const getActivePerformanceFilterKeys = (
  filterData: ModelCatalogFilterStates,
): ModelCatalogFilterKey[] => {
  const activeLatencyKeys = ALL_LATENCY_FIELD_NAMES.filter((key) => filterData[key] !== undefined);
  return [...PERFORMANCE_FILTER_KEYS, ...activeLatencyKeys];
};
```

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
+      if (parsed) {
+        return `${parsed.metric} ${parsed.percentile}: ≤${value}ms`;
       }
-      default:
-        return value;
+      return `${filterKey}: ≤${value}ms`;
     }
+
+    return String(value);
   };
```
mturley: You can simply not check isLatencyFieldName here - earlier in this function you've already handled cases for `ModelCatalogStringFilterKey` and `ModelCatalogNumberFilterKey`, so when you hit these lines `filterKey` has been narrowed to be only the remaining possible values which are `LatencyMetricFieldName`. Passing that into `parseLatencyFieldName` works just fine.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
 
   return (
     <>
       {filtersToShow.map((filterKey) => {
-        // Only process string filter keys that are arrays
-        if (!isEnumMember(filterKey, ModelCatalogStringFilterKey)) {
+        const filterValue = filterData[filterKey as keyof typeof filterData];
```
mturley: This type assertion isn't necessary, `filterKey` is already a valid type to use as a key of `filterData`.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
+
+        // Skip if no value is set
+        if (filterValue === undefined || filterValue === null) {
```
mturley: You can simply use `!filterValue` here.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
+        if (Array.isArray(filterValue) && filterValue.length === 0) {
           return null;
         }
 
-        const categoryName = MODEL_CATALOG_FILTER_CATEGORY_NAMES[filterKey];
+        // Normalize to array for consistent handling
+        const filterValues = Array.isArray(filterValue) ? filterValue : [filterValue];
+
+        const categoryName = MODEL_CATALOG_FILTER_CATEGORY_NAMES[
+          filterKey as keyof typeof MODEL_CATALOG_FILTER_CATEGORY_NAMES
```
mturley: Again this type assertion isn't necessary, it should just work without it.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
         const newValues = currentValues.filter((v) => String(v) !== String(labelKey));
         setFilterData(categoryKey, newValues);
       }
+    } else {
+      // For number filters and latency fields, clear the value
+      setFilterData(categoryKey as keyof typeof filterData, undefined);
```
mturley: ah, I guess I was wrong that TS would be happy with the else case here. We should still avoid the assertion, we do need a clean way to check that the categoryKey here is a valid filter key.

I think similar to how you have `ALL_LATENCY_FIELD_NAMES` maybe we should also create an `ALL_CATALOG_FILTER_KEYS` of type `ModelCatalogFilterKey[]` that combines the values of the two enums for string and number keys along with the values from `ALL_LATENCY_FIELD_NAMES`. Then we can have a util `isCatalogFilterKey(key: string)` that checks against it. To avoid type assertions in it we can use String() casting to do something like:

```ts
const isCatalogFilterKey = (key: string): key is ModelCatalogFilterKey =>
  ALL_CATALOG_FILTER_KEYS.some((k) => String(k) === key);
```

Then you could just return early at the top of this function if categoryKey isn't a valid filter key.

Edit: I just remembered we have `getAllFilterKeys`, maybe we can use that instead of introducing `ALL_CATALOG_FILTER_KEYS`, but there might be value in both since the former splits out by subtype

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
     }
   };
 
   const handleClearCategory = (categoryKey: string) => {
     if (isEnumMember(categoryKey, ModelCatalogStringFilterKey)) {
       setFilterData(categoryKey, []);
+    } else {
+      // For number filters and latency fields, clear the value
+      setFilterData(categoryKey as keyof typeof filterData, undefined);
```
mturley: Same here, early return if it's not a valid category key and you shouldn't need this assertion.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -30,57 +34,99 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
-      case ModelCatalogStringFilterKey.TASK: {
-        return isEnumMember(value, ModelCatalogTask)
-          ? MODEL_CATALOG_TASK_NAME_MAPPING[value]
-          : value;
+    }
+
+    // Handle number filter keys
+    if (isEnumMember(filterKey, ModelCatalogNumberFilterKey)) {
+      switch (filterKey) {
+        case ModelCatalogNumberFilterKey.MIN_RPS:
```
mturley, line 93: Hmm this one is interesting because if we anticipate further numeric filters we will want this comparison. Maybe add an eslint-ignore-line here and a TODO comment to remove it if we add more number filter keys.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -302,3 +305,178 @@ describe('filtersToFilterQuery', () => {
...
   //   });
 });
+
+describe('hasFiltersApplied', () => {
+  const mockFormData = ({
+    tasks = [],
+    license = [],
+    provider = [],
+    language = [],
+    hardware_type = [],
```
mturley, line 638: These camelcase warnings can be ignored, if you like you can throw a `/* eslint-disable camelcase */` at the top of this file.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -141,27 +144,45 @@ export const filterHardwareConfigurationArtifacts = (
...
+      if (stringFilterKeys.includes(key as ModelCatalogStringFilterKey)) {
+        setFilterData(key as ModelCatalogStringFilterKey, []);
+      } else if (
+        numberFilterKeys.includes(key as ModelCatalogNumberFilterKey) ||
+        isLatencyFieldName(key)
+      ) {
+        setFilterData(key as keyof ModelCatalogFilterStates, undefined);
+      }
+    });
+    return;
```
mturley, line 158: I'm not sure why CI isn't catching the type assertions here.. we also want to avoid these. We can leverage `isEnumMember` imported from `mod-arch-core` here and use a catch-all else case for the non-string filters.


```suggestion
  if (filterKeys) {
    filterKeys.forEach((key) => {
      if (isEnumMember(key, ModelCatalogStringFilterKey)) {
        setFilterData(key, []);
      } else {
        setFilterData(key, undefined);
      }
    });
    return;
```

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -288,10 +289,21 @@ export const getSourceFromSourceId = (
...
+  const keysToCheck = filterKeys ?? (Object.keys(filterData) as ModelCatalogFilterKey[]);
+
+  return keysToCheck.some((key) => {
+    const value = filterData[key as keyof ModelCatalogFilterStates];
     if (Array.isArray(value)) {
       return value.length > 0;
     }
     return value !== undefined;
   });
+};
```
mturley, line 353: Let's avoid a type assertion here. I think you can use Object.entries to handle this a little differently.


```suggestion
export const hasFiltersApplied = (
  filterData: ModelCatalogFilterStates,
  filterKeys?: ModelCatalogFilterKey[],
): boolean =>
  Object.entries(filterData).some(([key, value]) => {
    if (filterKeys && !filterKeys.some((k) => k === key)) {
      return false;
    }
    if (Array.isArray(value)) {
      return value.length > 0;
    }
    return value !== undefined;
  });
```

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -326,17 +343,26 @@ export enum AllLanguageCode {
...
 /**
  * Display names for filter categories.
- * TODO: When performance filters are ready, switch this to be a Record<ModelCatalogFilterKey, string>
- * to include all ModelCatalogFilterKeys (ModelCatalogStringFilterKey | ModelCatalogNumberFilterKey).
- * This will allow separate filter category names for "Max latency (TTFT Mean)" and "Max latency (TTFT P99)" etc.
+ * Includes all ModelCatalogFilterKeys (ModelCatalogStringFilterKey | ModelCatalogNumberFilterKey | LatencyMetricFieldName).
  */
-export const MODEL_CATALOG_FILTER_CATEGORY_NAMES: Record<ModelCatalogStringFilterKey, string> = {
+export const MODEL_CATALOG_FILTER_CATEGORY_NAMES: Record<
+  ModelCatalogStringFilterKey | ModelCatalogNumberFilterKey | LatencyMetricFieldName,
```
mturley: Let's have this use the type `ModelCatalogFilterKey` which is the same, and that will force us to update this if we introduce new filter key types later.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -326,17 +343,26 @@ export enum AllLanguageCode {
...
   [ModelCatalogStringFilterKey.LANGUAGE]: 'Language',
   [ModelCatalogStringFilterKey.HARDWARE_TYPE]: 'Hardware type',
   [ModelCatalogStringFilterKey.USE_CASE]: 'Workload type',
+  // Number filter keys
+  [ModelCatalogNumberFilterKey.MIN_RPS]: 'Min RPS',
+  // Latency field names - all use "Max latency" as category name
+  ...Object.fromEntries(ALL_LATENCY_FIELD_NAMES.map((field) => [field, 'Max latency'])) as Record<
+    LatencyMetricFieldName,
+    string
+  >,
```
mturley: I don't see a clean way around this assertion, let's put a `// eslint-disable-next-line @typescript-eslint/consistent-type-assertions` before this line.

manaswinidas:
@Philip-Carneiro Cypress tests are failing here. Can you have a look?

manaswinidas:
Canceled the approve to let the workflows run one final time but the bot merged it first - the tests were running fine before the rebase so I don't think this merge should cause issues

PR #2013, opened 2025-12-16
Implement model catalog client for source preview

ppadti:
## Description
This PR aims to add implementation for source preview endpoint

https://github.com/user-attachments/assets/b6837fb9-122c-480b-b9aa-5b2c337b776e

## How Has This Been Tested?
Run the application in mock mode
  - see the while adding a new source, after filling the required filed - click on preview button to see the preview 
  - click on Manage source - when the form open you will see the preview.
  
Run the application with kind cluster
 - port forward and run the application with mock_MC_client=false

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/repositories/catalog_source_preview.go
+++ b/clients/ui/bff/internal/repositories/catalog_source_preview.go
@@ -39,7 +39,7 @@ func (a CatalogSourcePreview) CreateCatalogSourcePreview(client httpclient.HTTPC
 		"excludedModels": sourcePreviewPayload.ExcludedModels,
 	}
 
-	if sourcePreviewPayload.Type == "huggingface" {
+	if sourcePreviewPayload.Type == CatalogTypeHuggingFace {
 		if org, ok := sourcePreviewPayload.Properties["allowedOrganization"]; ok {
 			configData["allowedOrganization"] = org
 		}
@@ -57,7 +57,7 @@ func (a CatalogSourcePreview) CreateCatalogSourcePreview(client httpclient.HTTPC
 		return nil, fmt.Errorf("error writing config data: %w", err)
 	}
 
-	if sourcePreviewPayload.Type == "yaml" {
+	if sourcePreviewPayload.Type == CatalogTypeYaml {
 		if yamlContent, ok := sourcePreviewPayload.Properties["yaml"].(string); ok && yamlContent != "" {
 			catalogDataPart, err := writer.CreateFormFile("catalogData", "catalog.yaml")
 			if err != nil {


ppadti:
Hi @pboyd I just had a question - For HF preview, we're sending apiKey in the request config but model-catalog reads from [HF_API_KEY env var](https://github.com/kubeflow/model-registry/tree/main/catalog#huggingface-source-configuration) Is this intentional, or should preview accept apiKey from the request?

clients/ui/bff/internal/integrations/httpclient/http.go
```diff
@@ -111,7 +111,7 @@ func (c *HTTPClient) GET(url string) ([]byte, error) {
 	return body, nil
 }
 
-func (c *HTTPClient) POST(url string, body io.Reader) ([]byte, error) {
+func (c *HTTPClient) POST(url string, body io.Reader, contentType ...string) ([]byte, error) {
```
ederign: @ppadti when the content Type will be ...string?

Can we follow https://pkg.go.dev/net/http#Client.Post signature?

clients/ui/bff/internal/integrations/httpclient/http.go
```diff
@@ -120,7 +120,11 @@ func (c *HTTPClient) POST(url string, body io.Reader) ([]byte, error) {
 		return nil, err
 	}
 
-	req.Header.Set("Content-Type", "application/json")
+	ct := "application/json"
```
ederign: Also, I suggest having two versions of this method. one that we don't pass content type (and we use "application/json as a parameter) and another one that sets it by default. Would that fit your needs?

clients/ui/bff/internal/integrations/httpclient/http.go
ederign: (we can go either with 1 or 0)

clients/ui/bff/internal/integrations/httpclient/http.go
ppadti: yes, that works!!

clients/ui/bff/internal/integrations/httpclient/http.go
```diff
@@ -112,6 +113,10 @@ func (c *HTTPClient) GET(url string) ([]byte, error) {
 }
 
 func (c *HTTPClient) POST(url string, body io.Reader) ([]byte, error) {
+	return c.POSTWithContentType(url, body, "application/json")
+}
+
+func (c *HTTPClient) POSTWithContentType(url string, body io.Reader, contentType string) ([]byte, error) {
```
ederign: @ppadti one last quick change: can this method be called just POST (with content type as parameters)? In that way we 100% align with http package.

PR #2030, opened 2025-12-19
Double check our UI TS types for catalog artifacts against real data

Philip-Carneiro:
## Description

This PR addresses type safety issues in the Model Catalog TypeScript type definitions where `customProperties` was defined as required but could be `null` or `undefined` in actual API responses. This mismatch between type definitions and runtime data was causing UI crashes when optional properties were missing from the API response.

### Problem Statement

The original TypeScript types were implemented based on an ADR (Architecture Decision Record) and then patched as errors appeared, but a comprehensive review was never performed to ensure all types matched the actual API data structure. This led to:

1. **UI crashes**: The application crashed on the landing page when `customProperties` or nested properties (like `hardware_type`) were missing from API responses
2. **Type mismatches**: The BFF (Backend-For-Frontend) defines `customProperties` as a nullable pointer (`*map[string]openapi.MetadataValue`) while the frontend TypeScript types defined it as required
3. **Insufficient null/undefined handling**: Code was accessing nested properties without proper defensive checks

## How Has This Been Tested?
Manual tested to check if the behaviour stays the same, updated unit tests and mocks.

## Merge criteria:

- [ X **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

### DIFF
--- a/clients/ui/frontend/src/__mocks__/mockCatalogModelArtifactList.ts
+++ b/clients/ui/frontend/src/__mocks__/mockCatalogModelArtifactList.ts
@@ -191,35 +191,3 @@ export const mockParetoFilteredPerformanceArtifactList = (
   size: 2,
   nextPageToken: '',
 });
-
-/**
- * Mock artifact with missing customProperties to test defensive handling.
- * This simulates an API response where customProperties is undefined/null.
- */
-export const mockCatalogPerformanceMetricsArtifactWithoutCustomProperties =
-  (): CatalogPerformanceMetricsArtifact => ({
-    artifactType: CatalogArtifactType.metricsArtifact,
-    metricsType: MetricsType.performanceMetrics,
-    createTimeSinceEpoch: '1739210683000',
-    lastUpdateTimeSinceEpoch: '1739210683000',
-    // customProperties intentionally omitted
-  });
-
-/**
- * Mock artifact with partial customProperties (some fields missing).
- * This simulates an API response where only some fields are present.
- */
-export const mockCatalogPerformanceMetricsArtifactWithPartialCustomProperties =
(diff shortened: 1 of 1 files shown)


ppadti:
@Philip-Carneiro thanks for updating the types!! we might also need to fix other types as well. 
we have to double check the types source_list, model_list and for filter_options

clients/ui/frontend/src/__mocks__/mockCatalogModelArtifactList.ts
```diff
@@ -191,3 +191,35 @@ export const mockParetoFilteredPerformanceArtifactList = (
   size: 2,
   nextPageToken: '',
 });
+
+/**
+ * Mock artifact with missing customProperties to test defensive handling.
+ * This simulates an API response where customProperties is undefined/null.
+ */
+export const mockCatalogPerformanceMetricsArtifactWithoutCustomProperties =
+  (): CatalogPerformanceMetricsArtifact => ({
```
ppadti: Just a question - we are not using this mocks created anywhere, is that intentional?

Philip-Carneiro: I created in case that need to be used, that was intentional, but I can remove and re-create it in the future if necessary, I just followed the same of some other mocks that was created but not used anywhere

ppadti: hmm, may be we remove this unused mocks and add it when we need it? 
>some other mocks that was created but not used anywhere

We should also remove those dead code - will create an issue to remove those.

PR #2047, opened 2025-12-26
chore: update pre-commit config and formatting

### LINKED ISSUE #2018: Update pre-commit
Our [.pre-commit-config.yaml](https://github.com/kubeflow/model-registry/blob/6a83f48ad6105bc98746e0783f22b22f644447a3/.pre-commit-config.yaml) has not been updated in a long time. It also isn't in sync with what we use in Makefiles and CI.

For instance, the ruff version used in the pre-commit is very old and has some quirks.

We may want to simply remove it, but let's check if there's any tools from it that we want to move to Makefiles.

aniketpati1121:
This PR updates .pre-commit-config.yaml as requested in  fixes #2018 and applies the resulting formatting/lint fixes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 exclude: |
-  ^clients/python/patches/.*\.patch$
+  .*\.patch$
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
@@ -9,20 +9,24 @@ repos:
       - id: check-ast
       - id: check-case-conflict
       - id: check-docstring-first
-      - id: check-executables-have-shebangs
+      # - id: check-executables-have-shebangs
       - id: check-json
       - id: check-merge-conflict
       - id: check-symlinks
       - id: debug-statements
       - id: detect-private-key
+        exclude: '.*_test\.go$'
       - id: end-of-file-fixer
+        exclude: '^clients/python/src/mr_openapi'
       - id: trailing-whitespace
+        exclude: '^clients/python/src/mr_openapi'
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.7.0
     hooks:
       - id: ruff
         args: [--fix]
-        exclude: ^clients/python/patches/.*\.patch$
+        files: '^clients/python'
       - id: ruff-format
-        exclude: ^clients/python/patches/.*\.patch$
+        files: '^clients/python/src/mr_openapi'
+


clients/python/patches/asyncio-only.patch
pboyd, line 1: Most of the whitespace changes are harmless, but this one seems seems to break `make install`. Can we exclude patch files from the pre-commit hook?

The problem is that these patches are meant to apply to the unformatted files produced from the Open API generator.

pboyd (commented):
First off, thank you for the taking the time to look at this. It's very much appreciated, and I really like the idea of making our pre-commit hook actually be useful.

That said, I have a few concerns. The pre-commit hook is mostly unused right now, and it will always be optional (to be honest, I'd never noticed the config file before this PR), so we can't rely on it to enforce project-wide standards. We don't need it for that anyway, because we check those things with GHA. It can certainly be helpful as an early warning for impending GHA failures, but I think it should largely be confined to that. The only exception I see is some optional nice-to-have code quality items that don't quite hit the level of a project-wide standard (e.g. whitespace, debug statements), or that would be impractical after submission (e.g. private key check).

Specifically,
- Can you remove the YAML formatting check? Personally, I think the formatting of our YAML files is fine. But if we want to enforce a standardized formatting it should be done in GHA and only then would it make sense in the pre-commit hook. (Feel free to tell me it's out of scope, since it existed before--mostly bringing it up because of all the YAML changes)
- Can you revert changes from applying the pre-commit hook's changes? I don't have a problem with most of these changes, but since the pre-commit hook is optional I'd prefer that it only apply to changes made in the future by devs who opt-in.
- Do you know why there are Python formatting changes after running the diff? There's a [rule to enforce Python formatting in GHA](https://github.com/kubeflow/model-registry/blob/68412d988282cb354396cb6a1cca32f172bb5b12/.github/workflows/python-tests.yml#L62-L65) and `make tidy` has no difference on `main` right now, so there really shouldn't be any difference after running the pre-commit hooks on this branch.

tarilabs:
I also want to thank you for updating the precommit.
I'm agreeing with Paul comments above, I believe task #2018 called for an assessment before going to the implementation, ...this PR can be used indeed now as part of that assessment, and deciding first what is really helpful in the precommit.

aniketpati1121:
I've reverted all formatting changes and limited this PR to the pre-commit configuration only, keeping pre-commit as an optional, opt-in developer helper rather than applying it repo-wide.

.pre-commit-config.yaml
```diff
@@ -1,8 +1,9 @@
+exclude: |
+  ^clients/python/patches/.*\.patch$
```
pboyd: This should probably apply to all patch files:

```suggestion
  .*\.patch$
```

.pre-commit-config.yaml
```diff
@@ -11,20 +12,17 @@ repos:
...
     hooks:
       - id: ruff
         args: [--fix]
+        exclude: ^clients/python/patches/.*\.patch$
       - id: ruff-format
-  - repo: https://github.com/google/yamlfmt
-    rev: v0.10.0
-    hooks:
-      - id: yamlfmt
+        exclude: ^clients/python/patches/.*\.patch$
```
pboyd: It doesn't look like ruff does anything to the patch files. And, for better or worse, we only use `ruff check` for `clients/python` and only use `ruff format` for `clients/python/src/mr_openapi`. We probably should expand that, but for now, can we limit the scope?

```
    hooks:
      - id: ruff
        files: '^clients/python'
      - id: ruff-format
        files: '^clients/python/src/mr_openapi'
```

.pre-commit-config.yaml
```diff
@@ -11,20 +12,17 @@ repos:
       - id: check-executables-have-shebangs
       - id: check-json
       - id: check-merge-conflict
-        # TODO: (goverter 1.3.2) generated files are marked 777 (see: https://github.com/jmattheis/goverter/issues/128)
-      # - id: check-shebang-scripts-are-executable
       - id: check-symlinks
       - id: debug-statements
       - id: detect-private-key
```
pboyd, line 17: This one failed when I tried it because of a Go test. Can we exclude those?

```
      - id: detect-private-key
        exclude: '.*_test\.go$'
```

.pre-commit-config.yaml
```diff
@@ -11,20 +12,17 @@ repos:
...
       - id: end-of-file-fixer
       - id: trailing-whitespace
```
pboyd, line 21: `end-of-file-fixer` tries to fix a generated README.md, I'd suggest we exclude the generated python client files from these two:

```
      - id: end-of-file-fixer
        exclude: '^clients/python/src/mr_openapi'
      - id: trailing-whitespace
        exclude: '^clients/python/src/mr_openapi'
```

pboyd:
This looks great! Thanks for the changes.

/lgtm

@tarilabs, can you re-approve the GH workflows?

pboyd:
Thanks again, @aniketpati1121 

/approve

PR #2053, opened 2025-12-30
Model card performance filters toggle changes and hide filtered latency columns in hardware configuration table 

manaswinidas:
## Description

- Hide latency columns according to applied filters in the hardware configuration table 
- Validated Model card content changes according to performance filters toggle 
- Some mock data changes to test the latency filter changes

https://github.com/user-attachments/assets/1fe8c893-d352-4f02-8d5b-e6f8e32fb82e

## How Has This Been Tested?

- Switch on the toggle using the temp feature flag on the console. Switch on/off the toggle to see the content changes in the validated model cards
- Go to performance insights tab and try changing the latency metric and perentile values and check whether the other columns are hidden when certain filters are applied

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/mocks/static_data_mock.go
+++ b/clients/ui/bff/internal/mocks/static_data_mock.go
@@ -905,97 +905,97 @@ func performanceMetricsCustomProperties(customProperties map[string]openapi.Meta
 		},
 		"ttft_mean": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  35.48,
+				DoubleValue:  35.48818160947744,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"ttft_p90": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  51.56,
+				DoubleValue:  51.55777931213379,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"ttft_p95": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  61.27,
+				DoubleValue:  61.26761436462402,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"ttft_p99": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  72.96,
+				DoubleValue:  72.95823097229004,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"e2e_mean": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  1994.48,
+				DoubleValue:  1994.480013381083,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"e2e_p90": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  2644.60,
+				DoubleValue:  2644.604682922363,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"e2e_p95": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  2813.79,
+				DoubleValue:  2813.79246711731,
 				MetadataType: "MetadataDoubleValue",
 			},
 		},
 		"e2e_p99": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  3117.57,
+				DoubleValue:  3117.565155029297,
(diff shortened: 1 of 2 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
YuliaKrimerman, line 76: Now that you use useCatalogPerformanceArtifacts with performanceViewEnabled I think you should add `isValidated && performanceViewEnabled, // Only fetch if validated AND toggle is ON` here. This will ensure the req of "The /artifacts fetch per-card can be disabled when the toggle is off" and we will not make an unnecessary API call

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
```diff
@@ -39,13 +39,22 @@ type ModelCatalogCardBodyProps = {
   source: CatalogSource | undefined;
 };
 
+const descriptionStyle: React.CSSProperties = {
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  WebkitLineClamp: 4,
+  WebkitBoxOrient: 'vertical',
+  display: '-webkit-box',
+};
```
YuliaKrimerman: I think PatternFly has a Truncate component or use Content component with appropriate truncation props instead of custom inline

YuliaKrimerman (commented):
Great job here ,2 small things

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
manaswinidas: But we need the "View x benchmarks" hyperlink leading to performance insights tab of catalog details page even when the toggle is off for validated models

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
manaswinidas: Yes, you are right. Using the `TruncatedText` component from `mod-arch-core` library now

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
mturley: aha.. that's a good point, I was wrong to specify that we can avoid this fetch if the toggle is off I guess ☹️

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -905,97 +905,97 @@ func performanceMetricsCustomProperties(customProperties map[string]openapi.Meta
 		},
 		"ttft_mean": {
 			MetadataDoubleValue: &openapi.MetadataDoubleValue{
-				DoubleValue:  35.48818160947744,
```
mturley, line 908: Why change these values? This is what we get from the API (I believe) and we have logic in the frontend to round it to 2 decimal places, we want to make sure that still works.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -18,11 +20,51 @@ const HardwareConfigurationTable: React.FC<HardwareConfigurationTableProps> = ({
...
+  const filteredColumns = React.useMemo((): HardwareConfigColumn[] => {
+    if (!activeLatencyField) {
+      // No latency filter selected, show all columns
+      return hardwareConfigColumns;
+    }
+
+    // Extract the percentile suffix from the active filter (e.g., 'ttft_p90' -> 'p90', 'e2e_mean' -> 'mean')
+    const percentileSuffix = activeLatencyField.split('_').pop();
+    // Build the matching TPS field name (e.g., 'tps_p90', 'tps_mean')
+    const matchingTpsField = `${tpsPrefix}_${percentileSuffix}`;
```
mturley: You should be able to use the `parseLatencyFieldName` and `getLatencyFieldName` utils here instead of splitting and joining with _ directly.

manaswinidas:
>@manaswinidas we'll need to make sure we follow up after the filters are implemented on the landing page to make these cards change which latency property is being shown based on the user's latency filter selection. Technically you already could do that here if you want since that filter can be set on the details page and is available in context when rendering the cards, but it may make sense to make it part of the scope of implementing the filters. Just don't want to lose track of that detail since it was scoped as part of this story.


Yes, I have that on my radar  - I'm addressing that in the PR that adds the filters in the landing page

Issue #2055, opened 2026-01-03
Add standalone UI to docker-compose for local development


ederign:
**Is your feature request related to a problem? Please describe.**
Currently, there is no easy way to showcase the complete Model Registry feature set (UI + Model Registry + Model Catalog) without deploying to a Kubernetes cluster. Users evaluating Model Registry for standalone use cases, demos, or proofs-of-concept need a simple way to experience the full stack locally.

**Describe the solution you'd like**
Add the Model Registry UI service to docker-compose, enabling a single command (`make compose/up/postgres`) to spin up the complete stack. This allows users to:
- Explore Model Registry features without Kubernetes
- Run demos and showcases locally
- Evaluate standalone deployment scenarios
- Quickly prototype integrations

PR #2056, opened 2026-01-03
feat: enable standalone UI in docker-compose with configurable deployment mode

ederign:
## Description
Add deployment mode and style theme as parameters to the UI Dockerfile, enable `--mock-k8s-client=true` for standalone mode, and integrate the UI service into docker-compose for running the complete local development stack.
Closes https://github.com/kubeflow/model-registry/issues/2055 .

### Changes
- Add `DEPLOYMENT_MODE` and `STYLE_THEME` as build arguments to UI Dockerfile
- Enable `--mock-k8s-client=true` in the standalone Docker image
- Add `model-registry-ui` service to `docker-compose.yaml` (pre-built images)
- Add `model-registry-ui` service to `docker-compose-local.yaml` (builds from source)
- Use `extra_hosts: localhost:host-gateway` to route BFF connections to model-registry and model-catalog containers

After PR review: 
- Create dedicated Dockerfile.standalone with envtest support to avoid
  bloating production image with K8s test binaries (~150MB)
- Keep main Dockerfile clean for kubeflow/federated deployments
- Update docker-compose.yaml to use ui-standalone image
- Add explicit --mock-mr-client=false and --mock-mr-catalog-client=false
  flags for better readability
- Update Makefile and GitHub workflow to use new Dockerfile
- Document Dockerfile structure in README

### Usage
```bash
# Using pre-built images
make compose/up/postgres

**Endpoints:**
- UI: http://localhost:9000
- Model Registry: http://localhost:8080
- Model Catalog: http://localhost:8081

# Building from source
make compose/local/up/postgres
```
## How Has This Been Tested?
Verified UI successfully fetches data from both model-registry and model-catalog via the host-gateway routing.
make compose/local/up/postgres
make compose/up/postgres

https://github.com/user-attachments/assets/cfe54a73-5e2d-4593-b83f-19ca8991c05b

## Merge criteria:

- [ ]Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/docker-compose-local.yaml
+++ b/docker-compose-local.yaml
@@ -14,13 +14,15 @@ services:
   model-registry-ui:
     build:
       context: ./clients/ui
-      dockerfile: Dockerfile
+      dockerfile: Dockerfile.standalone
       args:
         DEPLOYMENT_MODE: standalone
         STYLE_THEME: mui-theme
     command:
       - --port=9000
       - --mock-k8s-client=true
+      - --mock-mr-client=false
+      - --mock-mr-catalog-client=false
       - --dev-mode=true
       - --deployment-mode=standalone
       - --log-level=DEBUG
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,11 +12,13 @@
 version: '3.8'
 services:
   model-registry-ui:
-    image: ghcr.io/kubeflow/model-registry/ui:latest
+    image: ghcr.io/kubeflow/model-registry/ui-standalone:latest
     pull_policy: always
     command:
       - --port=9000
       - --mock-k8s-client=true
+      - --mock-mr-client=false
+      - --mock-mr-catalog-client=false
       - --dev-mode=true
       - --deployment-mode=standalone
       - --log-level=DEBUG
--- a/.github/workflows/build-and-push-ui-images-standalone.yml
+++ b/.github/workflows/build-and-push-ui-images-standalone.yml
@@ -71,6 +71,7 @@ jobs:
       uses: docker/build-push-action@v6
       with:
         context: ./clients/ui
+        file: ./clients/ui/Dockerfile.standalone
         push: true
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/clients/ui/Dockerfile.standalone
+++ b/clients/ui/Dockerfile.standalone
@@ -47,12 +47,19 @@ COPY ${BFF_SOURCE_CODE}/internal/ internal/
 # Build the Go application
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bff ./cmd
 
+# Install setup-envtest and download K8s binaries for mock mode
+RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17
+RUN setup-envtest use 1.29.0 --bin-dir /envtest-bin -p path
+
 # Final stage
 # Use distroless as minimal base image to package the application binary
 FROM ${DISTROLESS_BASE_IMAGE}
 WORKDIR /
 COPY --from=bff-builder /usr/src/app/bff ./
 COPY --from=ui-builder /usr/src/app/dist ./static/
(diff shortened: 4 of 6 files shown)


ederign:
@tarilabs better late than never 😅

clients/ui/Dockerfile
```diff
@@ -22,7 +22,7 @@ COPY ${UI_SOURCE_CODE} /usr/src/app
 # Install the dependencies and build
 RUN npm cache clean --force
 RUN npm ci --omit=optional
-RUN npm run build:prod
+RUN DEPLOYMENT_MODE=${DEPLOYMENT_MODE} STYLE_THEME=${STYLE_THEME} npm run build:prod
```
lucferbux, line 25: Ok, this is a good catch, we were building the images passing the values as BUILD-ARGS, and since we always were just doing that for build time and we didn't care about runtime was just fine, but this is safer.

clients/ui/Dockerfile
```diff
@@ -47,12 +47,19 @@ COPY ${BFF_SOURCE_CODE}/internal/ internal/
 # Build the Go application
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o bff ./cmd
 
+# Install setup-envtest and download K8s binaries for mock mode
+RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17
```
lucferbux: Well, this is not good, at least fo the "main" dockerfile, `setup-envtest` downloads the full kubernetes control plane and bloats the resulting binary, this will polute the production image with things that are not needed in prod.

My recommendation here would be having a second Dockerfile for standalone if we need this  or try to conditionally run this command only in specific parts (I would go with the first one) but we cannot add this to the final image.

docker-compose-local.yaml
```diff
@@ -3,12 +3,38 @@
...
+  model-registry-ui:
+    build:
+      context: ./clients/ui
+      dockerfile: Dockerfile
+      args:
+        DEPLOYMENT_MODE: standalone
+        STYLE_THEME: mui-theme
+    command:
+      - --port=9000
+      - --mock-k8s-client=true
```
lucferbux, line 23: nit: we can add the model-catalog and model-registry mocks set by default to false to explicitly display it in the docker compose spec, that adds legibility (we are mocking k8s but not the rest of the services)

docker-compose.yaml
```diff
@@ -3,12 +3,34 @@
...
 # - a model-registry server connected to the PostgreSQL database.
 # - a model-catalog server
+# - a model-registry-ui (standalone mode)
 version: '3.8'
 services:
+  model-registry-ui:
+    image: ghcr.io/kubeflow/model-registry/ui:latest
+    pull_policy: always
+    command:
+      - --port=9000
```
lucferbux, line 18: Same thing here.

docker-compose.yaml
```diff
@@ -3,12 +3,34 @@
...
+# - a model-registry-ui (standalone mode)
 # or
 # - a PostgreSQL database,
 # - a model-registry server connected to the PostgreSQL database.
 # - a model-catalog server
+# - a model-registry-ui (standalone mode)
 version: '3.8'
 services:
+  model-registry-ui:
+    image: ghcr.io/kubeflow/model-registry/ui:latest
```
lucferbux: Wouldn't be better to use the [ui-standalone](https://github.com/kubeflow/model-registry/pkgs/container/model-registry%2Fui-standalone) image better here?

lucferbux (changes requested):
A couple of nits but there's the part of adding env-test to the dockerfile, we should avoid that at all costs, maybe we might wanna explore having an extra dockerfile or smth similar.

docker-compose.yaml
ederign: Fixed!

clients/ui/Dockerfile
ederign: Fixed!

ederign:
@lucferbux ready for another round!

ederign:
@tarilabs, when you have a chance, could you please take a look at it?

PR #2059, opened 2026-01-05
User server side-filtering for hw configuration table

manaswinidas:
## Description

- Use server-side filtering for hardware configuration table and related Cypress tests
- Add icon to `Validated` labels in catalog 

https://github.com/user-attachments/assets/7b2e95f3-55c7-43ed-8ba2-d9ba1581409b

Screenshot 2026-01-05 at 4 33 40 PM
Screenshot 2026-01-05 at 4 33 49 PM

## How Has This Been Tested?

Check whether the `filterQuery` in the payload corresponding to `http://localhost:9000/model-registry/api/v1/model_catalog/sources/sample-source/performance_artifacts/repo1%2Fgranite-8b-code-instruct` is populated with the user-selected filters 

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
@@ -1,9 +1,7 @@
 /* eslint-disable camelcase */
 import { modelCatalog } from '~/__tests__/cypress/cypress/pages/modelCatalog';
 import {
-  mockCatalogAccuracyMetricsArtifact,
   mockCatalogModel,
-  mockCatalogModelArtifact,
   mockCatalogModelArtifactList,
   mockCatalogModelList,
   mockCatalogPerformanceMetricsArtifact,
@@ -111,6 +109,8 @@ const initIntercepts = ({
     mockCatalogFilterOptionsList(),
   );
 
+  // The /performance_artifacts endpoint only returns performance metrics artifacts
+  // (no accuracy or model artifacts - those are filtered server-side)
   cy.interceptApi(
     `GET /api/:apiVersion/model_catalog/sources/:sourceId/performance_artifacts/:modelName`,
     {
@@ -163,9 +163,10 @@ const initIntercepts = ({
             },
           },
         }),
-        mockCatalogAccuracyMetricsArtifact({}),
-        mockCatalogModelArtifact({}),
       ],
+      pageSize: 10,
+      size: 3,
+      nextPageToken: '',
     },
   ).as('getCatalogSourceModelArtifacts');
 };


clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -1,31 +1,67 @@
...
+        targetRPS,
+        latencyProperty,
+        recommendations: true,
+      },
+      filterData,
+      filterOptions,
+    );
+
+  // Extract performance artifacts from the list
+  const performanceArtifacts = React.useMemo(
```
YuliaKrimerman: Since we're now using server-side filtering, is this client-side filtering step still necessary? The server should already be returning only performance metrics artifacts. If the server is guaranteed to return the correct type, this filtering might be redundant.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -204,6 +204,20 @@ const isArrayOfSelections = (
  */
 const KNOWN_LESS_THAN_IDS: string[] = ALL_LATENCY_FIELD_NAMES;
 
+// const KNOWN_MORE_THAN_IDS: FilterId[] = [ModelCatalogNumberFilterKey.RPS_MEAN]; // TODO: populate with filters that need to talk about "more" values
```
YuliaKrimerman: Is this uncommented intentionally?

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -1,31 +1,67 @@
...
+      ),
+    [performanceArtifactsList.items],
+  );
 
   React.useEffect(() => {
     setPerformanceFiltersChangedOnDetailsPage(false);
   }, [setPerformanceFiltersChangedOnDetailsPage]);
 
-  if (loadError) {
+  if (performanceArtifactsError) {
```
YuliaKrimerman, line 42: Is there a case where `performanceArtifactsLoaded` might be `false` but there's no error, and we want to show a loading state?

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
manaswinidas: I think we still need client-side filtering for `hardware_type` 🤔

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
manaswinidas: Yes, this was unintentional - I removed these

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
mturley: No, we should not be doing any client-side filtering. The hardware configuration filter should be handled by filterQuery, which we should be able to get for free since we're already passing a filterQuery based on the filterData in context, and the hardware config dropdown should be populating a filter value there for the `hardware_configuration` property (the combined count x type string). If that's not functional with this PR only we can follow up after the changes to hardware config column/filter rendering @Philip-Carneiro is working on, but we should still remove any client-side filtering here.

mturley: Also we shouldn't need `filterArtifactsByType` anymore since all the artifacts returned by /artifacts/performance are performance artifacts, unlike the regular /artifacts endpoint that also returned other types.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
mturley: @YuliaKrimerman I think that's already covered by passing `isLoading={!performanceArtifactsLoaded}` to HardwareConfigurationTable.

mturley (changes requested):
Already left some inline responses - we need to double check what values need to be passed to the /artifacts/performance endpoint to properly filter everything. Some things are their own URL params as already implemented, but some things use filterQuery (based on what comes from the filter_options endpoint, there are options there with `artifacts.*` prefixes, and when passing those to filterQuery on this endpoint we need to strip the artifacts.* prefix - but this is already done by `filtersToArtifactsFilterQuery` so we just need to make sure the proper keys are in place in the filter state in context).

mturley:
/approve
/lgtm

@ppadti to follow up on fixing filters that still don't work, let me know if you want to pair on that

PR #2071, opened 2026-01-07
feat(catalog/ui): add architecture field to model catalog details page

chambridge:
## Description
Add architecture display to the model catalog details view with comprehensive validation and testing. The architecture field extracts CPU architecture information from model artifacts and displays valid architectures (amd64, arm64, s390x, ppc64le) to users.

This is an improvement in user experience because many models are not multi-architecture and may only be "amd64", an associate of my was trying to deploy to an arm64 node and could not determine what was a valid model to deploy for their use case. The intent here is to provide further clarity to avoid deployment failures.

Features implemented:
- Architecture field display in ModelDetailsView UI component
- getArchitecturesFromArtifacts() utility function for extraction and validation
- CpuArchitecture enum and VALID_ARCHITECTURES constant for single source of truth
- JSON array parsing from model artifact custom properties
- Lowercase normalization and validation filtering
- Error logging for debugging invalid architecture JSON

image
image

Related to: https://github.com/opendatahub-io/model-metadata-collection/pull/85

## How Has This Been Tested?
Using Tilt deployment with demo data.

Test coverage:
- 24 unit tests for getArchitecturesFromArtifacts() utility
- 4 E2E Cypress tests for architecture display scenarios
- Test data added to YAML catalog files for validation

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/mocks/static_data_mock.go
+++ b/clients/ui/bff/internal/mocks/static_data_mock.go
@@ -2,6 +2,7 @@ package mocks
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"os"
@@ -224,10 +225,10 @@ func newCustomProperties() *map[string]openapi.MetadataValue {
 }
 
 func catalogCustomProperties() *map[string]openapi.MetadataValue {
-	return catalogCustomPropertiesWithVariant("", "FP16", "[\"amd64\", \"arm64\", \"s390x\", \"ppc64le\"]")
+	return catalogCustomPropertiesWithVariant("", "FP16")
 }
 
-func catalogCustomPropertiesWithVariant(variantGroupId string, tensorType string, architectures string) *map[string]openapi.MetadataValue {
+func catalogCustomPropertiesWithVariant(variantGroupId string, tensorType string) *map[string]openapi.MetadataValue {
 	result := map[string]openapi.MetadataValue{
 		"tensorflow": {
 			MetadataStringValue: &openapi.MetadataStringValue{
@@ -292,12 +293,6 @@ func catalogCustomPropertiesWithVariant(variantGroupId string, tensorType string
 		"model_type": {
 			MetadataStringValue: &openapi.MetadataStringValue{
 				StringValue:  "generative",
-				MetadataType: "MetadataStringValue",
-			},
-		},
-		"architecture": {
-			MetadataStringValue: &openapi.MetadataStringValue{
-				StringValue:  architectures,
 				MetadataType: "MetadataStringValue",
 			},
 		},
@@ -368,7 +363,7 @@ func GetCatalogModelMocks() []models.CatalogModel {
 		LicenseLink:      stringToPointer("https://www.apache.org/licenses/LICENSE-2.0.txt"),
 		Maturity:         stringToPointer("Technology preview"),
 		Language:         []string{"ar", "cs", "de", "en", "es", "fr", "it", "ja", "ko", "nl", "pt", "zh"},
-		CustomProperties: catalogCustomPropertiesWithVariant(graniteVariantGroupId, "FP16", "[\"amd64\", \"arm64\", \"s390x\", \"ppc64le\"]"),
+		CustomProperties: catalogCustomPropertiesWithVariant(graniteVariantGroupId, "FP16"),
 		Logo:             stringToPointer("data:image/svg+xml;base64,PHN2ZyBpZD0iTGF5ZXJfMSIgZGF0YS1uYW1lPSJMYXllciAxIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxOTIgMTQ1Ij48ZGVmcz48c3R5bGU+LmNscy0xe2ZpbGw6I2UwMDt9PC9zdHlsZT48L2RlZnM+PHRpdGxlPlJlZEhhdC1Mb2dvLUhhdC1Db2xvcjwvdGl0bGU+PHBhdGggZD0iTTE1Ny43Nyw2Mi42MWExNCwxNCwwLDAsMSwuMzEsMy40MmMwLDE0Ljg4LTE4LjEsMTcuNDYtMzAuNjEsMTcuNDZDNzguODMsODMuNDksNDIuNTMsNTMuMjYsNDIuNTMsNDRhNi40Myw2LjQzLDAsMCwxLC4yMi0xLjk0bC0zLjY2LDkuMDZhMTguNDUsMTguNDUsMCwwLDAtMS41MSw3LjMzYzAsMTguMTEsNDEsNDUuNDgsODcuNzQsNDUuNDgsMjAuNjksMCwzNi40My03Ljc2LDM2LjQzLTIxLjc3LDAtMS4wOCwwLTEuOTQtMS43My0xMC4xM1oiLz48cGF0aCBjbGFzcz0iY2xzLTEiIGQ9Ik0xMjcuNDcsODMuNDljMTIuNTEsMCwzMC42MS0yLjU4LDMwLjYxLTE3LjQ2YTE0LDE0LDAsMCwwLS4zMS0zLjQybC03LjQ1LTMyLjM2Yy0xLjcyLTcuMTItMy4yMy0xMC4zNS0xNS43My0xNi42QzEyNC44OSw4LjY5LDEwMy43Ni41LDk3LjUxLjUsOTEuNjkuNSw5MCw4LDgzLjA2LDhjLTYuNjgsMC0xMS42NC01LjYtMTcuODktNS42LTYsMC05LjkxLDQuMDktMTIuOTMsMTIuNSwwLDAtOC40MSwyMy43Mi05LjQ5LDI3LjE2QTYuNDMsNi40MywwLDAsMCw0Mi41Myw0NGMwLDkuMjIsMzYuMywzOS40NSw4NC45NCwzOS40NU0xNjAsNzIuMDdjMS43Myw4LjE5LDEuNzMsOS4wNSwxLjczLDEwLjEzLDAsMTQtMTUuNzQsMjEuNzctMzYuNDMsMjEuNzdDNzguNTQsMTA0LDM3LjU4LDc2LjYsMzcuNTgsNTguNDlhMTguNDUsMTguNDUsMCwwLDEsMS41MS03LjMzQzIyLjI3LDUyLC41LDU1LC41LDc0LjIyYzAsMzEuNDgsNzQuNTksNzAuMjgsMTMzLjY1LDcwLjI4LDQ1LjI4LDAsNTYuNy0yMC40OCw1Ni43LTM2LjY1LDAtMTIuNzItMTEtMjcuMTYtMzAuODMtMzUuNzgiLz48L3N2Zz4="),
 		Readme: stringToPointer(`---
 pipeline_tag: text-generation
@@ -722,7 +717,7 @@ Granite 3.1 Instruct Models are primarily finetuned using instruction-response p
 		Maturity:         stringToPointer("Generally Available"),
 		Language:         []string{"en"},
 		SourceId:         stringToPointer("sample-source"),
-		CustomProperties: catalogCustomPropertiesWithVariant(graniteVariantGroupId, "INT4", "[\"amd64\"]"),
+		CustomProperties: catalogCustomPropertiesWithVariant(graniteVariantGroupId, "INT4"),
 		Logo:             stringToPointer("data:image/svg+xml;base64,PHN2ZyBpZD0iTGF5ZXJfMSIgZGF0YS1uYW1lPSJMYXllciAxIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxOTIgMTQ1Ij48ZGVmcz48c3R5bGU+LmNscy0xe2ZpbGw6I2UwMDt9PC9zdHlsZT48L2RlZnM+PHRpdGxlPlJlZEhhdC1Mb2dvLUhhdC1Db2xvcjwvdGl0bGU+PHBhdGggZD0iTTE1Ny43Nyw2Mi42MWExNCwxNCwwLDAsMSwuMzEsMy40MmMwLDE0Ljg4LTE4LjEsMTcuNDYtMzAuNjEsMTcuNDZDNzguODMsODMuNDksNDIuNTMsNTMuMjYsNDIuNTMsNDRhNi40Myw2LjQzLDAsMCwxLC4yMi0xLjk0bC0zLjY2LDkuMDZhMTguNDUsMTguNDUsMCwwLDAtMS41MSw3LjMzYzAsMTguMTEsNDEsNDUuNDgsODcuNzQsNDUuNDgsMjAuNjksMCwzNi40My03Ljc2LDM2LjQzLTIxLjc3LDAtMS4wOCwwLTEuOTQtMS43My0xMC4xM1oiLz48cGF0aCBjbGFzcz0iY2xzLTEiIGQ9Ik0xMjcuNDcsODMuNDljMTIuNTEsMCwzMC42MS0yLjU4LDMwLjYxLTE3LjQ2YTE0LDE0LDAsMCwwLS4zMS0zLjQybC03LjQ1LTMyLjM2Yy0xLjcyLTcuMTItMy4yMy0xMC4zNS0xNS43My0xNi42QzEyNC44OSw4LjY5LDEwMy43Ni41LDk3LjUxLjUsOTEuNjkuNSw5MCw4LDgzLjA2LDhjLTYuNjgsMC0xMS42NC01LjYtMTcuODktNS42LTYsMC05LjkxLDQuMDktMTIuOTMsMTIuNSwwLDAtOC40MSwyMy43Mi05LjQ5LDI3LjE2QTYuNDMsNi40MywwLDAsMCw0Mi41Myw0NGMwLDkuMjIsMzYuMywzOS40NSw4NC45NCwzOS40NU0xNjAsNzIuMDdjMS43Myw4LjE5LDEuNzMsOS4wNSwxLjczLDEwLjEzLDAsMTQtMTUuNzQsMjEuNzctMzYuNDMsMjEuNzdDNzguNTQsMTA0LDM3LjU4LDc2LjYsMzcuNTgsNTguNDlhMTguNDUsMTguNDUsMCwwLDEsMS41MS03LjMzQzIyLjI3LDUyLC41LDU1LC41LDc0LjIyYzAsMzEuNDgsNzQuNTksNzAuMjgsMTMzLjY1LDcwLjI4LDQ1LjI4LDAsNTYuNy0yMC40OCw1Ni43LTM2LjY1LDAtMTIuNzItMTEtMjcuMTYtMzAuODMtMzUuNzgiLz48L3N2Zz4="),
 	}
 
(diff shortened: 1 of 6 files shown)


clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
```diff
@@ -25,6 +28,91 @@ describe('Model Catalog Details Page', () => {
...
+
+  it('does not show architecture field when no architectures are available', () => {
+    modelCatalog.findLoadingState().should('not.exist');
+    modelCatalog.findModelCatalogDetailLink().first().click();
+    // Architecture field should not exist when no valid architectures
+    modelCatalog.findModelArchitecture().should('not.exist');
+  });
+
+  it('shows architecture field with valid architectures', () => {
+    // Override the artifacts intercept with architecture data
```
manaswinidas: **Potential race condition in intercept setup**

The `interceptArtifactsList` is called inside the test body after the `describe` block's `beforeEach` has already set up default intercepts. Depending on timing, the click might happen before this new intercept is registered.

Consider restructuring to ensure the intercept is set up before any navigation:

```typescript
it('shows architecture field with valid architectures', () => {
  // Set up intercept first
  interceptArtifactsList({
    items: [
      mockCatalogModelArtifact({
        customProperties: {
          architecture: {
            string_value: '["amd64", "arm64", "s390x"]',
            metadataType: ModelRegistryMetadataType.STRING,
          },
        },
      }),
    ],
    size: 1,
    pageSize: 10,
    nextPageToken: '',
  });

  // Then navigate
  cy.visit('/model-catalog'); // or use cy.reload() to ensure fresh state
  modelCatalog.findLoadingState().should('not.exist');
  modelCatalog.findModelCatalogDetailLink().first().click();
  // ...
});
```

Or use separate `describe` blocks with different `beforeEach` setups for architecture-specific tests.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -833,6 +839,306 @@ describe('hasFiltersApplied', () => {
...
+          metadataType: ModelRegistryMetadataType.STRING,
+        },
+      },
+    });
+    const result = getArchitecturesFromArtifacts([artifact]);
+    expect(result).toHaveLength(4);
+    expect(result).toContain('amd64');
+    expect(result).toContain('arm64');
+    expect(result).toContain('s390x');
+    expect(result).toContain('ppc64le');
```
manaswinidas, line 1163: **Missing test: Duplicate architecture handling**

The test suite doesn't verify behavior when the same architecture appears multiple times in the input (possibly with different casing). This could lead to duplicate values in the output.

Consider adding a test case:
```typescript
it('should deduplicate architectures when duplicates exist', () => {
  const artifact = mockCatalogModelArtifact({
    customProperties: {
      architecture: {
        string_value: '["amd64", "AMD64", "amd64"]',
        metadataType: ModelRegistryMetadataType.STRING,
      },
    },
  });
  const result = getArchitecturesFromArtifacts([artifact]);
  expect(result).toEqual(['amd64']); // Should return only one instance
});
```

With the current implementation, this would return `['amd64', 'amd64', 'amd64']` which may not be the intended behavior.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -833,6 +839,306 @@ describe('hasFiltersApplied', () => {
...
+          string_value: '',
+          metadataType: ModelRegistryMetadataType.STRING,
+        },
+      },
+    });
+    const result = getArchitecturesFromArtifacts([artifact]);
+    expect(result).toEqual([]);
+  });
+
+  it('should return empty array when architecture property has wrong metadata type', () => {
```
manaswinidas, line 895: **Missing test: Architecture aliases (x86_64, aarch64)**

Common architecture aliases like `x86_64` (for `amd64`) and `aarch64` (for `arm64`) are frequently used in container images and model artifacts, but they're not tested here.

Consider adding a test case to document the expected behavior:
```typescript
it('should handle common architecture aliases', () => {
  const artifact = mockCatalogModelArtifact({
    customProperties: {
      architecture: {
        string_value: '["x86_64", "aarch64"]',
        metadataType: ModelRegistryMetadataType.STRING,
      },
    },
  });
  const result = getArchitecturesFromArtifacts([artifact]);
  // Option 1: If aliases should be normalized
  expect(result).toEqual(['amd64', 'arm64']);
  // Option 2: If aliases are intentionally not supported
  expect(result).toEqual([]);
});
```

Currently `x86_64` and `aarch64` would be filtered out as "invalid" architectures. If this is intentional, consider adding this test to document the behavior. If not, the `VALID_ARCHITECTURES` set and/or an alias mapping should be updated.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -833,6 +839,306 @@ describe('hasFiltersApplied', () => {
...
+
+  it('should return empty array when no model artifact exists', () => {
+    const metricsArtifact: CatalogArtifacts = {
+      artifactType: CatalogArtifactType.metricsArtifact,
+      metricsType: MetricsType.performanceMetrics,
+    };
+    const result = getArchitecturesFromArtifacts([metricsArtifact]);
+    expect(result).toEqual([]);
+  });
+
```
manaswinidas, line 856: **Missing test: Verify console.warn is called for invalid JSON**

The implementation logs a warning when JSON parsing fails (`console.warn('Failed to parse architecture JSON:', ...)`), but this behavior isn't verified in tests.

Consider adding:
```typescript
it('should log a warning when architecture JSON is invalid', () => {
  const warnSpy = jest.spyOn(console, 'warn').mockImplementation();
  const artifact = mockCatalogModelArtifact({
    customProperties: {
      architecture: {
        string_value: 'not valid json',
        metadataType: ModelRegistryMetadataType.STRING,
      },
    },
  });
  
  getArchitecturesFromArtifacts([artifact]);
  
  expect(warnSpy).toHaveBeenCalledWith(
    'Failed to parse architecture JSON:',
    'not valid json',
    expect.any(Error)
  );
  warnSpy.mockRestore();
});
```

This ensures the debugging behavior is preserved across refactors.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
   artifacts.some((artifact) => artifact.artifactType === CatalogArtifactType.modelArtifact);
 
+/**
+ * Extracts and validates architecture values from the model artifact's custom properties.
+ * The architecture custom property should be a JSON-encoded array of architecture strings.
+ * Only valid architecture types (amd64, arm64, s390x, ppc64le) are returned.
+ *
+ * @param artifacts Array of catalog artifacts to search
+ * @returns Array of valid architecture strings, or empty array if none found or invalid
+ */
```
manaswinidas, line 116: **Code Duplication: Repeated artifact lookup pattern**

This `artifacts.find()` pattern is duplicated with the existing `getModelArtifactUri` function (lines 88-90 in this file). Consider extracting a shared helper function:

```typescript
const findModelArtifact = (artifacts: CatalogArtifacts[]) =>
  artifacts.find(
    (artifact) => artifact.artifactType === CatalogArtifactType.modelArtifact,
  );
```

Then both functions can use it:
```typescript
export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
  const modelArtifact = findModelArtifact(artifacts);
  return modelArtifact?.uri || '';
};

export const getArchitecturesFromArtifacts = (artifacts: CatalogArtifacts[]): string[] => {
  const modelArtifact = findModelArtifact(artifacts);
  // ...rest of the logic
};
```

This reduces duplication and creates a single source of truth for artifact lookup logic.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+  if (!architectureString) {
+    return [];
+  }
+
+  try {
+    const parsed = JSON.parse(architectureString);
+
+    if (!Array.isArray(parsed)) {
+      return [];
+    }
```
manaswinidas, line 136: **Performance: Double iteration and missing deduplication**

Two issues here:

1. **Double iteration**: The array is iterated twice (`.filter()` then `.map()`), each calling `toLowerCase()`. This can be optimized to a single pass.

2. **Missing deduplication**: If the input contains `["amd64", "AMD64"]`, the output would be `["amd64", "amd64"]` (duplicates).

Consider using `.reduce()` for a single-pass solution with deduplication:

```typescript
const seen = new Set<string>();
return parsed.reduce<string[]>((acc, item) => {
  if (typeof item === 'string') {
    const normalized = item.toLowerCase();
    if (VALID_ARCHITECTURES.has(normalized) && !seen.has(normalized)) {
      seen.add(normalized);
      acc.push(normalized);
    }
  }
  return acc;
}, []);
```

Or a simpler approach with `Set`:
```typescript
return [...new Set(
  parsed
    .filter((item): item is string => typeof item === 'string')
    .map((item) => item.toLowerCase())
    .filter((item) => VALID_ARCHITECTURES.has(item))
)];
```

While the performance impact is minimal for small arrays, deduplication is important for correctness.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -226,8 +226,24 @@ export enum CatalogModelCustomPropertyKey {
...
+/**
+ * Valid CPU architecture types
+ */
+export enum CpuArchitecture {
+  AMD64 = 'amd64',
+  ARM64 = 'arm64',
+  S390X = 's390x',
+  PPC64LE = 'ppc64le',
+}
+
```
manaswinidas: **Missing Architecture Aliases**

The `VALID_ARCHITECTURES` set only includes canonical names, but common aliases are frequently used in container images and OCI artifacts:

| Canonical | Common Aliases |
|-----------|---------------|
| `amd64` | `x86_64`, `x64` |
| `arm64` | `aarch64` |

Consider adding alias support:

```typescript
export const ARCHITECTURE_ALIASES: Record<string, CpuArchitecture> = {
  'x86_64': CpuArchitecture.AMD64,
  'x64': CpuArchitecture.AMD64,
  'aarch64': CpuArchitecture.ARM64,
};

export const normalizeArchitecture = (arch: string): CpuArchitecture | undefined => {
  const lower = arch.toLowerCase();
  if (VALID_ARCHITECTURES.has(lower)) {
    return lower as CpuArchitecture;
  }
  return ARCHITECTURE_ALIASES[lower];
};
```

This would allow model artifacts with `["x86_64", "aarch64"]` to be correctly displayed as `amd64, arm64`.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+ *
+ * @param artifacts Array of catalog artifacts to search
+ * @returns Array of valid architecture strings, or empty array if none found or invalid
+ */
+export const getArchitecturesFromArtifacts = (artifacts: CatalogArtifacts[]): string[] => {
+  const modelArtifact = artifacts.find(
+    (artifact) => artifact.artifactType === CatalogArtifactType.modelArtifact,
+  );
+
+  if (!modelArtifact?.customProperties) {
```
manaswinidas: **Code Smell: eslint-disable comments**

Multiple `eslint-disable` comments suggest potential type definition issues:

1. `@typescript-eslint/dot-notation` - Could be avoided by using a typed key from `CatalogModelCustomPropertyKey.ARCHITECTURE` instead of the string literal `'architecture'`.

2. `@typescript-eslint/no-unnecessary-condition` - This suggests the type definitions might not accurately reflect nullable states.

Consider:
```typescript
import { CatalogModelCustomPropertyKey } from '~/concepts/modelCatalog/const';

const architectureProp = modelArtifact.customProperties[CatalogModelCustomPropertyKey.ARCHITECTURE];
```

This would leverage the enum you've already added and potentially resolve the linting issues while improving type safety.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsView.tsx
```diff
@@ -71,6 +72,9 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
     ? getCustomPropString(model.customProperties, CatalogModelCustomPropertyKey.SIZE)
     : '';
 
+  // Extract architectures from artifacts
```
manaswinidas: **Performance: Missing memoization**

This computation runs on every render. While the cost is low, consider memoizing to prevent unnecessary recalculations when unrelated state changes:

```typescript
const architectures = React.useMemo(
  () => (artifactLoaded ? getArchitecturesFromArtifacts(artifacts.items) : []),
  [artifactLoaded, artifacts.items]
);
```

This is consistent with React best practices for derived state computations, especially in components that may re-render frequently.

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -295,6 +295,12 @@ func catalogCustomPropertiesWithVariant(variantGroupId string, tensorType string
 				MetadataType: "MetadataStringValue",
 			},
 		},
+		"architecture": {
+			MetadataStringValue: &openapi.MetadataStringValue{
+				StringValue:  architectures,
+				MetadataType: "MetadataStringValue",
```
manaswinidas: **Magic String: Hardcoded property key**

The property key `"architecture"` is hardcoded as a string literal. Consider defining a constant for consistency across the codebase:

```go
const ArchitecturePropertyKey = "architecture"
```

This would align with best practices and make it easier to maintain if the property key ever needs to change.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+
+    return parsed
+      .filter((item) => typeof item === 'string' && VALID_ARCHITECTURES.has(item.toLowerCase()))
+      .map((item) => item.toLowerCase());
+  } catch (error) {
+    // Invalid JSON - return empty array
+    // eslint-disable-next-line no-console
+    console.warn('Failed to parse architecture JSON:', architectureString, error);
+    return [];
+  }
```
manaswinidas, line 157: **Type Safety: Return type could be more specific**

The function returns `string[]` but could return `CpuArchitecture[]` for better type safety. This would ensure consumers of this function know they're working with validated architecture values:

```typescript
export const getArchitecturesFromArtifacts = (artifacts: CatalogArtifacts[]): CpuArchitecture[] => {
  // ...
  return parsed
    .filter((item): item is string => typeof item === 'string')
    .map((item) => item.toLowerCase())
    .filter((item): item is CpuArchitecture => VALID_ARCHITECTURES.has(item));
};
```

This provides compile-time guarantees that the returned values are valid architectures.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -226,8 +226,24 @@ export enum CatalogModelCustomPropertyKey {
   VALIDATED_ON = 'validated_on',
   TENSOR_TYPE = 'tensor_type',
   SIZE = 'size',
+  ARCHITECTURE = 'architecture',
```
manaswinidas, line 229: **Inconsistency: Enum defined but not fully utilized**

The `ARCHITECTURE = 'architecture'` is added to the enum here, but in `modelCatalogUtils.ts` the code still uses the string literal `'architecture'` with an eslint-disable comment:

```typescript
// eslint-disable-next-line @typescript-eslint/dot-notation
const architectureProp = modelArtifact.customProperties['architecture'];
```

For consistency, this enum value should be imported and used:

```typescript
import { CatalogModelCustomPropertyKey } from '~/concepts/modelCatalog/const';

const architectureProp = modelArtifact.customProperties[CatalogModelCustomPropertyKey.ARCHITECTURE];
```

This would also eliminate the need for the eslint-disable comment.

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -224,10 +224,10 @@ func newCustomProperties() *map[string]openapi.MetadataValue {
 }
 
 func catalogCustomProperties() *map[string]openapi.MetadataValue {
-	return catalogCustomPropertiesWithVariant("", "FP16")
+	return catalogCustomPropertiesWithVariant("", "FP16", "[\"amd64\", \"arm64\", \"s390x\", \"ppc64le\"]")
```
manaswinidas: **Function Signature: Consider using variadic or struct parameter**

The function now has 3 string parameters which can be confusing at call sites. Consider:

1. Using a struct for optional parameters:
```go
type CatalogCustomPropertiesOptions struct {
    VariantGroupId string
    TensorType     string
    Architectures  string
}
```

2. Or making architectures a slice for type safety:
```go
func catalogCustomPropertiesWithVariant(variantGroupId, tensorType string, architectures []string) *map[string]openapi.MetadataValue {
    archJson, _ := json.Marshal(architectures)
    // ...
}
```

This prevents errors like passing an invalid JSON string and makes the API more self-documenting.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+
+  // eslint-disable-next-line @typescript-eslint/dot-notation
+  const architectureProp = modelArtifact.customProperties['architecture'];
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!architectureProp || architectureProp.metadataType !== ModelRegistryMetadataType.STRING) {
+    return [];
+  }
```
manaswinidas, line 129: **Unnecessary Block: Redundant empty string check**

This check is redundant. The previous check `if (!architectureProp || ...)` already handles the case where `architectureProp` is falsy. And if `string_value` is an empty string `""`, `JSON.parse("")` will throw a `SyntaxError` which is already caught in the try-catch block below.

You can remove this block:
```typescript
// Remove this - it's covered by the try-catch
if (!architectureString) {
  return [];
}
```

The empty string case would be handled by the catch block, making this check unnecessary. The only reason to keep it would be to avoid the console.warn for empty strings, which could be addressed differently:

```typescript
try {
  if (!architectureString) {
    return [];
  }
  const parsed = JSON.parse(architectureString);
  // ...
}
```

Or simply let it fall through to the catch (empty string is technically invalid JSON anyway).

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+
+  const architectureString = architectureProp.string_value;
+
+  if (!architectureString) {
+    return [];
+  }
+
```
manaswinidas: **Unnecessary Block: Separate array check**

This `if (!Array.isArray(parsed))` check could be combined with the filter logic for cleaner code:

```typescript
try {
  const parsed = JSON.parse(architectureString);
  
  // Handle both non-array and array cases in one flow
  const items = Array.isArray(parsed) ? parsed : [];
  
  return items
    .filter((item) => typeof item === 'string' && VALID_ARCHITECTURES.has(item.toLowerCase()))
    .map((item) => item.toLowerCase());
} catch (error) {
  // ...
}
```

Or even more concise with nullish coalescing:
```typescript
const items = Array.isArray(parsed) ? parsed : [];
```

This reduces the number of early returns and makes the happy path clearer.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -226,8 +226,24 @@ export enum CatalogModelCustomPropertyKey {
...
 }
 
+/**
```
manaswinidas: **Unused Enum: CpuArchitecture enum values not directly used**

The `CpuArchitecture` enum is defined but its individual values (`AMD64`, `ARM64`, etc.) are never directly referenced in the code. The code only uses `VALID_ARCHITECTURES` Set which is derived from the enum values.

Consider whether you need the enum at all:

**Option 1: Keep enum, use it for type safety (recommended)**
```typescript
export const getArchitecturesFromArtifacts = (artifacts: CatalogArtifacts[]): CpuArchitecture[] => {
  // ... and filter returns CpuArchitecture values
};
```

**Option 2: Remove enum, just use the Set with literal values**
```typescript
export const VALID_ARCHITECTURES = new Set(['amd64', 'arm64', 's390x', 'ppc64le'] as const);
export type CpuArchitecture = typeof VALID_ARCHITECTURES extends Set<infer T> ? T : never;
```

The current implementation has an enum that adds no value since it's only used to populate the Set. Either leverage the enum for type safety or simplify to just the Set.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+  try {
+    const parsed = JSON.parse(architectureString);
+
+    if (!Array.isArray(parsed)) {
+      return [];
+    }
+
+    return parsed
+      .filter((item) => typeof item === 'string' && VALID_ARCHITECTURES.has(item.toLowerCase()))
```
manaswinidas: **Unnecessary: Console.warn in catch block**

The `console.warn` in the catch block logs detailed error information, but this may be unnecessary noise in production:

1. **Issue**: This will log every time a model has malformed architecture data, which could spam the console in production
2. **The error parameter is logged but not used meaningfully** - just passed through

Consider:

**Option 1: Remove the log entirely** (silent fail is documented in JSDoc)
```typescript
catch {
  return [];
}
```

**Option 2: Use conditional logging** (only in development)
```typescript
catch (error) {
  if (process.env.NODE_ENV === 'development') {
    console.warn('Failed to parse architecture JSON:', architectureString, error);
  }
  return [];
}
```

**Option 3: Use a proper logger if available**
```typescript
catch (error) {
  logger.debug('Failed to parse architecture JSON', { architectureString, error });
  return [];
}
```

The current approach adds runtime overhead for error logging that may not provide value in production.

manaswinidas:
Tested on kind cluster - able to see architecture. Unable to test in midstream (needs rebase)
Screenshot 2026-01-23 at 4 16 17 PM

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
 export const hasModelArtifacts = (artifacts: CatalogArtifacts[]): boolean =>
   artifacts.some((artifact) => artifact.artifactType === CatalogArtifactType.modelArtifact);
 
+/**
+ * Extracts and validates architecture values from the model artifact's custom properties.
+ * The architecture custom property should be a JSON-encoded array of architecture strings.
+ * Only valid architecture types (amd64, arm64, s390x, ppc64le) are returned.
```
mturley: @chambridge why do we need to validate against known architectures hard coded on the frontend? we don't do this for any other fields, and this adds maintenance overhead if more architectures are added on the backend. can we not just show whatever's there?

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -99,6 +101,55 @@ export const getModelArtifactUri = (artifacts: CatalogArtifacts[]): string => {
...
+export const getArchitecturesFromArtifacts = (artifacts: CatalogArtifacts[]): string[] => {
+  const modelArtifact = artifacts.find(
+    (artifact) => artifact.artifactType === CatalogArtifactType.modelArtifact,
+  );
+
+  if (!modelArtifact?.customProperties) {
+    return [];
+  }
+
+  // eslint-disable-next-line @typescript-eslint/dot-notation
```
mturley: Let's not disable any eslint rules unless absolutely necessary, and instead figure out why the rule is failing. In this case, you can either use dot notation `.architecture` or use `CatalogModelCustomPropertyKey.ARCHITECTURE` instead of the hard coded string.

If we do have a good reason for disabling an eslint rule we need a comment explaining why.

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
chambridge: Not addressing this since Mike said we shouldn't be validating architectures in the frontend, so that code will be removed.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
chambridge: sure, that's fine. I'll remove.

PR #2075, opened 2026-01-08
Performance filters for catalog landing page

manaswinidas:
## Description

- Added performance filters for catalog landing page when toggle is switched ON
- Workload type dropdown is now single-select instead of multi-select
- Chips under the filters that only show when we change the default filters(more cosmetic changes upcoming in a follow up PR)
- Fetching and using default filter values from `namedQueries` and filter reset behavior 
- Model card content changes from TTFT to ITL/E2E based on filter changes
- Min RPS is now Max RPS
- Unit and Cypress test fixes 

https://github.com/user-attachments/assets/d9c91bc7-4e8f-4514-9522-3221e6a97d18

## How Has This Been Tested?

Setup kind cluster - switch ON performance filters toggle and play around the filters and check whether the correct model cards data is displayed

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
@@ -56,12 +56,21 @@ const HardwareConfigurationFilterToolbar: React.FC<HardwareConfigurationFilterTo
     return null;
   }
 
+  // Custom clear filters button with test ID for Cypress tests
+  const customLabelGroupContent = onResetAllFilters && hasVisibleChips && (
+    <ToolbarItem>
+      <Button variant="link" onClick={onResetAllFilters} data-testid="clear-all-filters-button">
+        Clear all filters
+      </Button>
+    </ToolbarItem>
+  );
+
   return (
     <Toolbar
       {...(onResetAllFilters
         ? {
             clearAllFilters: onResetAllFilters,
-            clearFiltersButtonText: hasVisibleChips ? 'Clear all filters' : '',
+            customLabelGroupContent,
           }
         : {})}
     >
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
@@ -17,6 +17,7 @@ import ModelCatalogCard from '~/app/pages/modelCatalog/components/ModelCatalogCa
 import {
   getSourceFromSourceId,
   hasFiltersApplied,
+  getBasicFiltersOnly,
 } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
 import EmptyModelCatalogState from '~/app/pages/modelCatalog/EmptyModelCatalogState';
 import ScrollViewOnMount from '~/app/shared/components/ScrollViewOnMount';
@@ -37,15 +38,23 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
     filterOptionsLoaded,
     filterOptionsLoadError,
     catalogSources,
+    performanceViewEnabled,
   } = React.useContext(ModelCatalogContext);
   const filtersApplied = hasFiltersApplied(filterData);
+
+  // When performance view is disabled, exclude performance filters from API queries
+  // Memoize to prevent infinite re-fetching
+  const effectiveFilterData = React.useMemo(
+    () => (performanceViewEnabled ? filterData : getBasicFiltersOnly(filterData)),
+    [performanceViewEnabled, filterData],
+  );
 
   const { catalogModels, catalogModelsLoaded, catalogModelsLoadError } = useCatalogModelsBySources(
     '',
     selectedSourceLabel === 'All models' ? undefined : selectedSourceLabel,
     10,
     searchTerm,
-    filterData,
+    effectiveFilterData,
     filterOptions,
   );
 
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
@@ -14,6 +14,7 @@ import {
   decodeParams,
   getActiveLatencyFieldName,
 } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+import { parseLatencyFieldName } from '~/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils';
 import {
   applyFilterValue,
   getDefaultFiltersFromNamedQuery,
@@ -26,14 +27,15 @@ const PerformanceInsightsView = (): React.JSX.Element => {
     filterData,
     filterOptions,
     filterOptionsLoaded,
+    performanceViewEnabled,
     setPerformanceFiltersChangedOnDetailsPage,
     setFilterData,
   } = React.useContext(ModelCatalogContext);
 
   // Apply default performance filters on mount if they don't have values yet
-  // This handles the case where user navigates to details page without enabling toggle first
+  // Only apply defaults if performance view toggle is enabled
   React.useEffect(() => {
-    if (!filterOptionsLoaded || !filterOptions?.namedQueries) {
+    if (!filterOptionsLoaded || !filterOptions?.namedQueries || !performanceViewEnabled) {
       return;
     }
 
@@ -52,11 +54,15 @@ const PerformanceInsightsView = (): React.JSX.Element => {
     }
     // Only run on mount when filterOptions become available
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [filterOptionsLoaded]);
+  }, [filterOptionsLoaded, performanceViewEnabled]);
 
   // Get performance-specific filter params for the /performance_artifacts endpoint
   const targetRPS = filterData[ModelCatalogNumberFilterKey.MAX_RPS];
-  const latencyProperty = getActiveLatencyFieldName(filterData);
+  // Get full filter key and convert to short property key for the catalog API
+  const latencyFieldName = getActiveLatencyFieldName(filterData);
+  const latencyProperty = latencyFieldName
+    ? parseLatencyFieldName(latencyFieldName).propertyKey
+    : undefined;
 
   // Fetch performance artifacts from server with filtering/sorting/pagination
   const [performanceArtifactsList, performanceArtifactsLoaded, performanceArtifactsError] =
--- a/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
@@ -291,8 +291,50 @@ export const filtersToFilterQuery = (
 };
 
 /**
+ * Find the server filter key for a given filter ID.
+ * Handles the case where local state uses short keys but server uses fully qualified keys.
+ */
+const findServerFilterKey = (
+  filterId: string,
+  filters: CatalogFilterOptions | undefined,
+): string | undefined => {
+  if (!filters) {
+    return undefined;
+  }
+
+  if (filterId in filters) {
+    return filterId;
+  }
+
+  for (const suffix of ['.string_value', '.double_value', '.int_value', '.array_value']) {
+    const key = `artifacts.${filterId}${suffix}`;
+    if (key in filters) {
+      return key;
+    }
+  }
+  return undefined;
+};
+
+/**
+ * Convert a server filter key to the format expected by the filterQuery.
+ * Strips the 'artifacts.' prefix if present.
+ */
+const convertToFilterQueryKey = (serverFilterKey: string): string => {
+  if (serverFilterKey.startsWith(ARTIFACTS_FILTER_PREFIX)) {
+    return serverFilterKey.substring(ARTIFACTS_FILTER_PREFIX.length);
+  }
+  return serverFilterKey;
+};
+
+/**
+ * Check if a filter ID is a latency field name.
+ */
+const isLatencyFieldName = (id: string): id is LatencyMetricFieldName =>
+  ALL_LATENCY_FIELD_NAMES.some((name) => name === id);
+
+/**
  * Converts filter data into a filter query string for the /artifacts/performance endpoint.
- * Only includes filters that have the 'artifacts.' prefix.
+ * Only includes filters that have the 'artifacts.' prefix and strips that prefix in the output.
  * RPS is NOT included in filterQuery - it's passed as targetRPS param instead.
  */
 export const filtersToArtifactsFilterQuery = (
@@ -300,24 +342,23 @@ export const filtersToArtifactsFilterQuery = (
   options: CatalogFilterOptionsList,
 ): string => {
   const serializedFilters: string[] = Object.entries(filterData)
-    .filter(([filterId]) => {
-      // Only include artifact-specific filters (those with artifacts. prefix)
-      // RPS is excluded - it's passed as targetRPS param, not in filterQuery
-      if (filterId === ModelCatalogNumberFilterKey.MAX_RPS) {
-        return false;
-      }
-      // Include any filter with the artifacts. prefix
-      return filterId.startsWith(ARTIFACTS_FILTER_PREFIX);
-    })
+    .filter(
+      ([filterId]) =>
+        filterId === ModelCatalogStringFilterKey.HARDWARE_TYPE ||
+        filterId === ModelCatalogStringFilterKey.USE_CASE ||
+        isLatencyFieldName(filterId),
+    )
     .map(([filterId, data]) => {
       if (typeof data === 'undefined') {
         return '';
       }
 
-      // Filter IDs match backend filter_options keys directly, no conversion needed
+      const serverFilterKey = findServerFilterKey(filterId, options.filters);
+      const queryKey = serverFilterKey ? convertToFilterQueryKey(serverFilterKey) : filterId;
+
       const filterOption =
-        options.filters && isFilterIdInMap(filterId, options.filters)
-          ? options.filters[filterId]
+        serverFilterKey && options.filters && isFilterIdInMap(serverFilterKey, options.filters)
+          ? options.filters[serverFilterKey]
           : undefined;
 
       if (filterOption?.type === 'string' && Array.isArray(data)) {
@@ -325,16 +366,18 @@ export const filtersToArtifactsFilterQuery = (
           case 0:
             return '';
           case 1:
-            return eqFilter(filterId, data[0]);
+            return eqFilter(queryKey, data[0]);
           default:
-            return inFilter(filterId, data);
+            return inFilter(queryKey, data);
         }
       }
 
-      // Numeric filters - look up operator from namedQueries, fallback to '<'
-      if (isKnownNumericFilter(filterOption, filterId, data)) {
-        const operator = getNumericFilterOperator(options, filterId);
-        return `${filterId} ${operator} ${data}`;
+      if (
+        filterOption?.type === 'number' &&
+        isLatencyFieldName(filterId) &&
+        typeof data === 'number'
+      ) {
+        return `${queryKey}<${data}`; // e.g., ttft_p90.double_value<60
       }
 
       return '';
@@ -342,6 +385,30 @@ export const filtersToArtifactsFilterQuery = (
 
   const nonEmptyFilters = serializedFilters.filter((v) => !!v);
   return nonEmptyFilters.length === 0 ? '' : nonEmptyFilters.join(' AND ');
+};
+
+/**
+ * Returns a copy of filterData with only basic (non-performance) filters.
+ * Used when performance view is disabled to exclude performance filters from API queries.
+ */
+export const getBasicFiltersOnly = (
+  filterData: ModelCatalogFilterStates,
+): ModelCatalogFilterStates => {
+  const result: ModelCatalogFilterStates = {
+    ...filterData,
+    // Clear performance string filters
+    [ModelCatalogStringFilterKey.USE_CASE]: [],
+    [ModelCatalogStringFilterKey.HARDWARE_TYPE]: [],
+    // Clear performance number filter
+    [ModelCatalogNumberFilterKey.MAX_RPS]: undefined,
+  };
+
+  // Clear all latency fields
+  ALL_LATENCY_FIELD_NAMES.forEach((latencyKey) => {
+    result[latencyKey] = undefined;
+  });
+
+  return result;
 };
 
 export const getUniqueSourceLabels = (catalogSources: CatalogSourceList | null): string[] => {
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/support/constants/index.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/support/constants/index.ts
@@ -1,0 +1,1 @@
+export * from './modelCatalog';
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/support/constants/modelCatalog.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/support/constants/modelCatalog.ts
@@ -1,0 +1,60 @@
+/**
+ * Shared constants for Model Catalog Cypress tests.
+ * These test IDs and constants are used across multiple test files.
+ */
+
+/**
+ * Test IDs for performance filter components
+ */
+export const PERFORMANCE_FILTER_TEST_IDS = {
+  workloadType: 'workload-type-filter',
+  latency: 'latency-filter',
+  latencyContent: 'latency-filter-content',
+  latencyMetricSelect: 'latency-metric-select',
+  latencyPercentileSelect: 'latency-percentile-select',
+  latencyApply: 'latency-apply-filter',
+  latencyReset: 'latency-reset-filter',
+  maxRps: 'max-rps-filter',
+  maxRpsApply: 'max-rps-apply-filter',
+  hardwareTable: 'hardware-configuration-table',
+  clearAllFilters: 'clear-all-filters-button',
+} as const;
+
+/**
+ * Test IDs for model catalog card components
+ */
+export const MODEL_CARD_TEST_IDS = {
+  card: 'model-catalog-card',
+  detailLink: 'model-catalog-detail-link',
+  description: 'model-catalog-card-description',
+  benchmarkLink: 'validated-model-benchmark-link',
+  benchmarkNext: 'validated-model-benchmark-next',
+  benchmarkPrev: 'validated-model-benchmark-prev',
+  hardware: 'validated-model-hardware',
+  replicas: 'validated-model-replicas',
+  latency: 'validated-model-latency',
+} as const;
+
+/**
+ * Test IDs for model catalog details page components
+ */
+export const MODEL_DETAILS_TEST_IDS = {
+  tabs: 'model-details-page-tabs',
+  overviewTab: 'model-overview-tab',
+  performanceInsightsTab: 'performance-insights-tab',
+  overviewTabContent: 'model-overview-tab-content',
+  performanceInsightsTabContent: 'performance-insights-tab-content',
+  longDescription: 'model-long-description',
+} as const;
+
+/**
+ * Test IDs for alerts
+ */
+export const ALERT_TEST_IDS = {
+  performanceFiltersUpdated: 'performance-filters-updated-alert',
+} as const;
+
+/**
+ * Non-breaking space character used in table column headers
+ */
+export const NBSP = '\u00A0';
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogCard.cy.ts
@@ -3,6 +3,7 @@ import {
   setupModelCatalogIntercepts,
   type ModelCatalogInterceptOptions,
 } from '~/__tests__/cypress/cypress/support/interceptHelpers/modelCatalog';
+import { PERFORMANCE_FILTER_TEST_IDS } from '~/__tests__/cypress/cypress/support/constants';
 
 /**
  * Initialize intercepts for model catalog card tests.
@@ -167,6 +168,44 @@ describe('ModelCatalogCard Component', () => {
           modelCatalog.findValidatedModelBenchmarkPrev().click();
           modelCatalog.findValidatedModelHardware().should('contain.text', '33xRTX 4090');
         });
+      });
+    });
+  });
+
+  /**
+   * NOTE: Detailed latency filter interactions, workload type filter options,
+   * and filter reset behavior are comprehensively tested in modelCatalogTabs.cy.ts
+   * (on the Performance Insights tab). These tests focus on card-specific behavior.
+   */
+  describe('Performance Filters on Catalog Landing Page', () => {
+    beforeEach(() => {
+      initIntercepts({ useValidatedModel: true });
+      modelCatalog.visit({ enableTempDevCatalogAdvancedFiltersFeature: true });
+      cy.wait('@getCatalogSourceModelArtifacts');
+      modelCatalog.togglePerformanceView();
+      modelCatalog.findLoadingState().should('not.exist');
+    });
+
+    it('should display performance filters on catalog page when toggle is ON', () => {
+      cy.findByTestId(PERFORMANCE_FILTER_TEST_IDS.workloadType).should('be.visible');
+      cy.findByTestId(PERFORMANCE_FILTER_TEST_IDS.latency).should('be.visible');
+      cy.findByTestId(PERFORMANCE_FILTER_TEST_IDS.maxRps).should('be.visible');
+    });
+
+    it('should update card latency display when latency filter changes', () => {
+      // Card should show latency value
+      modelCatalog.findFirstModelCatalogCard().within(() => {
+        modelCatalog.findValidatedModelLatency().should('be.visible');
+      });
+
+      // Change latency filter
+      modelCatalog.openLatencyFilter();
+      modelCatalog.selectLatencyMetric('E2E');
+      modelCatalog.clickApplyFilter();
+
+      // Card should still show latency value (updated based on filter)
+      modelCatalog.findFirstModelCatalogCard().within(() => {
+        modelCatalog.findValidatedModelLatency().should('be.visible');
       });
     });
   });
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
@@ -1,89 +1,11 @@
-import {
-  mockCatalogModel,
-  mockCatalogModelArtifactList,
-  mockCatalogModelList,
-  mockCatalogSource,
-  mockCatalogSourceList,
-} from '~/__mocks__';
 import { modelCatalog } from '~/__tests__/cypress/cypress/pages/modelCatalog';
 import { mockModelRegistry } from '~/__mocks__/mockModelRegistry';
-import type { CatalogSource } from '~/app/modelCatalogTypes';
-import { MODEL_CATALOG_API_VERSION } from '~/__tests__/cypress/cypress/support/commands/api';
-import { mockCatalogFilterOptionsList } from '~/__mocks__/mockCatalogFilterOptionsList';
-
-type HandlersProps = {
-  sources?: CatalogSource[];
-  modelsPerCategory?: number;
-};
-
-const initIntercepts = ({
-  sources = [mockCatalogSource({}), mockCatalogSource({ id: 'source-2', name: 'source 2' })],
-  modelsPerCategory = 4,
-}: HandlersProps) => {
-  cy.interceptApi(
-    `GET /api/:apiVersion/model_catalog/sources`,
-    {
-      path: { apiVersion: MODEL_CATALOG_API_VERSION },
-    },
-    mockCatalogSourceList({
-      items: sources,
-    }),
-  );
-
-  sources.forEach((source) => {
-    source.labels.forEach((label) => {
-      cy.interceptApi(
-        `GET /api/:apiVersion/model_catalog/models`,
-        {
-          path: { apiVersion: MODEL_CATALOG_API_VERSION },
-          query: { sourceLabel: label },
-        },
-        mockCatalogModelList({
-          items: Array.from({ length: modelsPerCategory }, (_, i) =>
-            mockCatalogModel({
-              name: `${label.toLowerCase()}-model-${i + 1}`,
-              // eslint-disable-next-line camelcase
-              source_id: source.id,
-            }),
-          ),
-        }),
-      );
-    });
-  });
-
-  cy.interceptApi(
-    `GET /api/:apiVersion/model_catalog/sources/:sourceId/models/:modelName`,
-    {
-      path: {
-        apiVersion: MODEL_CATALOG_API_VERSION,
-        sourceId: 'source-2',
-        modelName: 'sample%20category%201-model-1',
-      },
-    },
-    mockCatalogModel({}),
-  );
-
-  cy.interceptApi(
-    `GET /api/:apiVersion/model_catalog/sources/:sourceId/artifacts/:modelName`,
-    {
-      path: {
-        apiVersion: MODEL_CATALOG_API_VERSION,
-        sourceId: 'source-2',
-        modelName: 'sample%20category%201-model-1',
-      },
-    },
-    mockCatalogModelArtifactList({}),
-  );
-
-  cy.interceptApi(
-    `GET /api/:apiVersion/model_catalog/models/filter_options`,
-    {
-      path: { apiVersion: MODEL_CATALOG_API_VERSION },
-      query: { namespace: 'kubeflow' },
-    },
-    mockCatalogFilterOptionsList(),
-  );
-};
+import {
+  setupModelCatalogIntercepts,
+  setupValidatedModelIntercepts,
+  interceptArtifactsList,
+  interceptPerformanceArtifactsList,
+} from '~/__tests__/cypress/cypress/support/interceptHelpers/modelCatalog';
 
 describe('Model Catalog Details Page', () => {
   beforeEach(() => {
@@ -92,7 +14,7 @@ describe('Model Catalog Details Page', () => {
       mockModelRegistry({ name: 'modelregistry-sample' }),
     ]).as('getModelRegistries');
 
-    initIntercepts({});
+    setupModelCatalogIntercepts({});
     modelCatalog.visit();
   });
 
@@ -102,5 +24,70 @@ describe('Model Catalog Details Page', () => {
     modelCatalog.findBreadcrumb().should('exist');
     modelCatalog.findDetailsProviderText().should('be.visible');
     modelCatalog.findDetailsDescription().should('exist');
+  });
+});
+
+/**
+ * NOTE: Performance Insights Tab Navigation, Hardware Configuration Table,
+ * Workload Type Filter, and Latency Filter tests are covered in modelCatalogTabs.cy.ts.
+ * This file focuses on filter state management across pages.
+ */
+
+describe('Model Catalog Details Page - Filter State Management', () => {
+  beforeEach(() => {
+    cy.intercept('GET', '/model-registry/api/v1/model_registry*', [
+      mockModelRegistry({ name: 'modelregistry-sample' }),
+    ]).as('getModelRegistries');
+
+    // Use validated model intercepts which include performance artifacts
+    setupValidatedModelIntercepts({});
+    interceptArtifactsList();
+    interceptPerformanceArtifactsList();
+  });
+
+  it('should persist filter state when navigating between Overview and Performance Insights tabs', () => {
+    modelCatalog.visit({ enableTempDevCatalogAdvancedFiltersFeature: true });
+    modelCatalog.findLoadingState().should('not.exist');
+    modelCatalog.togglePerformanceView();
+    modelCatalog.findLoadingState().should('not.exist');
+
+    modelCatalog.findModelCatalogDetailLink().first().click();
+    modelCatalog.clickPerformanceInsightsTab();
+
+    // Change a filter
+    modelCatalog.findWorkloadTypeFilter().click();
+    modelCatalog.selectWorkloadType('code_fixing');
+
+    // Navigate to Overview tab
+    modelCatalog.clickOverviewTab();
+    modelCatalog.findOverviewTabContent().should('be.visible');
+
+    // Navigate back to Performance Insights
+    modelCatalog.clickPerformanceInsightsTab();
+
+    // Filter should still show the selected value (capitalized as "Code Fixing")
+    modelCatalog.findWorkloadTypeFilter().should('contain.text', 'Code Fixing');
+  });
+
+  it('should sync filter changes back to catalog page', () => {
+    modelCatalog.visit({ enableTempDevCatalogAdvancedFiltersFeature: true });
+    modelCatalog.findLoadingState().should('not.exist');
+    modelCatalog.togglePerformanceView();
+    modelCatalog.findLoadingState().should('not.exist');
+
+    modelCatalog.findModelCatalogDetailLink().first().click();
+    modelCatalog.clickPerformanceInsightsTab();
+
+    // Change a filter on details page
+    modelCatalog.findWorkloadTypeFilter().click();
+    modelCatalog.selectWorkloadType('rag');
+
+    // Go back to catalog page
+    cy.go('back');
+    cy.go('back');
+    modelCatalog.findLoadingState().should('not.exist');
+
+    // The alert should show indicating filters were updated
+    modelCatalog.findPerformanceFiltersUpdatedAlert().should('be.visible');
   });
 });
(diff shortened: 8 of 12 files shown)


clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -110,12 +130,170 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
-  const setPerformanceViewEnabled = React.useCallback((enabled: boolean) => {
-    setBasePerformanceViewEnabled(enabled);
-    if (!enabled) {
-      setPerformanceFiltersChangedOnDetailsPage(false);
+  /**
+   * Applies filter values from a named query to the filter state.
+   * Maps backend field names (e.g., 'artifacts.use_case.string_value') to frontend filter keys.
+   * Handles special values like 'max' by resolving them from filter ranges.
+   */
+  const applyNamedQueryDefaults = React.useCallback(
```
mturley, line 137: This function duplicates a lot of logic in your `getDefaultFiltersFromNamedQuery` function in performanceFilterUtils.ts. Can we deduplicate it?

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+  Object.entries(namedQuery).forEach(([fieldName, fieldFilter]) => {
+    // Handle artifacts.* prefix - check both with and without prefix
+    const isUseCase =
+      fieldName === 'artifacts.use_case.string_value' || fieldName === 'use_case.string_value';
+    const isHardwareType =
+      fieldName === 'artifacts.hardware_type.string_value' ||
+      fieldName === 'hardware_type.string_value';
+    const isRps =
+      fieldName === 'artifacts.requests_per_second.double_value' ||
+      fieldName === 'requests_per_second.double_value';
```
mturley: We probably want to avoid hard-coded strings for these. Also do we really need to check both with and without the prefix? Looking at the filter_options response on ui-green-3 I'm only seeing these with the artifacts.* prefix.

Also, I'm noticing a direction we're moving in that we might want to discuss and possibly reconsider. The filter keys we have in ModelCatalogStringFilterKey / ModelCatalogNumberFilterKey (which are the keys for our ModelCatalogFilterStates object) were originally supposed to be exactly how they appear in filter_options, so we could just use them 1:1 when constructing the filterQuery string for the API. Now we have all this logic here and in other utils for converting between frontend and backend filter keys. That feels confusing and fragile to me... I wonder if it's avoidable? maybe not, but it could be worth looking at.

That is complicated by the fact that we now have the artifacts.* prefix and the string_value / double_value suffixes for sure, but I imagined we would change the keys we use on the frontend to match fully, so `ModelCatalogStringFilterKey.USE_CASE` would be the full string `artifacts.use_case.string_value` for example. Note that the suffix only matters for filters based on custom properties, we'd still have `license` and `language` be that plain key for example - but the UI logic ideally wouldn't care about the difference.

I wonder if changing our enums to use the full keys as they exist in filter_options we can avoid a lot of the special logic.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+
+    // Check if it's a latency field by extracting the field name from artifacts.*.double_value
+    const latencyMatch = fieldName.match(/^artifacts\.([a-z0-9_]+)\.double_value$/);
```
mturley: Yeah, for example if our latency fields are fully e.g. `artifacts.ttft_p90.double_value` we don't need a special regex here, we can just pass that full string into `isLatencyMetricField` and have that work. Note that I think we can still leave alone the LatencyMetric and LatencyPercentile enums, it's still useful to have those be e.g. `ttft` and `p90`, but we could change the `LatencyMetricFieldName` type to:
```ts
export type LatencyMetricFieldName = `artifacts.${Lowercase<LatencyMetric>}_${Lowercase<LatencyPercentile>}.double_value`;
```
and then we'd have to change the `getLatencyFieldName` and `parseLatencyFieldName` to match that.

Does this approach make sense? it may feel weird to use these full keys in the UI state but I think having it match the language of the API simplifies things.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -8,7 +8,7 @@ export enum ModelCatalogStringFilterKey {
 }
 
 export enum ModelCatalogNumberFilterKey {
-  MIN_RPS = 'rps_mean',
+  MAX_RPS = 'rps_mean',
```
mturley: RPS is a weird special case here. As you've done, we do need to pull its default from `artifacts.requests_per_second.double_value` in the named query, so maybe we should change this value to be `artifacts.requests_per_second.double_value` (I'm not sure where rps_mean came from to begin with). But we still don't want it included in our filterQuery because it's passed as the special `targetRPS` property when we fetch the perf artifacts. That goes for both `filtersToFilterQuery` and `filtersToArtifactsFilterQuery` I think, we never want to include RPS in a filterQuery for either request, only as targetRPS in the artifacts request. Right now I think you only filter it out in `filtersToArtifactsFilterQuery`.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -217,45 +206,76 @@ const eqFilter = (k: string, v: string) => `${k}=${wrapInQuotes(v)}`;
...
 /**
  * Converts filter data into a filter query string for the /models endpoint.
- * Supports string filters (equality/IN), and numeric filters (greater than for RPS, less than for latency).
+ * Supports string filters (equality/IN) and numeric filters (less-than for latency/RPS).
+ * Performance filters are mapped to their artifacts.* prefixed backend keys.
  */
 export const filtersToFilterQuery = (
   filterData: ModelCatalogFilterStates,
   options: CatalogFilterOptionsList,
 ): string => {
```
mturley, line 237: There is some duplication between filtersToFilterQuery and filtersToArtifactsFilterQuery. I wonder if these can be combined into one function that just has a param for whether it is filtering models or artifacts. It should be using the same logic for constructing the filter string and the only difference is which filters are included.

This gets easier if we are making my other suggested change to include the `artifacts.*` prefix as part of the UI state's filter keys for those filters.
* When we're getting the filterQuery for filtering models, we want to include everything (except the RPS filter).
* When we're getting the filterQuery for filtering artifacts, we want to exclude everything that doesn't have the `artifacts.*` prefix.
  * I believe in this case we also want to then strip that prefix off the key when we put it into the filterQuery string. That's because `filter_options` refers to everything from the perspective of filtering models, and the artifacts.* prefix tells the API "filter models by this property on their associated artifacts". Stripping it when we filter artifacts just means "filter by this property". I think it makes sense for the artifacts.* prefix to be present everywhere the key is used in the UI except right when it is placed in the filterQuery for filtering artifacts.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -110,12 +130,170 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+  const setPerformanceViewEnabled = React.useCallback(
+    (enabled: boolean) => {
+      setBasePerformanceViewEnabled(enabled);
+      if (enabled) {
+        // Apply default performance filters from namedQueries if available
+        const defaultQuery = filterOptions?.namedQueries?.['default-performance-filters'];
+        if (defaultQuery) {
+          applyNamedQueryDefaults(defaultQuery);
+        }
+      } else {
```
mturley, line 195: There's a possible race condition here - if the user turns on the toggle before the filterOptions are loaded, the default filters won't get applied. Let's disable the toggle while filterOptions are still loading. (the Switch in ModelPerformanceViewToggleCard). We should be able to pull the loading status of the filterOptions out of ModelCatalogContext there.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -110,12 +130,170 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+    // Clear all performance filters
+    baseSetFilterData(ModelCatalogStringFilterKey.USE_CASE, []);
+    baseSetFilterData(ModelCatalogStringFilterKey.HARDWARE_TYPE, []);
+    baseSetFilterData(ModelCatalogNumberFilterKey.MAX_RPS, undefined);
+    ALL_LATENCY_FIELD_NAMES.forEach((fieldName) => {
+      baseSetFilterData(fieldName, undefined);
+    });
+
+    // Then apply performance defaults from namedQueries if available
+    const defaultQuery = filterOptions?.namedQueries?.['default-performance-filters'];
```
mturley: We have this `"default-performance-filters"` query name as a hard-coded string in a few places, we should maybe lift it out as a constant.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+    backendFieldName = 'artifacts.use_case.string_value';
+  } else if (filterKey === ModelCatalogStringFilterKey.HARDWARE_TYPE) {
+    backendFieldName = 'artifacts.hardware_type.string_value';
+  } else if (filterKey === ModelCatalogNumberFilterKey.MAX_RPS) {
+    backendFieldName = 'artifacts.requests_per_second.double_value';
+  } else if (isLatencyField) {
+    backendFieldName = `artifacts.${filterKeyStr}.double_value`;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
mturley: I don't think disabling this is necessary here. However it won't matter if we're getting rid of the mapping between frontend and backend field names.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxRpsFilter.tsx
```diff
@@ -77,15 +91,17 @@ const MinRpsFilter: React.FC<MinRpsFilterProps> = ({ performanceArtifacts }) =>
...
 
   const toggle = (toggleRef: React.Ref<MenuToggleElement>) => (
     <MenuToggle
       ref={toggleRef}
+      data-testid="max-rps-filter"
       onClick={() => setIsOpen(!isOpen)}
       isExpanded={isOpen}
-      style={{ minWidth: '200px', width: 'fit-content' }}
+      isFullHeight
+      style={{ minWidth: '200px', width: 'fit-content', height: '56px' }}
```
mturley, line 101: Why the hard-coded height? Is that avoidable?

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogPerformanceFiltersAlert.cy.ts
```diff
@@ -36,79 +39,108 @@ const initIntercepts = ({
...
+          : undefined;
+      const name = i === 0 ? 'validated-model' : `all-models-model-${i + 1}`;
+      return mockCatalogModel({
+        name,
+        // eslint-disable-next-line camelcase
+        source_id: 'sample-source',
+        customProperties,
+      });
+    }),
+  });
```
mturley: There is a lot of duplicated logic between `initIntercepts` functions across test files. It's probably fine for this PR, but we may want to log something to deduplicate this stuff later. Claude's summary of the duplicated code:

Image

Image

Image

clients/ui/frontend/src/__mocks__/mockCatalogFilterOptionsList.ts
```diff
@@ -11,6 +11,22 @@ import {
...
+      value: 'max', // 'max' means use the max value from the range in filters
+    },
+    'artifacts.requests_per_second.double_value': {
+      operator: FilterOperator.LESS_THAN_OR_EQUAL,
+      value: 'max', // 'max' means use the max value from the range in filters (300 in mock)
+    },
+  },
   high_performance_gpu: {
     'hardware_type.string_value': { operator: FilterOperator.IN, value: ['H100-80', 'A100-80'] },
     'requests_per_second.double_value': {
```
mturley: Perhaps out of scope for this PR review because I see these lines were added in a previous PR, but - where are these additional namedQueries coming from? The only one I see when fetching filter_options on ui-green-3 is the `default-performance-filters`. I don't think we need the extra ones for our purposes, we can probably just remove them.

However, looking here I see in these extra ones the properties don't have the `artifacts.*` prefix on them. Is that why you had logic to check for the property keys both with and without that prefix? I think if the API is including them both ways that's a bug. If it turns out we do need to use these additional queries we should raise that with them.

clients/ui/frontend/src/__mocks__/mockCatalogFilterOptionsList.ts
```diff
@@ -89,7 +105,7 @@ export const mockCatalogFilterOptionsList = (
         UseCaseOptionValue.RAG,
       ],
     },
-    [ModelCatalogNumberFilterKey.MIN_RPS]: {
+    [ModelCatalogNumberFilterKey.MAX_RPS]: {
```
mturley, line 115: We need to look at the actual return value of the filter_options request from the API and update our CatalogFilterOptions type and mock data to match here - I believe just like in the namedQueries, all the performance filters now have the artifacts.* prefix and .string_value or .double_value suffix. This is what I'm seeing from the API:

Image

Actually, if we are changing our filter keys for state to include the prefixes and suffixes I think this type will reflect that correctly, since it's derived from those. Every filter has one true key used everywhere, it's just that some of them are changing to have a prefix and suffix now.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -110,12 +130,170 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+    baseSetFilterData(ModelCatalogStringFilterKey.LICENSE, []);
+    baseSetFilterData(ModelCatalogStringFilterKey.LANGUAGE, []);
+
+    // Clear all performance filters
+    baseSetFilterData(ModelCatalogStringFilterKey.USE_CASE, []);
+    baseSetFilterData(ModelCatalogStringFilterKey.HARDWARE_TYPE, []);
+    baseSetFilterData(ModelCatalogNumberFilterKey.MAX_RPS, undefined);
+    ALL_LATENCY_FIELD_NAMES.forEach((fieldName) => {
+      baseSetFilterData(fieldName, undefined);
+    });
```
mturley: If we add more filters later we'll need to make sure they're added here as well. Maybe that's fine for now since this PR is getting large, but ideally we could have this loop over all string filter keys, all number filter keys and all latency keys and reset them all. I guess that only works when we are trying to reset them all, for the case when the toggle is being disabled we'd need some way to filter that to only reset the performance filters. Something to think about later perhaps.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -110,12 +130,170 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+        } else {
+          baseSetFilterData(ModelCatalogNumberFilterKey.MAX_RPS, undefined);
+        }
+      } else if (isLatencyMetricFieldName(filterKeyStr)) {
+        if (hasDefault && typeof value === 'number') {
+          baseSetFilterData(filterKeyStr, value);
+        } else {
+          baseSetFilterData(filterKeyStr, undefined);
+        }
+      }
```
mturley: It seems like this logic is very similar to logic in `getDefaultFiltersFromNamedQuery`, is there a way we can avoid duplicating special cases for specific filter keys and centralize the logic?

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+        : typeof fieldFilter.value === 'string'
+          ? [fieldFilter.value]
+          : [];
+    return { hasDefault: true, value: values };
+  }
+
+  if (filterKey === ModelCatalogNumberFilterKey.MAX_RPS || isLatencyMetricFieldName(filterKeyStr)) {
+    const resolvedValue = resolveFilterValue(filterOptions, backendFieldName, fieldFilter.value);
+    return { hasDefault: true, value: resolvedValue };
+  }
```
mturley: It seems like this duplicates logic too. Can we not just use `getDefaultFiltersFromNamedQuery` and pull the single default value out of it that we need? It would be great to avoid having several places in the code where we have special cases for `if (filterKey === ...)`.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -30,13 +31,20 @@ const ModelCatalog: React.FC = () => {
...
 
+    // Clear all number filters
     Object.values(ModelCatalogNumberFilterKey).forEach((filterKey) => {
       setFilterData(filterKey, undefined);
     });
+
+    // Clear all latency filters
+    ALL_LATENCY_FIELD_NAMES.forEach((filterKey) => {
+      setFilterData(filterKey, undefined);
+    });
```
mturley: Hmm, yeah here's another place where we have logic to try to clear all filters. Maybe we can centralize what we have here and what you added in `resetPerformanceFiltersToDefaults`, at least for the part where we reset all the filters.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -138,13 +140,25 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
                   </ToolbarItem>
                 </ToolbarGroup>
               </ToolbarToggleGroup>
-              {onResetAllFilters && hasActiveFilters && (
+              {onResetAllFilters && hasActiveFilters && !performanceViewEnabled && (
```
mturley: Isn't this hiding the chip bar entirely when the performance view is disabled? I think we want the chip bar to appear regardless, but we can change what is included in `filtersToShow` based on whether the toggle is enabled.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -217,45 +206,76 @@ const eqFilter = (k: string, v: string) => `${k}=${wrapInQuotes(v)}`;
...
+      return 'artifacts.hardware_type.string_value';
+    case ModelCatalogStringFilterKey.USE_CASE:
+      return 'artifacts.use_case.string_value';
+    case ModelCatalogNumberFilterKey.MAX_RPS:
+      return 'artifacts.requests_per_second.double_value';
+    default:
+      // Non-performance filters use the key as-is
+      return frontendKey;
+  }
+};
```
mturley: Yeah, ideally this is totally unnecessary. Let's use the backend keys on the frontend.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -75,8 +83,8 @@ const HardwareConfigurationFilterToolbar: React.FC<HardwareConfigurationFilterTo
   }
 
   const handleClearAllFilters = () => {
-    // Only clear performance-related filters, not the basic catalog filters
-    clearAllFilters(setFilterData, filtersToShow);
+    // Reset performance filters to default values from namedQueries
+    resetPerformanceFiltersToDefaults();
```
mturley: So this is tricky. HardwareConfigurationFilterToolbar is used on both the landing page and details page now, but when you're on the details page and you click "reset all filters" we don't want to accidentally also reset the basic filters (non-performance-related filters that are only present on the landing page). Clicking reset should only clear the filters you see on the page you're looking at. That's why we had `filtersToShow` passed to `clearAllFilters` before. We'll need this component to be aware of which page it's on and properly either reset everything or just reset performance filters.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -15,21 +21,90 @@ import {
...
+  ModelCatalogStringFilterKey.HARDWARE_TYPE,
+  ModelCatalogNumberFilterKey.MAX_RPS,
+];
+
+/**
+ * Check if a filter key is a performance filter (should reset to default instead of clear)
+ */
+const isPerformanceFilter = (filterKey: ModelCatalogFilterKey): boolean =>
+  PERFORMANCE_FILTER_KEYS.includes(filterKey) ||
+  ALL_LATENCY_FIELD_NAMES.some((name) => name === filterKey);
```
mturley: These PERFORMANCE_FILTER_KEYS are also duplicated in the HardwareConfigurationFilterToolbar component. We should maybe lift them and the `BASIC_FILTER_KEYS` that are also there into a constants file to be used in both places.

This `isPerformanceFilter` util also might be useful to lift out to a utils file - in my earlier comment about changing the "reset filters" logic to map over all the filter keys instead of hard-coding a reset for each filter, you could use it to make sure you're only resetting the ones you want to in the cases where you need to reset only performance filters.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -120,11 +212,44 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
+                  >
+                    {labelText}
+                  </Label>
+                );
+              })}
+            </LabelGroup>
+          );
+        }
+
+        // For basic filters, use standard ToolbarFilter
```
mturley: I'm not sure this method of using our own LabelGroup for certain filters will work. I'm noticing strange rendering behavior here when applying both basic filters and perf filters:

* The custom LabelGroups you have for performance filters appear to the right of the filters and wrap onto the next line, and are on a line separate from the basic filters.
* Sometimes, I get an extra "Clear all filters" link at the top that doesn't do anything... i'm actually not sure how that happened, I got it in this screenshot but I can't figure out how to trigger it again... if we can't repro that we could ignore it and see if it comes up again as a future bug
* I don't think you actually added the custom undo icon here, the custom ones are still just appearing with X icons.

Image

I think we need to use the standard ToolbarFilter for all of these (just exclude perf filters from it if they are set to defaults or the toggle is off). That leaves the matter of changing the icon for those... if we can't, honestly I think that's ok and we could just address it as a future bug. It could at least be out of scope of this PR and opened as another bug in the epic that we maybe punt out.

But in order to change the icon of an X button in the standard ToolbarFilter:
* The proper way would be to add support in PatternFly for a custom close icon to be passed in as a new prop. [I started a thread about that here](https://redhat-internal.slack.com/archives/C04JMHKSD9C/p1765979389666809), and it's something we should consider following up on when there's time. But that's not going to happen before we need to release this.
* A while ago I messed around and came up with a very hacky but workable temporary solution. You're not going to like this, but it might be worth a try. You'd need to adapt this to change the `filterHasDefaultByName` function so that, given a filter name as rendered in the UI, it returns whether that filter has a default present in the namedQuery. when I was testing it I just had it change the icon on the Task filter. The hard-coded SVG here is copied from the PF RedoIcon. https://github.com/kubeflow/model-registry/pull/2076/changes

Either way, let's push this out to another PR. For this PR let's just use the standard ToolbarFilter for everything, we can deal with the incorrect icon after.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -96,7 +98,7 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
       <StackItem>
         <Toolbar
           key={`toolbar-${hasActiveFilters}`}
-          {...(onResetAllFilters
+          {...(onResetAllFilters && !performanceViewEnabled
```
mturley: If we're using the standard ToolbarFilter I think you'd want to change this back. The only difference here when the toggle is turned on is the filters included in the `filtersToShow` prop of ModelCatalogActiveFilters.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -81,11 +81,11 @@ export const filterHardwareConfigurationArtifacts = (
       }
     }
 
-    // Min RPS Filter
-    const minRpsFilter = filterState[ModelCatalogNumberFilterKey.MIN_RPS];
-    if (minRpsFilter !== undefined) {
+    // Max RPS Filter - filter out artifacts that exceed the max RPS requirement
+    const maxRpsFilter = filterState[ModelCatalogNumberFilterKey.MAX_RPS];
+    if (maxRpsFilter !== undefined) {
       const rpsPerReplica = getDoubleValue(artifact.customProperties, 'requests_per_second');
```
mturley: We should no longer need this function at all. All filtering of artifacts happens on the server now. It looks like it's already no longer used, can we remove it and its tests?

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -217,45 +206,76 @@ const eqFilter = (k: string, v: string) => `${k}=${wrapInQuotes(v)}`;
...
 
-    // Numeric filters: less-than for latency, greater-than for RPS
+    // Handle numeric filters (less-than for latency and RPS)
     if (isKnownLessThanValue(filterOption, filterId, data)) {
-      return `${filterId} < ${data}`;
-    }
-
-    if (isKnownGreaterThanValue(filterOption, filterId, data)) {
-      return `${filterId} > ${data}`;
+      return `${backendKey} < ${data}`;
```
mturley: Can we maybe look up the operator that is supplied in the defaultQuery, and fall back to `<` if this filter isn't there? maybe that's too much.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+    return 'artifacts.hardware_type.string_value';
+  }
+  if (filterKey === ModelCatalogNumberFilterKey.MAX_RPS) {
+    return 'artifacts.requests_per_second.double_value';
+  }
+  if (isLatencyMetricFieldName(filterKeyStr)) {
+    return `artifacts.${filterKeyStr}.double_value`;
+  }
+  return null;
+};
```
mturley: Isn't this a duplicate of `getBackendFilterKey`? we shouldn't need either anymore if we make the changes I'm suggesting.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+  // Check if it's an artifacts.* prefixed latency field
+  const artifactMatch = ALL_LATENCY_FIELD_NAMES.find(
+    (name) => fieldName === `artifacts.${name}.double_value`,
+  );
+  if (artifactMatch) {
+    return artifactMatch;
+  }
+  return null;
+};
```
mturley, line 59: We shouldn't need this, if we're changing the keys in our state/types to have the `artifacts.${name}.double_value` around them, we should be able to just use the existing `parseLatencyFieldName`.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Gets the default value for a performance filter from namedQueries.
+ * Used to determine if a filter chip should be shown (only when value differs from default).
+ */
+export const getPerformanceFilterDefaultValue = (
+  filterOptions: CatalogFilterOptionsList | null,
+  filterKey: keyof ModelCatalogFilterStates,
+): string | number | string[] | undefined => {
+  const defaultQuery = filterOptions?.namedQueries?.['default-performance-filters'];
```
mturley: This seems redundant with `getSingleFilterDefault`, which in turn has redundant logic to `getDefaultFiltersFromNamedQuery`. It seems like we should be able to handle all the "get all the default filter values (and maybe use one of them)" from a single function.

clients/ui/frontend/src/app/pages/modelCatalog/utils/validatedModelUtils.ts
```diff
@@ -20,6 +32,24 @@ export type PerformanceMetrics = {
...
+): LatencyMetricsMap => {
+  const result: LatencyMetricsMap = {};
+  ALL_LATENCY_FIELD_NAMES.forEach((fieldName) => {
+    const value = customProperties?.[fieldName]?.double_value;
+    if (value !== undefined) {
+      result[fieldName] = value;
+    }
+  });
+  return result;
 };
```
mturley, line 60: This will look a bit different if we make the changes I'm suggesting and `ALL_LATENCY_FIELD_NAMES` contains strings that have `artifacts.*.double_value` in them. Rather than trying to parse those out, maybe instead just loop over the LatencyMetric and LatencyPercentile enums (nest one loop inside the other), and use each combination of the two to get the `fieldName` you use to look up things in `customProperties`.

Edit: see my next comment - if you then want to use the fully qualified `artifacts.*.double_value` key as the key when you are assigning `result[fieldName]`, you should be able to use `getLatencyFieldName` and pass it the metric and percentile to get that.

clients/ui/frontend/src/app/pages/modelCatalog/utils/validatedModelUtils.ts
```diff
@@ -34,8 +64,35 @@ export const extractPerformanceMetrics = (
...
+export const getLatencyValue = (
+  latencyMetrics: ValidatedModelMetrics['latencyMetrics'],
+  fieldName: LatencyMetricFieldName | undefined,
+): number | undefined => {
+  if (!fieldName) {
+    // Default to ttft_mean if no field specified
+    return latencyMetrics.ttft_mean;
+  }
+  return latencyMetrics[fieldName];
+};
```
mturley: Hmm, maybe this one would get complicated by having `artifacts.*.double_value` embedded in LatencyMetricFieldName. Maybe we want to fully embrace those and have the `latencyMetrics` object you create with `extractLatencyMetrics` also use that full key as its keys. Literally everywhere in the UI where we combine latency metric+percentile into a single key it would have `artifacts.*.double_value` around it. If you think that's too far, we could revisit it but then we need to have some notion of the "latency key" and "latency filter key" which is maybe frustrating.

clients/ui/frontend/src/app/pages/modelCatalog/utils/validatedModelUtils.ts
```diff
@@ -34,8 +64,35 @@ export const extractPerformanceMetrics = (
...
+ */
+export const getLatencyLabel = (fieldName: LatencyMetricFieldName | undefined): string => {
+  if (!fieldName) {
+    return LatencyMetric.TTFT;
+  }
+  const [prefix] = fieldName.split('_');
+  const metric = Object.values(LatencyMetric).find((m) => m.toLowerCase() === prefix);
+  return metric || LatencyMetric.TTFT;
+};
+
```
mturley: I don't think this is necessary, right? If you already have the fieldName you should be able to just use `parseLatencyFieldName(fieldName)?.metric || LatencyMetric.TTFT`.

clients/ui/frontend/src/app/modelCatalogTypes.ts
```diff
@@ -241,6 +241,9 @@ export type CatalogFilterOptions = ModelCatalogStringFilterOptions & {
 } & {
   // Allow additional latency metric field names
   [key in LatencyMetricFieldName]?: CatalogFilterNumberOption;
+} & {
+  // Allow additional artifact filter keys from backend (e.g., artifacts.hardware_type.string_value)
+  [key: string]: CatalogFilterStringOption<string> | CatalogFilterNumberOption | undefined;
```
mturley: I'm not sure we want this. If CatalogFilterOptions has `[key: string]` in it, any key is valid and we lose our type safety of making sure we're accessing filter options that really exist. The example you give (`artifacts.hardware_type.string_value`) should just be one of our `ModelCatalogStringFilterOptions`, which should happen automatically if we're adding `artifacts.` and `.string_value` to it in `ModelCatalogStringFilterKey`.

mturley (changes requested):
All right, this is a big one. I'm happy to meet if you want to pair on it or have any questions.

Note: I posted these comments while jumping around the diff a lot, so you may want to read them in the order they appear in the Conversation tab instead of just scrolling through the diff.

In addition to the comments inline:
* I'm testing in mock mode and when I turn on the performance view toggle, I'm not seeing the filter defaults applied. They still just don't have values until I select one. Similarly, on the details page I'm not seeing the filters populated with default values. We shouldn't be able to ever see the performance filter controls with no filter value anymore (except hardware type I guess). I think we need to adjust our mock data to make sure this behavior works in mock mode.
* I did a temporary sync to ODH from this branch for testing, if it would be useful for you to test with as well. There were a couple conflicts to resolve there. https://github.com/opendatahub-io/odh-dashboard/pull/5937. Testing locally in ODH:
  * I do see the defaults populated correctly when I turn on the toggle on the landing page, but the defaults don't populate on the details page. I think that's something we just missed here - regardless of the state of the toggle on the landing page, when the user reaches the details page the filters in the hardware configs table there should show their defaults. So if the user reaches that page and the performance filters in state still don't have values we should populate them with defaults - but if they already have values we should keep them.
  * On the landing page, after the defaults are populated, if no models match my filters and I click the "reset all filters" button that appears in the empty state of the main page body, it is clearing all the filter values instead of the new behavior of clearing basic filters but resetting the performance ones to defaults. We need to make sure no matter how the user resets filters it is using this new behavior.
    * It looks like this is happening because we have separate special `resetAllFilters` behavior inside the `clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx` component, which is used for the `handleFilterReset` function in both ModelCatalogSourceLabelSelectorNavigator and ModelCatalogGalleryView there.
  * When there is a latency filter value applied that came from the defaults, it has too many decimal places. We should use the `formatLatency` util when rendering the current value in LatencyFilter like we do in ModelCatalogCardBody and HardwareConfigurationTableRow.
      Image

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxRpsFilter.tsx
manaswinidas: This has been mentioned in the designs

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
```diff
@@ -109,16 +111,35 @@ const LatencyFilter: React.FC<LatencyFilterProps> = ({ performanceArtifacts }) =
...
+    if (performanceArtifacts.length > 0) {
+      return getSliderRange({
+        performanceArtifacts,
+        getArtifactFilterValue: (artifact) =>
+          getDoubleValue(artifact.customProperties, propertyKey),
+        fallbackRange: FALLBACK_LATENCY_RANGE,
+        shouldRound: true,
+      });
+    }
+    // Fall back to filterOptions from context
```
mturley, line 132: I think we can't try to get the slider range from the artifacts at all since we may not have all of them in memory when we paginate the request. We have to just get it from filterOptions always.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxRpsFilter.tsx
```diff
@@ -18,33 +18,44 @@ import {
...
         getArtifactFilterValue: (artifact) =>
-          getDoubleValue(artifact.customProperties, 'requests_per_second'),
+          getDoubleValue(artifact.customProperties, PerformancePropertyKey.REQUESTS_PER_SECOND),
         fallbackRange: FALLBACK_RPS_RANGE,
-      }),
-    [performanceArtifacts],
-  );
+      });
+    }
+    // Fall back to filterOptions from context
```
mturley, line 46: Same here. Also we have duplicate code here, we can probably factor out a function for finding the range and use it in both LatencyFilter and MaxRpsFilter.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -9,81 +9,70 @@ import {
...
 
-  const handleClearAllFilters = () => {
-    // Only clear performance-related filters, not the basic catalog filters
-    clearAllFilters(setFilterData, filtersToShow);
-  };
+  // Custom clear filters button with test ID for Cypress tests
+  const customLabelGroupContent = onResetAllFilters && hasVisibleChips && (
+    <ToolbarItem>
+      <Button variant="link" onClick={onResetAllFilters} data-testid="clear-all-filters-button">
+        Clear all filters
```
mturley, line 63: Do we need this if we're using the PF ToolbarFilter for all our chips? I thought that rendered a reset link for us?

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -15,21 +15,41 @@ import {
...
   const handleRemoveFilter = (categoryKey: string, labelKey: string) => {
     if (!isCatalogFilterKey(categoryKey)) {
       return;
     }
+
+    // For performance filters when performance view is enabled, reset to default instead of clearing
+    if (performanceViewEnabled && isPerformanceFilterKey(categoryKey)) {
+      resetSinglePerformanceFilterToDefault(categoryKey);
+      return;
+    }
```
mturley, line 51: This is duplicated below, and also it is changing behavior based on whether the toggle is enabled in a place that also affects the details page which should ignore the toggle and always reset to defaults. I think instead we need something like an `includePerformanceFilters` prop for this component and have the landing page pass it in if `performanceViewEnabled` but have the details page just pass it in as `true`.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogActiveFilters.tsx
```diff
@@ -46,6 +66,13 @@ const ModelCatalogActiveFilters: React.FC<ModelCatalogActiveFiltersProps> = ({ f
...
```
mturley, line 74: Duplicate code

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -29,20 +31,16 @@ const ModelCatalog: React.FC = () => {
...
     setSearchTerm('');
-    resetAllFilters();
-  }, [resetAllFilters]);
+    // When performance view is enabled, clear basic filters but reset
+    // performance filters to defaults instead of clearing them completely
+    if (performanceViewEnabled) {
+      clearBasicFiltersAndResetPerformanceToDefaults();
+    } else {
+      clearAllFilters();
+    }
```
mturley, line 42: Here's another place where we are spreading this conditional reset logic around. I think if we change the approach so that we are always resetting performance filters to defaults we can avoid that. The perf view toggle can just impact how we use state rather than how we reset it - if the toggle is off we exclude perf filters from filterQuery and the chip bar, which allows us to leave them alone in state. We can consider a perf filter being undefined in state to be an invalid state (only happens when initially loading filter options).

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
```diff
@@ -37,15 +38,23 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
...
+    performanceViewEnabled,
   } = React.useContext(ModelCatalogContext);
   const filtersApplied = hasFiltersApplied(filterData);
 
+  // When performance view is disabled, exclude performance filters from API queries
+  // Memoize to prevent infinite re-fetching
+  const effectiveFilterData = React.useMemo(
+    () => (performanceViewEnabled ? filterData : getBasicFiltersOnly(filterData)),
+    [performanceViewEnabled, filterData],
+  );
```
mturley, line 50: Ok yeah there you go, we're already halfway there with excluding those filters based on the toggle regardless of what's in state.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -5,21 +5,64 @@ import HardwareConfigurationTable from '~/app/pages/modelCatalog/components/Hard
...
+    filterData,
+    filterOptions,
+    filterOptionsLoaded,
+    performanceViewEnabled,
+    setPerformanceFiltersChangedOnDetailsPage,
+    setFilterData,
+  } = React.useContext(ModelCatalogContext);
+
+  // Apply default performance filters on mount if they don't have values yet
+  // Only apply defaults if performance view toggle is enabled
```
mturley, line 36: We want to apply defaults regardless of the toggle, the toggle shouldn't affect this details page.

clients/ui/frontend/src/app/pages/modelCatalog/utils/hardwareConfigurationFilterUtils.ts
```diff
@@ -4,122 +4,35 @@ import {
...
-export const parseLatencyFieldName = (
-  fieldName: LatencyMetricFieldName,
-): { metric: LatencyMetric; percentile: LatencyPercentile } | null => {
-  const [prefix, suffix] = fieldName.split('_');
-  const metric = Object.values(LatencyMetric).find((m) => m.toLowerCase() === prefix);
-  const percentile = Object.values(LatencyPercentile).find((p) => p.toLowerCase() === suffix);
-  return metric && percentile ? { metric, percentile } : null;
-};
+// Re-export parseLatencyFilterKey as parseLatencyFieldName for backward compatibility
+export const parseLatencyFieldName = parseLatencyFilterKey;
```
mturley, line 18: This seems like a bad idea.... can we make sure we call functions the same name everywhere they're used?

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -219,57 +227,73 @@ const inFilter = (k: string, values: string[]) =>
...
-  // eg. filterQuery=rps_mean > 1 AND license IN ('mit','apache-2.0') AND ttft_mean < 10
+  // eg. filterQuery=license IN ('mit','apache-2.0') AND artifacts.hardware_type.string_value='H100'
   return nonEmptyFilters.length === 0 ? '' : nonEmptyFilters.join(' AND ');
 };
 
+/**
+ * Find the server filter key for a given filter ID.
+ * Handles the case where local state uses short keys but server uses fully qualified keys.
+ */
 const findServerFilterKey = (
```
mturley, line 297: It still seems strange that we have a concept of server and local keys, I don't think this comment is exactly correct since we'er not using short keys in local state anymore right?

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -291,21 +315,27 @@ const findServerFilterKey = (
...
+/**
+ * Convert a server filter key to the format expected by the filterQuery.
+ * Strips the 'artifacts.' prefix if present.
+ */
 const convertToFilterQueryKey = (serverFilterKey: string): string => {
-  const prefix = 'artifacts.';
-  if (serverFilterKey.startsWith(prefix)) {
-    return serverFilterKey.substring(prefix.length);
+  if (serverFilterKey.startsWith(ARTIFACTS_FILTER_PREFIX)) {
+    return serverFilterKey.substring(ARTIFACTS_FILTER_PREFIX.length);
```
mturley, line 324: This is specific to artifacts fetches, because artifacts. should still be there in the /models filterQuery. so this comment is a little misleading.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -291,21 +315,27 @@ const findServerFilterKey = (
...
 const isLatencyFieldName = (id: string): id is LatencyMetricFieldName =>
   ALL_LATENCY_FIELD_NAMES.some((name) => name === id);
 
 /**
  * Converts filter data into a filter query string for the /artifacts/performance endpoint.
  * Only includes filters that have the 'artifacts.' prefix and strips that prefix in the output.
- * RPS is NOT included in filterQuery for artifacts - it's passed as targetRPS param instead.
+ * RPS is NOT included in filterQuery - it's passed as targetRPS param instead.
  */
 export const filtersToArtifactsFilterQuery = (
```
mturley, line 340: We're still duplicating a lot of logic between filtersToFilterQuery and filtersToArtifactsFilterQuery. All the logic for how the string gets constructed is the same, we should make them just one function with an additional `isArtifactsQuery` argument. If that's true, it should include only filters that start with .artifacts and strip that prefix off of them. but otherwise it can use the same code.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -357,6 +387,30 @@ export const filtersToArtifactsFilterQuery = (
...
+    [ModelCatalogNumberFilterKey.MAX_RPS]: undefined,
+  };
+
+  // Clear all latency fields
+  ALL_LATENCY_FIELD_NAMES.forEach((latencyKey) => {
+    result[latencyKey] = undefined;
+  });
+
+  return result;
+};
```
mturley, line 412: I think instead of hard coding which filters are perf filters here we can probably rewrite this to make use of the `isPerformanceFilterKey` util.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+    } else {
+      setFilterData(ModelCatalogStringFilterKey.HARDWARE_TYPE, []);
+    }
+  } else if (filterKey === ModelCatalogNumberFilterKey.MAX_RPS) {
+    setFilterData(
+      ModelCatalogNumberFilterKey.MAX_RPS,
+      typeof value === 'number' ? value : undefined,
+    );
+  } else if (isLatencyMetricFieldName(filterKey)) {
+    setFilterData(filterKey, typeof value === 'number' ? value : undefined);
```
mturley, line 57: It seems like maybe we should be able to avoid special cases for applying specific filters and instead use their type (string vs number) but it's probably not worth worrying about this right now.

clients/ui/frontend/src/app/pages/modelCatalog/utils/performanceFilterUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+      result[ModelCatalogStringFilterKey.USE_CASE] = validValues;
+    } else if (isHardwareType) {
+      const values =
+        fieldFilter.operator === FilterOperator.IN && Array.isArray(fieldFilter.value)
+          ? fieldFilter.value.filter((v): v is string => typeof v === 'string')
+          : typeof fieldFilter.value === 'string'
+            ? [fieldFilter.value]
+            : [];
+      result[ModelCatalogStringFilterKey.HARDWARE_TYPE] = values;
+    } else if (isRps) {
```
mturley, line 128: Another place where we have special logic for each filter, if we add more perf filters later we'll have to make sure we handle them in all these places. at some point making this more generic or centralized would be nice.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -54,14 +86,65 @@ export const getLatencyFieldName = (
...
-  LatencyMetric,
-).flatMap((metric) =>
-  Object.values(LatencyPercentile).map((percentile) => getLatencyFieldName(metric, percentile)),
+export const ALL_LATENCY_FILTER_KEYS: LatencyFilterKey[] = Object.values(LatencyMetric).flatMap(
+  (metric) =>
+    Object.values(LatencyPercentile).map((percentile) => getLatencyFilterKey(metric, percentile)),
 );
 
+// Alias for backward compatibility
+export const ALL_LATENCY_FIELD_NAMES = ALL_LATENCY_FILTER_KEYS;
```
mturley, line 117: Again it feels strange for this to exist in the codebase with 2 names especially since one will now be misleading given the new distinction between field names and filter keys.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -54,14 +86,65 @@ export const getLatencyFieldName = (
...
+
+// Type guard to check if a string is a valid LatencyFilterKey
+export const isLatencyFilterKey = (value: string): value is LatencyFilterKey =>
+  ALL_LATENCY_FILTER_KEYS.some((name) => name === value);
+
+// Alias for backward compatibility
+export const isLatencyMetricFieldName = isLatencyFilterKey;
```
mturley, line 124: Same here, let's remove these aliases and use only one name for each thing across the code.

mturley (approved):
Approving because the bulk of it is working and we can unblock testing, but I have some more feedback to follow up on in another PR.

In addition to the inline comments, some things I noted while testing:
* Filters in the hw table should always use defaults, regardless of the toggle on the landing page
	* Perf filters should always have a value in state, even if the toggle is off. When the toggle is off we just make those filters not be passed in the filterQuery or appear as chips.
	* Code that clears perf filters should always just set to defaults instead. We can probably simplify a lot of the functions in ModelCatalogContext here, there need not be a difference between resetting vs setting to defaults for perf filters.
* Resetting filters on details page should not clear them to empty, it should set them to default
* Clear filters link should disappear when all filters are either empty or default
	* Look at what's going on with the PF ToolbarFilter code that renders that link? Make sure we aren't rendering a redundant one anymore
* Low priority: filter toolbar should wrap if necessary to prevent horizontal scroll, on both pages

mturley:
/approve
/lgtm

See above review though, let's follow up to address those things.

PR #2081, opened 2026-01-09
Default sorting on the Model Catalog when Toggle is On

YuliaKrimerman:
## Description

Adds sorting to the Model Catalog UI with two options:
- **Recent publish**: Sorts by `LAST_UPDATE_TIME` descending (default when performance view is off)
- **Lowest latency**: Sorts by the active latency metric (e.g., `artifacts.ttft.double_value`) ascending (default when performance view is on)
### Unit Tests
- **ModelCatalogContext.spec.tsx** (8 tests):
- **ModelCatalogSortDropdown.spec.tsx** (11 tests):

### Manual Testing
- Verified sort dropdown appears only when performance view is enabled
- Verified default sort changes when toggling performance view
- Verified sort options are correctly enabled/disabled based on context
- Verified API parameters are correctly passed to backend
- Verified console logs show correct request/response data ( see screenshots)
Screenshot 2026-01-07 at 11 49 17 AM
Screenshot 2026-01-07 at 11 49 36 AM

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogSortDropdown.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogSortDropdown.tsx
@@ -22,9 +22,9 @@ const ModelCatalogSortDropdown: React.FC<ModelCatalogSortDropdownProps> = ({
   const [isOpen, setIsOpen] = React.useState(false);
 
   const activeLatencyField = getActiveLatencyFieldName(filterData);
-  // When performance view is enabled, always allow "Lowest latency" to be selectable
-  // The API will handle fallback to recent publish if no latency field is active
-  const isLatencySortDisabled = !performanceViewEnabled && activeLatencyField === undefined;
+  // Disable latency sort if performance view is disabled or there's no active latency field
+  // Without an active latency field, sorting by latency would fallback to sorting by publish date
+  const isLatencySortDisabled = !performanceViewEnabled || activeLatencyField === undefined;
 
   // Hide dropdown when performance view is disabled
   if (!performanceViewEnabled) {
--- a/clients/ui/frontend/src/concepts/modelCatalog/const.ts
+++ b/clients/ui/frontend/src/concepts/modelCatalog/const.ts
@@ -576,13 +576,11 @@ export enum ModelCatalogSortOption {
   LOWEST_LATENCY = 'lowest_latency',
 }
 
-// API sort order constants
-export const SORT_ORDER = {
-  ASC: 'ASC',
-  DESC: 'DESC',
-} as const;
+export enum SortOrder {
+  ASC = 'ASC',
+  DESC = 'DESC',
+}
 
-// API sort field constants
-export const SORT_FIELD = {
-  LAST_UPDATE_TIME: 'LAST_UPDATE_TIME',
-} as const;
+export enum SortField {
+  LAST_UPDATE_TIME = 'LAST_UPDATE_TIME',
+}
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogCardBody.tsx
@@ -26,6 +26,7 @@ import {
   ModelCatalogNumberFilterKey,
   LatencyMetric,
   parseLatencyFilterKey,
+  SortOrder,
 } from '~/concepts/modelCatalog/const';
 import { useCatalogPerformanceArtifacts } from '~/app/hooks/modelCatalog/useCatalogPerformanceArtifacts';
 import {
@@ -88,7 +89,7 @@ const ModelCatalogCardBody: React.FC<ModelCatalogCardBodyProps> = ({
         // If a latency filter is applied, sort artifacts on the card by lowest latency.
         ...(latencyFieldName && {
           orderBy: stripArtifactsPrefix(latencyFieldName),
-          sortOrder: 'ASC',
+          sortOrder: SortOrder.ASC,
         }),
       },
       performanceViewEnabled ? filterData : undefined,
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/__tests__/ModelCatalogSortDropdown.spec.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/__tests__/ModelCatalogSortDropdown.spec.tsx
@@ -173,7 +173,7 @@ describe('ModelCatalogSortDropdown', () => {
       });
     });
 
-    it('should enable "Lowest latency" option when performance view is enabled', async () => {
+    it('should disable "Lowest latency" option when there is no active latency field', async () => {
       mockGetActiveLatencyFieldName.mockReturnValue(undefined);
 
       render(
@@ -187,7 +187,8 @@ describe('ModelCatalogSortDropdown', () => {
 
       await waitFor(() => {
         const lowestLatencyOption = screen.getByTestId('sort-option-lowest-latency');
-        expect(lowestLatencyOption).not.toHaveAttribute('disabled');
+        const button = lowestLatencyOption.querySelector('button');
+        expect(button).toBeDisabled();
       });
     });
 
@@ -277,6 +278,9 @@ describe('ModelCatalogSortDropdown', () => {
     });
 
     it('should call setSortBy when "Lowest latency" is selected', async () => {
+      // Mock an active latency field so the option is enabled
+      mockGetActiveLatencyFieldName.mockReturnValue('artifacts.ttft_p90.double_value');
+
       const TestComponent: React.FC = () => {
         const { sortBy } = React.useContext(ModelCatalogContext);
         const [selectedValue, setSelectedValue] = React.useState<string>('');
--- a/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
@@ -24,6 +24,8 @@ import {
   isPerformanceStringFilterKey,
   PERFORMANCE_FILTER_KEYS,
   ModelCatalogSortOption,
+  SortOrder,
+  SortField,
 } from '~/concepts/modelCatalog/const';
 import { CatalogSourceStatus } from '~/concepts/modelCatalogSettings/const';
 
@@ -248,8 +250,8 @@ export const getSortParams = (
 ): { orderBy: string; sortOrder: string } => {
   const effectiveSortBy = getEffectiveSortBy(sortBy, performanceViewEnabled);
   const recentPublishSort = {
-    orderBy: 'LAST_UPDATE_TIME',
-    sortOrder: 'DESC',
+    orderBy: SortField.LAST_UPDATE_TIME,
+    sortOrder: SortOrder.DESC,
   } as const;
 
   if (effectiveSortBy === ModelCatalogSortOption.RECENT_PUBLISH) {
@@ -267,7 +269,7 @@ export const getSortParams = (
   // This matches the filter key format used in filterQuery, so we can use it directly
   return {
     orderBy: activeLatencyField,
-    sortOrder: 'ASC', // Lowest first (ascending)
+    sortOrder: SortOrder.ASC, // Lowest first (ascending)
   };
 };
 
--- a/clients/ui/package-lock.json
+++ b/clients/ui/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "ui",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}


clients/ui/frontend/src/app/context/modelCatalog/__tests__/ModelCatalogContext.spec.tsx
```diff
@@ -0,0 +1,210 @@
...
+        getCatalogModelPerformanceArtifacts: jest.fn(),
+        getAllCatalogSources: jest.fn(),
+        getCatalogFilterOptions: jest.fn(),
+        createCatalogSourcePreview: jest.fn(),
+      },
+    },
+    jest.fn(),
+  ]),
+}));
+
```
manaswinidas, line 67: 🔴 **Major Duplication**: This entire mock block (~50 lines, lines 10-53) is duplicated verbatim in `ModelCatalogSortDropdown.spec.tsx`.

**Recommendation**: Extract these mocks into a shared test utility file:

```typescript
// __tests__/mocks/modelCatalogTestMocks.ts
export const setupModelCatalogMocks = () => {
  jest.mock('mod-arch-core', () => ({...}));
  jest.mock('react-router-dom', () => ({...}));
  jest.mock('~/app/hooks/modelCatalog/useCatalogSources', () => ({...}));
  // ... other mocks
};
```

Then in both test files:
```typescript
import { setupModelCatalogMocks } from '../mocks/modelCatalogTestMocks';
setupModelCatalogMocks();
```

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -188,8 +194,23 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+        });
+      } else {
         setPerformanceFiltersChangedOnDetailsPage(false);
+        // When toggle is off, set default sort to recent_publish
+        setSortBy((currentSortBy) => {
+          if (currentSortBy === null || currentSortBy === ModelCatalogSortOption.LOWEST_LATENCY) {
+            return ModelCatalogSortOption.RECENT_PUBLISH;
+          }
+          return currentSortBy;
+        });
```
manaswinidas: 🔴 **DRY Violation**: This sort default logic is duplicated. The same logic exists in:
1. Here in the context (lines 197-213)
2. `getEffectiveSortBy()` in `modelCatalogUtils.ts`
3. Inline in `ModelCatalogGalleryView.tsx`

**Recommendation**: Reuse `getEffectiveSortBy` from utils:

```typescript
import { getEffectiveSortBy } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';

const setPerformanceViewEnabled = React.useCallback(
  (enabled: boolean) => {
    // ... existing code
    const defaultSort = getEffectiveSortBy(null, enabled);
    const oppositeDefault = getEffectiveSortBy(null, !enabled);
    setSortBy((currentSortBy) => 
      currentSortBy === null || currentSortBy === oppositeDefault
        ? defaultSort
        : currentSortBy
    );
  },
  [resetPerformanceFiltersToDefaults],
);
```

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
```diff
@@ -49,13 +53,42 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
...
+
+    // effectiveSortBy must be LOWEST_LATENCY at this point
+    const latencyField = getActiveLatencyFieldName(filterData);
+    if (!latencyField) {
+      // Fallback to recent publish if no latency field is available
+      return {
+        orderBy: SORT_FIELD.LAST_UPDATE_TIME,
+        sortOrder: SORT_ORDER.DESC,
+      };
+    }
```
manaswinidas: 🔴 **Duplication & Performance**: This `sortParams` logic duplicates fallback handling and has a performance issue.

**Issues:**
1. The fallback `{ orderBy: SORT_FIELD.LAST_UPDATE_TIME, sortOrder: SORT_ORDER.DESC }` appears twice
2. The entire `filterData` object is in `useMemo` dependencies, but only latency fields are used - causing unnecessary recalculations when any filter changes

**Recommendation**: Extract to a utility function and optimize:

```typescript
// In modelCatalogUtils.ts
export const getSortParams = (
  sortBy: ModelCatalogSortOption | null,
  performanceViewEnabled: boolean,
): { orderBy: string; sortOrder: string } => {
  const effectiveSortBy = getEffectiveSortBy(sortBy, performanceViewEnabled);
  const recentPublishSort = { orderBy: SORT_FIELD.LAST_UPDATE_TIME, sortOrder: SORT_ORDER.DESC };
  
  if (effectiveSortBy === ModelCatalogSortOption.RECENT_PUBLISH) {
    return recentPublishSort;
  }
  
  const latencyField = getActiveLatencyFieldName(filterData);
  return latencyField
    ? { orderBy: `artifacts.${latencyField}.double_value`, sortOrder: SORT_ORDER.ASC }
    : recentPublishSort;
};
```

Then here:
```typescript
const activeLatencyField = React.useMemo(
  () => getActiveLatencyFieldName(filterData),
  [filterData]
);
const sortParams = React.useMemo(
  () => getSortParams(sortBy, performanceViewEnabled, { ...filterData }),
  [sortBy, performanceViewEnabled, activeLatencyField] // Use activeLatencyField for stable deps
);
```

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -568,3 +568,19 @@ export enum ModelDetailsTab {
...
+}
+
+// API sort order constants
+export const SORT_ORDER = {
+  ASC: 'ASC',
+  DESC: 'DESC',
+} as const;
+
+// API sort field constants
+export const SORT_FIELD = {
```
manaswinidas: 🟡 **Inconsistent Pattern**: The new constants use two different patterns:
- `ModelCatalogSortOption` uses `enum`
- `SORT_ORDER` and `SORT_FIELD` use `as const` objects

**Recommendation**: For consistency with the codebase (which uses enums extensively), consider using enums. Also add type aliases for better type safety:

```typescript
export enum SortOrder {
  ASC = 'ASC',
  DESC = 'DESC',
}

export enum SortField {
  LAST_UPDATE_TIME = 'LAST_UPDATE_TIME',
}
```

Or if keeping `as const`, add type aliases:
```typescript
export type SortOrderType = typeof SORT_ORDER[keyof typeof SORT_ORDER];
export type SortFieldType = typeof SORT_FIELD[keyof typeof SORT_FIELD];
```

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogSortDropdown.tsx
```diff
@@ -0,0 +1,98 @@
...
+const ModelCatalogSortDropdown: React.FC<ModelCatalogSortDropdownProps> = ({
+  performanceViewEnabled,
+}) => {
+  const { sortBy, setSortBy, filterData } = React.useContext(ModelCatalogContext);
+  const [isOpen, setIsOpen] = React.useState(false);
+
+  const activeLatencyField = getActiveLatencyFieldName(filterData);
+  // When performance view is enabled, always allow "Lowest latency" to be selectable
+  // The API will handle fallback to recent publish if no latency field is active
+  const isLatencySortDisabled = !performanceViewEnabled && activeLatencyField === undefined;
```
manaswinidas: 🟡 **Dead Code**: This condition contains dead code.

Since the component returns `null` when `!performanceViewEnabled` (line 30), this condition:
```typescript
const isLatencySortDisabled = !performanceViewEnabled && activeLatencyField === undefined;
```
Will never evaluate to `true` when the dropdown actually renders.

**Recommendation**: Simplify to:
```typescript
// The dropdown only renders when performanceViewEnabled is true
// so we only need to check if there's an active latency field
const isLatencySortDisabled = activeLatencyField === undefined;
```

However, the comment on line 25-26 says "always allow 'Lowest latency' to be selectable when performance view is enabled" which seems contradictory. Please clarify the intended behavior.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogSortDropdown.tsx
```diff
@@ -0,0 +1,98 @@
...
+  // Hide dropdown when performance view is disabled
+  if (!performanceViewEnabled) {
+    return null;
+  }
+
+  const isValidSortOption = (value: unknown): value is ModelCatalogSortOption =>
+    typeof value === 'string' &&
+    (value === ModelCatalogSortOption.RECENT_PUBLISH ||
+      value === ModelCatalogSortOption.LOWEST_LATENCY);
+
```
manaswinidas, line 38: 🟢 **Reuse Opportunity**: This type guard could be moved to `const.ts` for reuse elsewhere:

```typescript
// In const.ts
export const isModelCatalogSortOption = (value: unknown): value is ModelCatalogSortOption =>
  Object.values(ModelCatalogSortOption).includes(value as ModelCatalogSortOption);
```

This pattern is consistent with existing type guards like `isLatencyFilterKey` in the same file.

clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogModelsBySource.ts
```diff
@@ -116,14 +135,16 @@ export const useCatalogModelsBySources = (
     sourceLabel,
     filterData,
     filterOptions,
+    sortBy,
+    sortOrder,
   ]);
 
   React.useEffect(() => {
     setAllItems([]);
     setTotalSize(0);
```
manaswinidas, line 149: 🟡 **UX Consideration**: Every time sort changes, all items are cleared immediately. This causes a flash of empty content before new data loads.

Consider adding a loading state or keeping the old items visible during the transition (optimistic UI pattern). This is especially noticeable on slower connections.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
```diff
@@ -49,13 +53,42 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
...
+      return {
+        orderBy: SORT_FIELD.LAST_UPDATE_TIME,
+        sortOrder: SORT_ORDER.DESC, // Most recent first
+      };
+    }
+
+    // effectiveSortBy must be LOWEST_LATENCY at this point
+    const latencyField = getActiveLatencyFieldName(filterData);
+    if (!latencyField) {
+      // Fallback to recent publish if no latency field is available
```
manaswinidas: 🟢 **Minor**: This template literal pattern appears in multiple places. Consider a utility:

```typescript
// In modelCatalogUtils.ts
export const getArtifactSortField = (field: string): string => 
  `artifacts.${field}.double_value`;
```

This is consistent with the existing pattern for artifact field paths in the codebase.

manaswinidas:
Can you check we are sorting using the same key as the filter key? I think it needs the artifacts.* prefix and double_value suffix if it's not getting those. Should be the same key we use for filtering.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -570,3 +570,19 @@ export enum ModelDetailsTab {
...
+  LAST_UPDATE_TIME: 'LAST_UPDATE_TIME',
+} as const;
```
mturley: These aren't used anywhere. We should change them to be enums (for consistency) and then use them in place of hard-coded 'ASC', 'DESC' and 'LAST_UPDATE_TIME' strings.

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogSortDropdown.tsx
mturley: The latency sort should be disabled if there is no active latency field, essentially the `&&` here should be a `||`. We'll never hit that if the default filter values are populated properly, but if there's a problem with that it would be better to not let the user try to sort by latency rather than have it appear to sort by latency but really sort by publish.

mturley (changes requested):
A couple minor comments, in addition to the fix I raised that we discussed: https://github.com/YuliaKrimerman/model-registry/pull/4

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -188,6 +195,17 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
...
+
+      // Update sort to default for the new toggle state, preserving user selection if it doesn't match the opposite default
+      const defaultSort = getEffectiveSortBy(null, enabled);
+      const oppositeDefault = getEffectiveSortBy(null, !enabled);
+      setSortBy((currentSortBy) => {
+        if (currentSortBy === null || currentSortBy === oppositeDefault) {
+          return defaultSort;
+        }
+        return currentSortBy;
+      });
```
mturley, line 207: I'm not sure exactly what this is trying to do. I don't see my selection preserved when using the toggle, it always goes to lowest latency when I turn the toggle on. I think that's the expected behavior?

PR #2111, opened 2026-01-16
Model Performance view should only show validated models

Philip-Carneiro:
## Description

When Model performance view toggle is enabled, it should only show the validated models.
image

## How Has This Been Tested?

It have been manually tested.

## Merge criteria:

- [x] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/mocks/model_catalog_client_mock.go
+++ b/clients/ui/bff/internal/mocks/model_catalog_client_mock.go
@@ -5,6 +5,7 @@ import (
 	"log/slog"
 	"math"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


ppadti:
Thanks @Philip-Carneiro for fixing this. 
I think the problem is that we are not passing filterQuery with /models endpoint, when the validated source is disabled and the toggle is turned ON - because the filter_options endpoint doesn't includes performance data on filters object(when that source is disabled), and we are using those to construct filterQuery in `serializeFilterEntry` utils. If we fix this util and pass the filterQuery with default values - we will see the filtered models when the toggle is turned ON, regardless of the sources present. Let me know if that makes sense.

PR #2112, opened 2026-01-16
Fix arbitrary changing max value of latency slider in details page

manaswinidas:
## Description

Now the max value is fixed even after applying filters and re-opening the latency dropdown in details page. This change is tested midstream.

https://github.com/user-attachments/assets/6eab408e-6db3-41b7-84dd-e2c941879c19

https://github.com/user-attachments/assets/318fddef-0e21-4553-92a3-b707ef1434a4

## How Has This Been Tested?

Go to details page(hardware configuration table) of model catalog of any model - change the slider value of the latency filter and apply filter - re-open the filter and check whether the max value stays the same(shouldn't change to the changed value)

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
@@ -114,11 +114,8 @@ const LatencyFilter: React.FC = () => {
   }, [currentActiveFilter]);
 
   const { minValue, maxValue, isSliderDisabled } = React.useMemo((): SliderRange => {
-    // Use full filter key for accessing filterOptions
     const filterKey = getLatencyFilterKey(localFilter.metric, localFilter.percentile);
 
-    // Always get range from filterOptions (which provides the full range across all artifacts)
-    // Don't use performanceArtifacts since we may not have all of them in memory when paginating
     const latencyFilter = filterOptions?.filters?.[filterKey];
     if (latencyFilter && 'range' in latencyFilter && latencyFilter.range) {
       return {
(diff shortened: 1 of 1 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
```diff
@@ -130,6 +127,22 @@ const LatencyFilter: React.FC = () => {
     return FALLBACK_LATENCY_RANGE;
   }, [localFilter.metric, localFilter.percentile, filterOptions]);
 
+  // Reset value to max when metric or percentile changes (range changes)
```
YuliaKrimerman, line 130: Why are we automatically resetting to maxValue. Shouldn't we try to preserve the user's intent by keeping their value if it's within the new range, or clamping it to the new range if it's outside?
This matters for UX - if I set E2E P90 to 20ms, then switch to TTFT P90 (which has a max of 50ms), I probably want it to stay at 20ms, not jump to 50ms.
We already have clampedValue at line 146-149 that does this exact clamping.Can we leverageit instead  of introducing a separate reset mechanism? The clamping should already handle cases where the value is outside the valid range.

manaswinidas: But because of the clamping .. we are losing the maxValue entirely .. what if the user wants to increase the maxValue - the error was when a user selects 20ms and max is 50ms - you apply the filter and then re-open the filter, the maxValue should still be 50ms and not 20ms(provided you're not changing the metric and percentile). If you change the metric and percentile, e.g.TTFT P90 to E2E P90 the maxValue changes accordingly coz the range changes entirely.

https://github.com/user-attachments/assets/c0f2e7da-8ba9-4447-b50d-9b8df6eaf5c3

YuliaKrimerman: Gotcha, yeah that makes, I just didn't dig into the full context of the original issue with it

PR #2124, opened 2026-01-19
Install kustomize if not present

fege:
## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/clients/python/Makefile
+++ b/catalog/clients/python/Makefile
@@ -53,8 +53,8 @@ endif
 # Kustomize overlay for E2E testing
 KUSTOMIZE_E2E := ../../../manifests/kustomize/options/catalog/overlays/e2e
 LOCALBIN := $(shell pwd)/bin
-# Use local kustomize if available, otherwise fall back to system kustomize
-KUSTOMIZE ?= $(shell if [ -f $(LOCALBIN)/kustomize ]; then echo $(LOCALBIN)/kustomize; else echo kustomize; fi)
+KUSTOMIZE := $(LOCALBIN)/kustomize
+KUSTOMIZE_VERSION ?= 5.5.0
 
 .PHONY: install
 install: generate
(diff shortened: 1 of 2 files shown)


catalog/clients/python/Makefile
```diff
@@ -187,11 +190,24 @@ deploy-namespace:
 	@echo "Creating namespace: $(CATALOG_NAMESPACE)"
 	kubectl create namespace "$(CATALOG_NAMESPACE)" --dry-run=client -o yaml | kubectl apply -f -
 
+# Install kustomize if not already present
+.PHONY: deploy-kustomize
+deploy-kustomize:
+	@if [ ! -f $(KUSTOMIZE) ]; then \
```
pboyd, line 196: Is the intention for this to always install kustomize, or only install it if necessary?

It looks like this will install it in `$LOCALBIN` even if kustomize is on the user's path. I'm not sure if this is the intended behavior or not.

pboyd (approved):
One question, but..

/lgtm

catalog/clients/python/Makefile
Al-Pragliola: that's what we do in general in the root makefile, no? I think we should remove the fallback to system kustomize and just use the one in the bin folder

fege: `KUSTOMIZE ?= $(shell if [ -f $(LOCALBIN)/kustomize ]; then echo $(LOCALBIN)/kustomize; else echo kustomize; fi)` is setting `KUSTOMIZE` to the one in the user path as fallback and if that is setup is not going to install anything. if kustomize is not present at all yes it is going to be installed in `$LOCALBIN`

pboyd: @fege, the `else` block in that line sets `$KUSTOMIZE` to `kustomize`, but `-f` (in that case) only checks for a file called `kustomize` in the current directory so I wouldn't expect that to ever be true.

fege: @pboyd got it thanks, I did implement @Al-Pragliola 's suggestion in my third commit, hopefully ;)

pboyd: /lgtm

PR #2132, opened 2026-01-19
Added edge case for re-enabling source on the MC settings page

YuliaKrimerman:
## Description

Fixed a UI feedback issue where re-enabling a catalog source after disabling it showed "-" instead of "Starting" status, providing no feedback that the source was loading.

**Root Cause:**
When a catalog source is re-enabled:
- ConfigMap updates immediately: `enabled: true`
- Database/API response lags behind: `status: DISABLED` (until loading completes)
- Frontend only showed "Starting" for new sources (status=null), not for re-enabled sources (status=DISABLED)

**Solution:**
Updated `CatalogSourceStatus.tsx` to handle the ConfigMap/API mismatch. The component now shows "Starting" status when:
- ConfigMap indicates `enabled: true` (checked at line 24)
- AND API still returns `status: DISABLED` (caught in switch statement)

**Code Changes:**
- Leverages existing early return when `enabled: false` (line 24-26)
- When `DISABLED` case is reached in switch statement, we know `enabled: true` due to early return
- Simplified logic to return "Starting" label directly without redundant conditional check

**User Experience:**
- ✅ First time enabling: Shows "Starting" → "Available"/"Failed" (already worked)
- ✅ Re-enabling after disable: Shows "Starting" → "Available"/"Failed" (now fixed)
- ✅ Disabled source: Shows "-" (unchanged)

## How Has This Been Tested?

- Manually tested the re-enable flow:
  1. Created a catalog source
  2. Disabled it (status shows "-")
  3. Re-enabled it (status now shows "Starting" instead of "-")
- Verified disabled sources still show "-"
- Verified first-time enable still shows "Starting"
- Code passes linter checks with no errors

## Merge criteria:

If you have UI changes

  - _Explanation: This is a UI timing/state synchronization fix. The existing component relies on early returns for control flow, and the fix leverages that existing logic. Manual testing confirms the fix works correctly._

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogSettings.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogSettings.cy.ts
@@ -469,6 +469,20 @@ describe('Catalog Source Configs Table', () => {
       row.shouldHaveValidationStatus('Starting');
     });
 
+    it('should show "Starting" status when re-enabling source (enabled=true but status=disabled)', () => {
+      const reenablingSource = mockCatalogSource({
+        id: 'hf-google',
+        name: 'HuggingFace Google',
+        status: 'disabled',
+      });
+      setupMocks([reenablingSource], { catalogs: [huggingFaceSource] }); // huggingFaceSource has enabled=true
+      modelCatalogSettings.visit();
+      const row = modelCatalogSettings.getRow('HuggingFace Google');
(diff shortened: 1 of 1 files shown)


mturley:
@YuliaKrimerman can we explore maybe adding some tests here? we could see if AI can generate a handful of cypress tests to check that we're showing the right thing in the status column for various cases of what's present in the configmap and in the API. if that takes too much of your time we could open an issue to improve testing coverage instead.

YuliaKrimerman:
@mturley @manaswinidas the tests are there for other cases of the status updates, I added another test case that covers this scenario

PR #2133, opened 2026-01-20
Parse port from base url

### LINKED ISSUE #2110: Parse port from base url

I would like to be able to pass the base url for the server including the port as a single string. This is helpful if I have this string in a variable and want to pass it through rather than splitting it apart to separate the port.

Instead of having to specify the port separately
```
registry = ModelRegistry(
    server_address="http://localhost",
    port=9000,
    is_secure=False
)
```
I would like to be able to include it in the url
```
registry = ModelRegistry(
    server_address="http://localhost:9000",
    is_secure=False
)
```

The two examples above should be treated the same.

If the `port` kwarg is specified **and** a port is given in `server_address` at the same time this should probably cause an error.

--

**Additional context**

--

osamaahmed17:
Port is now provided in the server address.
Fixes #2110 
## Description
The port is now included in the URL, so we don’t need to split it separately. 

**Changed this**
```
client = ModelRegistry(
        server_address="http://localhost",
        port=8080,
        author="test_author",
        is_secure=False,
        # user_token=None -> ... Let it read from Env var
    )
```
**To this**
```
client = ModelRegistry(
        server_address="http://localhost:8080",
        author="test_author",
        is_secure=False,
        # user_token=None -> ... Let it read from Env var
    )
```

For secure connections 
secure connections are handled using Python’s `urllib.parse` module.

### DIFF
--- a/clients/python/tests/test_connection.py
+++ b/clients/python/tests/test_connection.py
@@ -18,11 +18,11 @@ def test_connection_args():
 
     # Secure connection tests
     client = ModelRegistryAPIClient.secure_connection(
-        server_address, port, user_token="token"
+        server_address, port, user_token="token"  # noqa: S106
     )
     assert client.config.host == host
 
     client = ModelRegistryAPIClient.secure_connection(
-        server_address_with_port, port, user_token="token"
+        server_address_with_port, port, user_token="token"  # noqa: S106
     )
     assert client.config.host == server_address_with_port


osamaahmed17:
@jonburdo , @pboyd , and @Al-Pragliola kindly review the pull request when you get the time.  Do let me know if anything needs to be added or removed.

pboyd:
Hi @osamaahmed17, thanks for the PR.

I might be missing something here, but I don't see anything to explicitly support putting the port number in `server_address` already.

It looks like it may work for an [`insecure_connection`](https://github.com/kubeflow/model-registry/blob/main/clients/python/src/model_registry/core.py#L68) (if the connection works with a trailing `:` in the URL). But I don't see how it could work for a [`secure_connection`](https://github.com/kubeflow/model-registry/blob/main/clients/python/src/model_registry/core.py#L41).

osamaahmed17:
@pboyd  We may use url parse module  to handle secure connections.

osamaahmed17:
@pboyd and @Al-Pragliola, do let me know if this **urllib.parse** module looks good and safe enough to be integrated. This is the only approach I could find to parse the url for secure connection.

https://docs.python.org/3/library/urllib.parse.html

pboyd:
@osamaahmed17, `url.parse` is fine. The code looks good to me.

Would you mind adding a test case for the old-style call with an explicit `port` argument? We'll need to support that version for a while and I want to make sure that we don't break it in the future.

clients/python/src/model_registry/core.py
```diff
@@ -56,9 +57,12 @@ def secure_connection(
             user_token: The PEM-encoded user token as a string.
             custom_ca: The path to a PEM-
         """
+        parsed_url = urlparse(server_address if "//" in server_address else f"//{server_address}")
```
jonburdo:
```suggestion
        parsed_url = urlparse(server_address)
```
Do we need `server_address if "//" in server_address else f"//{server_address}"`? It doesn't hurt, but I believe it's unnecessary complexity.

I understand it allows us to correctly recognize the port in something like `localhost:8080` as well as `http://localhost:8080` but if I'm not mistaken, the schema (`http://`) will always be necessary.

Same below.

osamaahmed17: Okay 👍

osamaahmed17:
@pboyd, I've added a test case. Please let me know if this looks correct.

PR #2148, opened 2026-01-26
Add ordering tests for catalog e2e from downstream

lugi0:
## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/clients/python/tests/sorting_utils.py
+++ b/catalog/clients/python/tests/sorting_utils.py
@@ -1,5 +1,6 @@
 """Utility functions for sorting tests."""
 
+import pytest
 from typing import Any
 
 
@@ -53,7 +54,7 @@ def validate_items_sorted_correctly(items: list[dict], field: str, order: str) -
         True if items are sorted correctly, False otherwise
     """
     if len(items) < 2:
-        return True
+        pytest.fail("List has fewer than 2 items, double check the data you are passing to this function")
 
     values = [get_field_value(item, field) for item in items]
 


catalog/clients/python/tests/test_ordering.py
```diff
@@ -131,6 +133,49 @@ def test_name_ordering_pagination_maintains_order(self, api_client: CatalogAPICl
...
+        """
+        response = api_client.get_models(order_by=order_by, sort_order=sort_order)
+
+        assert isinstance(response, dict), f"Response is not a dict: {type(response)}"
+        assert "items" in response, f"Response missing 'items' field: {response.keys()}"
+
+        items = response["items"]
+        if len(items) < 2:
+            pytest.skip(f"Need at least 2 models to test {order_by} sorting")
+
```
fege: we load the data, so we know that we have more than 1 model, I would fail the test instead of skipping if we have less models

lugi0: Removed

catalog/clients/python/tests/sorting_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        items: List of items to validate
+        field: Field name to check sorting on (ID, CREATE_TIME, LAST_UPDATE_TIME)
+        order: Sort order (ASC or DESC)
+
+    Returns:
+        True if items are sorted correctly, False otherwise
+    """
+    if len(items) < 2:
+        return True
```
dbasunag: If we are adding the tests, we have control of the number of items. We should not use data that can cause two items in the list for sorting tests. Otherwise, we would be silently passing sorting tests without real testing and since there would be no d/s tests to catch the issue, we can hide bugs.

lugi0: While I agree with the concept of not using data that makes little to no sense in the tests as long as we are in control of it, this is a function that returns True if a list of items is sorted correctly. If there are fewer than 2 items, the list is sorted correctly by definition, so IMHO this is the correct behaviour of the function. 
I can add a warning here that says something along the lines of `List has fewer than 2 items, are you sure this was the intended data for the call?`

dbasunag: Can we consider failing the test instead? My main worry is, no one is going to look at the log the the tests are passing. If it fails, that forces us to take a look and change the problematic test(s).

lugi0: Done

dbasunag:
Thank you @lugi0 .
/lgtm

PR #2158, opened 2026-01-26
Fix Reset buttons behavior for Latency and Max RPS filters 

Taj010:
## Description

Initially when the rest button was applied in the Latency and Max RPS filters it would reapply default values and trigger filter updates instead of undoing unapplied changes. The Reset Button now restores the filter values that were active when the dropdown was opened, without applying any changes. 

## How Has This Been Tested?

Manual Testing

Verified in browser that in the Model Catalog  when the Model performance view toggle is enabled and inside the performance insights tab for latency filter and MaxRPS filter
- when the respective filter is applied and dropdown is opened again if the values has been changed (without applying them) and the Reset button is clicked it should restore to applied values and do not refresh. 
- When the dropdown is opened but no values have been changed and adjusted it should simply show the current applied values and not go back to default values. 

Automated Tests
* Cypress E2E tests: Adjusted the `modelCatalog/modelCatalogTabs.cy.ts `file to remove the the old 'should reset to default latency filter (TTFT P90) when filter is reset' and adjust it to where the reset should restore applied values test.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
@@ -102,33 +102,17 @@ const LatencyFilter: React.FC = () => {
     return defaultFilterState;
   });
 
-  const appliedFilterStateRef = React.useRef<LatencyFilterState | null>(null);
-
-  // Update local filter when active filter changes
-  React.useEffect(() => {
-    if (currentActiveFilter) {
-      setLocalFilter({
-        metric: currentActiveFilter.metric,
-        percentile: currentActiveFilter.percentile,
-        value: currentActiveFilter.value,
-      });
-    }
-  }, [currentActiveFilter]);
-
-  // Capture the applied filter state when menu opens
   React.useEffect(() => {
     if (isOpen) {
-      // Capture the current filter state when menu opens
-      if (currentActiveFilter) {
-        appliedFilterStateRef.current = {
-          metric: currentActiveFilter.metric,
-          percentile: currentActiveFilter.percentile,
-          value: currentActiveFilter.value,
-        };
-      } else {
-        appliedFilterStateRef.current = defaultFilterState;
-      }
-      setLocalFilter(appliedFilterStateRef.current);
+      // Use currentActiveFilter or defaultFilterState
+      const initialState = currentActiveFilter
+        ? {
+            metric: currentActiveFilter.metric,
+            percentile: currentActiveFilter.percentile,
+            value: currentActiveFilter.value,
+          }
+        : defaultFilterState;
+      setLocalFilter(initialState);
     }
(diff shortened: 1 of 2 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
```diff
@@ -113,6 +115,23 @@ const LatencyFilter: React.FC = () => {
...
+  // Capture the applied filter state when menu opens
+  React.useEffect(() => {
+    if (isOpen) {
+      // Capture the current filter state when menu opens
+      if (currentActiveFilter) {
+        appliedFilterStateRef.current = {
+          metric: currentActiveFilter.metric,
+          percentile: currentActiveFilter.percentile,
+          value: currentActiveFilter.value,
+        };
```
mturley: I don't think you need a new way to capture the applied filter, can't you just use `currentActiveFilter` when resetting? because that comes directly from the page's full filter state and only changes when the user clicks Apply.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/MaxRpsFilter.tsx
```diff
@@ -10,28 +10,29 @@ import {
...
   const { value: rpsFilterValue, setValue: setRpsFilterValue } =
     useCatalogNumberFilterState(filterKey);
-  const { getPerformanceFilterDefaultValue } = React.useContext(ModelCatalogContext);
   const [isOpen, setIsOpen] = React.useState(false);
 
   const { minValue, maxValue, isSliderDisabled } = MAX_RPS_RANGE;
 
   const [localValue, setLocalValue] = React.useState<number>(() => rpsFilterValue ?? maxValue);
 
+  const appliedValueRef = React.useRef<number | undefined>(undefined);
```
mturley: Same here, I don't think we need a separate appliedValueRef if we have `rpsFilterValue` coming from context. It doesn't change until the user clicks apply.

clients/ui/frontend/src/app/pages/modelCatalog/components/globalFilters/LatencyFilter.tsx
```diff
@@ -102,33 +102,17 @@ const LatencyFilter: React.FC = () => {
...
-  // Update local filter when active filter changes
-  React.useEffect(() => {
-    if (currentActiveFilter) {
-      setLocalFilter({
-        metric: currentActiveFilter.metric,
-        percentile: currentActiveFilter.percentile,
-        value: currentActiveFilter.value,
-      });
-    }
-  }, [currentActiveFilter]);
```
mturley, line 116: With the removal of this, I think technically if some other code updates the active filter in context we'll no longer sync that change down to this local state unless the dropdown is open, but that's probably fine since we don't have any need to do that currently.

mturley (commented):
/lgtm
/approve

PR #2160, opened 2026-01-27
feat: add image signer

jonburdo:
This adds an image signer to sign and verify OCI container images using sigstore.

## Description

We'll add end-to-end tests as a follow-up but for now the tests avoid relying on the cosign binary and mainly test that we pass the correct options/args.

## How Has This Been Tested?

Tested with a private securesign instance, by imported `ImageSigner`, and passing:
```
identity_token
fulcio_url
rekor_url
tuf_url
oidc_issuer
```
Then tested sign and verify with a small test image on an oci registry.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/signing/image_signer.py
+++ b/clients/python/src/model_registry/signing/image_signer.py
@@ -1,0 +1,225 @@
+"""Cosign signing and verification for container images."""
+
+import os
+import shutil
+import subprocess
+import sys
+from functools import partial
+from pathlib import Path
+
+from model_registry.signing.exceptions import InitializationError, SigningError, VerificationError
+
+
+class CommandRunner:
+    """Command execution wrapper for running CLI tools."""
+
+    def __init__(self, **kwargs):
+        """Initialize CommandRunner with default subprocess.run arguments.
+
+        Args:
+            **kwargs: Additional default keyword arguments to pass to subprocess.run
+        """
+        self._run = partial(subprocess.run, check=True, capture_output=True, text=True, **kwargs)
+
+    def run(self, cmd: list[str]) -> subprocess.CompletedProcess:
+        """Run a CLI command and handle output.
+
+        Args:
+            cmd: Command and arguments as a list
+
+        Returns:
+            CompletedProcess instance
+
+        Raises:
+            subprocess.CalledProcessError: If the command fails
+        """
+        result = self._run(cmd)
+        if result.stderr:
+            print(result.stderr, file=sys.stderr)
+        return result
+
+
+class ImageSigner:
+    """Tool for signing and verifying container images."""
+
+    def __init__(
+        self,
+        tuf_url: str | None = None,
+        root: str | None = None,
+        root_checksum: str | None = None,
+        identity_token_path: str | os.PathLike[str] | None = None,
+        fulcio_url: str | None = None,
+        rekor_url: str | None = None,
+        certificate_identity: str | None = None,
+        oidc_issuer: str | None = None,
+    ):
+        """Initialize ImageSigner tool.
+
+        Args:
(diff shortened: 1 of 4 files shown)


clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Args:
+            tuf_url: TUF mirror URL (optional, overrides instance default)
+            root: Root JSON URL (optional, overrides instance default)
+            root_checksum: Root JSON checksum (optional, overrides instance default)
+        """
+        # Remove existing sigstore directory
+        sigstore_dir = Path.home() / ".sigstore"
+        if sigstore_dir.exists():
+            shutil.rmtree(sigstore_dir)
```
tarilabs: I suppose this is helpful for making sure in case any pre-existing root signing is available this actually replaces, but this is indeed "slate clean" any user-defined pre-existing root.

how do we warn this operation is not meant to be invoked lightheartedly? 

for example, in the async-upload job there is no problem in invoking this method, in a jupyter notebook likely it's not a problem for the data scientist to invoke this method, but this is a potentially descructive operation for a DevOps/AI Engineer persona.

clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if root_checksum is not None:
+            cmd.extend(["--root-checksum", root_checksum])
+
+        self.runner.run(cmd)
+
+    def sign(
+        self,
+        image: str,
+        identity_token: str | None = None,
```
tarilabs: instead of passing identity token via variable, should we point to the _path_ containing the token so we don't have to manage sensible information?

jonburdo:
Thanks @tarilabs! I'll make the changes to this one

clients/python/src/model_registry/signing/image_signer.py
jonburdo: I've adjusted so that an error will be raised instead by default. To overwrite any existing dir, the user can pass `force=True` (which we can do in the async job)

clients/python/src/model_registry/signing/image_signer.py
jonburdo: Changed to require that this be a path and rename to `identity_token_path`

clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Validate identity token path if provided
+        if identity_token_path is not None and not os.path.exists(identity_token_path):
+            msg = f"Identity token file not found: {identity_token_path}"
+            raise FileNotFoundError(msg)
+
+        cmd = ["cosign", "sign", "-y"]
+
+        if identity_token_path is not None:
+            cmd.extend(["--identity-token", os.fspath(identity_token_path)])
```
tarilabs, line 181: chef's kiss

```
    --identity-token='':
	identity token to use for certificate from fulcio. the token or a path to a file containing the token is
	accepted.
```

this way, one less problem "for us". thank you.

clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Raises:
+            FileExistsError: If sigstore directory exists and force=False
+        """
+        # Check for existing sigstore directory
+        sigstore_dir = self._get_sigstore_dir()
+        if sigstore_dir.exists():
+            if not force:
+                msg = f"Sigstore directory already exists: {sigstore_dir}. Use force=True to overwrite."
+                raise FileExistsError(msg)
```
tarilabs, line 118: perfect, thanks

PR #2165, opened 2026-01-28
Add namespace selector to register form

ppadti:
## Description
This PR aims to add a namespace selector the register form

https://github.com/user-attachments/assets/8234314c-d87e-464e-9476-3dce6b64e150

## How Has This Been Tested?
1. Go to register model/register new version form and check the namespace selector is present.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterAndStoreFields.tsx
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterAndStoreFields.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { UpdateObjectAtPropAndValue } from 'mod-arch-shared';
-import { Button, FormGroup, Popover } from '@patternfly/react-core';
-import { HelpIcon } from '@patternfly/react-icons';
+import { FormGroup } from '@patternfly/react-core';
+
 import { useThemeContext } from 'mod-arch-kubeflow';
 import FormSection from '~/app/pages/modelRegistry/components/pf-overrides/FormSection';
 import NamespaceSelector from '~/app/standalone/NamespaceSelector';
@@ -32,6 +32,7 @@ const RegisterAndStoreFields = <D extends RegistrationCommonFormData>({
       onSelect={handleNamespaceSelect}
       selectedNamespace={formData.namespace}
       isFullWidth
+      ignoreMandatoryNamespace
     />
   );
 
@@ -43,27 +44,9 @@ const RegisterAndStoreFields = <D extends RegistrationCommonFormData>({
 
       */}
       {isMUITheme ? (
-        <FormSection title="Project">{namespaceSelectorElement}</FormSection>
+        <FormSection title="Namespace">{namespaceSelectorElement}</FormSection>
       ) : (
-        <FormGroup
-          label="Project"
-          data-testid="namespace-form-group"
-          isRequired
-          labelHelp={
-            <Popover
-              data-testid="namespace-help-popover"
-              bodyContent="This list includes only projects that you and the selected model registry have permission to access. To request access to a new or existing project, contact your administrator."
-            >
-              <Button
-                variant="plain"
-                data-testid="namespace-help-button"
-                aria-label="More info on project field"
-                className="pf-v6-u-p-xs"
-                icon={<HelpIcon />}
-              />
-            </Popover>
-          }
-        >
+        <FormGroup label="Project" data-testid="namespace-form-group" isRequired>
           {namespaceSelectorElement}
         </FormGroup>
       )}
--- a/clients/ui/frontend/src/app/standalone/NamespaceSelector.tsx
+++ b/clients/ui/frontend/src/app/standalone/NamespaceSelector.tsx
@@ -10,6 +10,8 @@ interface NamespaceSelectorProps {
   selectedNamespace?: string;
   placeholderText?: string;
   isFullWidth?: boolean;
+  isGlobalSelector?: boolean;
+  ignoreMandatoryNamespace?: boolean;
 }
 
 const NamespaceSelector: React.FC<NamespaceSelectorProps> = ({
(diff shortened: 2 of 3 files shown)


clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelRegistryView/registerAndStoreFields.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.findRegisterToggleButton().click();
+  }
+
+  selectRegisterAndStoreMode() {
+    this.findRegisterAndStoreToggleButton().click();
+  }
+
+  shouldShowPlaceholder(placeholder = 'Select a namespace') {
+    this.findNamespaceSelector()
+      .find('span.pf-v6-c-menu-toggle__text')
```
Taj010: Would it be better to use `cy.findByText(placeholder)` here since its more reliable than a CSS selector

clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelRegistryView/registerAndStoreFields.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.findNamespaceSelector().click();
+    namespaces.forEach((namespace) => {
+      cy.findByRole('option', { name: namespace }).should('exist');
+    });
+    this.findNamespaceSelector().click();
+    return this;
+  }
+
+  shouldShowSelectedNamespace(name: string) {
+    this.findNamespaceSelector().find('span.pf-v6-c-menu-toggle__text').should('have.text', name);
```
Taj010: same here on using` cy.findByText(placeholder)..`

clients/ui/frontend/src/app/standalone/NamespaceSelector.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const NamespaceSelector: React.FC<NamespaceSelectorProps> = ({
+  placeholderText,
+  onSelect,
+  className,
+  isDisabled: externalDisabled,
+  selectedNamespace,
+  isFullWidth,
+}) => {
+  const { namespaces, preferredNamespace, updatePreferredNamespace } = useNamespaceSelector();
```
mturley: @ppadti I don't think we can use the `preferredNamespace` and `updatePreferredNamespace` from `useNamespaceSelector`. It hooks into the mod arch context and updates the global preferred namespace, which is also used for the namespace selector in the header of the whole app. I think we need to be using local state. However, we can still piggyback on the namespace data fetching that hook does, so I think we can just pull `namespaces` out of this hook and then have our own `useState` for which one is selected.

So since this component is reused, we probably either need a prop like `isStandalone` that makes it not update the preferred namespace, or we need to entirely lift `useNamespaceSelector` out and handle the differences of how different consumers update the state outside the component.

clients/ui/frontend/src/app/standalone/NamespaceSelector.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { config } = useModularArchContext();
+
+  // Check if mandatory namespace is configured
+  const isMandatoryNamespace = Boolean(config.mandatoryNamespace);
```
mturley, line 31: For this use case, we don't want to disable the selector if there's a mandatory namespace. Even if we aren't allowed to switch the whole app to focus on different namespaces, we should be able to choose any namespace we want for the purpose of copy jobs. I think maybe we need a new prop like `ignoreMandatoryNamespace` that we only use for this new use case.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterAndStoreFields.tsx
```diff
@@ -15,35 +19,79 @@ const RegisterAndStoreFields = <D extends RegistrationCommonFormData>({
...
-);
+  return (
+    <>
+      TODO job name field here
+      {/*
+      TODO use the K8sNameResourceField component here for the job name.
+
+      */}
+      {isMUITheme ? (
+        <FormSection title="Project">{namespaceSelectorElement}</FormSection>
```
mturley: We need to not use the Project terminology upstream, this is a Namespace.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterAndStoreFields.tsx
```diff
@@ -15,35 +19,79 @@ const RegisterAndStoreFields = <D extends RegistrationCommonFormData>({
...
+      ) : (
+        <FormGroup
+          label="Project"
+          data-testid="namespace-form-group"
+          isRequired
+          labelHelp={
+            <Popover
+              data-testid="namespace-help-popover"
+              bodyContent="This list includes only projects that you and the selected model registry have permission to access. To request access to a new or existing project, contact your administrator."
```
mturley: Regardless of upstream or downstream, we currently have no good way to check that a project/namespace has been given access to a regisry. We can remove this and just show this selector the same way in both themes for now. We'll open a followup story to implement this check and add the associated help text, because it needs some investigation.

clients/ui/frontend/src/app/standalone/NamespaceSelector.tsx
ppadti: yeah, but I am using `preferredNamespace` and  updating `updatePreferredNamespace` conditionally for the navbar selector. But forgot to update it with new prop and still using placeholder text for this condition, sorry. 

I will update it with a explicit prop, let me know if that's acceptable.

mturley: Oh I didn't notice that, ok. thanks, looking

mturley:
Thanks, that addresses my comments @ppadti !
/lgtm
/approve

PR #2167, opened 2026-01-29
Add model transfer jobs table with BFF mocks

manaswinidas:
## Description

- Add BFF layer with mock data for model transfer jobs table 
- Add route and kebab action leading to model transfer jobs table(behind feature flag)
- Add header, columns, filters, pagination and empty state for model transfer jobs table 
- Cypress tests haven't been added but some unit tests are tweaked (will be added in subsequent PRs)
- Tested midstream
Screenshot 2026-01-29 at 5 08 17 PM
Screenshot 2026-01-29 at 5 10 38 PM
Screenshot 2026-01-29 at 5 10 51 PM
Screenshot 2026-01-29 at 5 16 16 PM

## How Has This Been Tested?

Go to "View model transfer jobs" in the kebab action in registered models list and check whether you see the model transfer table with BFF mock data. Check the filters etc.

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/ModelTransferJobs.tsx
```diff
@@ -0,0 +1,51 @@
...
+      loadError={jobsLoadError}
+      loaded={jobsLoaded}
+      provideChildrenPadding
+      headerContent={
+        <ModelRegistrySelectorNavigator
+          getRedirectPath={(modelRegistryName) => modelTransferJobsUrl(modelRegistryName)}
+        />
+      }
+    >
+      <ModelTransferJobsListView jobs={jobs.items} isLoading={!jobsLoaded} />
```
mturley: I don't think we need an `isLoading` on this component since we already pass the same thing to `loaded` of ApplicationsPage, so we'll never render this with isLoading true.

clients/ui/frontend/src/app/types.ts
```diff
@@ -226,4 +226,104 @@ export type ModelRegistryAPIs = {
...
+};
+
+export type ModelTransferJobList = ModelRegistryListParams & { items: ModelTransferJob[] };
+
+export type GetListModelTransferJobs = (opts: APIOptions) => Promise<ModelTransferJobList>;
+
+export type GetModelTransferJobsByRegisteredModel = (
+  opts: APIOptions,
+  registeredModelId: string,
+) => Promise<ModelTransferJobList>;
```
mturley: It looks like this type isn't used. Is that something we should remove or is it intended to be implemented later?

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -2267,3 +2267,129 @@ func CreateCatalogSourcePreviewMockWithFilter(filterStatus string, pageSize int,
...
+			Bucket: "destination-bucket",
+			Key:    "models/my-model-3/v1",
+			Region: "us-west-2",
+		},
+		UploadIntent:             models.ModelTransferJobUploadIntentCreateVersion,
+		RegisteredModelId:        "1",
+		RegisteredModelName:      "my-model-3",
+		ModelVersionId:           "1",
+		ModelVersionName:         "v1",
+		Project:                  "project - 1",
```
mturley: We need to be using the concept of Namespace rather than Project upstream.

clients/ui/frontend/src/app/types.ts
```diff
@@ -226,4 +226,104 @@ export type ModelRegistryAPIs = {
...
+  modelVersionName?: string;
+  modelArtifactId?: string;
+  modelArtifactName?: string;
+  project?: string;
+  author?: string;
+  status: ModelTransferJobStatus;
+  createTimeSinceEpoch: string;
+  lastUpdateTimeSinceEpoch: string;
+  errorMessage?: string;
+};
```
mturley, line 320: The jobs here under the hood are going to be [k8s batch/v1 Job resources](https://kubernetes.io/docs/concepts/workloads/controllers/job/). Unless we plan to have the BFF translate object types, this should extend `K8sResourceCommon` and its structure will need to match examples like: https://github.com/kubeflow/model-registry/blob/main/jobs/async-upload/samples/sample_job_s3_to_oci.yaml

so we'd have id/name/description under metadata, and source/destination will be in secrets... hmm. I guess we don't want to be fetching the credential part of the secrets on the frontend though. So is the intention to indeed have a mapping of this type to the underlying k8s stuff within the BFF? that makes sense to me, but I just want to confirm the intent.

mturley (changes requested):
One small thing - we just need to use `namespace` instead of `project`. I also was going to ask for using k8s types directly here but realized that's probably a bad idea, so my other comment is just confirming your intent here.

PR #2168, opened 2026-01-29
feat: add model signer

jonburdo:
This implements OMS signing to produce a signal .sig file for any repo or collection of files.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/poetry.lock
+++ b/clients/python/poetry.lock
@@ -274,6 +274,19 @@ doc = ["doc8", "sphinx (>=7.0.0)", "sphinx-autobuild", "sphinx-autodoc-typehints
 test = ["dateparser (==1.*)", "pre-commit", "pytest", "pytest-cov", "pytest-mock", "pytz (==2021.1)", "simplejson (==3.*)"]
 
 [[package]]
+name = "asn1crypto"
+version = "1.5.1"
+description = "Fast ASN.1 parser and serializer with definitions for private keys, public keys, certificates, CRL, OCSP, CMS, PKCS#3, PKCS#7, PKCS#8, PKCS#12, PKCS#5, X.509 and TSP"
+optional = true
+python-versions = "*"
+groups = ["main"]
+markers = "extra == \"signing\""
+files = [
+    {file = "asn1crypto-1.5.1-py2.py3-none-any.whl", hash = "sha256:db4e40728b728508912cbb3d44f19ce188f218e9eba635821bb4b68564f8fd67"},
+    {file = "asn1crypto-1.5.1.tar.gz", hash = "sha256:13ae38502be632115abf8a24cbe5f4da52e3b5231990aff31123c805306ccb9c"},
+]
+
+[[package]]
 name = "async-timeout"
 version = "4.0.3"
 description = "Timeout context manager for asyncio programs"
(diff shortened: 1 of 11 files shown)


clients/python/poetry.lock
```diff
@@ -1120,6 +1487,44 @@ files = [
...
+    {file = "importlib_resources-5.13.0.tar.gz", hash = "sha256:82d5c6cca930697dbbd86c93333bb2c2e72861d4789a11c2662b933e5ad2b528"},
+]
+
+[package.extras]
+docs = ["furo", "jaraco.packaging (>=9.3)", "jaraco.tidelift (>=1.4)", "rst.linker (>=1.9)", "sphinx (>=3.5)", "sphinx-lint"]
+testing = ["pytest (>=6)", "pytest-black (>=0.3.7) ; platform_python_implementation != \"PyPy\"", "pytest-checkdocs (>=2.4)", "pytest-cov", "pytest-enabler (>=2.2)", "pytest-mypy (>=0.9.1) ; platform_python_implementation != \"PyPy\"", "pytest-ruff"]
+
+[[package]]
+name = "in-toto-attestation"
+version = "0.9.3"
```
tarilabs, line 1509: note: we should try to bump this in-toto-attestation to >1.0.0 if possible (I dont' see any lockfile for rh-image-signing)

clients/python/src/model_registry/signing/trust_manager.py
```diff
@@ -0,0 +1,354 @@
...
+            if "checkpointKeyId" in obj:
+                obj["logId"] = obj.pop("checkpointKeyId")
+            for key, value in obj.items():
+                obj[key] = TrustManager._transform_checkpoint_to_logid(value)
+        elif isinstance(obj, list):
+            for i, item in enumerate(obj):
+                obj[i] = TrustManager._transform_checkpoint_to_logid(item)
+        return obj
+
+    def _create_trust_config(
```
jonburdo, line 261: This generates trust config that will pass the pydantic model checks, similar to: https://github.com/securesign/model-transparency/blob/a182e42d91fb7bc9012c236ad46a37fe3ffb5ddf/generate-trust-config.sh

clients/python/pyproject.toml
```diff
@@ -27,11 +27,14 @@ nest-asyncio = "^1.6.0"
 huggingface-hub = { version = ">=0.20.1,<1.4.0", optional = true }
 olot = { version = "^0.1.6", optional = true }
 boto3 = { version = "^1.37.34", optional = true }
+rh-model-signing = { version = "0.0.3", optional = true }
```
jonburdo, line 30: Pinning the latest pypi release for the moment. We can simplify some of our code (i.e. our custom `sign_sigstore.Signer`) after the next release

tarilabs:
> Note the verify step will need to have the model.sig outside of the model dir, since it's not one of the files that needs to be checked itself.

Mmm this sounds like a bug.

iirc the model-signing should have ootb the following ignore:

```
       "ignore_paths": [
         ".gitattributes",
         ".github",
         "model.sig",
         ".gitignore",
         ".git"
       ]
```

Given a .sig you should be able to check the ootb ignores with:

```
cat model.sig | jq -r '.dsseEnvelope.payload' | base64 -d | jq '.predicate.serialization.ignore_paths'
```

or similar.

This however is a good reminder probably we should consider sign with `--ignore-paths ` with an empty dir of `.cache` where the model files are contained, as KServe creates that. Or, simply rely on the fact the Upstream operator has the additional flag now thanks to our friends at TAS: https://github.com/sigstore/model-validation-operator/blob/1ff8f893d33495d4424f64dcf3686d40e2f3ef18/api/v1alpha1/modelvalidation_types.go#L42

jonburdo:
> iirc the model-signing should have ootb the following ignore:

Looks like it's [just the git paths](https://github.com/securesign/model-transparency/blob/a182e42d91fb7bc9012c236ad46a37fe3ffb5ddf/src/model_signing/hashing.py#L270-L275)

I've fixed by adding `signature_path` as an absolute path to `config._hashing_config._ignored_paths` and tested that this works. If it's under `model_path` we'll get `"model.sig"` (or whatever `signature_path` relative to `model_path` is) added. If it's outside of `model_path` this will be ignored.

jonburdo:
And I've added `ignore_paths` which can be used to add `.cache`:
`sign(..., ignore_paths=[".cache"])`

Should be set to review now :)

clients/python/poetry.lock
jonburdo: For the moment with `rh-model-signing==0.0.3`, I think we're stuck with this

jonburdo:
Note: `poetry.lock` accounts for a chunk of the changes.

The main entrypoint here is `ModelSigner` in:
`clients/python/src/model_registry/signing/model_signer.py`

Which is supported by:
`clients/python/src/model_registry/signing/trust_manager.py`

Fyi:
```
$ git diff main --stat -- ':!clients/python/poetry.lock'
 clients/python/noxfile.py                                  |   4 +-
 clients/python/pyproject.toml                              |   3 ++
 clients/python/src/model_registry/signing/__init__.py      |  23 +++++++-
 clients/python/src/model_registry/signing/exceptions.py    |  16 +++---
 clients/python/src/model_registry/signing/model_signer.py  | 370 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 clients/python/src/model_registry/signing/sign_sigstore.py | 107 ++++++++++++++++++++++++++++++++++++
 clients/python/src/model_registry/signing/token.py         |  47 ++++++++++++++++
 clients/python/src/model_registry/signing/trust_manager.py | 354 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 clients/python/tests/test_signing.py                       |  80 +++++++++++++++++++++++++++
 clients/python/tests/test_token.py                         | 106 ++++++++++++++++++++++++++++++++++++
 10 files changed, 1100 insertions(+), 10 deletions(-)
```

clients/python/src/model_registry/signing/sign_sigstore.py
```diff
@@ -0,0 +1,107 @@
...
+import pathlib
+
+from model_signing._signing.sign_sigstore import _DEFAULT_CLIENT_ID, _DEFAULT_CLIENT_SECRET
+from model_signing._signing.sign_sigstore import Signer as BaseSigner
+from sigstore import models as sigstore_models
+from sigstore import oidc as sigstore_oidc
+from sigstore import sign as sigstore_signer
+
+
+class Signer(BaseSigner):
```
jonburdo, line 13: Modified `__init__` and `_get_identity_token` slightly to apply this change: https://github.com/securesign/model-transparency/pull/35/changes/4e6f879b50b074a1bfc983b10c87441ff817943e

This is needed when verifying with an already produced token (i.e. service account token) to avoid attempting to hit the OIDC endpoints again.

Once the next version of `rh-model-signing` is released we can remove this entire file and stick with the ootb sigstore `Signer`

clients/python/src/model_registry/signing/token.py
```diff
@@ -0,0 +1,47 @@
...
+        token: JWT token string in format header.payload.signature
+
+    Returns:
+        Dictionary of claims from the token payload
+
+    Raises:
+        ValueError: If token format is invalid or decoding fails
+    """
+    try:
+        payload = token.split(".")[1]
```
fege: is this robust enough ? the hardcoded position is what I am referring to

clients/python/src/model_registry/signing/trust_manager.py
```diff
@@ -0,0 +1,354 @@
...
+        except Exception as e:
+            if isinstance(e, (InitializationError, FileExistsError)):
+                raise
+            msg = f"Trust initialization failed: {e}"
+            raise InitializationError(msg) from e
+
+    def _try_direct_trust_updater(self, tuf_url: str) -> dict | None:
+        """Try TrustUpdater without bootstrap (auto-detect mode only)."""
+        try:
+            from sigstore._internal.tuf import TrustUpdater
```
fege: should we move to the top of the file?

clients/python/src/model_registry/signing/trust_manager.py
```diff
@@ -0,0 +1,354 @@
...
+        else:
+            logger.info("No checksum provided (skipping validation)")
+
+    def _bootstrap_trust_updater(self, tuf_url: str, root_data: bytes) -> dict:
+        """Bootstrap TrustUpdater with root.json and return trusted_root."""
+        with tempfile.TemporaryDirectory() as tmpdir:
+            bootstrap_root = Path(tmpdir) / "root.json"
+            bootstrap_root.write_bytes(root_data)
+
+            from sigstore._internal.tuf import TrustUpdater
```
fege: same as above

clients/python/tests/test_signing.py
```diff
@@ -346,3 +346,83 @@ def test_verify_raises_verification_error_on_command_failure(self, mocker):
...
                 certificate_identity="https://kubernetes.io/namespaces/test/serviceaccounts/default"
             )
+
+
+class TestModelSigner:
+    """Test ModelSigner class."""
+
+    def test_get_trust_config_path_with_tuf_url(self, tmp_path):
+        """Test get_trust_config_path returns correct path when tuf_url is set."""
+        from model_registry.signing import ModelSigner
```
fege: move at the top of the file or at least in the class, it is repeated in several tests

clients/python/tests/test_signing.py
```diff
@@ -346,3 +346,83 @@ def test_verify_raises_verification_error_on_command_failure(self, mocker):
...
+            tuf_url="https://tuf.example.com",
+            cache_dir=tmp_path
+        )
+
+        result = signer.get_trust_config_path()
+
+        # Should return path within cache_dir
+        assert result.parent.parent == tmp_path
+        assert result.name == "trust_config.json"
+        assert "https%3A%2F%2Ftuf.example.com" in str(result)
```
fege, line 373: isn't enough to check the presence of `tuf.example.com` ?

clients/python/tests/test_signing.py
```diff
@@ -346,3 +346,83 @@ def test_verify_raises_verification_error_on_command_failure(self, mocker):
...
+    def test_ensure_trust_initialized_creates_config_if_missing(self, tmp_path, mocker):
+        """Test _ensure_trust_initialized calls initialize if config doesn't exist."""
+        from model_registry.signing import ModelSigner
+
+        signer = ModelSigner(
+            tuf_url="https://tuf.example.com",
+            fulcio_url="https://fulcio.example.com",
+            rekor_url="https://rekor.example.com",
+            tsa_url="https://tsa.example.com",
+            oidc_issuer="https://oidc.example.com",
```
fege: they can be const,  `tuf_url` is also used in different tests

clients/python/src/model_registry/signing/token.py
tarilabs: > A JWT is represented as the concatenation of the Encoded JWT Header, the JWT Second Part, and the JWT Third Part, in that order, with the parts being separated by period ('.') characters.

https://openid.net/specs/draft-jones-json-web-token-07.html#:~:text=A%20JWT%20is%20represented%20as%20the,JWE%20used%20to%20represent%20the%20JWT.

clients/python/tests/test_signing.py
```diff
@@ -346,3 +346,83 @@ def test_verify_raises_verification_error_on_command_failure(self, mocker):
...
+        config_path.write_text('{"test": "config"}')
+
+        # Mock initialize method
+        mock_initialize = mocker.patch.object(signer, "initialize")
+
+        # Call _ensure_trust_initialized
+        signer._ensure_trust_initialized()
+
+        # Should NOT have called initialize
+        mock_initialize.assert_not_called()
```
tarilabs, line 427: thank you

clients/python/src/model_registry/signing/model_signer.py
```diff
@@ -0,0 +1,370 @@
...
+        """Verify a signed model.
+
+        Args:
+            model_path: Path to model directory
+            signature_path: Path to signature file (default: model_path/signature_filename)
+            certificate_identity: Expected certificate identity (uses instance default if not provided)
+            oidc_issuer: OIDC issuer (uses instance default if not provided)
+
+        Returns:
+            True if verification passed, False otherwise
```
tarilabs: I don't think it ever return `False` , but only raises if signature does not pass verification with an error either here or the underlying signer?

clients/python/src/model_registry/signing/sign_sigstore.py
tarilabs: thank you for making this extra effort!

clients/python/src/model_registry/signing/token.py
fege: thanks @tarilabs

clients/python/src/model_registry/signing/model_signer.py
```diff
@@ -0,0 +1,370 @@
...
+
+class ModelSigner:
+    """Tool for signing and verifying ML models with Sigstore."""
+
+    def __init__(
+        self,
+        tuf_url: str | None = None,
+        root: str | None = None,
+        root_checksum: str | None = None,
+        identity_token_path: PathLike | None = None,
```
dbasunag, line 35: Do we need a way to pass the token of the current user that is going to do the signing? Thinking in the context of an openshift cluster.

jonburdo: Yes, good call out. I'm thinking to do this in a followup. We need to provide an easy auto-config to read from env vars and default to getting the token from `/var/run/secrets/kubernetes.io/serviceaccount/token`.

We do this in the model registry client: https://github.com/kubeflow/model-registry/blob/717d70a680835deebe311e3444e53d6e0e136b9a/clients/python/src/model_registry/_client.py#L74

clients/python/tests/test_signing.py
jonburdo: We're url-quoting so that we can create a single file-system safe directory name for a local cache similar to what sigstore-python does here: https://github.com/sigstore/sigstore-python/blob/664693f74814f158a8402b80244ebea0a4f3d357/sigstore/_internal/tuf.py#L50

So we're testing that this quoting works right too

jonburdo (commented):
Thanks for the reviews @tarilabs @fege @dbasunag ! Much appreciated

Latest changes: https://github.com/kubeflow/model-registry/compare/cf75eaf5f9cb5ca7ab8164534ac59d3ad9472dfe..8eb867f7a0600b206eab1c66dd26692a4cc155df

clients/python/src/model_registry/signing/model_signer.py
dbasunag: From test side, we need to be able to pass user token and turn off certificate validations as at one point we may want to run these tests against live clusters for shift left.

jonburdo: For testing with any token, the token can be written to a temp file and that filepath passed in.

I'm not sure about certificate validations.
1. For `cosign` binary I see there's an option `cosign sign --allow-insecure-registry <image-reference>`
2. For `sigstore-python` (used by `rh-model-signing`) I don't see an option right off
Ideally it seems like we should be testing on clusters with valid certificates, though I may be missing context here

tarilabs: I concur with Jon here, we're just integrating Sigstore tooling (cosign, model-signing) so we can only deal with what those tooling supports.

clients/python/src/model_registry/signing/token.py
```diff
@@ -0,0 +1,59 @@
...
+        the Encoded JWT Header, the JWT Second Part, and the JWT Third Part, in that
+        order, with the parts being separated by period ('.') characters."
+        https://openid.net/specs/draft-jones-json-web-token-07.html
+    """
+    try:
+        # Validate JWT has exactly 3 parts (header.payload.signature)
+        parts = token.split(".")
+        if len(parts) != 3:
+            msg = f"Invalid JWT format: expected 3 parts separated by '.', got {len(parts)}"
+            raise ValueError(msg)
```
tarilabs, line 30: 😆 I appreciate the extra mile at the risk of this looking a bit of "over-defensive" function

clients/python/src/model_registry/signing/model_signer.py
```diff
@@ -0,0 +1,368 @@
...
+            msg = f"Signing failed: {e}"
+            raise SigningError(msg) from e
+
+    def verify(  # noqa: C901
+        self,
+        model_path: PathLike,
+        signature_path: PathLike | None = None,
+        certificate_identity: str | None = None,
+        oidc_issuer: str | None = None,
+    ) -> None:
```
tarilabs, line 294: thanks for this update based on actual function behaviour.
(and py docs below)

tarilabs (commented):
/lgtm

thanks a lot @jonburdo for the updates, I leave it for the other reviewers to chime in

clients/python/src/model_registry/signing/token.py
jonburdo: It was an easy change, and I do like to have more useful error messages (i.e. users understands "oh my jwt is invalid" rather than "huh a random IndexError")

jonburdo:
I believe I addressed @fege comments, but please comment if you see anything I missed, and I can address in a followup!

With the "lgtm" and approval from @tarilabs and @dbasunag I've done ahead and self-approved triggering the merge. Thanks everyone!

PR #2183, opened 2026-02-03
Add envtest mocks and client logic for model transfer job

ppadti:
## Description
This is the first pass to unblock UI work.
Adds CRUD endpoints for model transfer jobs in mock mode. Follow-up PR will add the actual K8s job creation logic.

## How Has This Been Tested?
Go to model registry and open view model transfer job page and see the jobs table

curl cmd:
GET:

```
curl -X GET "http://localhost:4000/api/v1/model_registry/model-registry/model_transfer_jobs?namespace=kubeflow" \
  -H "Content-Type: application/json" \
  -H "kubeflow-userid: user@example.com" 
```

CREATE:

```
curl -X POST "http://localhost:4000/api/v1/model_registry/model-registry/model_transfer_jobs?namespace=bella-namespace" \
  -H "Content-Type: application/json" \
  -H "kubeflow-userid: user@example.com" \
  -d '{"data": {"name": "my-job", "source": {"type": "s3"}, "destination": {"type": "oci"}, "uploadIntent": "create_model"}}'

```

PATCH

```
curl -X PATCH "http://localhost:4000/api/v1/model_registry/model-registry/model_transfer_jobs/transfer-job-002?namespace=bella-namespace" \
  -H "Content-Type: application/json" \
  -H "kubeflow-userid: user@example.com" \
  -d '{"description": "Updated"}'
```

DELETE

```
curl -X DELETE "http://localhost:4000/api/v1/model_registry/model-registry/model_transfer_jobs/transfer-job-001?namespace=bella-namespace" \
  -H "kubeflow-userid: user@example.com"
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/api/model_transfer_job_handler.go
+++ b/clients/ui/bff/internal/api/model_transfer_job_handler.go
@@ -45,7 +45,7 @@ func (app *App) GetAllModelTransferJobsHandler(w http.ResponseWriter, r *http.Re
 	}
 }
 
-func (app *App) CreateModelTransferJobHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+func (app *App) CreateModelTransferJobHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	ctx := r.Context()
 
 	namespace, ok := ctx.Value(constants.NamespaceHeaderParameterKey).(string)
@@ -124,7 +124,7 @@ func (app *App) UpdateModelTransferJobHandler(w http.ResponseWriter, r *http.Req
 	// 	Data: updatedModelTransferJob,
 	// }
 
-	err = app.WriteJSON(w, http.StatusCreated, map[string]string{"status": "updated"}, nil)
+	err = app.WriteJSON(w, http.StatusOK, map[string]string{"status": "updated"}, nil)
 	if err != nil {
 		app.serverErrorResponse(w, r, fmt.Errorf("error writing JSON"))
 		return
--- a/clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
+++ b/clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
@@ -357,6 +357,9 @@ func (kc *SharedClientLogic) CreateModelTransferJob(ctx context.Context, namespa
 
 	_, err := kc.Client.BatchV1().Jobs(namespace).Create(ctx, job, metav1.CreateOptions{})
 
+	// TODO: After creating the Job, patch ConfigMap and Secrets to add ownerReferences
+	// pointing to the Job's UID so they are garbage collected when the Job is deleted.
+
 	if err != nil {
 		sessionLogger.Error("failed to create job",
 			"namespace", namespace,
@@ -387,7 +390,7 @@ func (kc *SharedClientLogic) UpdateModelTransferJob(ctx context.Context, namespa
 		return err
 	}
 
-	// TODO: add logic to construct the job to update the model transfer job
+	// TODO: Add logic to construct the job to update the model transfer job
 
 	_, err = kc.Client.BatchV1().Jobs(namespace).Update(ctx, modelTransferJob, metav1.UpdateOptions{})
 
--- a/clients/ui/bff/internal/api/model_catalog_settings_handler.go
+++ b/clients/ui/bff/internal/api/model_catalog_settings_handler.go
@@ -88,7 +88,7 @@ func (app *App) GetCatalogSourceConfigHandler(w http.ResponseWriter, r *http.Req
 
 }
 
-func (app *App) CreateCatalogSourceConfigHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+func (app *App) CreateCatalogSourceConfigHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	ctx := r.Context()
 
 	namespace, ok := ctx.Value(constants.NamespaceHeaderParameterKey).(string)


clients/ui/api/openapi/mod-arch.yaml
```diff
@@ -2405,6 +2517,124 @@ components:
...
+      description: The type of source for a transfer job.
+      type: string
+      enum:
+        - s3
+        - uri
+    ModelTransferJobDestinationType:
+      description: The type of destination for a transfer job.
+      type: string
+      enum:
+        - oci
```
ppadti, line 2530: Just wanted to double check this here - we will have only s3 and uri as source and oci will be the destination ? 
because I wanted to updated the types being added in model_transfer_job.go file

mturley: @ppadti correct, for now those are the only supported sources/destinations. I believe the spec technically allows us to specify more but they aren't supported yet and we won't expose them in the UI.

clients/ui/bff/internal/api/model_transfer_job_handler.go
```diff
@@ -42,3 +44,125 @@ func (app *App) GetAllModelTransferJobsHandler(w http.ResponseWriter, r *http.Re
...
+		app.serverErrorResponse(w, r, err)
+		return
+	}
+
+	// TODO: uncomment the following when we implement the actual logic
+	// modelTransferJob := ModelTransferJobEnvelope{
+	// 	Data: updatedModelTransferJob,
+	// }
+
+	err = app.WriteJSON(w, http.StatusCreated, map[string]string{"status": "updated"}, nil)
```
mturley: I think for a PATCH you want `http.StatusOK` (200) instead of `StatusCreated` (201).

clients/ui/api/openapi/mod-arch.yaml
mturley: I think it's ok to add these extra types in the bff and we can just not use them from the UI yet. either way works.

clients/ui/bff/internal/repositories/model_registry.go
```diff
@@ -134,11 +136,120 @@ func (m *ModelRegistryRepository) ResolveServerAddress(clusterIP string, httpPor
...
+func (m *ModelRegistryRepository) DeleteModelTransferJob(ctx context.Context, client k8s.KubernetesClientInterface, namespace string, jobId string) error {
+	err := client.DeleteModelTransferJob(ctx, namespace, jobId)
+	if err != nil {
+		return fmt.Errorf("failed to delete model transfer job %s: %w", jobId, err)
+	}
+	return nil
+}
+
+// TODO: These functions convert the minimum required fields for now. Improve these to convert all the necessary fields
+func convertModelToK8sJob(payload models.ModelTransferJob, namespace string) *batchv1.Job {
```
mturley, line 187: Claude is telling me this is incomplete, but I imagine that's intentional based on the comment. Just copying its comment here for my own future reference so we can make sure we get them all later.

```
Incomplete conversion in convertK8sJobToModel
model_registry.go:198

The function doesn't populate:

Source and Destination fields
UploadIntent field
LastUpdateTimeSinceEpoch field
ErrorMessage from job conditions

Example fix for UploadIntent:
UploadIntent: models.ModelTransferJobUploadIntent(annotations["modelregistry.kubeflow.org/upload-intent"]),
```

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
```diff
@@ -324,3 +324,153 @@ func (kc *SharedClientLogic) UpdateCatalogSourceConfig(
...
+			"namespace", namespace,
+			"error", err,
+		)
+		return &batchv1.JobList{}, fmt.Errorf("failed to list model transfer job: %w", err)
+	}
+
+	return modelTransferJobList, nil
+}
+
+func (kc *SharedClientLogic) CreateModelTransferJob(ctx context.Context, namespace string, job *batchv1.Job) error {
```
mturley, line 355: We don't necessarily need to implement it in this PR, but we will need the client to attach ownerReferences pointing to the new job onto the dependent resources (configmap, secrets) or they will be orphaned when the job is deleted. I'm not sure where the best place would be for that, but can we include a TODO comment somewhere? I think the workflow will need to be: create secrets and configmap, create job, patch secrets/configmap to add an ownerReference to the job's uid, all as part of the handler of one frontend request.

clients/ui/bff/internal/api/model_transfer_job_handler.go
```diff
@@ -42,3 +44,125 @@ func (app *App) GetAllModelTransferJobsHandler(w http.ResponseWriter, r *http.Re
 		app.serverErrorResponse(w, r, err)
 	}
 }
+
+func (app *App) CreateModelTransferJobHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
mturley: Looks like the `ps httprouter.Params` param here is unused, not sure if that's a problem. We could change it to _ like in `GetAllModelTransferJobsHandler`.

clients/ui/bff/internal/repositories/model_registry.go
ppadti: yeah, thats right. I am just passing the minimum required fields for the table for now - This function will be updated later in the follow up PR.

clients/ui/bff/internal/integrations/kubernetes/shared_k8s_client.go
ppadti: yeah, i just created the job to make it work in mock mode, I will add a TODO so that we don't forget.

PR #2188, opened 2026-02-04
Always pass perf filters to card artifacts call for correct benchmark counts regardless of toggle state

manaswinidas:
## Description

Now the benchmark counts are same whether the performance filters toggle is ON or OFF

Added a Cypress test and edited mocks 

## How Has This Been Tested?

Tested midstream - On the model catalog landing page cards, benchmark counts on “View N benchmarks” links should be the same both without the toggle enabled and with the toggle enabled (while filters are set to default), and be the same as the number of benchmarks as seen on the details page with default filters applied.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogPerformanceFiltersApi.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalogSettings/modelCatalogPerformanceFiltersApi.cy.ts
@@ -177,6 +177,7 @@ describe('Model Catalog Performance Filters API Behavior', () => {
         // If latency filter is active, verify short property key format (e.g., 'ttft_p90')
         if (url.includes('latencyProperty=')) {
           const latencyMatch = url.match(/latencyProperty=([^&]+)/);
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
           expect(latencyMatch).to.not.be.null;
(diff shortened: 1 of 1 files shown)


ppadti:
I tested this locally, but I am not seeing the the default perf filters being passed on performance endpoint on load.  I think Its because on load, filterData in context is empty so no filterQuery is being passed. 

Screenshot 2026-02-05 at 2 26 11 PM

mturley (changes requested):
+1 to @ppadti 's comment - right now the default perf filter values don't end up in state until the first time the perf view toggle is turned on. we might just want to change that behavior so we put those perf filter defaults in state as soon as the page loads. That should be safe now that we have excluded those elsewhere when the toggle is off.

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -284,6 +284,13 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
     [baseSetFilterData, isOnDetailsPage],
   );
 
+  // Apply default performance filters when filterOptions first loads
+  React.useEffect(() => {
+    if (filterOptionsLoaded && filterOptions?.namedQueries) {
+      resetPerformanceFiltersToDefaults();
+    }
```
mturley: Hm, I wonder if we should only have this trigger if there are no perf filters in state (or maybe if at least one of the filters present in namedQueries has no value in state) so we don't accidentally reset them to defaults when filter options are fetched again

PR #2194, opened 2026-02-04
Implement the Hardware Configuration  table to sort by ascending order

Taj010:
## Description

Fixed the current default sorting in Performance Insights. The default order of records in the Hardware Configuration table is now sorted in ASC order by the selected Metric & Percentile. 

## How Has This Been Tested?

This has been manually tested by checking if the selected metric values are sorted by ascending order. 

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -31,6 +32,33 @@ const HardwareConfigurationTable: React.FC<HardwareConfigurationTableProps> = ({
...
+  const sortedData = React.useMemo(() => {
+    if (!sortColumn) {
+      return performanceArtifacts;
+    }
+    const sortFn = sortColumn.sortable;
+    if (typeof sortFn !== 'function') {
+      return performanceArtifacts;
+    }
+    return [...performanceArtifacts].toSorted((a, b) => sortFn(a, b, sortColumn.field));
+  }, [sortColumn, performanceArtifacts]);
```
YuliaKrimerman: The Table component already has sorting built-in via defaultSortColumn. Why are we pre-sorting the data AND setting defaultSortColumn? This feels like we're doing the work twice. Can we just rely on the Table's sorting with defaultSortColumn={sortIndex} and not pre-sort?

clients/ui/frontend/src/app/pages/modelCatalog/components/useHardwareConfigColumns.ts
```diff
@@ -96,33 +119,59 @@ export const useHardwareConfigColumns = (
...
+    if (newIndex !== -1) {
+      setSortIndex(newIndex);
+      setSortDirection('asc');
     }
+  }, [activeLatencyField, manageColumnsResult, stickyColumns, manageableColumns]);
 
-    // Add the matching TPS column (if not already present)
-    if (!newVisibleIds.includes(matchingTpsPropertyKey)) {
-      newVisibleIds.push(matchingTpsPropertyKey);
+  // Ensure sort is set correctly when columns are ready (handles initial mount case)
```
Taj010, line 137: I ended up using two useEffect hooks to keep the sorting in sync. The first one handles the immediate visibility change when a filter is picked, and the second one acts as a 'safety check' to make sure the sortIndex actually points to the right column after the table re-renders.

I tried combining them, but it kept breaking the sort during transitions. This is the best I could come up with, but let me know if you think there's a cleaner way to handle the state dependency

mturley: @Taj010 what we might want to do is, instead of keeping sortIndex in state, keep a sortColumnField in state (referring to the `field` of a column from the columns array, which serves as a column identifier). We'd have to translate that to an index when passing it into the table (by looking it up in the filtered/reordered columns array with a [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)), and we'd have to translate it back in the callback when the user changes the sort (by grabbing the column by index and putting its `field` in our state). but that way, no matter what happens with columns being reordered, the sort will naturally stay accurate as things re-render. Indexes in state when order can change is a minefield for bugs, let's avoid it.

Honestly, it would be better if the table itself used column ids instead of numeric indexes, but we can still use them as the source of truth as long as we have this sort of translation layer.

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
mturley: The pre-sorting was removed, but as for not using `defaultSortColumn` I think the useEffect approach for the initial sort column makes sense since the filter state might still be initializing and we might not know the latency field we want to sort by. So we need the effect either way, and it doesn't really matter if we also set defaultSortColumn on the off chance we already know the field we need, since the effect will run anyway.

mturley:
Awesome!! Thanks @Taj010 , LGTM.

/lgtm
/approve

PR #2202, opened 2026-02-06
Some more changes to ensure tests runs from testops image in midstream

dbasunag:
## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/clients/python/tests/conftest.py
+++ b/catalog/clients/python/tests/conftest.py
@@ -113,7 +113,7 @@ _MAX_PARENT_LEVELS = 10
 
 
 @pytest.fixture(scope="session")
-def root(request) -> Path | None:
+def root(request) -> Path:
     """Get repository root directory.
 
     Navigates up from catalog/clients/python to find the repo root.
@@ -291,7 +291,7 @@ def local_testdata_dir() -> Path:
 
 
 @pytest.fixture(scope="session")
-def test_catalog_data(pytestconfig: pytest.Config) -> dict | None:
+def test_catalog_data(pytestconfig: pytest.Config) -> dict:
     """Load test catalog data used by E2E tests.
 
     Returns:
--- a/catalog/clients/python/tests/test_q_search.py
+++ b/catalog/clients/python/tests/test_q_search.py
@@ -115,6 +115,7 @@ class TestModelSearch:
         assert "items" in response
 
         if kind_cluster:
+            # Validate against test data only for Kind clusters
             is_valid, errors = validate_search_results_against_test_data(
                 api_response=response,
                 search_term=search_term,


catalog/clients/python/tests/conftest.py
```diff
@@ -122,18 +122,19 @@ def root(request) -> Path:
...
-        if (current / ".git").exists():
-            return current
-        current = current.parent
-    # Fail explicitly if repo root not found
-    msg = (
-        f"Could not find repository root (.git directory) starting from "
-        f"{request.config.rootpath}. Searched {_MAX_PARENT_LEVELS} levels up."
-    )
-    raise RuntimeError(msg)
+    if kind_cluster:
```
jonburdo: I think we really don't want to change this fixture if possible, but if we do, we should ideally still have all paths lead to an explicit `return something` (i.e. `return None` at the bottom outside the if-statement).

Or it could be better (just my opinion here) to simply add at the top of the function body:
```
if not kind cluster:
    return None
```

dbasunag: Basically in the image we added for the testops in midstream, there is no .git. So whichever test (all the q_search ones) uses this fixture, they fail when run from the container image. I can definitely change it to `if not kind_cluster`, if you want.

catalog/clients/python/tests/conftest.py
```diff
@@ -290,14 +291,15 @@ def local_testdata_dir() -> Path:
...
     """
-    test_catalog_path = (
-        root / "manifests" / "kustomize" / "options" / "catalog" / "overlays" / "e2e" / "test-catalog.yaml"
-    )
-    with open(test_catalog_path) as f:
-        return yaml.safe_load(f)
+    if kind_cluster:
+        test_catalog_path = (
+            Path(pytestconfig.rootpath) / "../../.." / "manifests" / "kustomize" / "options" / "catalog" / "overlays" / "e2e" / "test-catalog.yaml"
+        )
```
lugi0: This seems fairly prone to breaking - can we implement it in a somewhat more flexible way? A couple suggestions from claude are:
```
# search for file
catalog_files = list(root.rglob("**/e2e/test-catalog.yaml"))
# handle more than one found using the expected path, or use the one found
```

or 
```
# Allow override via environment variable
catalog_path_env = os.getenv("TEST_CATALOG_PATH")
# if not envvar, use expected path
```

catalog/clients/python/tests/conftest.py
```diff
@@ -290,14 +291,15 @@ def local_testdata_dir() -> Path:
...
 
     Returns:
         Dictionary containing the test catalog YAML data.
     """
-    test_catalog_path = (
-        root / "manifests" / "kustomize" / "options" / "catalog" / "overlays" / "e2e" / "test-catalog.yaml"
-    )
-    with open(test_catalog_path) as f:
-        return yaml.safe_load(f)
+    if kind_cluster:
```
lugi0: how do we handle the case in which this is not a kind cluster? this function is a no-op in that case, if this is called only in kind envs the initial if is not needed

dbasunag: in non-kind cluster, the test data does not apply, instead, they work with the real catalog data on the cluster.

catalog/clients/python/tests/conftest.py
dbasunag: Unless we move tests around, it should not break. I did see similar pattern in multiple places in the repo, so any change in path would require them all to be addressed (e.g. applying the scripts, manifests etc.). I would let @fege decide a path forward here.

catalog/clients/python/tests/conftest.py
lugi0: Then do we need this if clause?

dbasunag: I can definitely remove it and skip the validation in the test instead. Let me do that instead.

catalog/clients/python/tests/conftest.py
fege: even if I wrote the initial Path I agree with @lugi0 , maybe we can fix this now and open a followup if there are other places to update

catalog/clients/python/tests/conftest.py
```diff
@@ -290,14 +291,14 @@ def local_testdata_dir() -> Path:
...
-def test_catalog_data(root: Path) -> dict:
+def test_catalog_data(pytestconfig: pytest.Config) -> dict | None:
     """Load test catalog data used by E2E tests.
 
     Returns:
         Dictionary containing the test catalog YAML data.
     """
     test_catalog_path = (
-        root / "manifests" / "kustomize" / "options" / "catalog" / "overlays" / "e2e" / "test-catalog.yaml"
+        Path(pytestconfig.rootpath) / "../../.." / "manifests" / "kustomize" / "options" / "catalog" / "overlays" / "e2e" / "test-catalog.yaml"
```
jonburdo: Imo this is perfect for now.

Note for future iteration:
If we do want to make this more robust, my suggestion would be to set this var the same way, but:
1. Make `test_catalog_path` its own fixture, constant, or function
2. (maybe) If we want to guard against rearranging the repo, do this by adding an assertion that we are in the subdirectories we assume we are in (i.e. that `pytestconfig.rootpath` ends with components `catalog/clients/python/tests`) 
3. Call `Path.resolve(strict=True)` on this. This will normalize `/repo/root/a/b/c/../../../x/y` into `/repo/root/x/y`, and the `strict=True` will ensure it exists (giving us a fast fail).
4. nit: The `Path()` call can just be followed by a single string containing the rest:

```suggestion
        Path(pytestconfig.rootpath) / "../../../manifests/kustomize/options/catalog/overlays/e2e/test-catalog.yaml"
```

catalog/clients/python/tests/conftest.py
```diff
@@ -113,7 +113,7 @@ def pytest_report_teststatus(report, config):
 
 
 @pytest.fixture(scope="session")
-def root(request) -> Path:
+def root(request) -> Path | None:
```
jonburdo: Should still just be `Path` I think? We're either returning a `Path` or raising an exception

catalog/clients/python/tests/test_q_search.py
```diff
@@ -108,18 +108,20 @@ def test_q_parameter_no_results(
...
-            api_response=response,
-            search_term=search_term,
-            catalog_data=test_catalog_data,
-        )
+        if kind_cluster:
+            is_valid, errors = validate_search_results_against_test_data(
+                api_response=response,
+                search_term=search_term,
+                catalog_data=test_catalog_data,
+            )
```
jonburdo, line 123: When skipping part of a test, maybe a comment mentioning why would be worthwhile

dbasunag: Sure. I will do that in other places (in this file) too.

jonburdo (approved):
/lgtm
/approve

PR #2211, opened 2026-02-10
feat(signing): add Signer entrypoint and env var support

jonburdo:
This adds a single unified class for signing operations. This includes signing and verifying for both models and container images.

This also adds configuration via environment variables so users can simply call `Signer()` instead of passing all of the arguments explicitly.

## Description

This makes it easy to configure via environment variables and then simply:
```python
from model_registry.signing import Signer

signer = Signer()

signer.sign_model(model_dir)
signer.verify_model(model_dir)

signer.sign_image(image_uri)
signer.verify_image(image_uri)
```

The env vars you would set if not passing explicit args:
```
SIGNING_TUF_URL=https://tuf.example.com
SIGNING_FULCIO_URL=https://fulcio.example.com.com
SIGNING_REKOR_URL=https://rekor.example.com.com
SIGNING_TSA_URL=https://tsa.example.com.com
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/signing/config.py
+++ b/clients/python/src/model_registry/signing/config.py
@@ -5,26 +5,30 @@ from __future__ import annotations
 import os
 from collections.abc import Callable, Iterable, Sequence
 from pathlib import Path
-from typing import TypeAlias, TypeVar, overload
+from typing import Final, TypeAlias, TypeVar, overload
 
 from pydantic import BaseModel
 from typing_extensions import Self
+
+from .token import decode_jwt_payload, extract_client_id, k8s_sub_to_certificate_identity
 
 PathLike: TypeAlias = str | os.PathLike[str]
 
 T = TypeVar("T")
 U = TypeVar("U")
 
+DEFAULT_IGNORE_PATHS: Final[tuple[PathLike]] = (".cache",)
+
 
 class SigningConfig(BaseModel):
     """Configuration for model and image signing operations.
 
-    Loads configuration from environment variables with SIGNING_ prefix.
+    Loads configuration from environment variables with SIGSTORE_ prefix.
     All fields are optional to allow flexible configuration.
 
     Example:
-        >>> # Load from environment variables
-        >>> config = SigningConfig.from_env()
+        >>> # Create with intelligent defaults and env var fallback
+        >>> config = SigningConfig.create()
         >>> # Or provide values directly
         >>> config = SigningConfig(
         ...     tuf_url="https://tuf.sigstore.dev",
@@ -48,7 +52,7 @@ class SigningConfig(BaseModel):
     ignore_paths: Sequence[Path] | None = None
 
     @classmethod
-    def from_env(
+    def create(  # noqa: C901
         cls,
         tuf_url: str | None = None,
         root_url: str | None = None,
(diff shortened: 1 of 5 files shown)


clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -178,14 +208,21 @@ def sign(  # noqa: C901
         cmd = ["cosign", "sign", "-y"]
 
         if identity_token_path is not None:
-            cmd.extend(["--identity-token", os.fspath(identity_token_path)])
+            # Read token content from file (cosign doesn't actually accept file paths despite docs)
+            with open(identity_token_path) as f:
+                token_content = f.read().strip()
+            cmd.extend(["--identity-token", token_content])
```
jonburdo, line 216: I thought this worked when I tested it before by simply passing the path, but now it doesn't seem to be working despite the docs:
```
$ cosign sign -h | grep -A 2 -- --identity-token
    --identity-token='':
	identity token to use for certificate from fulcio. the token or a path to a file containing the token is accepted.
```

tarilabs: could it be the case of the SIGNING_ prefix?

clients/python/src/model_registry/signing/config.py
```diff
@@ -0,0 +1,240 @@
...
+            ignore_paths: Paths to ignore during signing
+
+        Returns:
+            SigningConfig instance with resolved values
+        """
+        # Resolve 4 URLs from env vars
+        resolved_tuf_url = resolve(tuf_url, "SIGSTORE_TUF_URL")
+        resolved_fulcio_url = resolve(fulcio_url, "SIGSTORE_FULCIO_URL")
+        resolved_rekor_url = resolve(rekor_url, "SIGSTORE_REKOR_URL")
+        resolved_tsa_url = resolve(tsa_url, "SIGSTORE_TSA_URL")
```
jonburdo, line 112: Updated to use `SIGSTORE_` for now. After looking into it other tools like `gitsign`, `sigstore-go`, use this

clients/python/src/model_registry/signing/config.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # Default identity_token_path to Kubernetes service account token
+        resolved_identity_token_path: Path | None = None
+        if identity_token_path is not None:
+            resolved_identity_token_path = Path(identity_token_path) if identity_token_path else None
+        else:
+            k8s_token_path = Path("/var/run/secrets/kubernetes.io/serviceaccount/token")
+            if k8s_token_path.exists():
+                resolved_identity_token_path = k8s_token_path
```
jonburdo, line 121: A lot of logic here and below in this one method. This should be refactored, but I didn't get to it yet

clients/python/src/model_registry/signing/image_signer.py
jonburdo: It's possible I was mistaken or maybe had a different version of `cosign` but I'm hitting issues with a bash test script, so I think we just need to pass the value

clients/python/src/model_registry/signing/config.py
tarilabs: thanks for the check on this!

clients/python/src/model_registry/signing/config.py
jonburdo: I will plan to fix this in a follow-up

PR #2214, opened 2026-02-10
Model Catalog chips enhancements

YuliaKrimerman:
## Description

Enhanced the filter chip display and behavior on the Model Catalog landing page to improve UX consistency and clarity:

### Manual Testing Required:
1. **Performance View Toggle**
   - Enable performance view toggle on Model Catalog page
   -  Verify performance filter dropdowns appear (Workload Type, Latency, Max RPS, Hardware)

2. **Workload Type Chip**
   - Select a workload type from dropdown
   -  Verify chip displays: `"Scenario: {Name} ({X} input / {Y} output tokens)"`
   -  Verify chip only appears when different from default
   -  Click chip and verify it resets to default (not cleared)
   - Verify fa-undo icon appears on chip (not X)

3. **Latency Chip Group**
   - Select different latency metric/percentile/threshold
   - Verify 3 separate chips appear: "Metric: X", "Percentile: Y", "Under Zms"
   -  Verify individual chips have NO close buttons
   - Verify group has one reset button with fa-undo icon
   -  Click any individual chip and verify entire latency filter resets to default
   -  Click group reset button and verify entire latency filter resets to default

4. **Max RPS Chip**
   -  Set Max RPS to a value
   -  Verify chip displays: `"Max RPS: {value}"`
   -  Verify chip only appears when different from default
   -  Click chip and verify it resets to default
   -  Verify fa-undo icon appears (not X)

5. **Hardware Chips**
   -  Select multiple hardware configurations
   - Verify each hardware chip has standard X icon (not fa-undo)
   - Click X on individual chip and verify only that chip is removed
   - Verify multiple hardware chips can be selected simultaneously

6. **Side Panel Filters (License, Provider, etc.)**
   - Select multiple licenses/providers
   - Verify chips use standard X icons
   - Verify individual chips can be removed

7. **Reset All Filters Button**
   - Apply various filters
   -  Click "Reset all defaults" button
   -  Verify multi-select filters are cleared (Hardware, License)
   -  Verify performance filters reset to defaults (Workload Type, Latency, Max RPS)

https://github.com/user-attachments/assets/56f620e3-1bd3-43b3-81e5-a3ceec954cb9

## Merge criteria:

If you have UI changes:
- [x] The developer has added tests or explained why testing cannot be added (Manual UI testing required - behavior involves PatternFly Toolbar chip interactions and CSS-based icon styling that are difficult to unit test)
- [x] Included any necessary screenshots or gifs if it was a UI change (Screenshots provided above showing chip formats and icons)
- [x] Verify that UI/UX changes conform the UX guidelines for Kubeflow (Changes follow PatternFly design patterns and existing Model Catalog UX conventions)

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
@@ -127,7 +127,7 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
                 clearAllFilters: handleClearAllFilters,
                 clearFiltersButtonText: 'Reset all filters',
               }
-            : {})}
+            : { clearAllFilters: undefined, clearFiltersButtonText: '' })}
         >
           <ToolbarContent rowWrap={{ default: 'wrap' }}>
             <Flex>


YuliaKrimerman:
@manaswinidas 
Screenshot 2026-02-17 at 12 12 14 PM
Fixed

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationFilterToolbar.tsx
```diff
@@ -193,6 +188,17 @@ const HardwareConfigurationFilterToolbar: React.FC<HardwareConfigurationFilterTo
...
+        {onResetAllFilters && hasVisibleChips && (
+          <ToolbarItem>
+            <Button
+              variant="link"
+              onClick={onResetAllFilters}
+              data-testid="clear-all-filters-button"
+            >
+              Reset all defaults
+            </Button>
+          </ToolbarItem>
```
mturley: Since you're using a combination of custom-rendered chips and ToolbarFilter chips, if both are rendered at the same time we have both this "Reset all defaults" link and the built-in "Clear all filters" one that is provided by Toolbar when there are active ToolbarFilters under it. If we're no longer relying only on the built-in ToolbarFilter component, we should force the built-in one to never render (I think we might be able to do that by passing empty values into `Toolbar`, like we conditionally did before above on line 100). If that doesn't work... we may need to abandon using the built in ToolbarFilter chips. The intention was to use only those because we wanted to avoid custom rendering, but it seems that's not possible. If we cause ourselves too much more pain here maybe we should entirely render only custom chips for everything.

Image

Another thing we lose by combining the built-in with custom chips is that they appear on two separate lines, and your custom ones are wrapping starting to the right of the toolbar controls which isn't right, they should all be below the toolbar.

Also, the chips not in a group are not vertically centered with the chips that are in a group. This is true in both themes but more visible in the MUI theme. The MUI theme also has color issues, with white icons on grey chips.

Finally, for your latency filter chip group, the reset icon is appearing outside the group and larger rather than within the group as it should, at the same size as the other icons.

PR #2228, opened 2026-02-13
add partially avail source status

adysenrothman:
This PR adds a new source status to the model catalog called "partially-available". This status is used to represent sources (specifically HF type for now) that are able to load some models but not all models. The status is replacing the previously used "error" status to better represent the state of the source. A follow up PR will be created that ensures these sources are present in the dashboard so that any successfully pulled models are not hidden.

## Merge criteria:

If you have UI changes

Tested locally with Kind/ Tilt
Screenshot 2026-02-13 at 9 24 05 AM

### DIFF
--- a/catalog/clients/python/src/catalog_openapi/models/catalog_source_status.py
+++ b/catalog/clients/python/src/catalog_openapi/models/catalog_source_status.py
@@ -19,13 +19,14 @@ from typing_extensions import Self
 
 
 class CatalogSourceStatus(str, Enum):
-    """Operational status of a catalog source. - `available`: The source is functioning correctly and models can be retrieved - `error`: The source is experiencing issues and cannot provide models - `disabled`: The source has been intentionally disabled
+    """Operational status of a catalog source. - `available`: The source is functioning correctly and models can be retrieved - `partially-available`: The source loaded some models successfully but encountered errors with others - `error`: The source is experiencing issues and cannot provide models - `disabled`: The source has been intentionally disabled
     """
 
     """
     allowed enum values
     """
     AVAILABLE = "available"
+    PARTIALLY_MINUS_AVAILABLE = "partially-available"
     ERROR = "error"
     DISABLED = "disabled"
 
--- a/catalog/internal/catalog/loader.go
+++ b/catalog/internal/catalog/loader.go
@@ -19,10 +19,29 @@ import (
 
 // Source status constants matching the OpenAPI enum values
 const (
-	SourceStatusAvailable = "available"
-	SourceStatusError     = "error"
-	SourceStatusDisabled  = "disabled"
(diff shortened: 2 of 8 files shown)


catalog/clients/python/src/catalog_openapi/models/catalog_source_status.py
```diff
@@ -19,13 +19,14 @@
...
 class CatalogSourceStatus(str, Enum):
-    """Operational status of a catalog source. - `available`: The source is functioning correctly and models can be retrieved - `error`: The source is experiencing issues and cannot provide models - `disabled`: The source has been intentionally disabled
+    """Operational status of a catalog source. - `available`: The source is functioning correctly and models can be retrieved - `partially-available`: The source loaded some models successfully but encountered errors with others - `error`: The source is experiencing issues and cannot provide models - `disabled`: The source has been intentionally disabled
     """
 
     """
     allowed enum values
     """
     AVAILABLE = "available"
+    PARTIALLY_MINUS_AVAILABLE = "partially-available"
```
jonburdo, line 29: Just a general note: The `MINUS` seems odd. Looks like this is auto-generated so probably not something we can fix (at least out-of-scope for this), and I see other similar examples (e.g. `MODEL_MINUS_ARTIFACT`)

catalog/internal/catalog/loader.go
```diff
@@ -19,11 +19,30 @@ import (
...
 
+// PartiallyAvailableError indicates that a source loaded some models successfully
+// but encountered errors with others.
+type PartiallyAvailableError struct {
+	FailedModels []string
+}
+
+func (e *PartiallyAvailableError) Error() string {
+	return fmt.Sprintf("Failed to fetch some models, ensure models exist and are accessible with given credentials. Failed models: %v", e.FailedModels)
+}
```
jonburdo, line 36: Possible future improvement: This may not be needed, but I wonder if we would want to include the failed models in the response in a more structured format. This could mean a new key in the json object. Or if we don't want to change the structure, simply converting this value to a json array before adding it to the error message could allow it to be parsed.

jonburdo (approved):
/lgtm
/approve

catalog/clients/python/src/catalog_openapi/models/catalog_source_status.py
adysenrothman: I noticed that too and disregarded because of model-artifact. I think that if it is changed to partially_available it is handled differently but don't necessarily think it is an issue, just weird openapi generated rules.

catalog/internal/catalog/loader.go
adysenrothman: Yeah good call out. I could see that being useful.

PR #2246, opened 2026-02-18
UI: Remove duplicate search label text in filter boxes

### LINKED ISSUE #2245: Bug(Model Catalog) : Search box doesn't need extra same text on placeholder and on border
The Search box have text of _Filter by name, description or provider_  which is on both places. First at border of box and one as placeholder.

We can reduce this duplication by removing the text on border and keep UI clean. We can make this changes to all Search Boxes to keep it clean

Attaching screenshot : 

Image

amirHdev:
## Summary
This PR removes duplicated helper text shown in search/filter inputs where the same text appeared both:
- in the field border/legend, and
- as placeholder text.
- Fixes #2245 

To keep the UI cleaner, we now keep only the placeholder text.

## Changes
- Removed `fieldLabel` from `ThemeAwareSearchInput` usages in:
  - model catalog search
  - model registry list filters
  - model versions list/archive filters
  - registered models archive filters
  - model transfer jobs filters

## Result
Search boxes no longer show duplicated label text on the border and inside the input.

## Testing
- Verified code changes are scoped to frontend search/filter components.
- Type-check command could not be executed in this environment (`tsc` not available).

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
@@ -136,8 +136,7 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
                 <ToolbarGroup variant="filter-group" gap={{ default: 'gapMd' }} alignItems="center">
                   <ToolbarItem>
                     <ThemeAwareSearchInput
-                      dara-testid="search-input"
-                      fieldLabel="Filter by name, description and provider"
+                      data-testid="search-input"
                       aria-label="Search with submit button"
                       className="toolbar-fieldset-wrapper"
                       placeholder="Filter by name, description and provider"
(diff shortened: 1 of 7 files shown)


clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -137,7 +137,6 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
                   <ToolbarItem>
                     <ThemeAwareSearchInput
                       dara-testid="search-input"
```
manaswinidas: unrelated to your changes - but yikes- typo in the test-id 
```suggestion
                      data-testid="search-input"
```
🙈 not sure why we didn't catch this earlier

amirHdev:
Thanks @manaswinidas 
Agree with that instinct. Since all current uses have fieldLabel removed, leaving the prop on ThemeAwareSearchInput is now dead weight and invites the duplication to come back. we can remove fieldLabel and keep FormFieldset or wrapper it if needed .
If you think i can make changes

manaswinidas:
Can you remove the `fieldLabel` from `ThemeAwareSearchInput` ?

amirHdev:
> Can you remove the `fieldLabel` from `ThemeAwareSearchInput` ?

Done

PR #2255, opened 2026-02-19
refactor(catalog): split catalog/internal/catalog

pboyd:
## Description

Reorganizes catalog code into three packages:
- `basecatalog`: shared config, BaseLoader, file monitoring, validation
- `modelcatalog`: model-specific loaders, filters, labels, sources
- `catalog`: facade that re-exports subpackage types

basecatalog provides:
- `SourceConfig` with a new `model_catalogs` field (`catalogs` exists, but is now deprecated)
- `BaseLoader` managing leader state, write tracking, file watchers
- Config validation detecting duplicate IDs
- Monitor for K8s ConfigMap hot-reload

modelcatalog contains:
- All model catalog implementations (DB, HuggingFace, YAML)
- Model filtering, license transforms, performance metrics
- `APIProvider` interface and test helpers

`catalog.go` re-exports public types to maintain API stability.

This is prep work to add code to load MCP catalogs.

## How Has This Been Tested?
On a local development cluster.

## Merge criteria:


catalog/internal/catalog/basecatalog/config.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// DEPRECATED: Use ModelCatalogs instead
+	// This field is maintained for backwards compatibility
+	Catalogs []Source `yaml:"catalogs,omitempty" json:"catalogs,omitempty"`
+}
+
+// GetModelCatalogs returns the merged list of model catalogs, combining the
+// new model_catalogs field with the deprecated catalogs field.
+// If there are ID conflicts, model_catalogs takes precedence.
+func (c *SourceConfig) GetModelCatalogs() []Source {
```
Al-Pragliola, line 41: in this function we use a map and convert to slice, in your opinion is this a problem if we lose the original ordering?

pboyd: Good point. I've updated this to preserve the order.

PR #2266, opened 2026-02-24
test: add sorting test for artifacts endpoint

fege:
Add order_by and sort_order parameters to the get_artifacts method to enable sorting.
Add test coverage for sorting by ID and custom property values

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/clients/python/src/model_catalog/_client.py
+++ b/catalog/clients/python/src/model_catalog/_client.py
@@ -12,6 +12,7 @@ from collections.abc import Callable
 from functools import wraps
 from typing import Any, TypeVar
 from urllib.parse import quote
+from catalog_openapi.models import OrderByField, SortOrder
 
 logger = logging.getLogger(__name__)
 
@@ -338,7 +339,6 @@ class CatalogAPIClient:
         Returns:
             Dict with models response.
         """
-        from catalog_openapi.models import OrderByField, SortOrder
 
         source_list = [source] if source else None
         page_size_str = str(page_size) if page_size is not None else None
@@ -393,7 +393,6 @@ class CatalogAPIClient:
         Returns:
             Dict with artifacts response.
         """
-        from catalog_openapi.models import SortOrder
 
         # Convert artifact_type to list format expected by OpenAPI client
         artifact_type_list = None
--- a/catalog/clients/python/tests/test_artifacts.py
+++ b/catalog/clients/python/tests/test_artifacts.py
@@ -12,7 +12,7 @@ To run these tests:
 import pytest
 
 from model_catalog import CatalogAPIClient, CatalogValidationError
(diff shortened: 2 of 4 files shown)


catalog/clients/python/tests/test_artifacts.py
```diff
@@ -428,4 +429,86 @@ def test_filter_artifacts_by_invalid_artifact_type(
...
+        response = api_client.get_artifacts(
+            source_id=source_id,
+            model_name=model_name,
+            order_by=order_by,
+            sort_order=sort_order,
+        )
+
+        items = response["items"]
+        assert len(items) >= 2, "Need at least 2 artifacts to validate sorting"
+        assert validate_custom_property_sorted(items, order_by, sort_order)
```
jonburdo: Just one thing to consider. If instead you simply have a function that does the sorting and returns a sorted list, then you can do `assert result == expected`. This may make it easier to debug if the test fails since the error will show the two lists and the diff instead of just an incorrect boolean.

But I also think it's okay as is! I appreciate the helper function so the test is cleaner

catalog/clients/python/src/model_catalog/_client.py
```diff
@@ -382,12 +384,17 @@ def get_artifacts(
...
+            order_by: Optional field to order by (ID, NAME, CREATE_TIME, LAST_UPDATE_TIME,
+                or custom property path like "accuracy.double_value").
+            sort_order: Optional sort order (ASC or DESC).
             page_size: Optional page size.
             next_page_token: Optional pagination token.
 
         Returns:
             Dict with artifacts response.
         """
+        from catalog_openapi.models import SortOrder
```
jonburdo: nit: Maybe this should be at the top?

fege: this is true but I followed the pattern used in `_client.py`

catalog/clients/python/tests/test_artifacts.py
fege: good point

catalog/clients/python/src/model_catalog/_client.py
fege: updated

catalog/clients/python/tests/test_artifacts.py
```diff
@@ -428,4 +429,86 @@ def test_filter_artifacts_by_invalid_artifact_type(
...
+        response = api_client.get_artifacts(
+            source_id=source_id,
+            model_name=model_name,
+            order_by=order_by,
+            sort_order=sort_order,
+        )
+
+        items = response["items"]
+        assert len(items) >= 2, "Need at least 2 artifacts to validate sorting"
+        assert items == sort_items_by_field(items, order_by, sort_order)
```
jonburdo, line 465: Thanks for the extra effort to do it this way!

PR #2270, opened 2026-02-24
Storage location in Model Version Page

YuliaKrimerman:
## Description

Adds a new **Storage location** section to the Model Version details page sidebar for models registered via the Register + Store flow. When a model artifact has `modelSourceKind: "transfer_job"` properties, the UI fetches the associated transfer job from the BFF and displays:

- **Storage location** — the OCI/S3 destination where the model is currently stored, with a popover explaining this is the "Model destination location" specified during registration
- **Model URI / Path** — the destination URI or path with a context-aware popover (OCI vs S3 vs URI)
- **Storage source details** — an expandable section showing the original model origin location (source type, namespace, and path) before the transfer
- **Stored timestamp** — a new "Stored" field in the metadata section showing when the artifact was successfully stored
- **No-access handling** — if the user doesn't have permission to the transfer job's namespace, a degraded view shows the namespace with a popover explaining how to request access

For models registered without Store (Register Only), the existing "Model location" section is preserved unchanged.
## How Has This Been Tested?

https://github.com/user-attachments/assets/820c55c7-f21f-46ea-ac9e-d22521eae6e0

- **Cypress tests** added for:
  - Storage location section visibility for transfer job artifacts
  - Destination connection info and Model URI content
  - Expandable source details with origin location and path
  - Stored timestamp presence
  - Model location section for non-transfer job artifacts (Register Only)

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionDetails.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionDetails.cy.ts
@@ -608,12 +608,12 @@ describe('Model version details', () => {
         modelVersionDetails.findStorageLocationTitle().should('contain.text', 'Storage location');
       });
 
-      it('should show destination connection info and Model URI', () => {
+      it('should show destination secret info and Model URI', () => {
         initStorageLocationIntercepts();
         modelVersionDetails.visit();
 
         modelVersionDetails
-          .findStorageConnectionDisplay()
+          .findStorageSecretDisplay()
           .first()
           .should('contain.text', 'OCI storage')
           .and('contain.text', 'my-project-1');
@@ -630,7 +630,7 @@ describe('Model version details', () => {
         modelVersionDetails.findStorageSourceDetails().find('button').first().click();
 
         modelVersionDetails
-          .findStorageConnectionDisplay()
+          .findStorageSecretDisplay()
           .last()
           .should('contain.text', 'S3 storage')
           .and('contain.text', 'my-project-1');
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
@@ -43,13 +43,13 @@ const NoAccessPopover: React.FC<{ namespace: string }> = ({ namespace }) => (
   <Popover
     headerContent={
       <>
-        You don&apos;t have access to the <strong>{namespace}</strong> Project.
+        You don&apos;t have access to the <strong>{namespace}</strong> namespace.
       </>
     }
     bodyContent={
       <>
         <Content component={ContentVariants.p}>
-          To request access to a new or existing project, contact your administrator.
+          To request access to a new or existing namespace, contact your administrator.
         </Content>
         <Content component={ContentVariants.p}>Your administrator might be:</Content>
         <List>
@@ -62,7 +62,7 @@ const NoAccessPopover: React.FC<{ namespace: string }> = ({ namespace }) => (
   >
     <Button
       variant="plain"
-      aria-label="More info about project access"
+      aria-label="More info about namespace access"
       data-testid="no-access-popover-button"
     >
       <OutlinedQuestionCircleIcon />
@@ -70,19 +70,19 @@ const NoAccessPopover: React.FC<{ namespace: string }> = ({ namespace }) => (
   </Popover>
 );
 
-const ConnectionDisplay: React.FC<{
+const SecretDisplay: React.FC<{
   secretName?: string;
   storageType?: StorageType;
   namespace: string;
   hasAccess?: boolean;
 }> = ({ secretName, storageType, namespace, hasAccess = true }) => (
-  <Content component={ContentVariants.p} data-testid="storage-connection-display">
+  <Content component={ContentVariants.p} data-testid="storage-secret-display">
     {(secretName || storageType) && (
       <>
         <strong>
           {secretName || (storageType && `${getStorageTypeLabel(storageType)} storage`)}
         </strong>{' '}
-        connection in{' '}
+        secret in{' '}
       </>
     )}
     <Icon size="sm" isInline>
@@ -100,7 +100,7 @@ const StorageLocationTitle: React.FC = () => (
       headerContent="Storage location"
       bodyContent={
         <Content component={ContentVariants.p}>
-          The connection currently being used to store this model and its artifacts. This is the{' '}
+          The secret currently being used to store this model and its artifacts. This is the{' '}
           <strong>Model destination location</strong> specified during registration.
         </Content>
       }
@@ -163,7 +163,7 @@ const StorageLocationSection: React.FC<StorageLocationSectionProps> = ({
           <StorageLocationTitle />
         </StackItem>
         <StackItem>
-          <ConnectionDisplay namespace={fallbackNamespace} hasAccess={false} />
+          <SecretDisplay namespace={fallbackNamespace} hasAccess={false} />
         </StackItem>
       </Stack>
     );
@@ -196,7 +196,7 @@ const StorageLocationSection: React.FC<StorageLocationSectionProps> = ({
       </StackItem>
 
       <StackItem>
-        <ConnectionDisplay
+        <SecretDisplay
           secretName={transferJob.destSecretName}
           storageType={destType}
           namespace={namespace}
@@ -230,7 +230,7 @@ const StorageLocationSection: React.FC<StorageLocationSectionProps> = ({
           <Stack hasGutter>
             <StackItem>
               <Content component={ContentVariants.p}>
-                Details of the connection used to store the model before it was registered.
+                Details of the secret used to store the model before it was registered.
               </Content>
             </StackItem>
 
@@ -238,9 +238,9 @@ const StorageLocationSection: React.FC<StorageLocationSectionProps> = ({
               <DescriptionList>
                 <DashboardDescriptionListGroup
                   title="Model origin location"
-                  popover="The connection that was used to store the model at the time it was registered."
+                  popover="The secret that was used to store the model at the time it was registered."
                 >
-                  <ConnectionDisplay
+                  <SecretDisplay
                     secretName={transferJob.sourceSecretName}
                     storageType={sourceType}
                     namespace={namespace}
--- a/clients/ui/frontend/src/app/utils.ts
+++ b/clients/ui/frontend/src/app/utils.ts
@@ -142,11 +142,11 @@ export const getStorageTypeLabel = (type: StorageType): string => {
 export const getModelUriPopoverContent = (destType: ModelTransferJobDestinationType): string => {
   switch (destType) {
     case ModelTransferJobDestinationType.OCI:
-      return 'The URI of the OCI connection that is currently being used as the model storage location.';
+      return 'The URI of the OCI secret that is currently being used as the model storage location.';
     case ModelTransferJobDestinationType.S3:
-      return 'The path of the S3-compatible object storage connection that is currently being used as the model storage location.';
+      return 'The path of the S3-compatible object storage secret that is currently being used as the model storage location.';
     default:
-      return 'The URI of the connection that is currently being used as the model storage location.';
+      return 'The URI of the secret that is currently being used as the model storage location.';
   }
 };
 
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelRegistryView/modelVersionDetails.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/pages/modelRegistryView/modelVersionDetails.ts
@@ -182,8 +182,8 @@ class ModelVersionDetails {
     return cy.findByTestId('storage-location-uri');
   }
 
-  findStorageConnectionDisplay() {
-    return cy.findAllByTestId('storage-connection-display');
+  findStorageSecretDisplay() {
+    return cy.findAllByTestId('storage-secret-display');
   }
 
   findStorageSourceDetails() {


clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -138,7 +138,7 @@ func GetModelArtifactMocks() []openapi.ModelArtifact {
 		CustomProperties:         newCustomProperties(),
 		Description:              stringToPointer("This artifact can do more than you would expect"),
 		ExternalId:               stringToPointer("1000001"),
-		Uri:                      stringToPointer("http://localhost/artifacts/1"),
+		Uri:                      stringToPointer("oci://registry.redhat.io/rhelai1/modelcar-granite-7b-starter:1.4.0"),
```
ppadti: I think we should not be using `redhat` - could you please change this to something generic

clients/ui/frontend/src/__mocks__/mockModelTransferJob.ts
```diff
@@ -14,16 +14,18 @@ export const mockModelTransferJob = (partial?: Partial<ModelTransferJob>): Model
...
     awsAccessKeyId: 'accessKeyId',
     awsSecretAccessKey: 'secretAccessKey',
   },
   destination: {
-    type: ModelTransferJobDestinationType.S3,
-    bucket: 'destination-bucket',
-    key: 'models/fraud-detection/v1',
-    region: 'us-west-2',
+    type: ModelTransferJobDestinationType.OCI,
+    uri: 'registry.redhat.io/rhelai1/modelcar-granite-7b-starter:1.4.0',
```
ppadti: same here.

clients/ui/frontend/src/app/hooks/useModelTransferJobForArtifact.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const callback = React.useCallback<FetchStateCallbackPromise<ModelTransferJob | null>>(
+    (opts) => {
+      if (!apiAvailable) {
+        return Promise.reject(new Error('API not yet available'));
+      }
+      if (!jobNamespace || !jobName) {
+        return Promise.resolve(null);
+      }
+      // TODO: Replace with a GET single job endpoint when the BFF supports it,
```
ppadti: [PR which adds GET endpoint](https://github.com/kubeflow/model-registry/pull/2281) is merged now - could you please rebase it.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+const StorageLocationSection: React.FC<StorageLocationSectionProps> = ({
+  fallbackNamespace,
+  transferJob,
+  transferJobLoaded,
+  transferJobError,
+}) => {
+  const [isSourceDetailsExpanded, setIsSourceDetailsExpanded] = React.useState(false);
+
+  // TODO: Replace string matching with structured error status codes from the BFF
```
ppadti: @mturley @YuliaKrimerman - I think here we need to get the list of namespace from `useNamespaceSelector` and find that the job's namespace is in the list to check for user access to that namespace here right?

clients/ui/frontend/src/app/api/service.ts
```diff
@@ -246,6 +246,18 @@ export const getListModelTransferJobs =
       },
     );
 
+export const getModelTransferJobByName =
+  (hostPath: string) =>
```
manaswinidas: Bug: `getModelTransferJobByName` doesn't accept or forward `queryParams`, unlike every other function in this file (e.g. `getListModelTransferJobs`, `createModelTransferJob`, `deleteModelTransferJob`). It only passes `{ namespace }` to `restGET`, dropping the standard `queryParameters` that get wired in via `useModelRegistryAPIState`.

Compare with `getListModelTransferJobs` just above:
```ts
export const getListModelTransferJobs =
  (hostPath: string, queryParams: Record<string, unknown> = {}) => // <-- takes queryParams
```

And the wiring in `useModelRegistryAPIState.tsx`:
```ts
getModelTransferJobByName: getModelTransferJobByName(path),           // <-- missing queryParameters
createModelTransferJob: createModelTransferJob(path, queryParameters), // <-- has it
```

If `queryParameters` carries auth tokens or routing parameters required by the BFF, this function will fail in production. Suggest aligning the signature:
```ts
export const getModelTransferJobByName =
  (hostPath: string, queryParams: Record<string, unknown> = {}) =>
  (opts: APIOptions, namespace: string, jobName: string): Promise<ModelTransferJob> =>
    handleRestFailures(
      restGET(hostPath, `/model_transfer_jobs/${encodeURIComponent(jobName)}`, { ...queryParams, namespace }, opts),
    )...
```
and updating the wiring to `getModelTransferJobByName(path, queryParameters)`.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -54,6 +57,12 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
   const modelArtifact = modelArtifacts.items.length ? modelArtifacts.items[0] : null;
   const { apiState } = React.useContext(ModelRegistryContext);
   const storageFields = uriToStorageFields(modelArtifact?.uri || '');
+  const transferJobParams = modelArtifact
+    ? modelSourcePropertiesToTransferJobParams(modelArtifact)
+    : null;
+  const isTransferJobSource = !!transferJobParams;
+  const [transferJob, transferJobLoaded, transferJobError] =
+    useModelTransferJobForArtifact(modelArtifact);
```
manaswinidas: Nit (DRY): `modelSourcePropertiesToTransferJobParams(modelArtifact)` is called here **and** again inside `useModelTransferJobForArtifact` (line 11 of that hook). Both calls run on every render with the same input.

Two options to deduplicate:
1. Have the hook accept pre-computed `TransferJobParams | null` instead of a raw `ModelArtifact`, so the parent computes once and passes down.
2. Have the hook return the computed `isTransferJobSource` boolean alongside the fetch state, so the parent doesn't need to call the converter at all.

clients/ui/frontend/src/app/utils.ts
```diff
@@ -118,3 +123,60 @@ export const getDoubleValue = <T extends ModelRegistryCustomProperties>(
   }
   return 0;
 };
+
+export const getStorageTypeLabel = (type: StorageType | string): string => {
```
manaswinidas: Type safety: The `| string` widening on `StorageType | string` (used here and in `getModelUriPopoverContent` / `getModelUriLabel` below) defeats TypeScript exhaustiveness checking. Any string value will silently fall through to the `default` case without a compiler warning.

Since callers pass `transferJob.destination.type` (typed as `ModelTransferJobDestinationType`) or `transferJob.source.type` (typed as `ModelTransferJobSourceType`), the `| string` shouldn't be needed. Removing it would let TypeScript catch unhandled enum members at compile time if new types are added in the future.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,243 @@
...
+  // once the BFF returns proper error types for permission issues.
+  const errorMessage = transferJobError?.message ?? '';
+  const hasAccessError =
+    !!transferJobError &&
+    (errorMessage.includes('403') ||
+      errorMessage.includes('forbidden') ||
+      errorMessage.includes('Forbidden'));
```
manaswinidas: Robustness: The string-matching heuristic here is fragile — it will miss cases like HTTP 401, non-English error messages, or error messages that happen to contain the substring "403" in a different context.

The TODO is good, but consider at minimum matching on a status code property if the error object carries one (e.g. `(transferJobError as any).statusCode === 403`), or defining a small `isAccessError(error)` utility that can be updated in one place when the BFF adds structured codes.

clients/ui/frontend/src/__mocks__/mockModelTransferJob.ts
```diff
@@ -14,16 +14,18 @@ export const mockModelTransferJob = (partial?: Partial<ModelTransferJob>): Model
...
   destination: {
-    type: ModelTransferJobDestinationType.S3,
-    bucket: 'destination-bucket',
-    key: 'models/fraud-detection/v1',
-    region: 'us-west-2',
+    type: ModelTransferJobDestinationType.OCI,
+    uri: 'quay.io/my-org/my-model:v1.0.0',
+    registry: 'quay.io',
+    username: 'username',
+    password: 'password',
```
manaswinidas: Breaking change: The default mock destination changed from `S3` to `OCI` and the source `key` changed. Any existing tests or downstream consumers that call `mockModelTransferJob()` without overrides and assert on `destination.type === 'S3'`, `destination.bucket`, `destination.key`, or `source.key` will break.

Consider keeping the original S3 defaults intact and instead creating a separate factory like `mockModelTransferJobOCI()` (similar to how `mockModelArtifactWithTransferJob` was added alongside `mockModelArtifact`). This way existing consumers are unaffected and OCI-specific tests use the purpose-built mock.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,243 @@
...
+  </Popover>
+);
+
+const ConnectionDisplay: React.FC<{
+  storageType: string;
+  namespace: string;
+  hasAccess?: boolean;
+}> = ({ storageType, namespace, hasAccess = true }) => (
+  <Content component={ContentVariants.p} data-testid="storage-connection-display">
+    {storageType && (
```
ppadti: I think here we should be showing the name of the source secret name /dest secret name rather than the source type.

ppadti:
Also, @YuliaKrimerman - could you please update the model_transfer_jobs.go file - I missed this before

from 
```
data["ModelArtifact.model_source_kind"] = "Job"
data["ModelArtifact.model_source_class"] = "async-upload"
data["ModelArtifact.model_source_group"] = "batch/v1"
data["ModelArtifact.model_source_name"] = jobName
```

to 
```
data["ModelArtifact.model_source_kind"] = "transfer_job"
data["ModelArtifact.model_source_class"] = "async-upload"
data["ModelArtifact.model_source_group"] = namespace
data["ModelArtifact.model_source_name"] = jobName
```

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -149,6 +149,9 @@ func GetModelArtifactMocks() []openapi.ModelArtifact {
 		StoragePath:              stringToPointer("/artifacts/1"),
 		ModelFormatVersion:       stringToPointer("1.0.0"),
 		ServiceAccountName:       stringToPointer("service-1"),
+		ModelSourceKind:          stringToPointer("transfer_job"),
```
manaswinidas, line 152: **Blocker:** The mock now sets `ModelSourceKind: "transfer_job"`, but the actual `model_transfer_jobs.go` backend file still uses `"Job"` (per ppadti's earlier comment). Without that backend change, the frontend's `modelSourceKind === 'transfer_job'` check will never match against real (non-mocked) data, making this feature non-functional in production.

Please include the `model_transfer_jobs.go` update in this PR:
```go
// from
data["ModelArtifact.model_source_kind"] = "Job"
data["ModelArtifact.model_source_group"] = namespace  // was missing

// to
data["ModelArtifact.model_source_kind"] = "transfer_job"
data["ModelArtifact.model_source_group"] = namespace
```

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,269 @@
...
+  if (!error) {
+    return false;
+  }
+  const { message } = error;
+  return (
+    message.includes('403') ||
+    message.includes('forbidden') ||
+    message.includes('Forbidden') ||
+    message.includes('not have access') ||
+    message.includes(jobNamespace)
```
manaswinidas: Now that `useNamespaceSelector` is the primary access-check mechanism (line 140-143), consider whether this string-matching fallback can be removed entirely.

In particular, `message.includes(jobNamespace)` (this line) is a false-positive risk — any error message that happens to contain the namespace string (e.g., `"Failed to connect to service in my-project-1"`) will be misclassified as an access error, hiding the real failure from the user.

If you want to keep a belt-and-suspenders approach, at least remove the `message.includes(jobNamespace)` check, and consider using a structured error status code if/when the BFF supports it.

clients/ui/frontend/src/app/utils.ts
```diff
@@ -118,3 +123,60 @@ export const getDoubleValue = <T extends ModelRegistryCustomProperties>(
...
+    case ModelTransferJobDestinationType.OCI:
+      return 'OCI';
+    case ModelTransferJobSourceType.URI:
+      return 'URI';
+    default:
+      return type;
+  }
+};
+
+export const getModelUriPopoverContent = (destType: StorageType): string => {
```
manaswinidas: Nit (type narrowing): `getModelUriPopoverContent` and `getModelUriLabel` below are only ever called with `destType` (a `ModelTransferJobDestinationType`), but they accept the broader `StorageType` union. This means source-only types like `ModelTransferJobSourceType.URI` would silently fall through to the default case.

Narrowing the parameter to `ModelTransferJobDestinationType` would let TypeScript catch unhandled destination types at compile time if new ones are added:

```ts
export const getModelUriPopoverContent = (destType: ModelTransferJobDestinationType): string => {
```

Also, the default case returns `"The URI of the URI connection..."` which reads awkwardly — consider a more generic fallback like `"The URI of the connection currently being used as the model storage location."`.

clients/ui/frontend/src/app/utils.ts
```diff
@@ -118,3 +123,60 @@ export const getDoubleValue = <T extends ModelRegistryCustomProperties>(
   }
   return 0;
 };
+
+export const getStorageTypeLabel = (type: StorageType): string => {
```
manaswinidas, line 127: Nit: The new utility functions (`getStorageTypeLabel`, `getModelUriPopoverContent`, `getModelUriLabel`, `getDestinationUri`, `getSourcePath`) are only exercised indirectly through Cypress tests. Consider adding lightweight unit tests — especially for edge cases like missing `key`/`uri` fields, or when new enum values are added to `StorageType`.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,262 @@
...
+    (!userHasNamespaceAccess && !!transferJobError) || isAccessError(transferJobError);
+
+  if (!transferJobLoaded) {
+    return (
+      <Bullseye>
+        <Spinner size="lg" />
+      </Bullseye>
+    );
+  }
+
```
manaswinidas, line 150: Bug (Critical): `useFetchState` does **not** call `setLoaded(true)` when the promise rejects — `loaded` stays `false` while `loadError` is set. Because the `!transferJobLoaded` check (line 139) returns a spinner **before** the error check (line 146), any non-access API failure (500, network error, timeout) results in an **infinite spinner** that the user cannot escape without a full page reload.

The error alert on line 146 is unreachable in this scenario.

Fix: check for errors before (or alongside) the loading state:
```tsx
if (transferJobError && !hasAccessError) {
  return (
    <Alert variant="danger" isInline title="Failed to load storage location">
      {transferJobError.message}
    </Alert>
  );
}

if (!transferJobLoaded) {
  return (
    <Bullseye><Spinner size="lg" /></Bullseye>
  );
}
```

clients/ui/frontend/src/app/hooks/useModelTransferJobForArtifact.ts
```diff
@@ -0,0 +1,30 @@
...
+  transferJobParams: TransferJobParams | null,
+): FetchState<ModelTransferJob | null> => {
+  const { api, apiAvailable } = useModelRegistryAPI();
+  const jobNamespace = transferJobParams?.jobNamespace;
+  const jobName = transferJobParams?.jobName;
+
+  const callback = React.useCallback<FetchStateCallbackPromise<ModelTransferJob | null>>(
+    (opts) => {
+      if (!apiAvailable) {
+        return Promise.reject(new Error('API not yet available'));
```
manaswinidas: Bug (High): `useFetchState` silently ignores `NotReadyError` (via `isCommonStateError`) but stores plain `Error` objects as `loadError`. This rejection with a plain `Error` means:

1. The error is stored as `loadError` — combined with the loading-check-first issue in `StorageLocationSection`, the user sees an infinite spinner while the API initializes.
2. Even after `apiAvailable` flips to `true` and the callback re-runs successfully, there's a flash where the error state is visible.

Use `NotReadyError` from `mod-arch-core` instead, which is the standard pattern for "not yet ready" states:
```ts
import { NotReadyError } from 'mod-arch-core';
// ...
if (!apiAvailable) {
  return Promise.reject(new NotReadyError('API not yet available'));
}
```

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,262 @@
...
+  transferJobLoaded,
+  transferJobError,
+}) => {
+  const [isSourceDetailsExpanded, setIsSourceDetailsExpanded] = React.useState(false);
+  const { namespaces = [] } = useNamespaceSelector();
+
+  const userHasNamespaceAccess = namespaces.some((ns) => ns.name === fallbackNamespace);
+  const hasAccessError =
+    (!userHasNamespaceAccess && !!transferJobError) || isAccessError(transferJobError);
+
```
manaswinidas: Bug (High): The first condition `(!userHasNamespaceAccess && !!transferJobError)` treats **any** error as an access error when the user's namespace list doesn't include `fallbackNamespace`. If the BFF returns a 500 or the network is down, `transferJobError` is truthy, so `hasAccessError` becomes `true` and the user sees "You don't have access to {namespace}" — completely masking the real failure.

Consider requiring that a real error is specifically an access error, or that there's no error at all:
```ts
const hasAccessError =
  !userHasNamespaceAccess && (!transferJobError || isAccessError(transferJobError));
```
This way: no namespace access + no error (or an access error) = show degraded view, but no namespace access + a 500 = show the actual error.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -54,6 +57,12 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
   const modelArtifact = modelArtifacts.items.length ? modelArtifacts.items[0] : null;
   const { apiState } = React.useContext(ModelRegistryContext);
   const storageFields = uriToStorageFields(modelArtifact?.uri || '');
+  const transferJobParams = modelArtifact
+    ? modelSourcePropertiesToTransferJobParams(modelArtifact)
+    : null;
+  const isTransferJobSource = !!transferJobParams;
+  const [transferJob, transferJobLoaded, transferJobError] =
```
manaswinidas: Nit (Medium): The `refresh` function (4th element of `FetchState`) is destructured away here. If the transfer job fetch fails, the user has no way to retry without reloading the entire page. Other hooks in this file capture their refresh functions (e.g. `refreshModelVersion`, `refreshModelArtifacts`).

Consider capturing `refresh` and either passing it to `StorageLocationSection` to render a retry button on error, or wiring it into `refreshModelArtifacts`.

clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -786,9 +786,9 @@ func buildModelMetadataConfigMap(generateNamePrefix, namespace string, payload m
 		data["ModelArtifact.id"] = payload.ModelArtifactId
 	}
 
-	data["ModelArtifact.model_source_kind"] = "Job"
+	data["ModelArtifact.model_source_kind"] = "transfer_job"
 	data["ModelArtifact.model_source_class"] = "async-upload"
-	data["ModelArtifact.model_source_group"] = "batch/v1"
+	data["ModelArtifact.model_source_group"] = namespace
 	data["ModelArtifact.model_source_name"] = jobName
```
manaswinidas, line 883: Backwards compatibility: Both `model_source_kind` (`"Job"` → `"transfer_job"`) and `model_source_group` (`"batch/v1"` → `namespace`) changed semantics. Any transfer jobs already created before this change will have the old values stored in their ConfigMaps/artifacts. The frontend now checks `modelSourceKind === 'transfer_job'` and uses `modelSourceGroup` as the namespace to fetch the job — so existing records will silently fail (the kind won't match, or `"batch/v1"` will be used as a namespace).

Is there a migration path for existing data, or is the transfer job feature new enough that no production data exists with the old values?

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -284,18 +300,30 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
...
+                  {isTransferJobSource && transferJob && (
+                    <DashboardDescriptionListGroup
+                      title="Stored"
+                      popover="When the model artifact was successfully stored in the specified model location."
+                      isEmpty={!transferJob.lastUpdateTimeSinceEpoch}
+                      contentWhenEmpty="Unknown"
+                    >
+                      <ModelTimestamp timeSinceEpoch={transferJob.lastUpdateTimeSinceEpoch} />
+                    </DashboardDescriptionListGroup>
+                  )}
```
manaswinidas, line 327: Nit: `lastUpdateTimeSinceEpoch` is a generic "last modified" timestamp on the transfer job — it will change if the job is updated for any reason after storage completes (e.g. metadata edits, status transitions). The popover says "when the model artifact was successfully stored," which implies a precise completion time.

Is there a dedicated completion/finish timestamp on the transfer job (e.g. from the job's status or conditions) that would more accurately represent when storage finished?

clients/ui/bff/internal/repositories/model_transfer_jobs.go
ppadti: I think its fine as its still behind the feature flag and not yet released.

clients/ui/frontend/src/app/hooks/useModelTransferJobForArtifact.ts
```diff
@@ -0,0 +1,30 @@
...
+      }
+      if (!jobNamespace || !jobName) {
```
ppadti, line 19: when jobName and jobNamespace not present we should return  `return Promise.reject(new NotReadyError('No jobName or jobNamespace'));`

manaswinidas:
> Please add Cypress tests: for no-access view and any errors ?

This is not covered yet

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,274 @@
...
+        <StackItem>
+          <ConnectionDisplay namespace={fallbackNamespace} hasAccess={false} />
+        </StackItem>
+      </Stack>
+    );
+  }
+
+  if (!transferJob) {
+    return null;
+  }
```
manaswinidas, line 183: If loading succeeds with no error and no access issue but `transferJob` is `null`, this silently renders nothing — the user gets no feedback. Consider showing a fallback like an empty state or informational message (e.g. "Storage information unavailable") instead of returning `null`.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,274 @@
...
+  </Title>
+);
+
+const isAccessError = (error: Error | undefined): boolean => {
+  if (!error) {
+    return false;
+  }
+  const { message } = error;
+  return message.includes('403') || message.includes('forbidden') || message.includes('Forbidden');
+};
```
manaswinidas: Now that `useNamespaceSelector` is the primary access-check mechanism, consider whether this string-matching fallback adds value or just risk. It will miss 401s and could false-positive on unrelated error messages containing `"403"`. If keeping it as defense-in-depth, a structured status code check (e.g. `(error as any).statusCode === 403`) would be safer.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelRegistry/modelVersionDetails.cy.ts
```diff
@@ -559,4 +561,146 @@ describe('Model version details', () => {
...
+    it('should not show Model location section for transfer job artifacts', () => {
+      initStorageLocationIntercepts();
+      modelVersionDetails.visit();
+
+      cy.contains('Model location').should('not.exist');
+    });
+  });
+
+  describe('Storage location error states', () => {
+    const transferJobArtifact = mockModelArtifactWithTransferJob({
```
manaswinidas: Nit (DRY): `transferJobArtifact` is defined identically here and in the `Storage location (Register + Store)` describe block above. The artifact intercept setup is also duplicated between `initStorageLocationIntercepts` and `initErrorIntercepts`. Consider hoisting both to a shared scope.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,274 @@
...
+  if (!transferJob) {
+    return null;
+  }
+
+  const destType = transferJob.destination.type;
+  const sourceType = transferJob.source.type;
+  const namespace = transferJob.namespace || '';
+  const destinationUri = getDestinationUri(transferJob);
+  const sourcePath = getSourcePath(transferJob);
+  const sourceLabel =
```
manaswinidas: Nit (DRY): This inline `sourceLabel` logic mirrors what `getModelUriLabel` does for destinations in `utils.ts`. Consider extracting a parallel `getSourceLabel` utility for consistency and reuse.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -284,18 +301,30 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
...
                     <ModelTimestamp timeSinceEpoch={mv.createTimeSinceEpoch} />
                   </DashboardDescriptionListGroup>
+                  {isTransferJobSource && transferJob && (
+                    <DashboardDescriptionListGroup
+                      title="Stored"
+                      popover="When the model artifact was successfully stored in the specified model location."
+                      isEmpty={!transferJob.lastUpdateTimeSinceEpoch}
+                      contentWhenEmpty="Unknown"
+                    >
+                      <ModelTimestamp timeSinceEpoch={transferJob.lastUpdateTimeSinceEpoch} />
```
manaswinidas, line 325: Nit: `lastUpdateTimeSinceEpoch` is a generic "last modified" timestamp that changes on any job update (metadata edits, status transitions), but the popover says "when the model artifact was successfully stored." Is there a dedicated completion timestamp on the transfer job that would be more accurate here?

YuliaKrimerman: This is from the mocks

YuliaKrimerman:
> > Please add Cypress tests: for no-access view and any errors ?
> 
>>This is not covered yet

@manaswinidas 
with isAccessError removed, the no-access view is now purely driven by useNamespaceSelector. In Cypress (where namespaces are empty), a 404/null response triggers the no-access view. Both 500 and 403 show the generic error alert with retry — which is the correct behavior since we can't distinguish access errors from server errors without structured status codes.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,271 @@
...
+  List,
+  ListItem,
+  Title,
+  Bullseye,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon, FolderIcon } from '@patternfly/react-icons';
+import { DashboardDescriptionListGroup, InlineTruncatedClipboardCopy } from 'mod-arch-shared';
+import { useNamespaceSelector } from 'mod-arch-core';
```
mturley: Bug: `useNamespaceSelector` (from `mod-arch-core`) returns a **filtered** namespace list — when `MANDATORY_NAMESPACE` is set, it only returns that single namespace. This means the access check on line 129 (`namespaces.some(ns => ns.name === fallbackNamespace)`) will incorrectly return `false` whenever the transfer job's namespace differs from the mandatory namespace, showing the "no access" degraded view even when the user has full access.

This is the same class of bug we just fixed in #2326, where `NamespaceSelectorField` was switched from `useNamespaceSelector` to the local `useNamespaces()` hook (which calls the BFF's `/api/v1/namespaces` endpoint directly and returns **all** namespaces the user has access to, bypassing the mandatory namespace filter).

This should use `useNamespaces()` instead:
```ts
import { useNamespaces } from '~/app/hooks/useNamespaces';
// ...
const [namespaces, namespacesLoaded, namespacesLoadError] = useNamespaces();
const userHasNamespaceAccess = namespaces.some((ns) => ns.name === fallbackNamespace);
```

Additionally, the loading and error states from the hook are not being used. While namespaces are loading, `namespaces` will be an empty array, so `userHasNamespaceAccess` will be `false` and the component could briefly flash the "no access" view before namespaces finish loading. Consider showing a spinner (or folding namespace loading into the existing `!transferJobLoaded` spinner) while `namespacesLoaded` is `false`.

clients/ui/frontend/src/app/api/service.ts
```diff
@@ -259,6 +259,23 @@ export const getModelTransferJob =
...
     });
 
+export const getModelTransferJobByName =
+  (hostPath: string, queryParams: Record<string, unknown> = {}) =>
+  (opts: APIOptions, namespace: string, jobName: string): Promise<ModelTransferJob> =>
+    handleRestFailures(
+      restGET(
+        hostPath,
+        `/model_transfer_jobs/${encodeURIComponent(jobName)}`,
+        { ...queryParams, namespace },
```
mturley: Bug: The query parameter name is wrong. The BFF's `getRequiredJobNamespace` reads `r.URL.Query().Get("jobNamespace")` ([model_transfer_job_handler.go:339](https://github.com/kubeflow/model-registry/blob/main/clients/ui/bff/internal/api/model_transfer_job_handler.go#L339)), but this line sends `namespace` as the key.

This causes every request to fail with `"missing required query parameter: jobNamespace"`.

Fix:
```ts
{ ...queryParams, jobNamespace: namespace },
```

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/StorageLocationSection.tsx
```diff
@@ -0,0 +1,273 @@
...
+        connection in{' '}
+      </>
+    )}
+    <Icon size="sm" isInline>
+      <FolderIcon />
+    </Icon>{' '}
+    {namespace}
+    {!hasAccess && <NoAccessPopover namespace={namespace} />}
+  </Content>
+);
```
mturley, line 94: Naming: This is an upstream Kubeflow repo, so we should use Kubernetes terminology rather than OpenShift terminology:

- **"Project" → "namespace"**: Lines 46, 52, 65 use "Project" (OpenShift term). Should be "namespace" (k8s term).
  - `You don't have access to the {namespace} Project.` → `You don't have access to the {namespace} namespace.`
  - `To request access to a new or existing project` → `To request access to a new or existing namespace`
  - `aria-label="More info about project access"` → `aria-label="More info about namespace access"`

- **"connection" → "secret"**: The component name `ConnectionDisplay` and user-facing text throughout this section use "connection" (OpenShift/ODH term for a k8s Secret with specific annotations). In upstream Kubeflow context, these should refer to "secret" instead.
  - Line 85: `connection in` → `secret in`
  - Line 103: `The connection currently being used to store this model` → `The secret currently being used to store this model`
  - Line 233: `Details of the connection used to store the model before it was registered.` → `Details of the secret used to store the model before it was registered.`
  - Line 241: `The connection that was used to store the model at the time it was registered.` → `The secret that was used to store the model at the time it was registered.`
  - Component/testId naming (`ConnectionDisplay`, `storage-connection-display`) should also be updated for consistency.

clients/ui/frontend/src/app/utils.ts
```diff
@@ -118,3 +123,69 @@ export const getDoubleValue = <T extends ModelRegistryCustomProperties>(
...
+};
+
+export const getModelUriPopoverContent = (destType: ModelTransferJobDestinationType): string => {
+  switch (destType) {
+    case ModelTransferJobDestinationType.OCI:
+      return 'The URI of the OCI connection that is currently being used as the model storage location.';
+    case ModelTransferJobDestinationType.S3:
+      return 'The path of the S3-compatible object storage connection that is currently being used as the model storage location.';
+    default:
+      return 'The URI of the connection that is currently being used as the model storage location.';
```
mturley: Same naming issue here — "connection" should be "secret" for upstream Kubeflow:
- `The URI of the OCI connection that is currently being used...` → `The URI of the OCI secret that is currently being used...`
- `The path of the S3-compatible object storage connection...` → `The path of the S3-compatible object storage secret...`
- `The URI of the connection that is currently being used...` → `The URI of the secret that is currently being used...`

PR #2274, opened 2026-02-24
Add retry functionality for failed model transfer jobs

mturley:
## Description

This PR implements the retry functionality for failed model transfer jobs.

Screenshot 2026-02-26 at 3 18 41 PM

Screenshot 2026-02-26 at 3 18 46 PM

### BFF Changes

Fixes the `UpdateModelTransferJob` function to properly preserve metadata when retrying a failed job:

**Problem**: When a user retries a failed transfer job, the existing implementation was not recovering important metadata fields from the original job, causing the retry to lose information like Author, Description, model format, and custom properties.

**Solution**: Added recovery logic to read metadata from:
- Job annotations (Author, Description)
- ConfigMap data (VersionDescription, SourceModelFormat, SourceModelFormatVersion)
- ConfigMap JSON (ModelCustomProperties, VersionCustomProperties)

**Credential Cloning**: When retrying a job without providing new credentials, the BFF clones the credentials from the old job's secrets into **new Secret objects**. This ensures that deleting the old job (via the "Delete failed job" checkbox) doesn't affect the new job's secrets.

**Compatibility**: Updated to use snake_case ConfigMap field names (`model_format_name`, `model_format_version`) per PR #2281.

### Frontend Changes

Implements the UI for retrying failed transfer jobs:

- **RetryJobModal** component with:
  - Auto-generated retry job name (adds `-2`, `-3`, etc. suffix)
  - Editable resource name with K8s DNS-1123 validation
  - Checkbox to delete the failed job after successful retry
  - Error handling and loading states

- **Retry button** next to the "Failed" status label in the jobs table (kebab menu has only Delete action)
- **API integration** with BFF PATCH endpoint supporting `deleteOldJob` query param

### Screenshots

The retry button appears next to the Failed status label. Clicking it opens the retry modal where users can:
1. See the auto-generated new job name
2. Optionally edit the resource name
3. Choose whether to delete the failed job
4. Confirm the retry operation

## How Has This Been Tested?

**BFF Tests**:
- Added new test context `TestModelTransferJob retry metadata preservation` with tests:
  - Verifies PATCH with only new name preserves source, destination, and metadata from old job
  - Verifies PATCH clones credentials from old job into new secrets (creating new Secret objects, not reusing existing ones)
- All existing BFF tests pass: `make test` in `clients/ui/bff/`

**Frontend Tests**:
- Added 17 unit tests for RetryJobModal component covering:
  - Modal rendering and content
  - Auto-generated job name with suffix logic
  - Form validation for K8s names
  - Retry button interaction
  - Error handling and loading states
- TypeScript compilation passes
- ESLint passes
- All unit tests pass: `npm run test:unit` (508 tests)

## Merge criteria:

- [x] Automated tests are provided as part of the PR
- [x] Code changes follow the kubeflow contribution guidelines.

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/repositories/model_transfer_jobs.go
+++ b/clients/ui/bff/internal/repositories/model_transfer_jobs.go
@@ -26,6 +26,22 @@ var (
 	ErrModelRegistryNotFound = errors.New("model registry not found in the selected namespace")
 )
 
+// recoverFromAnnotation copies annotation value to target if target is empty
+func recoverFromAnnotation(target *string, annotations map[string]string, key string) {
+	if *target == "" {
+		*target = annotations[key]
+	}
+}
+
+// recoverEnumFromAnnotation copies annotation value to target enum if target is empty
+func recoverEnumFromAnnotation[T ~string](target *T, annotations map[string]string, key string) {
+	if *target == "" {
+		if val := annotations[key]; val != "" {
+			*target = T(val)
+		}
+	}
+}
+
 func (m *ModelRegistryRepository) GetAllModelTransferJobs(ctx context.Context, client k8s.KubernetesClientInterface, namespace string, modelRegistryID string) (*models.ModelTransferJobList, error) {
 	if modelRegistryID == "" {
 		return &models.ModelTransferJobList{Items: []models.ModelTransferJob{}, Size: 0, PageSize: 0}, nil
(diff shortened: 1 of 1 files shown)


ppadti:
tested locally - when I clicked on edit resource name, the resource name field doesn't appear.

https://github.com/user-attachments/assets/a9fbb610-612a-4d65-83c4-e9b88e86d3fb

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+  };
+
+  return (
+    <Modal isOpen onClose={onClose} variant="small" data-testid="retry-job-modal">
+      <ModalHeader title="Retry model transfer job?" />
+      <ModalBody>
+        <Stack hasGutter>
+          <StackItem>
+            A new transfer job will be created for the{' '}
+            <strong>{job.modelVersionName}</strong> model version.
```
ppadti: I think based on intent we should be changing this `<strong>{job.modelVersionName}</strong>` to model name/version name?

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+  Stack,
+  StackItem,
+  TextInput,
+  Modal,
+  ModalBody,
+  ModalHeader,
+  ModalFooter,
+} from '@patternfly/react-core';
+import { ModelTransferJob } from '~/app/types';
+import ResourceNameDefinitionTooltip from '~/concepts/k8s/ResourceNameDefinitionTootip';
```
ppadti: typo in the file name: `/ResourceNameDefinitionTootip` -> `/ResourceNameDefinitionTooltip`

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
mturley: Good call, fixing.

mturley:
@ppadti 

> tested locally - when I clicked on edit resource name, the resource name field doesn't appear.

oops. at one point it was using the existing K8sNameDescriptionField component instead of trying to write its own, and at some AI iteration I failed to re-review it. AI growing pains, sorry. Fixing it

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+    initialData: { name: generatedName, k8sName: generatedName },
+    editableK8sName: true,
+  });
+  const [deleteOldJob, setDeleteOldJob] = React.useState(true);
+  const [isRetrying, setIsRetrying] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
+
+  const { k8sName } = fieldData;
+  const isNameValid =
+    !k8sName.state.invalidCharacters && !k8sName.state.invalidLength && k8sName.value.length > 0;
```
ppadti, line 71: @mturley - one issue we might encounter here is that - the job name length would be 253 and when we create a job it will create pods and attach job name as label (which should be 63) - we might get an error in creating a pod if the job name exceeds 63.

mturley: Good catch, thanks. Pushed a fix.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+const RetryJobModal: React.FC<RetryJobModalProps> = ({ job, onClose, onRetry }) => {
+  const generatedName = generateRetryJobName(job.name);
+  const { data: fieldData, onDataChange } = useK8sNameDescriptionFieldData({
+    initialData: { name: generatedName },
+    editableK8sName: true,
+    maxK8sNameLength: 63,
+  });
+  const [deleteOldJob, setDeleteOldJob] = React.useState(true);
+  const [isRetrying, setIsRetrying] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
```
manaswinidas, line 61: **Bug: Name length validation mismatch with BFF.**

`maxK8sNameLength` is set to 63 here, but the BFF (`model_transfer_jobs.go` line 322) rejects job names over **50 characters**:

```go
if len(newJobName) > 50 {
    return nil, fmt.Errorf("%w: job name must be 50 characters or less", ErrJobValidationFailed)
}
```

This means names between 51-63 characters will pass client-side validation but fail on the server, giving the user a confusing error.

Since the reason for using 63 is that job names become pod labels (DNS-1123 label limit), the BFF limit should probably be raised to 63 to match. Alternatively, this could be lowered to 50 -- but 50 isn't a standard K8s boundary so 63 seems more correct.

clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -402,6 +402,12 @@ func (m *ModelRegistryRepository) UpdateModelTransferJob(
 	if newPayload.ModelArtifactId == "" {
 		newPayload.ModelArtifactId = oldAnnotations["modelregistry.kubeflow.org/model-artifact-id"]
 	}
+	if newPayload.Author == "" {
+		newPayload.Author = oldAnnotations["modelregistry.kubeflow.org/author"]
+	}
+	if newPayload.Description == "" {
+		newPayload.Description = oldAnnotations["modelregistry.kubeflow.org/description"]
+	}
```
manaswinidas: Nit (DRY): The check-empty-then-copy-from-annotation pattern is repeated ~15 times in this function. A small helper like `recoverFromAnnotation(target *string, annotations map[string]string, key string)` would reduce the repetition.

clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -461,9 +467,43 @@ func (m *ModelRegistryRepository) UpdateModelTransferJob(
...
+			}
+		}
+		if newPayload.VersionCustomProperties == nil {
+			if val, ok := oldConfigMap.Data["ModelVersion.customProperties"]; ok && val != "" {
+				var props map[string]interface{}
+				if err := json.Unmarshal([]byte(val), &props); err == nil {
+					newPayload.VersionCustomProperties = props
+				}
+			}
 		}
```
manaswinidas, line 494: Two things:
1. **Silent failure**: If ConfigMap JSON is malformed, `Unmarshal` error is silently swallowed — no log line. Could make missing custom properties on retry hard to debug. Consider adding `logger.Warn("failed to unmarshal custom properties", ...)`.
2. **DRY**: The `ModelCustomProperties` and `VersionCustomProperties` blocks are structurally identical. A helper like `recoverJSONMapFromConfigMap(target *map[string]interface{}, data map[string]string, key string)` would eliminate the duplication.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+ * e.g., "my-job" -> "my-job-2", "my-job-2" -> "my-job-3"
+ */
+const generateRetryJobName = (originalName: string): string => {
+  const numericSuffixMatch = originalName.match(/^(.+)-(\d+)$/);
+  if (numericSuffixMatch) {
+    const [, baseName, numStr] = numericSuffixMatch;
+    return `${baseName}-${parseInt(numStr, 10) + 1}`;
+  }
+  return `${originalName}-2`;
+};
```
manaswinidas: Nit: `generateRetryJobName` doesn't check if the result exceeds `maxK8sNameLength` (63). A 62-char job name becomes 65 chars after appending `-2`, immediately failing validation. Consider truncating the base name before appending the suffix.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+  });
+  const [deleteOldJob, setDeleteOldJob] = React.useState(true);
+  const [isRetrying, setIsRetrying] = React.useState(false);
+  const [error, setError] = React.useState<Error | undefined>();
+
+  // Trigger initial k8sName generation from name on mount
+  React.useEffect(() => {
+    onDataChange('name', generatedName);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
manaswinidas, line 67: Nit: Triggering `onDataChange` in a mount-only `useEffect` with ESLint suppression is a workaround to kick off k8sName auto-generation. If `useK8sNameDescriptionFieldData` is refactored later, this could silently break. Ideally the hook would support generating `k8sName` from `initialData.name` during initialization.

clients/ui/frontend/src/app/api/service.ts
```diff
@@ -260,13 +260,18 @@ export const createModelTransferJob =
...
+    data: Partial<ModelTransferJob>,
+    additionalQueryParams?: Record<string, unknown>,
+  ): Promise<ModelTransferJob> =>
     handleRestFailures(
       restPATCH(
         hostPath,
         `/model_transfer_jobs/${jobId}`,
         assembleModArchBody(data),
-        queryParams,
+        { ...queryParams, ...additionalQueryParams },
```
manaswinidas, line 274: Nit: Spread merge means `additionalQueryParams` can silently overwrite base params like `namespace`. Low risk today with a single caller, but could cause subtle bugs if more callers are added.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
ppadti: yes, this is a known issue - I am fixing it currently...

clients/ui/frontend/src/app/api/service.ts
mturley: I think maybe this is okay because we might want to be able to override some of the base params per-request. I agree it's subtle though. Not sure exactly the better pattern. Since this pattern is already present elsewhere I'll leave it alone for now.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
mturley: > Ideally the hook would support generating k8sName from initialData.name during initialization

This is how it worked before, but then if the user makes changes to the name before they click "edit resource name" their changes won't be reflected in the resource name field. There's probably a better way to achieve that though, will take a look.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelTransferJobs/RetryJobModal.tsx
manaswinidas: @ppadti Make sure to follow up - thanks!

PR #2277, opened 2026-02-25
Add endpoint, types and API hooks for MCP server catalog

ppadti:
## Description
This PR aims to add endpoint, types, mocks and api hooks for mcp server catalog.
Need to add unit test for hooks - will be done in follow-up PR.

## How Has This Been Tested?
run the application in mock mode and try this curl commands:

List MCP servers

`curl -s -H "kubeflow-userid: user@example.com" \  "http://localhost:4000/api/v1/model_catalog/mcp_servers?namespace=kubeflow"`

MCP filter options
`curl -s -H "kubeflow-userid: user@example.com" \  "http://localhost:4000/api/v1/model_catalog/mcp_servers_filter_options?namespace=kubeflow"`

Get a MCP server
`curl -s -H "kubeflow-userid: user@example.com" \  "http://localhost:4000/api/v1/model_catalog/mcp_servers/prometheus-mcp?namespace=kubeflow"`

MCP server tools
`curl -s -H "kubeflow-userid: user@example.com" \  "http://localhost:4000/api/v1/model_catalog/mcp_servers/prometheus-mcp/tools?namespace=kubeflow"`

modified source endpoint

`curl -s "http://localhost:4000/api/v1/model_catalog/sources?namespace=kubeflow&assetType=models" -H "kubeflow-userid: [user@example.com](mailto:user@example.com)"`

`curl -s "http://localhost:4000/api/v1/model_catalog/sources?namespace=kubeflow&assetType=mcp_servers" -H "kubeflow-userid: [user@example.com](mailto:user@example.com)"`

```
curl -X GET "http://localhost:4000/api/v1/model_catalog/sources?namespace=kubeflow" \
  -H "kubeflow-userid: user@example.com"
```

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/bff/internal/api/mcp_servers_catalog_handler.go
+++ b/clients/ui/bff/internal/api/mcp_servers_catalog_handler.go
@@ -12,7 +12,6 @@ import (
 )
 
 type McpServerListEnvelope Envelope[*models.McpServerList, None]
-type McpServerFilterOptionEnvelope Envelope[*models.FilterOption, None]
 type McpServerFilterOptionsListEnvelope Envelope[*models.FilterOptionsList, None]
 type McpServerEnvelope Envelope[*models.McpServer, None]
 type McpServerToolsListEnvelope Envelope[*models.McpToolList, None]
--- a/clients/ui/bff/internal/repositories/mcp_server_catalog.go
+++ b/clients/ui/bff/internal/repositories/mcp_server_catalog.go
@@ -23,7 +23,7 @@ type McpServerCatalog struct {
 	McpServerCatalogInterface
 }
 
-func (a *CatalogSources) GetAllMcpServers(client httpclient.HTTPClientInterface, pageValues url.Values) (*models.McpServerList, error) {
+func (a *McpServerCatalog) GetAllMcpServers(client httpclient.HTTPClientInterface, pageValues url.Values) (*models.McpServerList, error) {
 	responseData, err := client.GET(UrlWithPageParams(mcpServerPath, pageValues))
 
 	if err != nil {
@@ -40,7 +40,7 @@ func (a *CatalogSources) GetAllMcpServers(client httpclient.HTTPClientInterface,
 
 }
 
-func (a *CatalogSources) GetMcpServersFilter(client httpclient.HTTPClientInterface) (*models.FilterOptionsList, error) {
+func (a *McpServerCatalog) GetMcpServersFilter(client httpclient.HTTPClientInterface) (*models.FilterOptionsList, error) {
 	responseData, err := client.GET(mcpFilterOptionPath)
 
 	if err != nil {
@@ -55,7 +55,7 @@ func (a *CatalogSources) GetMcpServersFilter(client httpclient.HTTPClientInterfa
 	return &filters, nil
 }
 
-func (a *CatalogSources) GetMcpServer(client httpclient.HTTPClientInterface, serverId string) (*models.McpServer, error) {
+func (a *McpServerCatalog) GetMcpServer(client httpclient.HTTPClientInterface, serverId string) (*models.McpServer, error) {
 	path, err := url.JoinPath(mcpServerPath, serverId)
 
 	if err != nil {
@@ -77,7 +77,7 @@ func (a *CatalogSources) GetMcpServer(client httpclient.HTTPClientInterface, ser
 	return &mcpServer, nil
 }
 
-func (a *CatalogSources) GetMcpServersTools(client httpclient.HTTPClientInterface, serverId string) (*models.McpToolList, error) {
+func (a *McpServerCatalog) GetMcpServersTools(client httpclient.HTTPClientInterface, serverId string) (*models.McpToolList, error) {
 	path, err := url.JoinPath(mcpServerPath, serverId, "tools")
 
 	if err != nil {
--- a/clients/ui/frontend/src/app/hooks/mcpServerCatalog/useMcpServersBySourceLabel.ts
+++ b/clients/ui/frontend/src/app/hooks/mcpServerCatalog/useMcpServersBySourceLabel.ts
@@ -12,6 +12,7 @@ type PaginatedMcpServerList = {
   isLoadingMore: boolean;
   hasMore: boolean;
   refresh: () => void;
+  loadMoreError?: Error;
 };
 
 type McpServers = {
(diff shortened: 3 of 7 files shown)


clients/ui/bff/internal/repositories/mcp_server_catalog.go
```diff
@@ -0,0 +1,100 @@
...
+
+type McpServerCatalogInterface interface {
+	GetAllMcpServers(client httpclient.HTTPClientInterface, pageValues url.Values) (*models.McpServerList, error)
+	GetMcpServersFilter(client httpclient.HTTPClientInterface) (*models.FilterOptionsList, error)
+	GetMcpServer(client httpclient.HTTPClientInterface, serverId string) (*models.McpServer, error)
+	GetMcpServersTools(client httpclient.HTTPClientInterface, serverId string) (*models.McpToolList, error)
+}
+
+type McpServerCatalog struct {
+	McpServerCatalogInterface
```
manaswinidas, line 23: **[High] `McpServerCatalog` struct is dead weight — methods defined on wrong receiver**

The `McpServerCatalog` struct embeds `McpServerCatalogInterface` but has no methods of its own. All four MCP methods (`GetAllMcpServers`, `GetMcpServersFilter`, `GetMcpServer`, `GetMcpServersTools`) are defined on `*CatalogSources`, not on `McpServerCatalog`.

Since `ModelCatalogClient` embeds both `CatalogSources` and `McpServerCatalog`, the MCP methods are already promoted from `CatalogSources` — the `McpServerCatalog` embedding contributes nothing.

Suggested fix: Either
1. Remove `McpServerCatalog` struct and its embedding from `ModelCatalogClient`, since `CatalogSources` already satisfies the interface, or
2. Move the methods to `McpServerCatalog` as the receiver to maintain proper domain separation (consistent with how `CatalogModels` and `CatalogSourcePreview` have their own methods and receivers).

Option 2 is preferable for maintainability — mixing MCP server concerns into `CatalogSources` muddies the domain boundary.

clients/ui/frontend/src/app/hooks/mcpServerCatalog/useMcpServersBySourceLabel.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (loaded && !error) {
+      setAllItems(firstPageData.items ?? []);
+      setTotalSize(firstPageData.size);
+      setNextPageTokenValue(firstPageData.nextPageToken);
+    }
+  }, [firstPageData, loaded, error]);
+
+  const loadMore = React.useCallback(async () => {
+    if (!nextPageTokenValue || isLoadingMore || !apiAvailable) {
+      return;
```
manaswinidas, line 103: **[High] Unhandled promise rejection in `loadMore`**

`loadMore` is an `async` function that re-throws inside its `catch` block. Since it's called directly (e.g. via a button click handler or similar) with no `.catch()` at the call site, this re-thrown error becomes an **unhandled promise rejection**, which can crash the app or silently fail.

```ts
} catch (err) {
  throw new Error(...);
}
```

Suggested fix: Instead of re-throwing, store the error in state so consumers can react to it:
```ts
const [loadMoreError, setLoadMoreError] = React.useState<Error | undefined>();

// in catch:
setLoadMoreError(new Error(`Failed to load more servers: ${err instanceof Error ? err.message : String(err)}`));
```

Or propagate it through the return value so the caller can handle it.

clients/ui/bff/internal/api/mcp_servers_catalog_handler.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/kubeflow/model-registry/ui/bff/internal/constants"
+	"github.com/kubeflow/model-registry/ui/bff/internal/integrations/httpclient"
+	"github.com/kubeflow/model-registry/ui/bff/internal/models"
+)
+
+type McpServerListEnvelope Envelope[*models.McpServerList, None]
```
manaswinidas, line 14: **[Medium] Unused type alias `McpServerFilterOptionEnvelope`**

`McpServerFilterOptionEnvelope` is declared here but never used anywhere in this PR or the codebase. Only `McpServerFilterOptionsListEnvelope` (line 15) is actually used in `GetMcpServersFiltersHandler`. This dead type should be removed to avoid confusion.

clients/ui/frontend/src/app/mcpServerCatalogTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+export type McpListParams = {
+  size: number;
+  pageSize: number;
+  nextPageToken: string;
+};
+
+export type McpServerList = McpListParams & { items?: McpServer[] };
+
+export type McpToolWithServer = {
+  serverId: string;
```
manaswinidas: **[Medium] Duplicated pagination type — `McpListParams` duplicates `ModelCatalogListParams`**

`McpListParams` is structurally identical to the existing `ModelCatalogListParams` in `modelCatalogTypes.ts`:

```ts
// modelCatalogTypes.ts (already exists)
export type ModelCatalogListParams = {
  size: number;
  pageSize: number;
  nextPageToken: string;
};
```

`McpServerList` and `McpToolList` should reuse `ModelCatalogListParams` instead of duplicating it. Consider either importing `ModelCatalogListParams` directly, or extracting a shared `PaginationParams` type that both domains reference.

clients/ui/frontend/src/app/hooks/mcpServerCatalog/useMcpServersBySourceLabel.ts
```diff
@@ -0,0 +1,157 @@
...
+      setNextPageTokenValue(response.nextPageToken);
+    } catch (err) {
+      throw new Error(
+        `Failed to load more servers: ${err instanceof Error ? err.message : String(err)}`,
+      );
+    } finally {
+      setIsLoadingMore(false);
+    }
+  }, [api, apiAvailable, buildMcpServerListParams, isLoadingMore, nextPageTokenValue]);
+
```
manaswinidas, line 126: **[Medium] State reset race condition — stale data flash on filter change**

There are two `useEffect` hooks that interact problematically:

1. **Data sync effect** (lines 69-75): Sets `allItems` from `firstPageData` whenever it changes.
2. **Reset effect** (lines 99-114): Clears state when filter/sort params change.

When filters change, both effects fire in the same render cycle. The reset effect clears state, but the data sync effect can re-apply stale `firstPageData` (which hasn't updated yet because the new fetch hasn't completed). This causes a brief flash of stale data before the fresh fetch resolves.

Suggested fix: Use a ref to track the current "query key" and discard results from stale queries:
```ts
const queryKeyRef = React.useRef(0);

// In reset effect: queryKeyRef.current += 1;
// In data sync effect: check queryKeyRef.current matches before applying
```

Alternatively, merge both effects into one that resets state and lets `useFetchState` handle the fresh fetch, avoiding the intermediate stale-data render.

manaswinidas:
Tested in mock mode- all endpoints seem to be returning correct mock data

clients/ui/bff/internal/repositories/mcp_server_catalog.go
ppadti: ops, I just copy pasted the handler from source and updated that - thanks, will fix it.

PR #2279, opened 2026-02-25
Add YAML preview drawer

Philip-Carneiro:
## Description

- Added a hyperlink to open a drawer with sample YAML in the field where we drag'nDrop a YAML file
- Added a drawer to "pop-up" from right with sample YAML
- Sample YAML retrieved from existing file in the path

Without PF-mode
image

With PF-mode
image

## How Has This Been Tested?

This was manually tested and created tests for it

## Merge criteria:

- [ x The developer has manually tested the changes and verified that the changes work.

If you have UI changes


clients/ui/frontend/config/webpack.common.js
```diff
@@ -165,10 +169,6 @@ module.exports = (env) => ({
           'sass-loader',
         ],
       },
-      {
-        test: /\.ya?ml$/,
-        use: 'js-yaml-loader',
-      },
     ],
   },
```
mturley, line 173: This is probably fine, but we should be cautious with any changes to webpack config. I wonder what we were using that relied on `js-yaml-loader`? we should also perhaps try testing this change in ODH before we merge it here (we can use the `--pr` flag of the sync script), since that is a different webpack build environment, and we should make sure we're not breaking any other functionality with this change.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/expectedYamlFormatContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  return bounds;
+}
+
+type ExpectedYamlFormatDrawerProps = {
+  isOpen: boolean;
+  onClose: () => void;
+};
+
+const ExpectedYamlFormatDrawer: React.FC<ExpectedYamlFormatDrawerProps> = ({ isOpen, onClose }) => {
```
mturley: This custom Drawer implementation is probably unnecessary. Did you try using the [PatternFly Drawer component](https://www.patternfly.org/components/drawer)?

Whenever we find ourselves writing a bunch of custom layout code like this (especially if it contains a lot of inline styles) we should look into whether we can use PatternFly components instead.

Also, React component files should have capitalized names for consistency.

clients/ui/frontend/config/webpack.common.js
Philip-Carneiro: inside the actual repo, this is not used, but need to check the ODH

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/YamlSection.tsx
```diff
@@ -17,12 +19,19 @@ import {
   HELP_TEXT,
 } from '~/app/pages/modelCatalogSettings/constants';
 
+const EXPECTED_FORMAT_LINK_TEXT = 'View expected file format';
```
manaswinidas: DRY: This is the same string as `EXPECTED_FORMAT_DRAWER_TITLE` in `ExpectedYamlFormatDrawer.tsx`. Either import and reuse that constant, or move both to `constants.tsx` which already centralizes `FORM_LABELS`, `VALIDATION_MESSAGES`, `HELP_TEXT`, etc.

clients/ui/Makefile
```diff
@@ -76,31 +76,31 @@ dev-bff-federated:
 
 .PHONY: docker-build
 docker-build:
-	$(CONTAINER_TOOL) build --build-arg DEPLOYMENT_MODE=kubeflow --build-arg STYLE_THEME=mui-theme -t ${IMG_UI} .
+	$(CONTAINER_TOOL) build -f Dockerfile --build-arg UI_SOURCE_CODE=clients/ui/frontend --build-arg BFF_SOURCE_CODE=clients/ui/bff --build-arg DEPLOYMENT_MODE=kubeflow --build-arg STYLE_THEME=mui-theme -t ${IMG_UI} ../..
```
manaswinidas: DRY: `--build-arg UI_SOURCE_CODE=clients/ui/frontend --build-arg BFF_SOURCE_CODE=clients/ui/bff` is repeated in all 7 docker targets. Extract to a variable:

```makefile
DOCKER_BUILD_ARGS := --build-arg UI_SOURCE_CODE=clients/ui/frontend --build-arg BFF_SOURCE_CODE=clients/ui/bff
```

Then use `$(DOCKER_BUILD_ARGS)` in each target.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ExpectedYamlFormatDrawer.tsx
```diff
@@ -0,0 +1,180 @@
...
+        </DrawerContent>
+      </Drawer>
+    </div>
+  );
+
+  return (
+    <>
+      {children}
+      {createPortal(overlay, document.body)}
+    </>
```
manaswinidas: The `createPortal` + fixed positioning + `useContainerBounds` approach is a lot of custom layout to drive a PatternFly `Drawer`. The PF `Drawer` is designed to sit in the component tree where it renders — wrapping `ManageSourceForm` at the page level (which you're already doing in `ManageSourcePage`) should be sufficient without needing a portal or manual bounds tracking. Could you try placing the `Drawer` directly in the tree without portaling?

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ExpectedYamlFormatDrawer.tsx
```diff
@@ -0,0 +1,180 @@
...
+        }
+      };
+      applyPanelMaxWidth();
+      const id = window.setTimeout(applyPanelMaxWidth, 0);
+      const observer = new MutationObserver(applyPanelMaxWidth);
+      observer.observe(el, { childList: true, subtree: true });
+      return () => {
+        window.clearTimeout(id);
+        observer.disconnect();
+      };
```
manaswinidas: Fragile: querying by internal PF class name `.pf-v6-c-drawer__panel` will break silently on PatternFly version upgrades. If the MUI theme needs max-width overrides, consider using PF's `widths` prop on `DrawerPanelContent` (which you're already using above) or a scoped CSS class instead of direct DOM manipulation via `querySelector` + `MutationObserver`.

clients/ui/frontend/src/app/pages/modelCatalogSettings/screens/ManageSourcePage.tsx
```diff
@@ -42,10 +44,36 @@ const ManageSourcePage: React.FC = () => {
...
+            style={{
+              position: 'relative',
+              flex: 1,
+              minHeight: 0,
+              display: 'flex',
+              flexDirection: 'column',
+              width: '100%',
+            }}
+          >
+            <ManageSourceForm
```
manaswinidas, line 63: Nit: The prop is named `onOpen...` but the callback toggles (`(prev) => !prev`). Either rename to `onToggleExpectedFormatDrawer` to match the behavior, or keep the `onOpen` name and always set `true` here (letting the drawer's `onClose` handle closing).

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/__tests__/ExpectedYamlFormatDrawer.spec.tsx
manaswinidas, line 1: Good happy-path coverage. Missing test scenarios:
- **MUI theme branch**: The `isMUITheme` code path (CSS variable override, `MutationObserver`) has no coverage.
- **Missing container element**: When `#primary-app-container` doesn't exist in the DOM, the drawer silently doesn't render — worth a test to verify graceful degradation.
- **Page-level integration**: No test for `ManageSourcePage` verifying the open/close state wiring between the drawer and the form's link button.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ExpectedYamlFormatDrawer.tsx
```diff
@@ -0,0 +1,180 @@
...
+      </DrawerActions>
+    </DrawerHead>
+    <DrawerPanelBody style={{ flex: 1, minHeight: 200, overflow: 'auto' }}>
+      <CodeBlock>
+        <CodeBlockCode>{sampleCatalogYamlContent}</CodeBlockCode>
+      </CodeBlock>
+    </DrawerPanelBody>
+  </DrawerPanelContent>
+);
+
```
manaswinidas: This ID isn't defined or referenced anywhere else in the codebase. If the element doesn't exist at runtime, `bounds` stays `null` and the drawer silently never renders with no user feedback. Consider either:
- Importing this ID from a shared constant where the container is actually defined
- Using a React ref passed from the parent instead of a DOM ID lookup
- At minimum, adding a `console.warn` when the container isn't found

clients/ui/frontend/config/webpack.common.js
```diff
@@ -234,9 +234,14 @@ module.exports = (env) => ({
...
   ],
   resolve: {
-    extensions: ['.js', '.ts', '.tsx', '.jsx'],
+    extensions: ['.js', '.ts', '.tsx', '.jsx', '.yaml'],
     alias: {
       '~': path.resolve(SRC_DIR),
+      ...(process.env.SAMPLE_CATALOG_YAML_PATH && {
+        '~/app/pages/modelCatalogSettings/sample-catalog.yaml': path.resolve(
+          process.env.SAMPLE_CATALOG_YAML_PATH,
+        ),
```
manaswinidas, line 244: This env var override is undocumented. A brief comment here (or in the README/contributing guide) explaining when and why a developer would set `SAMPLE_CATALOG_YAML_PATH` would help — e.g., "Set to override the bundled sample YAML with a custom path at build time."

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ExpectedYamlFormatDrawer.tsx
```diff
@@ -0,0 +1,112 @@
...
+  boxShadow: 'var(--pf-t--global--box-shadow--lg--left)',
+};
+
+type ExpectedYamlFormatDrawerProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  children: React.ReactNode;
+};
+
+const ExpectedYamlFormatDrawer: React.FC<ExpectedYamlFormatDrawerProps> = ({
```
YuliaKrimerman: PatternFly has a full Drawer component (@patternfly/react-core) with DrawerContent, DrawerPanelContent, DrawerContentBody, etc. that handles positioning, overlay, animation, focus management, and accessibility out of the box. This PR manually implements all of that with createPortal, custom position: absolute, zIndex: 400, width: 50%, and direct DOM manipulation via container.style.position = 'relative'. Was there a specific reason we chose to go with inline heavy implementation?

clients/ui/frontend/src/app/pages/modelCatalogSettings/screens/ManageSourcePage.tsx
```diff
@@ -42,10 +44,36 @@ const ManageSourcePage: React.FC = () => {
...
     >
-      <ManageSourceForm
-        existingSourceConfig={existingSourceConfig || undefined}
-        isEditMode={!isAddMode}
-      />
+      <ExpectedYamlFormatDrawer
+        isOpen={isExpectedFormatDrawerOpen}
+        onClose={() => setIsExpectedFormatDrawerOpen(false)}
+      >
+        <PageSection
```
YuliaKrimerman: A lot of inline styles we should avoid. 
The original   
<ManageSourceForm
        existingSourceConfig={existingSourceConfig || undefined}
         isEditMode={!isAddMode} should be preserved this simplicity. Adding a PageSection with component="main" also introduces a second <main> element on the page (PF's Page already renders <main>), which is an HTML/accessibility violation.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/YamlSection.tsx
```diff
@@ -75,23 +83,43 @@ const YamlSection: React.FC<YamlSectionProps> = ({ formData, setData }) => {
...
     <FormSection data-testid="yaml-section">
-      <FormGroup label={FORM_LABELS.YAML_CONTENT} isRequired fieldId="yaml-content">
-        <FormFieldset component={yamlInput} field="YAML" />
-        <FormHelperText>
-          <HelperText>
-            <HelperTextItem>{HELP_TEXT.YAML}</HelperTextItem>
-          </HelperText>
-        </FormHelperText>
-        {isYamlTouched && !isYamlContentValid && (
+      <div style={{ position: 'relative' }}>
```
YuliaKrimerman: We shouldn't use absolute positioning for layout. This approach is fragile -- the button will overlap the FormGroup label if the text is long enough, or if the viewport is narrow. Instead, we could use a Flex or Split component to position the label and the link side by side:
`<FormSection data-testid="yaml-section">
  <FormGroup
    label={
      <Flex justifyContent={{ default: 'justifyContentSpaceBetween' }}>
        <FlexItem>{FORM_LABELS.YAML_CONTENT}</FlexItem>
        {onToggleExpectedFormatDrawer && (
          <FlexItem>
            <Button
              variant="link"
              isInline
              onClick={onToggleExpectedFormatDrawer}
              data-testid="view-expected-yaml-format-link"
              icon={<OpenDrawerRightIcon />}
              iconPosition="end"
            >
              {EXPECTED_YAML_FORMAT_LABEL}
            </Button>
          </FlexItem>
        )}
      </Flex>
    }
    isRequired
    fieldId="yaml-content"
  >
    ...
  </FormGroup>
</FormSection>`

mturley:
+1 to @YuliaKrimerman 's comments. We need to make sure this is not duplicating complexity PatternFly can handle for us, and not using custom styles that could break when we upgrade PatternFly. I know the custom drawer implementation probably took considerable effort, but we should remove it and try to use all the PF Drawer components Yulia is referring to. There are examples in PF documentation.

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/ExpectedYamlFormatDrawer.tsx
Philip-Carneiro: @YuliaKrimerman @mturley 
The Patternfly drawer is not functional for this specific page.
This change is necessary for the following reasons:

- The drawer is not in the correct position relative to the page, but rather to the form, leaving a white space at the top until it reaches the height of the form.
- The footer overlaps the drawer.
- For non-PF mode, the drawer has 25% of the page width, making it very crooked and not correctly aligned with the proposed UX design.

clients/ui/frontend/src/app/pages/modelCatalogSettings/screens/ManageSourcePage.tsx
Philip-Carneiro: the page section was added to be able to make the position of the drawer inside of the application page stay in the correct position, this was necessary to the drawer stays in the inside of the container as the designs presented

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/YamlSection.tsx
Philip-Carneiro: as commented in other comments above

clients/ui/frontend/src/app/pages/modelCatalogSettings/components/YamlSection.tsx
```diff
@@ -75,23 +83,43 @@ const YamlSection: React.FC<YamlSectionProps> = ({ formData, setData }) => {
...
+            }}
+          >
+            {EXPECTED_YAML_FORMAT_LABEL}
+          </Button>
         )}
-      </FormGroup>
+      </div>
     </FormSection>
   );
 };
```
mturley, line 132: nit: The absolute positioning here (`position: relative` wrapper + `position: absolute` button) is fragile — the button will overlap the FormGroup label if the text is long or the viewport is narrow.

Consider using PF `Flex` in the `FormGroup` `label` prop instead, which keeps them in the same flow:

```tsx
<FormGroup
  label={
    <Flex justifyContent={{ default: "justifyContentSpaceBetween" }} alignItems={{ default: "alignItemsCenter" }}>
      <FlexItem>{FORM_LABELS.YAML_CONTENT}</FlexItem>
      {onToggleExpectedFormatDrawer && (
        <FlexItem>
          <Button
            variant="link"
            isInline
            onClick={onToggleExpectedFormatDrawer}
            data-testid="view-expected-yaml-format-link"
            icon={<OpenDrawerRightIcon />}
            iconPosition="end"
          >
            {EXPECTED_YAML_FORMAT_LABEL}
          </Button>
        </FlexItem>
      )}
    </Flex>
  }
  isRequired
  fieldId="yaml-content"
>
```

This eliminates the wrapper `<div>` and all three inline style props.

PR #2308, opened 2026-03-04
feat: add /ok-to-test workflow to approve fork PR runs

manaswinidas:
## Description

Add `gh-workflow-approve.yml` workflow to approve pending workflow runs for fork PRs, following the [kubeflow/pipelines](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/gh-workflow-approve.yml) pattern.

When external contributors open PRs from forks, their workflow runs require manual approval. This workflow automates that:
- Auto-approves workflow runs for trusted contributors (org members, owners, collaborators) via `author_association`
- Approves workflow runs when a reviewer adds the `ok-to-test` label (added by Prow when someone comments `/ok-to-test`)
- Re-approves on new commits (`synchronize`) if the `ok-to-test` label is already present

## How Has This Been Tested?

This is a GitHub Actions workflow — it cannot be tested locally. The workflow logic follows the same pattern already running in production in `kubeflow/pipelines`. Testing will occur on the first fork PR after merge by having a reviewer comment `/ok-to-test`.

## Merge criteria:

### DIFF
--- a/.github/workflows/gh-workflow-approve.yml
+++ b/.github/workflows/gh-workflow-approve.yml
@@ -3,105 +3,69 @@ name: Approve Workflow Runs
 permissions:
   actions: write
   contents: read
-  pull-requests: write
 
 on:
-  issue_comment:
-    types: [created]
   pull_request_target:
     types:
       - labeled
       - synchronize
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.number }}
   cancel-in-progress: true
 
 jobs:
   ok-to-test:
-    if: |
-      (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '/ok-to-test')) ||
-      (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'ok-to-test'))
+    if: contains(github.event.pull_request.labels.*.name, 'ok-to-test') || github.event_name == 'pull_request_target'
     runs-on: ubuntu-latest
     continue-on-error: true
 
     steps:
-      - name: Checkout repository
-        if: github.event_name == 'issue_comment'
-        uses: actions/checkout@v6
-
-      - name: Check if commenter is in OWNERS file
-        if: github.event_name == 'issue_comment'
-        id: owners-check
+      - name: Check if author is a Kubeflow GitHub member
+        id: membership-check
         env:
-          COMMENTER: ${{ github.event.comment.user.login }}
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-workflow-approve.yml
```diff
@@ -0,0 +1,107 @@
...
+      - synchronize
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number }}
+  cancel-in-progress: true
+
+jobs:
+  ok-to-test:
+    if: |
+      (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '/ok-to-test')) ||
```
jonburdo: Do we need to actually have this check? prow should add the `ok-to-test` label when a reviewer comments `/ok-to-test`. I see other components have similar files that are all a little bit different, but they mostly just seem to check for the label.

.github/workflows/gh-workflow-approve.yml
```diff
@@ -0,0 +1,107 @@
...
+    continue-on-error: true
+
+    steps:
+      - name: Checkout repository
+        if: github.event_name == 'issue_comment'
+        uses: actions/checkout@v6
+
+      - name: Check if commenter is in OWNERS file
+        if: github.event_name == 'issue_comment'
+        id: owners-check
```
jonburdo: Similar to above, I'm thinking we can check the label and do the rest more similarly to how pipelines or sdk do this.

jonburdo:
The pipelines file you linked might actually be a good one to follow closely. I see other ones usually do check the kubeflow membership using an api, or similar: https://github.com/kubeflow/pipelines/pull/12698

.github/workflows/gh-workflow-approve.yml
pboyd: I'll second this. I don't think we need anything special. (I know we had discussed limiting this to reviewers/approvers, but if kubeflow members is the org policy, then we should defer to that.)

.github/workflows/gh-workflow-approve.yml
manaswinidas: Yes, prow does this already. Removed now.

.github/workflows/gh-workflow-approve.yml
manaswinidas: Thanks for your review. The comments are now addressed

jonburdo:
Thanks again @manaswinidas! This will be great to have :smile: 

For reference: I see this now follows the pipelines [current workflow file](https://github.com/kubeflow/pipelines/blob/479369e8579f748966b00a5e8f9829f00cdad2d8/.github/workflows/gh-workflow-approve.yml), as mentioned, with only a small code comment/whitespace change.

/lgtm
/approve

PR #2310, opened 2026-03-04
BFF: read async-upload job image from ConfigMap

mturley:
## Description

In federated deployment mode, the BFF now reads the async-upload job container image URI from the `model-registry-ui-config` ConfigMap at runtime on each job creation/update request. This allows the image to be configured per-environment (upstream kubeflow, midstream ODH, downstream RHOAI) without code changes.

If the ConfigMap is absent or doesn't contain the expected key, the BFF falls back to the default hard-coded image (`ghcr.io/kubeflow/model-registry/job/async-upload:latest`). In non-federated modes (kubeflow, standalone), the ConfigMap is not consulted at all.

The ConfigMap is deployed by the dashboard manifests — see [odh-dashboard PR #6466](https://github.com/opendatahub-io/odh-dashboard/pull/6466).

### Design

Following the existing codebase pattern, `isFederatedMode` and `podNamespace` are threaded as parameters from the handler layer rather than stored as struct fields on `ModelRegistryRepository`. This keeps the repository layer mode-agnostic and avoids a second source of truth for `isFederatedMode`.

Additionally, `getModelRegistryAddress` now accepts `isFederatedMode` as a parameter instead of hardcoding `true` when calling `GetModelRegistryWithMode`. This fixes a bug where the registry address lookup always used federated mode regardless of the actual deployment configuration.

## How Has This Been Tested?

- `go build ./...` — compiles successfully
- `go test ./...` — all tests pass
- Unit tests for `resolveAsyncUploadImage` covering all branches: non-federated mode, empty namespace, ConfigMap missing, ConfigMap present with key set, key missing, key empty, and whitespace-only values

## Merge criteria:

### DIFF
--- a/clients/ui/bff/internal/repositories/model_transfer_jobs.go
+++ b/clients/ui/bff/internal/repositories/model_transfer_jobs.go
@@ -286,7 +286,7 @@ func (m *ModelRegistryRepository) createModelTransferJobResources(
 
 	logger := helper.GetContextLogger(ctx)
 
-	modelRegistryAddress, err := m.getModelRegistryAddress(ctx, client, namespace, modelRegistryID)
+	modelRegistryAddress, err := m.getModelRegistryAddress(ctx, client, namespace, modelRegistryID, isFederatedMode)
 	if err != nil {
 		return nil, err
 	}
@@ -613,8 +613,8 @@ func (m *ModelRegistryRepository) DeleteModelTransferJob(ctx context.Context, cl
 // getModelRegistryAddress returns the registry address for use in transfer job env.
 // It uses federated/external address when available (e.g. Route URL from Service annotation)
 // so the job pod can reach the registry via the ingress path when NetworkPolicy restricts direct ClusterIP access.
-func (m *ModelRegistryRepository) getModelRegistryAddress(ctx context.Context, client k8s.KubernetesClientInterface, namespace, modelRegistryID string) (string, error) {
-	modelRegistry, err := m.GetModelRegistryWithMode(ctx, client, namespace, modelRegistryID, true)
+func (m *ModelRegistryRepository) getModelRegistryAddress(ctx context.Context, client k8s.KubernetesClientInterface, namespace, modelRegistryID string, isFederatedMode bool) (string, error) {
+	modelRegistry, err := m.GetModelRegistryWithMode(ctx, client, namespace, modelRegistryID, isFederatedMode)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return "", ErrModelRegistryNotFound


clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -612,8 +620,27 @@ func (m *ModelRegistryRepository) getModelRegistryAddress(ctx context.Context, c
...
+			"configmap", asyncUploadConfigMapName, "error", err)
+		return DefaultAsyncUploadImage
+	}
+	if img, ok := cm.Data[asyncUploadConfigMapKey]; ok && img != "" {
+		return img
+	}
+	logger.Info("ConfigMap key not found, using default async-upload image",
+		"configmap", asyncUploadConfigMapName, "key", asyncUploadConfigMapKey)
+	return DefaultAsyncUploadImage
+}
```
manaswinidas: This method has multiple branches (non-federated, empty namespace, ConfigMap missing, key missing, key empty, happy path) and none of them are tested. All test changes in this PR are mechanical signature updates passing `false, ""`. Please add unit tests for at least: federated mode with ConfigMap present + key set, ConfigMap missing (fallback), ConfigMap present but key missing (fallback).

clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -612,8 +620,27 @@ func (m *ModelRegistryRepository) getModelRegistryAddress(ctx context.Context, c
...
+	logger := helper.GetContextLogger(ctx)
+	cm, err := client.GetConfigMap(ctx, m.PodNamespace, asyncUploadConfigMapName)
+	if err != nil {
+		logger.Info("ConfigMap not found, using default async-upload image",
+			"configmap", asyncUploadConfigMapName, "error", err)
+		return DefaultAsyncUploadImage
+	}
+	if img, ok := cm.Data[asyncUploadConfigMapKey]; ok && img != "" {
+		return img
+	}
```
manaswinidas: The `img != ""` check doesn't catch whitespace-only values (e.g. `"  "`). A misconfigured ConfigMap with a whitespace value would pass through and cause a confusing pod scheduling failure.

```suggestion
	if img, ok := cm.Data[asyncUploadConfigMapKey]; ok && strings.TrimSpace(img) != "" {
		return strings.TrimSpace(img)
	}
```

clients/ui/bff/internal/repositories/model_registry.go
```diff
@@ -11,10 +11,15 @@ import (
 )
 
 type ModelRegistryRepository struct {
+	IsFederatedMode bool
+	PodNamespace    string
 }
```
manaswinidas: suggestion: These fields are exported but only accessed within the `repositories` package. Consider making them unexported (`isFederatedMode`, `podNamespace`) to prevent external mutation.

clients/ui/bff/internal/repositories/model_transfer_jobs.go
```diff
@@ -612,8 +620,27 @@ func (m *ModelRegistryRepository) getModelRegistryAddress(ctx context.Context, c
...
+	logger.Info("ConfigMap key not found, using default async-upload image",
+		"configmap", asyncUploadConfigMapName, "key", asyncUploadConfigMapKey)
```
manaswinidas: suggestion: ConfigMap exists but the expected key is missing — this likely indicates a misconfiguration rather than an absent ConfigMap. `Warn` would be more appropriate than `Info` here to make it visible in operational monitoring.

ppadti (commented):
Just a minor suggestion, let me know what you think:
The existing codebase consistently passes isFederatedMode as a parameter from the handler layer (see middleware.go, model_registry_handler.go) so the repository layer stays mode-agnostic. Adding it as a struct field creates a second source of truth for the same concept. To stay consistent with the pattern, isFederatedMode should be threaded as a parameter with the handler passing app.config.DeploymentMode.IsFederatedMode()

mturley:
@ppadti I think I addressed your comment, let me know if that's what you meant

ppadti:
yeah the changes lgtm
But I think I have a made a mistake in one of my PR previously - could you just fix that. Sorry, I tested that midstream and directly copy pasted that in here, my bad. It also worked upstream due to the fallback logic so didn't notice that mistake, sorry...
the changes I wanted are - in `createModelTransferJobResources`: the `getModelRegistryAddress` should be passing this is `isFederatedMode` and in here the `true` should be replaced by  `isFederatedMode` https://github.com/kubeflow/model-registry/blob/ddcd237e6136aa3acaa1a99719a4c6895fee15a8/clients/ui/bff/internal/repositories/model_transfer_jobs.go#L605

PR #2286, opened 2026-02-26
changed selector icon and updated mod-arch packages

rsun19:
## Description

Changed the selector icon in the Model Registry page and updated the `mod-arch` packages

Screenshot 2026-02-26 at 4 45 09 PM

## How Has This Been Tested?

Manual testing

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [x] The commits have meaningful messages
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).
- [ ] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/model-registry/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

- [ ] The developer has added tests or explained why testing cannot be added.
- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

PR #2314, opened 2026-03-04
Add MCP Catalog landing page and route

Philip-Carneiro:
## Description

- Added new page, MCP Catalog
- Added filter and MCP sample cards
- Add routes
- Add sidebar filters
- Add search filter
- Connect to API hooks

Non-PF mode
Screenshot 2026-03-04 at 17 19 34

PF mode
Screenshot 2026-03-04 at 17 20 51

## How Has This Been Tested?

It was manually tested, and test files are created

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/mcpCatalog/types/mcpServer.ts
```diff
@@ -0,0 +1,11 @@
+export type McpServerDeploymentMode = 'Remote' | 'Local';
+
+export type McpServerMock = {
```
mturley: This type doesn't match the actual shape of the objects we'll be getting from the API, and it has `Mock` in the name. If you rebase your branch you'll find that https://github.com/kubeflow/model-registry/pull/2277 added a file at `clients/ui/frontend/src/app/mcpServerCatalogTypes.ts` with the real types the BFF will respond with. See the `McpServer` type there - that's what we should be using throughout this UI. At the very least for this PR let's change your mock data to use the real types. Looks like we have two redundant types files now too, maybe we use that other one and remove this one.

That PR also added the necessary hooks to fetch this data - perhaps that should be a followup PR, but I think it's in scope for your issue. You should be able to set up your McpCatalogContext so that it is actually fetching the source labels and MCP servers using the new hooks in `clients/ui/frontend/src/app/hooks/mcpServerCatalog/*`. The pattern should be the same as what we do for model catalog: fetch sources, identify the unique labels among them, use those to populate the source label selector (instead of the hard coded categories you have in McpCatalogSourceLabelSelector.tsx), and finally have the context fetch sources based on the selected label.

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
+
+export type McpCatalogPaginationState = {
+  page: number;
+  pageSize: number;
+  totalItems: number;
+};
+
+export type McpCatalogFiltersState = Record<string, unknown>;
```
mturley: This type is too loose - mapping to `unknown` means anything that accesses these filters isn't getting value types narrowed by key. Also, we should move this type into one of the types files instead of being here in the context file.

You already have `McpFilterCategoryKey`, isn't this state mapping the same keys to particular filter values? We should have a type similar to the `ModelCatalogFilterStates` type.

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export type McpCatalogCategoryId = 'all' | 'sample' | 'other';
```
mturley: We can't know what the categories will be since they are configurable by the admin. We just need to use `string` for the selected category - and I think we should be consistent with ModelCatalogContext and call it `selectedSourceLabel` because that's what it actually is. In general let's take another look comparing your stuff with model catalog and make sure we're matching existing patterns where possible because later on we may want to try to deduplicate a bunch of this code.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogStringFilter.tsx
```diff
@@ -0,0 +1,113 @@
+import { Button, Checkbox, Content, ContentVariants, SearchInput } from '@patternfly/react-core';
+import * as React from 'react';
+import { McpCatalogContext } from '~/app/context/mcpCatalog/McpCatalogContext';
+import type {
+  McpFilterCategoryKey,
+  McpCatalogFilterStringOption,
+} from '~/app/pages/mcpCatalog/types/mcpCatalogFilterOptions';
+import { MCP_FILTER_MAX_VISIBLE } from '~/app/pages/mcpCatalog/constants/mcpCatalogFilterOptions';
+
+function useMcpFilterState(filterKey: McpFilterCategoryKey) {
```
mturley: It feels weird for this hook to be part of this component since it is generic and could be used for other types of filters later.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogStringFilter.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type McpCatalogStringFilterProps = {
+  title: string;
+  filterKey: McpFilterCategoryKey;
+  filters: McpCatalogFilterStringOption | undefined;
+  showSearch?: boolean;
+};
+
+const McpCatalogStringFilter: React.FC<McpCatalogStringFilterProps> = ({
```
mturley, line 15: This component duplicates a lot of very generic stuff we have in ModelCatalogStringFilter. Really the only thing specific to model catalog in that component is the types and the context. It would be easy to lift out a generic CatalogStringFilter (maybe we should make a place for components that are common to both catalog pages like `clients/ui/frontend/src/app/shared/components/catalog`?) that just works with string types, and have ModelCatalogStringFilter and McpCatalogStringFilter be thin wrappers around it that just make its exposed types more narrow. We should also preserve the behavior of that component - for example, you have the `showSearch` boolean as a prop, but the original component always shows a search if there are more than 5 filter values.

I know we don't want to spend too much time trying to perfectly reuse code, but what I don't want to do is have the two catalogs diverge in implementation - small code reuse wins like this are a good opportunity.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogFilters.tsx
```diff
@@ -0,0 +1,26 @@
...
+import DeploymentModeFilter from '~/app/pages/mcpCatalog/components/globalFilters/DeploymentModeFilter';
+import SupportedTransportsFilter from '~/app/pages/mcpCatalog/components/globalFilters/SupportedTransportsFilter';
+import McpLicenseFilter from '~/app/pages/mcpCatalog/components/globalFilters/McpLicenseFilter';
+import LabelsFilter from '~/app/pages/mcpCatalog/components/globalFilters/LabelsFilter';
+import SecurityVerificationFilter from '~/app/pages/mcpCatalog/components/globalFilters/SecurityVerificationFilter';
+
+const McpCatalogFilters: React.FC = () => {
+  const { filters } = mockMcpCatalogFilterOptions;
+
+  const getFilterProps = () => filters;
```
mturley: This function isn't doing anything.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogFilters.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Stack } from '@patternfly/react-core';
+import { mockMcpCatalogFilterOptions } from '~/app/pages/mcpCatalog/mocks/mockMcpCatalogFilterOptions';
+import DeploymentModeFilter from '~/app/pages/mcpCatalog/components/globalFilters/DeploymentModeFilter';
+import SupportedTransportsFilter from '~/app/pages/mcpCatalog/components/globalFilters/SupportedTransportsFilter';
+import McpLicenseFilter from '~/app/pages/mcpCatalog/components/globalFilters/McpLicenseFilter';
+import LabelsFilter from '~/app/pages/mcpCatalog/components/globalFilters/LabelsFilter';
+import SecurityVerificationFilter from '~/app/pages/mcpCatalog/components/globalFilters/SecurityVerificationFilter';
+
+const McpCatalogFilters: React.FC = () => {
+  const { filters } = mockMcpCatalogFilterOptions;
```
mturley: We are also ready to fetch the filter options from the BFF now, but perhaps that's out of scope of this PR. Maybe coordinate with @YuliaKrimerman to see which of you will do that in the scope of your issues (@manaswinidas which issue was supposed to cover integrating the UI and BFF for these things?)

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const MCP_DETAILS_LINK_PLACEHOLDER = '#';
+
+type McpCatalogCardProps = {
+  server: McpServerMock;
+};
+
+const SecurityTag: React.FC<{ label: string }> = ({ label }) => (
+  <FlexItem>
+    <span className="pf-v5-u-display-inline-flex pf-v5-u-align-items-center">
```
mturley: We should not use v5 utility classes, only v6. Also, I think you should be able to use the PF Flex component here instead of your own span with classes.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+          justifyContent={{ default: 'justifyContentSpaceBetween' }}
+          style={{ gap: '4px' }}
+          className="pf-v6-u-mb-md"
+        >
+          <FlexItem>
+            <span
+              className="pf-v5-u-display-inline-block"
+              style={{ fontSize: '2rem', color: 'var(--pf-v5-global--default-color--200)' }}
+              aria-hidden
+            >
```
mturley, line 55: We need to use PF v6 things and not v5. That goes for the rest of this file as well. Also the inline styles in this file may be avoidable, we have utility classes for text size/styles/colors: https://www.patternfly.org/utility-classes/text

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const McpCatalogCard: React.FC<McpCatalogCardProps> = ({ server }) => {
+  const deploymentType =
+    server.deploymentMode === 'Local' ? McpCardIconType.LOCAL_TO_CLUSTER : McpCardIconType.REMOTE;
+  const deploymentConfig = getMcpCardIconConfig(deploymentType);
+
+  return (
+    <Card
+      isFullHeight
+      style={{ minHeight: '296.58px' }}
```
mturley: What's with this very specific magic number minHeight?

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalog.cy.ts
```diff
@@ -0,0 +1,48 @@
+import { mcpCatalog } from '~/__tests__/cypress/cypress/pages/mcpCatalog';
+
+describe('MCP Catalog Page', () => {
```
mturley, line 41: This file is under a modelCatalog directory, we should make a separate mcpCatalog mock tests directory

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+    setInputValue(searchTerm || '');
+  }, [searchTerm]);
+
+  const handleClearAllFilters = React.useCallback(() => {
+    if (hasFiltersAppliedValue) {
+      onResetAllFilters();
+    }
+  }, [hasFiltersAppliedValue, onResetAllFilters]);
+
+  const handleModelSearch = React.useCallback(() => {
```
mturley: This is named after "model", maybe it should just be generic `handleSearch`.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalog.cy.ts
manaswinidas: +1 to this

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export type McpCatalogContextType = {
```
manaswinidas: This should be in a types file - something like https://github.com/kubeflow/model-registry/blob/main/clients/ui/frontend/src/concepts/modelRegistry/types.ts if this is shared (I see this is exported)

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelRegistrySelector.tsx
```diff
@@ -150,7 +150,7 @@ const ModelRegistrySelector: React.FC<ModelRegistrySelectorProps> = ({
   return (
     <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
       <Icon iconSize="xl" isInline>
-        <TypedObjectIcon resourceType={ProjectObjectType.modelRegistryNavigator} />
+        <TypedObjectIcon resourceType={ProjectObjectType.registeredModels} />
```
manaswinidas: This was introduced in #2286 - any reason why we are changing this ?

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
manaswinidas: +1 to this

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
manaswinidas: +1 to this

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogFilters.tsx
manaswinidas: It should be done in this PR mostly - otherwise there was no need for waiting on BFF for this - I'm fine with merging with mocks but we have to follow up in another PR - why not use the hooks ?

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <ToolbarToggleGroup breakpoint="md" toggleIcon={<FilterIcon />}>
+                <ToolbarGroup variant="filter-group" gap={{ default: 'gapMd' }} alignItems="center">
+                  <ToolbarItem>
+                    <ThemeAwareSearchInput
+                      data-testid="mcp-catalog-search-input"
+                      aria-label="Search with submit button"
+                      className="toolbar-fieldset-wrapper"
+                      placeholder="Filter by name, description and provider"
+                      value={inputValue}
+                      style={{ minWidth: '600px' }}
```
manaswinidas, line 96: This style will overflow on mobile/other smaller devices

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogCard.tsx
Philip-Carneiro: This was the min card size in the mock LOL xD I just added it for now to see how that was going

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  const mcpServersResult = useMcpServersBySourceLabelWithAPI(apiState, {
+    sourceLabel: selectedSourceLabel,
+    pageSize: pagination.pageSize,
+    searchQuery,
+  });
+
+  const useMockData = !apiState.apiAvailable;
+  const apiHasItems = mcpServersResult.mcpServers.items.length > 0;
+  const isShowingMockItems = useMockData || !apiHasItems;
```
manaswinidas: The mock fallback is too aggressive: `!apiHasItems` (line 108) means even when the API is available and legitimately returns an empty list, we silently fall back to mock data. This masks real "no data" states.

Also, the mock server data lacks `license`, `tags`, and `transports` fields, so when mocks are active those sidebar filters always produce empty results — the filter options and mock servers are inconsistent with each other.

Suggestion: only fall back to mocks when `!apiState.apiAvailable`, not when the API returns zero items.

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+  );
+  const rawItems = useMockData
+    ? filteredMockItems
+    : apiHasItems
+      ? mcpServersResult.mcpServers.items
+      : filteredMockItems;
+  const itemsAfterSearch =
+    isShowingMockItems && searchQuery.trim().length > 0
+      ? filterMcpServersBySearchQuery(rawItems, searchQuery)
+      : rawItems;
```
manaswinidas: Perf: `filteredMockItems`, `rawItems`, and `itemsAfterSearch` are computed as bare variables on every render — not wrapped in `useMemo`. Since `itemsAfterSearch` is a new array reference each time, the `useMemo` on the next line that depends on it always recomputes too, defeating the memo.

With 5 mock items this is fine, but once real API data comes in these filter passes will run on every keystroke/state change. Consider memoizing the intermediate steps or collapsing them into the existing `useMemo`.

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogGalleryView.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <Stack hasGutter>
+        <McpCatalogCategorySection title="Servers" servers={items} />
+      </Stack>
+    );
+  }
+
+  return (
+    <Stack hasGutter>
+      {sourceLabels.map((label) => {
+        const sectionItems = items.filter((s) => s.source_id === label);
```
manaswinidas: Perf: `items.filter()` runs inline on every render for each source label (O(N×M) with N labels, M items), plus once more for uncategorized. Should be memoized with `useMemo` keyed on `items` and `sourceLabels`.

clients/ui/frontend/src/app/pages/mcpCatalog/components/McpCatalogFilters.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { mockMcpCatalogFilterOptions } from '~/app/pages/mcpCatalog/mocks/mockMcpCatalogFilterOptions';
+import DeploymentModeFilter from '~/app/pages/mcpCatalog/components/globalFilters/DeploymentModeFilter';
+import SupportedTransportsFilter from '~/app/pages/mcpCatalog/components/globalFilters/SupportedTransportsFilter';
+import McpLicenseFilter from '~/app/pages/mcpCatalog/components/globalFilters/McpLicenseFilter';
+import LabelsFilter from '~/app/pages/mcpCatalog/components/globalFilters/LabelsFilter';
+import SecurityVerificationFilter from '~/app/pages/mcpCatalog/components/globalFilters/SecurityVerificationFilter';
+
+const McpCatalogFilters: React.FC = () => {
+  const { filterOptions } = React.useContext(McpCatalogContext);
+  const filters = filterOptions?.filters ?? mockMcpCatalogFilterOptions.filters;
```
manaswinidas: `filterOptionsLoadError` and `filterOptionsLoaded` are both ignored here — if the API call fails, we silently fall back to mock filter options with no error shown. Model catalog handles this with a `<Spinner />` while loading and an `<Alert variant="danger">` on error (see `ModelCatalogFilters`). We should match that pattern.

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalog.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, [clearAllFilters]);
+
+  return (
+    <>
+      <ScrollViewOnMount shouldScroll scrollToTop />
+      <ApplicationsPage
+        title={MCP_CATALOG_TITLE}
+        description={MCP_CATALOG_SUBTITLE}
+        empty={false}
+        loaded
```
manaswinidas, line 39: `loaded` and `empty` are hardcoded here, so `ApplicationsPage` never shows its built-in loading/empty states. `catalogSourcesLoadError` is also never checked anywhere in the page tree — if catalog sources fail to load, the error is silently swallowed and mock data is shown instead.

Model catalog gates the entire page behind a `ModelCatalogCoreLoader` that checks `catalogSourcesLoaded` / `catalogSourcesLoadError` before rendering children. We should have equivalent handling here.

manaswinidas (commented):
Flagging error handling gaps — see inline comments. Errors from catalog sources and filter options are silently swallowed, falling back to mock data with no user-visible indication. Model catalog has established patterns for these (loader gate, spinner, alert) that we should match.

Philip-Carneiro:
> The filters for local and remote work great!
> 
> * Is this all the data we have for filters ? if yes, the BFF mock data needs to updated to include more,
> * also the BFF mock data needs to be updated to include license information as now it all leads to empty state
> * Search doesn't work
> 
> Image

Hi @manaswinidas 
These changes I will add it in another PR, because this one is big already. Thank you for the review

clients/ui/frontend/src/app/context/mcpCatalog/McpCatalogContext.tsx
```diff
@@ -0,0 +1,199 @@
...
+import type {
+  McpCatalogContextType,
+  McpCatalogPaginationState,
+} from '~/app/pages/mcpCatalog/types/mcpCatalogContext';
+import type { McpCatalogFiltersState } from '~/app/pages/mcpCatalog/types/mcpCatalogFilterOptions';
+import {
+  filterMcpServersByFilters,
+  filterMcpServersBySearchQuery,
+} from '~/app/pages/mcpCatalog/utils/mcpCatalogUtils';
+import { mockMcpServers } from '~/app/pages/mcpCatalog/mocks/mockMcpServers';
```
manaswinidas, line 21: **Non-blocking / follow-up:** `mockMcpServers` is statically imported and bundled in production. When `!apiReady`, mock data is shown with `mcpServersLoadError` set to `undefined` — silently masking API failures. Consider removing the mock fallback (show empty state or spinner while loading) or track as a follow-up to remove before GA.

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,134 @@
...
```
manaswinidas, line 96: **Non-blocking:** `minWidth: 600px` will overflow on smaller viewports. Consider `style={{ width: '100%', maxWidth: '600px' }}` or a responsive utility class.

clients/ui/frontend/src/app/pages/mcpCatalog/components/globalFilters/DeploymentModeFilter.tsx
```diff
@@ -0,0 +1,27 @@
...
+import McpCatalogStringFilter from '~/app/pages/mcpCatalog/components/McpCatalogStringFilter';
+import type { McpCatalogFilterOptions } from '~/app/pages/mcpCatalog/types/mcpCatalogFilterOptions';
+
+const filterKey = 'deploymentMode';
+
+type DeploymentModeFilterProps = {
+  filters?: McpCatalogFilterOptions;
+};
+
+const DeploymentModeFilter: React.FC<DeploymentModeFilterProps> = ({ filters }) => {
```
manaswinidas, line 12: **Non-blocking / DRY:** All five global filter components (`DeploymentModeFilter`, `LabelsFilter`, `McpLicenseFilter`, `SecurityVerificationFilter`, `SupportedTransportsFilter`) are structurally identical — only `filterKey` and `title` differ. Consider a single config-driven component to avoid five near-duplicate files.

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogGalleryView.tsx
```diff
@@ -0,0 +1,109 @@
...
+  if (!sourceLabel) {
+    return 'Other';
+  }
+  return sourceLabel.charAt(0).toUpperCase() + sourceLabel.slice(1).toLowerCase();
+}
+
+type McpCatalogGalleryViewProps = {
+  searchTerm: string;
+};
+
```
manaswinidas, line 17: **Nit:** `searchTerm` is declared in the props type but never read (filtering happens in context). Remove the prop and `McpCatalogGalleryViewProps` type to avoid dead code.

clients/ui/frontend/src/app/pages/mcpCatalog/mocks/mockMcpServers.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable camelcase */
+import type { McpServer } from '~/app/mcpServerCatalogTypes';
+
+export const mockMcpServers: McpServer[] = [
```
manaswinidas, line 4: **Non-blocking:** Mock servers lack `license`, `tags`, and `transports` fields, so when mock fallback is active the License/Labels/Transports sidebar filters always yield empty results. If mocks are kept, they should be consistent with `mockMcpCatalogFilterOptions`.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/mcpCatalog/mcpCatalog.cy.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  it('should display Deployment mode filter with Local and Remote options', () => {
+    mcpCatalog.visit();
+    mcpCatalog.findFilterCheckbox('deploymentMode', 'Local').should('be.visible');
+    mcpCatalog.findFilterCheckbox('deploymentMode', 'Remote').should('be.visible');
+  });
+
+  it('filter Show more should expand labels list', () => {
+    mcpCatalog.visit();
+    mcpCatalog.findFilterShowMore('labels').scrollIntoView();
```
manaswinidas, line 82: **Nit:** Three separate intercepts registered for the same MCP servers endpoint (lines 82-91). Only the most specific one should be needed.

manaswinidas:
/lgtm
/approve 

The comments highlighted will be covered in the follow up @Philip-Carneiro FYI

PR #2337, opened 2026-03-06
feat(async-upload): add model signing to upload job

jonburdo:
This adds the cosign binary to the Dockerfile for the async upload job. It also updates model-registry to >=0.3.7 and adds the signing extra. This will enable use of the model_registry.signing submodule for signing and verification of model artifacts.

It also adjust the async upload job to sign model files with OMS signing and to sign the OCI image if the destination is an OCI registry if sigstore server urls are provided.

## How Has This Been Tested?

I've adjusted the integration test that runs on kind to pass through the SIGSTORE_ env vars if provided. This allows running them with and without signing. It passes in tests against private sigstore servers using the OIDC-backed service account token as expected.

## Merge criteria:

If you have UI changes


jobs/async-upload/job/config.py
```diff
@@ -373,6 +378,35 @@ def str2bool(x):
...
 
+_SIGSTORE_ENV_VARS = ("SIGSTORE_TUF_URL", "SIGSTORE_FULCIO_URL", "SIGSTORE_REKOR_URL")
+
+
+def _resolve_signing(sign_flag: bool | None) -> bool:
+    """Resolve whether signing should be enabled.
+
+    If the --sign flag is explicitly set, use that value.
+    Otherwise, auto-detect from SIGSTORE environment variables:
+    - All set and non-empty → enabled
```
dbasunag, line 389: Should we not have the ability to explicitly request to sign or not? Can we have a use case, where a user wants to sign few specific model but not some?

jonburdo: This `--sign` option or the equivalent `MODEL_SYNC_SIGN` env var can be set to true or false to explicitly enable or disable signing. Typically it shouldn't be as there is generally no harm in adding a signature, especially if your environment is configured to do so.

This kubernetes job is run once per model, so each invocation can be passed its own value. If a frontend calling this job wants to have an option to enable/disable it can do so and pass the value to this variable.

jobs/async-upload/Dockerfile
```diff
@@ -2,6 +2,8 @@
 #  Asynchronous Model-Sync Job image for Kubeflow Model Registry (https://github.com/kubeflow/model-registry/issues/1108)
 ###############################################################################
 
+FROM quay.io/securesign/cli-cosign@sha256:a8289d488491991d454a32784de19476f2c984917eb7a33b4544e55512f2747c AS cosign
```
tarilabs, line 5: thanks for using digest version instead of tags

jobs/async-upload/README.md
```diff
@@ -171,6 +173,29 @@ spec:
...
 
+## Signing
+
+The job can sign model files and OCI images using sigstore servers. Signing behavior is controlled by the `--sign` flag (env: `MODEL_SYNC_SIGN`):
+
+| Scenario | Behavior |
+| --- | --- |
+| `--sign` not set, all 3 `SIGSTORE_*` env vars present (`SIGSTORE_TUF_URL`, `SIGSTORE_FULCIO_URL`, `SIGSTORE_REKOR_URL`) | Signing **enabled** automatically |
+| `--sign` not set, none of the `SIGSTORE_*` env vars present | Signing **skipped** silently |
+| `--sign` not set, only some `SIGSTORE_*` env vars present | Signing **skipped** with a warning listing the missing vars |
```
tarilabs: I think for completeness I would document for 
> `--sign` is set

jobs/async-upload/job/models.py
```diff
@@ -233,3 +233,5 @@ class AsyncUploadConfig(BaseModel):
     storage: StorageConfig = Field(default_factory=StorageConfig)
     registry: RegistryConfig
     metadata: ConfigMapMetadata | None = None  # Optional ConfigMap metadata
+    signing_enabled: bool = False
+    signing_identity_token_path: str | None = None
```
tarilabs, line 238: the default value for this path is set in the MR py client integration, correct? 
hence the default here is not needed (in place of the `None`)

preferred to double check

jonburdo: Yes, good question! `Signer(identity_token_path=None)` is functionally the same as `Signer()`.

I've tested that `Signer()` can be initialized with no args on a kubernetes cluster provided that:
a. A service account token is present at the default location
b. The 3 required `SIGSTORE_*` env vars are set (the 4th, TSA, is optional)

For reference, `Signer` will pass a `None` value for the token to `SigningConfig` [here](https://github.com/kubeflow/model-registry/blob/90c6c77cbb3e6b4dd136edb962e804e60dd33927/clients/python/src/model_registry/signing/signer.py#L67-L81) which will default to that default service account token location [here](https://github.com/kubeflow/model-registry/blob/90c6c77cbb3e6b4dd136edb962e804e60dd33927/clients/python/src/model_registry/signing/config.py#L119)

tarilabs: thanks! I think this:

> For reference, Signer will pass a None value for the token to SigningConfig [here](https://github.com/kubeflow/model-registry/blob/90c6c77cbb3e6b4dd136edb962e804e60dd33927/clients/python/src/model_registry/signing/signer.py#L67-L81) which will default to that default service account token location [here](https://github.com/kubeflow/model-registry/blob/90c6c77cbb3e6b4dd136edb962e804e60dd33927/clients/python/src/model_registry/signing/config.py#L119)

is worthy as code comment in line 237 above/on the right 🙃 or a shorter version just to "post-it note" this clarification

jonburdo:
Updated README to document when `--sign` is set.

Added comment to indicate token defaults to `/var/run/secrets/kubernetes.io/serviceaccount/token`

Squashed to clean git history. Diff: https://github.com/kubeflow/model-registry/compare/7b982961fa6d24605267143c2092c64c02d07fbd..130d455a827f1ae8d34b959c9a39f255c2d44754

jonburdo:
/approve

Thank you for the reviews @tarilabs @dbasunag!

Any additional feedback is welcome - happy to address in a follow-up. For now, given the lgtms I will proceed with this :)

PR #2338, opened 2026-03-09
Use shared Manage Columns from mod-arch-shared and adjust the logic in KF 

Taj010:
## Description

This PR switches the Model Registry UI to consume the shared Manage Columns implementation from mod-arch-shared instead of maintaining a local copy. The local imports are replaced with library exports while keeping the existing behavior and deleted the local copies that were originally copied from odh-dashboard

## How Has This Been Tested?

Ran `npm run test` to ensure no tests are broken. 
No new functionality added. 
## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/package-lock.json
+++ b/clients/ui/frontend/package-lock.json
@@ -19745,10 +19745,9 @@
       }
     },
     "node_modules/mod-arch-core": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/mod-arch-core/-/mod-arch-core-1.11.0.tgz",
-      "integrity": "sha512-QFHobWbeiE69D/RFCfNmbug64xHt/2lyGBP0kV4FvgyIQ2slpNuB8fWcjpmI549cD2QukPvaxsVAlou307hFWw==",
-      "license": "Apache-2.0",
+      "version": "1.11.1",
+      "resolved": "https://registry.npmjs.org/mod-arch-core/-/mod-arch-core-1.11.1.tgz",
+      "integrity": "sha512-eWALaKwdq7eunQ2E3PJD+mfhlKXfrEdyq803td4FseAn21K02QHJGAapcQzf9MA77m31YmDHB+y1sZN16NNnkA==",
       "dependencies": {
         "lodash-es": "^4.17.23",
         "tslib": "^2.6.2"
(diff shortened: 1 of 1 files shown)


manaswinidas:
Expanding on my review comment above:

The 4 retained test files (`ManageColumnSearchInput.spec.tsx`, `ManageColumnsModal.spec.tsx`, `useManageColumns.spec.ts`, `utils.spec.ts`) now only test `mod-arch-shared` internals — the same treatment applied to `ContentModal.spec.tsx` (correctly deleted) should apply here. The library should carry its own unit tests.

What's missing is Cypress coverage for the **app-specific wiring** that this PR touches — things the library can't test:

- Clicking the Columns icon opens the modal
- Toggling columns off/on updates the table headers
- Column visibility persists across navigation (localStorage key correctness)
- Latency filter interaction with column visibility (the `useEffect` in `useHardwareConfigColumns`)
- Sticky columns (Hardware configuration, Workload type) don't appear in the modal

These behaviors live in the app's integration layer and have no automated coverage today. Not necessarily a blocker for this PR, but worth tracking as a follow-up since the retained unit tests don't cover any of them either.

Taj010:
I removed the retained test files, and created a follow up for the missing cypress coverage

manaswinidas (changes requested):
A screenshot/screen-recording of "Customize columns" would be better - I ran this PR locally - the layout seems weird - the icon and the field should in the same line 
Screenshot 2026-03-11 at 12 42 49 PM

Taj010:
@manaswinidas Could the icon and the field not being in the same line be a pre-existing issue? I just moved the the logic from kf to mod-arch without any functional changes. 

Let me have a look again and check if we can do a easy fix for this.

Taj010:
I adjusted the columns layout it now looks like this -

image

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.scss
```diff
@@ -0,0 +1,9 @@
+/* PatternFly DragDropSort (variant="default") renders each Draggable item as a
+   plain div wrapping a DragButton + content.  That div lacks flex layout, so the
+   grip handle drops to its own line.  Force it to a single row. */
+[data-testid='hardware-config-manage-columns'] {
```
mturley: @Taj010 I think this will be a problem with any usage of this component in the MUI theme. we should put this style fix in the library itself, and make sure it doesn't affect patternfly theme. Would you prefer if we remove this style and merge this PR first, or hold this PR until we can update the library and get it released again? We could do it as part of your issue or as its own small followup issue. Up to you.

manaswinidas:
This PR has a merge conflict

manaswinidas:
I checked out this branch and main too both show this
Screenshot 2026-03-18 at 4 16 41 PM

checking midstream

manaswinidas:
Also, can you update the PR description ? I don't see this in the changes anymore -- "Explicit mod-arch dependency bump (important) : To consume the new Manage Columns APIs, this PR bumps the mod-arch stack to 1.10.1" Is this version bump being handled in another PR that I missed ?

manaswinidas:
Looks okay in midstream
Screenshot 2026-03-18 at 4 36 39 PM

Taj010:
updated the description to remove the explicit mod-arch dependency bump message. And the changes were not seen as the library version was not v1.11.1 which has the changes, adjusted this  it now.

PR #2339, opened 2026-03-09
Fix transfer job toast notifications showing false success

manaswinidas:
## Description

Previously, two toasts ("started" and "complete") fired simultaneously when a transfer job was created, with the "complete" toast misleadingly showing a green checkmark even though the job was still running and could later fail.

Now a single "started" (info) toast fires at creation time, and background polling detects the actual job outcome to show "succeeded" or "failed" when the job truly finishes.

## How Has This Been Tested?

Try creating a transfer model job while registering a model in a kind cluster - you should see one notification for "started" and another for "failed/succeeded".

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/context/TransferJobNotificationsContext.tsx
```diff
@@ -0,0 +1,129 @@
...
+        watchedJobsRef.current = watchedJobsRef.current.filter(
+          (j) => !resolvedIds.includes(j.jobId),
+        );
+        if (watchedJobsRef.current.length === 0) {
+          setPolling(false);
+        }
+      }
+    };
+
+    pollJobs();
```
YuliaKrimerman, line 112: The cleanup removes resolved jobs from the ref, but if ALL jobs complete on the first poll, watchedJobsRef.current will be empty and setPolling(false) will be called. Then the effect cleans up the interval. However, between the initial pollJobs() call on line 114 and the interval starting on line 115, there's a potential timing issue: if pollJobs resolves synchronously (in tests), setPolling(false) would trigger a re-render. The useEffect cleanup from the new render would clear the interval that was just created. In practice this works because setPolling(false) is async (batched by React), but it's something to watch for in tests with fake timers.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModel.tsx
```diff
@@ -86,14 +88,18 @@ const RegisterModel: React.FC = () => {
 
     // Branch based on registration mode
     if (formData.registrationMode === RegistrationMode.RegisterAndStore) {
-      registrationNotification.showRegisterAndStoreSubmitting(toastParams);
```
YuliaKrimerman, line 89: The old code showed a "submitting" toast BEFORE the API call, so the user got immediate feedback. The new code only shows the "started" toast AFTER the API call returns a transferJob. If the registerViaTransferJob API call takes a while, there's no user feedback between clicking "Create" and the API response. The isSubmitting state disables the button (line 82), and a spinner may show, but the previous "job started" toast gave clear confirmation that something was happening. Was this change intentional? If not , consider keeping an immediate feedback toast (or rely on the form footer's loading state). Since the user navigates away immediately after success (navigate(modelRegistryUrl(mrName))), the "started" toast appears on the next page -- which is fine. But if the API call takes 5+ seconds, the user is staring at a disabled button with no toast.

mturley: I think the spinner on the disabled button while we're waiting for the submit is fine. That's how the register form already worked, and we also shouldn't notify that the job was submitted until we know it submitted successfully - we could fail to create the job in the first place, which would be rendered as an inline error while you're still on the form.

clients/ui/frontend/src/app/context/TransferJobNotificationsContext.tsx
manaswinidas: Good catch on the timing concern. I've addressed this in two ways:

1. **Switched to GET-by-ID polling** (per @mturley's suggestion) — the context now calls `GET /model_transfer_jobs/{job_name}` for each watched job individually instead of listing all jobs and searching with `.find()`. This removes the `registryGroups` grouping logic entirely and makes each poll a single targeted API call.

2. **Added a `cancelled` guard** — the effect cleanup now sets a `cancelled` flag, and `pollJobs` checks `if (cancelled) return` after awaiting, so stale completions from a previous effect cycle won't call `setPolling(false)` on an unmounted or re-rendered component.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/useRegistrationNotification.ts
ppadti, line 31: I think we can remove this success case as we are only calling  showAlert with AlertVariant.info and AlertVariant.danger

clients/ui/frontend/src/app/context/TransferJobNotificationsContext.tsx
```diff
@@ -54,59 +54,54 @@ export const TransferJobNotificationsProvider: React.FC<React.PropsWithChildren>
...
-      const resolvedIds: string[] = [];
+      const resolvedNames: string[] = [];
 
       await Promise.all(
-        Array.from(registryGroups, async ([registryName, registryJobs]) => {
-          const hostPath = `${URL_PREFIX}/api/${BFF_API_VERSION}/model_registry/${registryName}`;
-          const listJobs = getListModelTransferJobs(hostPath);
+        jobs.map(async (watched) => {
+          const hostPath = `${URL_PREFIX}/api/${BFF_API_VERSION}/model_registry/${watched.registryName}`;
+          const fetchJob = getModelTransferJob(hostPath);
```
ppadti, line 71: TransferJobNotificationsContext constructs its own getModelTransferJob call without passing the namespace query parameter that the BFF requires.
so getting error for GET
```
{
	"error": {
		"code": "400",
		"message": "missing required query parameter: namespace"
	}
}
```

clients/ui/frontend/src/app/hooks/useModelRegistryAPIState.tsx
```diff
@@ -46,6 +47,7 @@ const useModelRegistryAPIState = (
       patchRegisteredModel: patchRegisteredModel(path, queryParameters),
       patchModelVersion: patchModelVersion(path, queryParameters),
       patchModelArtifact: patchModelArtifact(path, queryParameters),
+      getModelTransferJob: getModelTransferJob(path, queryParameters),
```
ppadti, line 50: getModelTransferJob was added to the ModelRegistryAPIs type, useModelRegistryAPIState, and the test mock in updateTimestamps.test.ts. But nothing consumes it through apiState - TransferJobNotificationsContext directly import getModelTransferJob from service

PR #2367, opened 2026-03-10
fix: correct sort order and name filter in MCP catalog

pboyd:
## Description

The database layer expects `orderBy` and `sortOrder` in uppercase (`ASC`, `DESC`), but API values arrive lowercase. Uppercase both before forwarding to prevent silent sort failures.

MCP server names are stored as `base_name@version` composites, so filtering by `Context.name` fails for display names like `jira-mcp`. Map the `name` filter to the `base_name` context property in the entity mappings and `applyMCPServerListFilters`.

## How Has This Been Tested?
On a local dev cluster.

## Merge criteria:

### DIFF
--- a/catalog/clients/python/src/model_catalog/_client.py
+++ b/catalog/clients/python/src/model_catalog/_client.py
@@ -509,6 +509,7 @@ class CatalogAPIClient:
     @_handle_api_errors
     def get_mcp_servers(
         self,
+        name: str | None = None,
         include_tools: bool | None = None,
         page_size: int | None = None,
         next_page_token: str | None = None,
@@ -516,6 +517,7 @@ class CatalogAPIClient:
         """Get MCP servers from catalog.
 
         Args:
+            name: Filter by MCP server name.
             include_tools: Whether to include the tools array in each server result.
             page_size: Number of items per page.
             next_page_token: Token for pagination.
@@ -525,6 +527,7 @@ class CatalogAPIClient:
         """
         page_size_str = str(page_size) if page_size is not None else None
         response = self.mcp_api.find_mcp_servers(
+            name=name,
             include_tools=include_tools,
             page_size=page_size_str,
             next_page_token=next_page_token,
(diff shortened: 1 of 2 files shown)


catalog/internal/catalog/mcpcatalog/service/mcp_server.go
jonburdo, line 351: We're still using `Context.name` so for `jira-mcp@1.0.0` I think we'd match it with:
- `jira-mcp` - expected
- `1.0.0` - expected
- `jira-mcp@1.0.0` - maybe not expected if this format is an implementation detail
- `mcp@1` - maybe not expected

Maybe it's fine or we an revisit - just wanted to point it out

fege:
Hey @pboyd, I've added a Python E2E test that validates the name filter fix. Could you add this to the PR? I did try to push but I do not have permission 

**1. `catalog/clients/python/src/model_catalog/_client.py`** — add `name` parameter to `get_mcp_servers()`:

```diff
     @_handle_api_errors
     def get_mcp_servers(
         self,
+        name: str | None = None,
         include_tools: bool | None = None,
         page_size: int | None = None,
         next_page_token: str | None = None,
     ) -> dict[str, Any]:
         """Get MCP servers from catalog.

         Args:
+            name: Filter by MCP server name.
             include_tools: Whether to include the tools array in each server result.
             page_size: Number of items per page.
             next_page_token: Token for pagination.

         Returns:
             Dict with MCP servers list and pagination info.
         """
         page_size_str = str(page_size) if page_size is not None else None
         response = self.mcp_api.find_mcp_servers(
+            name=name,
             include_tools=include_tools,
             page_size=page_size_str,
             next_page_token=next_page_token,
         )
```

**2. `catalog/clients/python/tests/test_mcp_servers.py`** — add test after `test_mcp_server_pagination_fields`:

```python
    def test_filter_by_name(
        self,
        api_client: CatalogAPIClient,
        suppress_ssl_warnings: None,
    ):
        """Test filtering MCP servers by exact name."""
        target_name = "calculator"
        response = api_client.get_mcp_servers(name=target_name)
        items = response.get("items", [])
        assert len(items) == 1, f"Expected 1 server named '{target_name}', got {len(items)}"
        assert items[0]["name"] == target_name
```

This test will fail without the name filter fix (since `name=calculator` returns empty when the DB stores `calculator@2.0.0`), and pass with it.

PR #2368, opened 2026-03-10
fix: catalog “Register & store” behavior for OCI vs HF models

Taj010:
## Description

This PR addresses to fix catalog “Register & store” behavior so that OCI-backed catalog models only support Register (toggle hidden), while HF/URI catalog models still show the Register / Register and store toggle and correctly render source/destination fields when using Register and store.

## Merge criteria:

- [] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
@@ -78,8 +78,10 @@ const RegistrationCommonFormSections = <D extends RegistrationCommonFormData>({
       setData('namespace', '');
     }
     if (registrationMode === RegistrationMode.RegisterAndStore && isCatalogOciSource) {
-      // From catalog: prevent OCI => OCI by forcing back to "Register"
+      // From catalog: prevent OCI => OCI by forcing back to "Register".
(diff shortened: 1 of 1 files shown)


clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegistrationCommonFormSections.tsx
```diff
@@ -72,7 +77,11 @@ const RegistrationCommonFormSections = <D extends RegistrationCommonFormData>({
       setData('registrationMode', RegistrationMode.Register);
       setData('namespace', '');
     }
-  }, [isRegistryStorageFeatureAvailable, registrationMode, setData]);
+    if (registrationMode === RegistrationMode.RegisterAndStore && isCatalogOciSource) {
+      // From catalog: prevent OCI => OCI by forcing back to "Register"
+      setData('registrationMode', RegistrationMode.Register);
```
mturley, line 83: We may want to also clear the `namespace` value here like the above logic does. Also, I almost asked to remove this since it isn't something we should be able to hit ("Register" is the default mode and there's no way to initially render the form in "Register and store" mode, so hiding the mode toggle is sufficient), but this does protect us if that changes in the future. Maybe we should add a comment to that effect.

mturley (commented):
One tiny nit

PR #2385, opened 2026-03-13
feat(catalog): add includedServers/excludedServers filtering to MCP sources

pboyd:
## Description

Filter MCP servers by glob patterns using includedServers and excludedServers on MCP source entries, mirroring the existing includedModels/excludedModels behavior on model sources.

## How Has This Been Tested?
Unit tests and on a local kind cluster

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/basecatalog/name_filter.go
+++ b/catalog/internal/catalog/basecatalog/name_filter.go
@@ -1,0 +1,151 @@
+package basecatalog
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+// NameFilter encapsulates include/exclude pattern matching for names.
+type NameFilter struct {
+	included []*compiledPattern
+	excluded []*compiledPattern
+}
+
+type compiledPattern struct {
+	re *regexp.Regexp
+}
+
+func newCompiledPattern(field string, idx int, raw string) (*compiledPattern, error) {
+	value := strings.TrimSpace(raw)
+	if value == "" {
+		return nil, fmt.Errorf("%s[%d]: pattern cannot be empty", field, idx)
+	}
+
+	// Convert a simple glob (only supporting '*') into a regexp.
+	var b strings.Builder
+	b.WriteString("(?i)^")
+	for _, r := range value {
+		if r == '*' {
+			b.WriteString(".*")
+			continue
+		}
+		b.WriteString(regexp.QuoteMeta(string(r)))
+	}
+	b.WriteString("$")
+
+	re, err := regexp.Compile(b.String())
+	if err != nil {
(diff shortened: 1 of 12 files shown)


Al-Pragliola:
Genuine question: Don't we also need to add the new fields to the OpenAPI spec? (/sources endpoint) @pboyd

pboyd:
> Genuine question: Don't we also need to add the new fields to the OpenAPI spec? (/sources endpoint) @pboyd

That's a good question, I almost added them to the response, but I decided against it. I know we have `includedModels`/`excludedModels` in the response for model catalog sources, but (only in hindsight, of course) I'm not sure that was the right call either. I'd be happy to be persuaded otherwise, but here's my reasoning:

- The definition for the config and the API have never been equivalent, and really shouldn't be. The API needs status fields and the config needs connection details. IMO, these filters are only a configuration detail.
- `includeModels`/`excludedModels` are specific to model catalog but sources are more general now. I couldn't think of a nice general way to extend this and adding `includedServers` and `excludedServers` sounds like it would quickly spiral into `included<ThirdThing>`/`excluded<ThirdThing>`, etc.
- Other than consistency with model catalog, I don't see a good reason to add `includedServers`/`excludedServers` to the API. The admin functionality in the dashboard will need these values, but they also need other fields from the configmap so it doesn't matter that much if we add them. And normal API consumers can't do anything with the values (after all, they will only see results that passed the filters--what could they apply them to?)

Al-Pragliola:
I completely agree with points 1 and 2, point 3 was my only concern:

> but they also need other fields from the configmap so it doesn't matter that much if we add them

Do you mean that is something that we will try to cover in future work?

Al-Pragliola:
Either way,

/lgtm

pboyd:
> I completely agree with points 1 and 2, point 3 was my only concern:
> 
> > but they also need other fields from the configmap so it doesn't matter that much if we add them
> 
> Do you mean that is something that we will try to cover in future work?

No, I mean that the dashboard's admin functionality currently has to retrieve the configmap because the sources endpoint doesn't return everything (`properties` and `type`, for instance, aren't returned). So if they have to check the configmap anyway, they can get `includedServers` and `excludedServers` at the same time.

Al-Pragliola:
> > I completely agree with points 1 and 2, point 3 was my only concern:
> > > but they also need other fields from the configmap so it doesn't matter that much if we add them
> > 
> > 
> > Do you mean that is something that we will try to cover in future work?
> 
> No, I mean that the dashboard's admin functionality currently has to retrieve the configmap because the sources endpoint doesn't return everything (`properties` and `type`, for instance, aren't returned). So if they have to check the configmap anyway, they can get `includedServers` and `excludedServers` at the same time.

Oh I didn't know, then I agree also on the third and last point 😁

PR #2392, opened 2026-03-14
fix: Readability of Code and better accessibility

HarshwardhanPatil07:
## Description

- it reduces nesting and makes the JSX cleaner to read.
- id="source-type-readonly" — better accessibility, consistent with how source-name has an id
- data-testid="source-type-readonly" — makes the read-only value directly targetable in tests (previously untestable without traversing the DOM)
- className="pf-v6-u-display-block pf-v6-u-mt-sm" — adds a top margin and block display so the read-only text aligns visually with how the radio buttons are laid out below it in non-edit mode

## How Has This Been Tested?

1. Adding a catalog source.
2. Clicking the "Manage source" button on the catalog sources table.
3. Verified the **Source type** field is fully visible with correct label in the modal.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalogSettings/components/SourceDetailsSection.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalogSettings/components/SourceDetailsSection.tsx
@@ -55,6 +55,11 @@ const SourceDetailsSection: React.FC<SourceDetailsSectionProps> = ({
     />
   );
 
+  const sourceTypeLabel =
+    formData.sourceType === CatalogSourceType.HUGGING_FACE
+      ? SOURCE_TYPE_LABELS.HUGGING_FACE
+      : SOURCE_TYPE_LABELS.YAML;
+
   return (
     <FormSection>
       <FormGroup label={FORM_LABELS.NAME} isRequired fieldId="source-name">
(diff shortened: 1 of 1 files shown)


HarshwardhanPatil07:
@manaswinidas updated the description of PR

manaswinidas:
Thanks for the contribution, @HarshwardhanPatil07. A few thoughts on this PR:

1. **Readability**: The inline ternary (3 lines between two constants) is already straightforward. Extracting it to a variable doesn't meaningfully reduce nesting or improve clarity — it just moves the logic elsewhere.

2. **Accessibility**: Adding an `id` to the `<span>` doesn't improve accessibility on its own. It would only matter if another element referenced it via `aria-labelledby` or `aria-describedby`, which nothing in this PR does.

3. **`data-testid`**: This is the one potentially useful addition, but no tests are added here to use it. It would be better introduced alongside the tests that need it.

4. **Styling (`pf-v6-u-display-block pf-v6-u-mt-sm`)**: In edit mode, only the read-only text renders (radio buttons are hidden), so there's no visual misalignment to fix. No before/after screenshots are provided to justify the change.

Overall, I don't think this PR is necessary — the changes are cosmetic with no functional or accessibility improvement. I'd recommend closing this.

HarshwardhanPatil07:
@manaswinidas Thank you for that information

PR #2400, opened 2026-03-16
Add labels endpoint, update ReadMe, openAPI spec and fix bugs with All server section

ppadti:
## Description
This is PR aims to update the Readme and openAPI. Updated the /labels endpoint to take assetType as a queryParam. 
fix the bugs with All server section to be consistent with model catalog

## How Has This Been Tested?
Run the MCP catalog with kind cluster and make sure everything works as expected. 
few issues found with backend - starting a thread now
## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/pages/mcpCatalog.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/pages/mcpCatalog.ts
@@ -50,8 +50,8 @@ class McpCatalog {
     return cy.findByTestId(`mcp-catalog-card-${serverId}`);
   }
 
-  findCardDetailsLink() {
-    return cy.findByTestId('mcp-catalog-card-detail-link-');
+  findCardDetailsLink(serverId: string) {
+    return cy.findByTestId(`mcp-catalog-card-detail-link-${serverId}`);
   }
 
   findCardDescription(serverId: string) {
--- a/clients/ui/frontend/src/app/pages/mcpCatalog/utils/mcpCatalogUtils.ts
+++ b/clients/ui/frontend/src/app/pages/mcpCatalog/utils/mcpCatalogUtils.ts
@@ -47,7 +47,7 @@ const FRONTEND_TO_BACKEND_FILTER_KEY: Record<string, string> = Object.fromEntrie
   Object.entries(BACKEND_TO_FRONTEND_FILTER_KEY).map(([backend, frontend]) => [frontend, backend]),
 );
 
-const wrapInQuotes = (v: string): string => `'${v}'`;
+const wrapInQuotes = (v: string): string => `'${v.replace(/'/g, "''")}'`;
 
 export function mcpFiltersToFilterQuery(filters: McpCatalogFiltersState): string {
   const clauses: string[] = [];
--- a/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
@@ -336,7 +336,7 @@ export const getSortParams = (
   };
 };
 
-const wrapInQuotes = (v: string): string => `'${v}'`;
+const wrapInQuotes = (v: string): string => `'${v.replace(/'/g, "''")}'`;
 
 const eqFilter = (k: string, v: string) => `${k}=${wrapInQuotes(v)}`;
 const inFilter = (k: string, values: string[]) =>


clients/ui/frontend/src/app/pages/mcpCatalog/utils/mcpCatalogUtils.ts
```diff
@@ -54,107 +43,25 @@ export const hasMcpFiltersApplied = (
...
-  });
-}
+const FRONTEND_TO_BACKEND_FILTER_KEY: Record<string, string> = Object.fromEntries(
+  Object.entries(BACKEND_TO_FRONTEND_FILTER_KEY).map(([backend, frontend]) => [frontend, backend]),
+);
 
-function isMcpTransport(s: string): s is McpTransportType {
-  return s === 'stdio' || s === 'sse' || s === 'http';
-}
+const wrapInQuotes = (v: string): string => `'${v}'`;
```
manaswinidas: Bug: `wrapInQuotes` doesn't escape single quotes in values. A filter value like `O'Reilly` produces `tags='O'Reilly'` — malformed query that will error or return wrong results.

```suggestion
const wrapInQuotes = (v: string): string => `'${v.replace(/'/g, "''")}'`;
```

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpServerToolsSection.tsx
```diff
@@ -36,6 +36,8 @@ const getAccessTypeConfig = (
       return { text: 'read/write', color: 'orange' };
     case 'execute':
       return { text: 'execute', color: 'purple' };
+    default:
+      return { text: 'read-only', color: 'blue' };
```
manaswinidas, line 40: Adding a `default` here silently swallows unknown access types (renders them as "read-only"). The previous code with no `default` let TypeScript's exhaustiveness checking catch missing cases at compile time.

Consider a `never` guard instead:
```ts
default: {
  const _exhaustive: never = accessType;
  return { text: 'read-only', color: 'blue' };
}
```
This way you still get a compile error if a new access type is added to the union.

clients/ui/frontend/src/__tests__/cypress/cypress/pages/mcpCatalog.ts
```diff
@@ -50,18 +50,14 @@ class McpCatalog {
     return cy.findByTestId(`mcp-catalog-card-${serverId}`);
   }
 
-  findCardDetailLink(serverId: string) {
-    return cy.findByTestId(`mcp-catalog-card-detail-link-${serverId}`);
+  findCardDetailsLink() {
+    return cy.findByTestId('mcp-catalog-card-detail-link-');
   }
```
manaswinidas: This always looks for `data-testid="mcp-catalog-card-detail-link-"` (trailing dash, no ID) — it will never match a real element. The `serverId` param was dropped but the test ID pattern still needs one.

Either remove this dead method or restore the parameter:
```ts
findCardDetailsLink(serverId: string) {
  return cy.findByTestId(`mcp-catalog-card-detail-link-${serverId}`);
}
```

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -670,34 +670,36 @@ export const findLabelData = (
...
-  // but we handle it gracefully
   if (sourceLabel === SourceLabel.other) {
-    return 'Other models'; // fallback for the special null case
+    return otherFallback;
   }
 
-  return generateCategoryName(sourceLabel);
+  return sourceLabel.toLowerCase().endsWith(categorySuffix)
+    ? sourceLabel
+    : `${sourceLabel} ${categorySuffix}`;
```
manaswinidas, line 702: Behavior regression: the old fallback used `generateCategoryName()` which humanized labels (`"red_hat"` → `"Red Hat Models"`). This new path produces `"red_hat models"` (raw snake_case + suffix). Since `getLabelDisplayName` is shared by both model catalog and MCP catalog pages, this affects existing model catalog rendering too.

clients/ui/bff/internal/api/mcp_servers_catalog_handler.go
```diff
@@ -27,7 +27,12 @@ func (app *App) GetAllMcpServersHandler(w http.ResponseWriter, r *http.Request,
 	mcpServers, err := app.repositories.ModelCatalogClient.GetAllMcpServers(client, r.URL.Query())
 
 	if err != nil {
-		app.serverErrorResponse(w, r, err)
+		var httpErr *httpclient.HTTPError
+		if errors.As(err, &httpErr) {
+			app.errorResponse(w, r, httpErr)
+		} else {
+			app.serverErrorResponse(w, r, err)
+		}
```
manaswinidas, line 35: This 5-line error-handling block is now copy-pasted in all 4 handlers (`GetAllMcpServersHandler`, `GetMcpServersFiltersHandler`, `GetMcpServerHandler`, `GetMcpServersToolsHandler`). Consider extracting a helper:
```go
func (app *App) handleHTTPError(w http.ResponseWriter, r *http.Request, err error) {
    var httpErr *httpclient.HTTPError
    if errors.As(err, &httpErr) {
        app.errorResponse(w, r, httpErr)
    } else {
        app.serverErrorResponse(w, r, err)
    }
}
```

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogAllServersView.tsx
```diff
@@ -0,0 +1,63 @@
...
+        />
+      ))}
+      {hasSourcesWithoutLabelsValue && (
+        <McpCatalogCategorySection
+          key="other-servers"
+          label={SourceLabel.other}
+          searchTerm={searchTerm}
+          pageSize={4}
+          onShowMore={handleShowMoreCategory}
+        />
```
manaswinidas, line 57: Each `McpCatalogCategorySection` independently calls `useMcpServersBySourceLabelWithAPI`, so N labels = N parallel API calls on mount. `McpCatalogSourceLabelBlocks` also re-derives the same `filterEnabledCatalogSources → getUniqueSourceLabels → orderLabelsByPriority` chain. Consider computing this once in the context provider to avoid duplicating the work.

clients/ui/bff/internal/api/catalog_sources_handler_test.go
```diff
@@ -149,3 +149,69 @@ var _ = Describe("TestGetAllCatalogSourcesHandler", func() {
...
+			actual, rs, err := setupApiTest[CatalogLabelListEnvelope](http.MethodGet, "/api/v1/model_catalog/labels?namespace=kubeflow&assetType=models", nil, kubernetesMockedStaticClientFactory, requestIdentity, "kubeflow")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("should return model catalog labels")
+			Expect(rs.StatusCode).To(Equal(http.StatusOK))
+			Expect(actual.Data).NotTo(BeNil())
+			Expect(actual.Data.Size).To(Equal(expected.Data.Size))
+			Expect(actual.Data.PageSize).To(Equal(expected.Data.PageSize))
+			Expect(len(actual.Data.Items)).To(Equal(len(expected.Data.Items)))
+			Expect(actual.Data.Items).To(Equal(expected.Data.Items))
```
manaswinidas, line 193: Nit: The default (no `assetType`) and `assetType=models` test cases both hit `GetCatalogLabelListMock()` in the mock and assert identical values. They're effectively the same test — consider differentiating the mock data so they actually verify the filtering path.

manaswinidas:
Tested on a `kind` cluster 
Screenshot 2026-03-19 at 9 29 42 PM
Screenshot 2026-03-19 at 9 29 56 PM
Screenshot 2026-03-19 at 9 30 24 PM

The filters and search are all working fine - one thing I'm concerned about is missing metadata on the MCP cards specially these - is the backend not returning these fields? sorry if this has been discussed earlier

Screenshot 2026-03-19 at 9 36 21 PM

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpServerToolsSection.tsx
ppadti: But backend is sending read-only as default one, so aligned it with backend.

clients/ui/bff/internal/api/catalog_sources_handler_test.go
ppadti: That is the expectation. with assetType=models and with no assetType it should return models data

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
ppadti: I double checked, generateCategoryName never did any humanization or case transformation.

clients/ui/bff/internal/api/mcp_servers_catalog_handler.go
ppadti: Good catch — this pattern is actually pre-existing and repeated 13 times across 6 handler files (registered_models_handler.go, model_versions_handler.go, model_artifacts_handler.go, artifacts_handler.go, catalog_source_preview_handler.go, and mcp_servers_catalog_handler.go). The MCP handlers just followed the existing convention. Can we have a separate followup PR to cleanup by extracting is as a helper?

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogAllServersView.tsx
ppadti: > Each McpCatalogCategorySection independently calls useMcpServersBySourceLabelWithAPI, so N labels = N parallel API calls on mount.

This is expected behaviour, and aligned with what model catalog is doing.

> also re-derives the same filterEnabledCatalogSources → getUniqueSourceLabels → orderLabelsByPriority chain. Consider computing this once in the context provider to avoid duplicating the work.

We are using React.useMemo so it's only recomputed when dependencies change.

ppadti:
> Tested on a `kind` cluster Screenshot 2026-03-19 at 9 29 42 PM Screenshot 2026-03-19 at 9 29 56 PM Screenshot 2026-03-19 at 9 30 24 PM
> 
> The filters and search are all working fine - one thing I'm concerned about is missing metadata on the MCP cards specially these - is the backend not returning these fields? sorry if this has been discussed earlier
> 
> Screenshot 2026-03-19 at 9 36 21 PM

hmm... I missed this. This is a bug in backend code, will mention this to them

Issue #2384, opened 2026-03-13
Use real MCP servers for the MCP catalog demo


pboyd:
The MCP catalog demo data has fake AI-generated servers when there are perfectly good real servers we could use instead.

---

> I think this URL is wrong - returns 404
>
> Perhaps we wanna use this one here: https://github.com/containers/kubernetes-mcp-server  ?

_Originally posted by @matzew in https://github.com/kubeflow/model-registry/pull/2301#discussion_r2925840062_

RashmiAg27:
Hi, I'd like to work on this!

RashmiAg27:
assign

pboyd:
@RashmiAg27, it looks like @Deepak8858 already sent a PR for this. But that PR is updating the BFF mock data, instead of the [demo manifests](https://github.com/kubeflow/model-registry/tree/main/manifests/kustomize/options/catalog/overlays/demo). I don't think there's anything wrong with doing both, but the manifests were the intention of this issue. Can you two work out how to divide this?

RashmiAg27:
Hi @pboyd, thank you! I'll take the demo manifests part and open a PR soon.

RashmiAg27:
Hi @pboyd! While going through the demo manifest files, I noticed neither Gemini nor OpenAI have official MCP server.So should i entirely remove this link or replace with closest well-known community alternatives?

pboyd:
Yes, feel free. That file is literally just AI-hallucinated MCP server metadata. It's there so we could get _something_ back from the API. It's junk, remove whatever you need to.

PR #2430, opened 2026-03-18
fix: use real MCP servers for catalog demo

Deepak8858:
Replaces fake AI-generated or generic MCP servers with real ones in the catalog demo data. Specifically:\n\n1. Updated Kubernetes MCP Server to use .\n2. Replaced the generic 'Standalone MCP Server' with 'Google Maps MCP Server'.\n\nFixes #2384


pboyd:
Hi @Deepak8858, thanks for the PR. I'll leave it for someone more knowledgeable than me on the BFF to review, but this doesn't really solve #2384, because that was specifically about the [demo manifests](https://github.com/kubeflow/model-registry/tree/main/manifests/kustomize/options/catalog/overlays/demo), not BFF mock data.

btw, you'll need to add the "Sign-off" line to your commit message to pass the DCO check (`git commit --amend -as; git push -f` is probably all you need, but click the link for the failing check for the instructions).

/ok-to-test

pboyd:
This was fixed in #2441 instead.

PR #2434, opened 2026-03-18
feat: extract shared infrastructure into internal/platform layer

Al-Pragliola:
## Description

Introduces internal/platform/, a new package tree that separates reusable infrastructure from model-registry-specific internals, then migrates the model registry to use it.

### Problem

The catalog (and any future registry) must import packages from the model registry's internal/ tree — internal/db/, internal/datastore/, internal/apiutils/, etc. These packages mix generic database infrastructure with model-registry-specific logic. This makes the model registry the platform itself rather than a consumer of it, and makes it impossible to add a new registry without depending on model-specific code.

   ### What this does

  1. Creates internal/platform/ with all shared infrastructure: database abstractions (entity model, filtering, query building, repository, schema, scopes, migrations), datastore connectors, error types, TLS config, proxy/health infrastructure, server middleware, and API utilities.
  2. Migrates the model registry to import from internal/platform/ — converts shared packages (db/models/, db/filter/, db/service/, db/types/) to thin re-export shims, updates all internal consumers (core/, converter/, mapper/, proxy/, server/, etc.), and deletes the now-redundant originals.
  3. Migrates the catalog — all catalog files switch from internal/db/, internal/datastore/, and pkg/api imports to their internal/platform/ equivalents. The catalog no longer depends on any model-registry-specific internal package.
  4. Updates tooling — gorm-gen, Makefile, and scripts/remove_gorm_defaults.sh updated to reference the new platform migration and schema paths.

   ### Design constraint

   The platform layer has zero imports from model-specific packages. No file in internal/platform/ references internal/core/, internal/db/models/, internal/converter/, internal/server/openapi/, or internal/defaults/.

## How Has This Been Tested?

  - go build ./... passes from root and catalog
  - go test ./... -short passes
  - go vet ./... clean (pre-existing participle parser tag warnings only)
  - No file in internal/platform/ imports from model-specific packages
  - Existing functionality unchanged — pure structural refactor

## Merge criteria:


catalog/internal/db/filter/entity_mappings.go
```diff
@@ -86,32 +86,18 @@ func (c *catalogEntityMappings) IsChildEntity(entityType filter.RestEntityType)
...
-	case "name":
-		// If value already contains ":", treat as full namespaced name — exact match only.
-		if strings.Contains(strVal, ":") {
-			return nil, false
-		}
-		// Match namespaced form so model name without source prefix still finds the row.
-		return "%:" + escapeLike(strVal), true
-	default:
-		return nil, false
-	}
```
pboyd, line 114: This reverts the change from #2424. Assuming that was unintentional?

internal/platform/datastore/repos.go
```diff
@@ -3,23 +3,13 @@ package datastore
 import "reflect"
 
 // Spec is the specification for the datastore.
-// Each entry
```
pboyd, line 6: These doc comments aren't perfect, but I think they do have some value. Can we keep them? Looks like they've been removed from a other few files too.

internal/platform/db/repository/generic_repository.go
```diff
@@ -0,0 +1,598 @@
...
+				r.setCreateTime(&schemaEntity, now)
+			}
+		} else {
+			if existingUpdateTime == 0 {
+				r.setLastUpdateTime(&schemaEntity, now)
+			}
+			if existingCreateTime == 0 {
+				r.setCreateTime(&schemaEntity, now)
+			}
+		}
```
pboyd: I know this was just copied here, but I took some AI help with this review and so I'll pass along the finding. The if/else branches here are identical.


```suggestion
		if existingUpdateTime == 0 {
			r.setLastUpdateTime(&schemaEntity, now)
		}
		if existingCreateTime == 0 {
			r.setCreateTime(&schemaEntity, now)
		}
```

internal/platform/db/scopes/paginate.go
```diff
@@ -166,7 +150,6 @@ func DecodeCursor(token string) (*Cursor, error) {
 		return nil, err
 	}
 
-	// Split only on the first ":" so the value can contain colons (e.g. stored names "sourceId:modelName").
```
pboyd, line 169: A lot of the removed comments are junk, but this one seems like a helpful explanation of something subtle.

internal/server/middleware/validation.go
```diff
@@ -12,39 +12,31 @@ import (
...
 }
 
 // ValidationMiddleware validates all query parameters and request body for unsafe characters
 func ValidationMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// Parse query parameters
 		queryParams, err := url.ParseQuery(r.URL.RawQuery)
 		if err != nil {
-			// If we can't parse query parameters, let the request continue
-			// The parsing error will be handled elsewhere
```
pboyd, line 42: This is another helpful comment.

internal/tls/config.go
```diff
@@ -90,8 +90,6 @@ func (c *TLSConfig) BuildTLSConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
-// parseCipherSuites parses a colon-separated list of cipher suite names
-// and returns the corresponding cipher suite IDs
```
pboyd, line 94: Not the best doc comment. But it at least gives some indication of the expected input format.

pboyd (commented):
Looks good, in general. I didn't go line-by-line through the added files because I assume they've just been copied. Let me know if there's anything I should pay particular attention to.

I think there's some lost code (at least #2424), and some of the removed comments were probably helpful (mostly, I agree with removing them--but a few have information that's not obvious from the code).

Al-Pragliola:
> Looks good, in general. I didn't go line-by-line through the added files because I assume they've just been copied. Let me know if there's anything I should pay particular attention to.
> 
> I think there's some lost code (at least #2424), and some of the removed comments were probably helpful (mostly, I agree with removing them--but a few have information that's not obvious from the code).


`internal/db/filter/` was a single package that mixed generic query-building with model-registry-specific
   entity mappings. I split it so the platform side defines interfaces (RestEntityType, EntityMappingFunctions, EqualityExpander in a new entity_types.go) and the query builder accepts those via injection instead of hardcoding defaultEntityMappings. The model-registry-specific mappings stay in internal/db/filter/. This is the change that actually enables other registries to plug in their own entity types

`proxy/health.go` was extracted from readiness.go. The generic health check plumbing (HealthChecker, DatabaseHealthChecker, GeneralReadinessHandler) moved to platform, and readiness.go kept only the model-registry-specific readiness logic

`errors/errors.go` + `pkg/api/error.go` — sentinel errors moved to platform, and pkg/api re-exports them as var ErrX = platformerrors.ErrX. Worth a sanity check that errors.Is still works through the re-export (it does — same pointer identity — but it's the kind of thing that's easy to get wrong)

`db/drivers/drivers.go` groups the MySQL and Postgres blank imports into one package so consumers import _ "internal/platform/db/drivers" instead of blank-importing each backend separately. It's a small readability win — someone new to the codebase can look at one place to see what backends are available

`db/entity/` files are extractions from `db/models/` — Properties, Pagination, Entity[T], BaseEntity[T], etc. The originals are now type alias shims. Straightforward but there's enough surface area that it's worth confirming nothing got lost in translation

pboyd (approved):
Thanks for the changes, @Al-Pragliola.

/lgtm

Al-Pragliola:
closed in favor of #2644

PR #2440, opened 2026-03-19
feat(csi): add per-namespace overlay for multi-profile deployments

### LINKED ISSUE #2227: Namespace-scoped StorageContainer for profiles and service accounts separation.
I am deploying a model registry in a kubeflow cluster. I am using profiles to separate users data and actions.
I also have kserve components deployed, and I want to use models from the kubeflow model registry.
To use the starageUri in the model registry reference format (model-registry://model/version), I am using the [model-registry-storage-initializer](https://github.com/kubeflow/model-registry/blob/v0.3.6/manifests/kustomize/options/csi/clusterstoragecontainer.yaml). I have noticed that these are cluster-scoped resources, and I cannot use them with multi-profiles because they reference the namespace-scoped service account (MODEL_REGISTRY_BASE_URL).
```
apiVersion: "serving.kserve.io/v1alpha1"
kind: ClusterStorageContainer
metadata:
  name: model-registry-storage-initializer
spec:
  container:
    name: storage-initializer
    image: ghcr.io/kubeflow/model-registry/storage-initializer:latest
    env:
    - name: MODEL_REGISTRY_BASE_URL
      value: "model-registry-service.kubeflow.svc.cluster.local:8080"
    resources:
      requests:
        memory: 100Mi
        cpu: 100m
      limits:
        memory: 1Gi
  supportedUriFormats:
    - prefix: model-registry://
```

namespace-scoped StorageContainer for profiles and service accounts separation.

Storm1289:
Closes #2227

## What's the problem?
`ClusterStorageContainer` is cluster-wide, so the CSI storage initializer
could only point to one Model Registry via `MODEL_REGISTRY_BASE_URL`.

In multi-profile Kubeflow setups, each namespace runs its own registry.
The old workaround was to embed the full internal Kubernetes DNS address:

  storageUri: "model-registry://model-registry-service.profile-alpha.svc.cluster.local:8080/iris/v1"

This was poor UX  users shouldn't need to know internal Kubernetes DNS
rules just to load a model. It also required cumbersome Kustomize overlays
as workarounds.

## What's the fix?
Users can now pass the target namespace as a simple query parameter:

  storageUri: "model-registry://iris/v1?namespace=profile-alpha"

The CSI initializer builds the correct internal URL automatically.

## What changed?

**`cmd/csi/internal/modelregistry/api_client.go`**
- Parses `?namespace=` from `storageUri`
- If found, dynamically builds:
  `<service-name>.<namespace>.<cluster-domain>:<port>`
- Added regex validation for `namespace` (`^[a-zA-Z0-9_-]+$`) and
  `port` (`^[0-9]+$`)  malformed values fall back to the cluster
  default safely
- Uses `net.SplitHostPort` and `strings.Cut` for IPv6-safe parsing
- If not found, falls back to `MODEL_REGISTRY_BASE_URL` as before

**`cmd/csi/main.go`**
- Added `MODEL_REGISTRY_SERVICE_NAME` env variable (default:
  `model-registry-service`)
- Added `MODEL_REGISTRY_CLUSTER_DOMAIN` env variable (default:
  `svc.cluster.local`) so custom infrastructure setups are supported

**`cmd/csi/internal/storage/modelregistry_provider.go`**
- Strips query parameters before parsing model and version name so
  `iris?namespace=profile-alpha` is never sent as the model name

**Docs updated**
- Removed the old `manifests/kustomize/overlays/per-namespace-csi`
  overlay  no longer needed since `?namespace=` replaces it natively
- Updated `cmd/csi/GET_STARTED.md` and `manifests/kustomize/README.md`
  to document the new `?namespace=` pattern

## Example
```yaml
# Before (complex)
storageUri: "model-registry://model-registry-service.profile-alpha.svc.cluster.local:8080/iris/v1"

# After (simple)
storageUri: "model-registry://iris/v1?namespace=profile-alpha"
```
Both route to: `model-registry-service.profile-alpha.svc.cluster.local:8080`

## Verification
- ✅ Unit tests: 8 new cases in `TestParseModelVersion` covering standard
  URIs, embedded-host URIs, and `?namespace=` combinations
- ✅ E2E tests: Scenarios 3 & 4 in `test/csi/e2e_test.sh` updated to
  use and validate the new `?namespace=` syntax
- ✅ Local build: `go build` passes inside `cmd/csi`

## Limitations
`MODEL_REGISTRY_BASE_URL` still acts as a fallback when no `?namespace=`
is provided.


pboyd:
Hi, @Storm1289, thanks for the PR.

This appears to document a work-around instead of addressing the issue. This work-around requires users to know the internal kubernetes URL, which is something that most users won't know.

The original request was to have a namespace-scoped CRD. And, while that would be ideal, it would require a change to KServe (it's their CRD, after all). I think it would be enough to extend the existing model-registry style URLs with some sort of namespace marker. For instance, supporting URLs of this form would meet the objective: `model-registry://iris/v1?namespace=my-user-namespace`.

Storm1289:
Hi @pboyd , Thank you for the review
I've made the changes , could you take another look. 
Let me know if anything else needs fixing.

cmd/csi/csi
pboyd, line 1: I'm guessing you added this by mistake. Can you remove it?

cmd/csi/internal/modelregistry/api_client.go
```diff
@@ -2,16 +2,38 @@ package modelregistry
...
+	u, err := url.Parse(storageUri)
+	if err == nil {
+		ns := u.Query().Get("namespace")
+		if ns != "" {
+			// Extract port from cfg.Host
+			hostParts := strings.Split(cfg.Host, ":")
+			port := "8080"
+			if len(hostParts) == 2 {
+				port = hostParts[1]
+			}
```
pboyd: I'll recommend you use [net.SplitHostPort](https://pkg.go.dev/net#SplitHostPort) for this:

```suggestion
			_, port, err := net.SplitHostPort(cfg.Host)
			if err != nil {
				port = "8080"
			}
```

cmd/csi/internal/modelregistry/api_client.go
```diff
@@ -2,16 +2,38 @@ package modelregistry
...
+			
+			newCfg := openapi.NewConfiguration()
+			newCfg.Host = fmt.Sprintf("%s.%s.svc.cluster.local:%s", serviceName, ns, port)
```
pboyd: The namespace and port should be validated.

- **namespace**: should only allow characters `[a-zA-Z0-9_-]`
- **port**: should only allow digits: `[0-9]`

And can `svc.cluster.local` be configurable? `svc.cluster.local` is the default, but it's possible for a cluster to configure this so we ought to be flexible.

cmd/csi/internal/storage/modelregistry_provider.go
```diff
@@ -123,6 +123,9 @@ func (p *ModelRegistryProvider) DownloadModel(modelDir string, modelName string,
 func (p *ModelRegistryProvider) parseModelVersion(storageUri string) (string, *string, error) {
 	var versionName *string
 
+	// Strip query parameters
+	storageUri = strings.Split(storageUri, "?")[0]
```
pboyd: This is fine, but you might want to consider `strings.Cut`, since you only need the first match:

```suggestion
	storageUri, _, _ = strings.Cut(storageUri, "?")
```

Storm1289:
Hi @pboyd , I have implemented the changes you suggested
Could you take another look

pboyd:
Thanks for the fixes, @Storm1289.

/lgtm

I'll leave it open a bit longer in case anyone else wants to review. But I'll merge it tomorrow if I don't hear anything.

pboyd:
Thanks again for working on this, @Storm1289.

/approve

PR #2465, opened 2026-03-24
Implement cursor-based pagination for Hardware Configuration table

Taj010:
## Description

This PR removes the temporary pageSize: '99999' workaround in Performance Insights and implements proper cursor pagination for the Hardware Configuration table. The paginated flow using `pageSize` + `nextPageToken` and the load more button to fetch and append the more items is added. 

## How Has This Been Tested?

Added Cypress coverage for pagination behavior in` modelCatalogTabs.cy.ts`

https://github.com/user-attachments/assets/a6cd1035-437a-4ab6-85f3-3482e5bca556

 _tested by setting HARDWARE_CONFIG_PAGE_SIZE = 10 but the default is set to 20_ 
 
## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
```diff
@@ -91,3 +97,162 @@ export const useCatalogPerformanceArtifacts = (
...
+      }
+      if (!sourceId) {
+        return Promise.reject(new Error('No source id'));
+      }
+      if (!modelName) {
+        return Promise.reject(new Error('No model name'));
+      }
+      if (!enabled) {
+        return Promise.reject(new Error('Fetching is disabled'));
+      }
```
mturley, line 158: These rejections should use `NotReadyError` instead of plain `Error`, matching the original `useCatalogPerformanceArtifacts` hook (lines 72-76). `useFetchState` treats `NotReadyError` specially — it does not set the error state, so the component waits silently while params are still resolving. Using plain `Error` here would cause the error alert to flash briefly on mount.

```suggestion
    (opts) => {
      if (!apiAvailable) {
        return Promise.reject(new Error('API not yet available'));
      }
      if (!sourceId) {
        return Promise.reject(new NotReadyError('No source id'));
      }
      if (!modelName) {
        return Promise.reject(new NotReadyError('No model name'));
      }
      if (!enabled) {
        return Promise.reject(new NotReadyError('Fetching is disabled'));
      }
```

clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
```diff
@@ -91,3 +97,162 @@ export const useCatalogPerformanceArtifacts = (
...
+      setNextPageToken(firstPageData.nextPageToken);
+    }
+  }, [firstPageData, loaded, error]);
+
+  const loadMore = React.useCallback(async () => {
+    if (!nextPageToken || isLoadingMore || !apiAvailable || !enabled) {
+      return;
+    }
+
+    setIsLoadingMore(true);
```
mturley, line 192: Nit: This is probably fine, but `isLoadingMore` is captured from state in this closure, so two rapid clicks could both read `isLoadingMore === false` before either `setState` is processed. PF's `isLoading` prop on the Button likely prevents this in practice, but a ref guard would be more bulletproof (e.g. `const isLoadingMoreRef = React.useRef(false);`, used alongside the isLoadingMore state to block `loadMore` from proceeding. You'd need both since the state version is required to trigger a re-render to show the loading indicator).

clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
```diff
@@ -91,3 +97,162 @@ export const useCatalogPerformanceArtifacts = (
...
+    isLoadingMore,
+    enabled,
+  ]);
+
+  React.useEffect(() => {
+    // Keep current rows while query is refetching to avoid a brief empty flash.
+    setNextPageToken('');
+    setIsLoadingMore(false);
+    setLoadMoreError(undefined);
+  }, [sourceId, modelName, normalizedParams, filterData, filterOptions, enabled]);
```
mturley, line 239: These two effects both manage the same state (`nextPageToken`, `isLoadingMore`, `loadMoreError`) and their interaction may be surprising. When filters/params change:

1. The second effect (line 227) fires and resets `nextPageToken`/`isLoadingMore`/`loadMoreError`, but intentionally keeps `allItems` stale (per the comment).
2. `useFetchState` triggers a refetch of the first page.
3. The first effect (line 176) fires when the new `firstPageData` arrives and overwrites `allItems`.

During the window between steps 1 and 3, `performanceArtifacts.items` holds stale data from the previous filter set, while `performanceArtifacts.size` (which reads directly from `firstPageData.size`) may already reflect the new response. A consumer reading both simultaneously could see inconsistent data (e.g. "showing 42 items" with `size: 5`).

If the intent is to avoid a flash of empty content, consider also resetting `allItems` to `[]` and showing a loading skeleton instead — this would be more consistent and avoids the stale-data window entirely. Alternatively, a comment explaining this intentional tradeoff would help future maintainers.

clients/ui/frontend/src/app/pages/modelCatalog/screens/PerformanceInsightsView.tsx
```diff
@@ -81,22 +84,23 @@ const PerformanceInsightsView: React.FC<PerformanceInsightsViewProps> = ({ model
...
+  // Fetch performance artifacts from server with filtering and cursor-based pagination
+  const {
+    performanceArtifacts,
+    performanceArtifactsLoaded,
+    performanceArtifactsLoadError: performanceArtifactsError,
+  } = usePaginatedCatalogPerformanceArtifacts(
+    decodedParams.sourceId || '',
+    encodeURIComponent(`${decodedParams.modelName}`),
+    {
+      targetRPS,
```
mturley, line 96: `orderBy` and `sortOrder` are not passed here, so the server returns items in its default order. The `HardwareConfigurationTable` then sorts them client-side via PatternFly's built-in table sorting.

Before this PR, that was fine because `pageSize: '99999'` fetched everything. But now with real pagination, client-side sorting of a partial dataset produces incorrect results — the sorted top-20 from the server (if sorted server-side) would likely be different from what the client computes by sorting only the first page.

The table's active sort column and direction need to be lifted up and passed as `orderBy` / `sortOrder` here, and when they change, the paginated hook should reset to page 1 and refetch.

mturley:
oof sorry for the noise, claude had trouble with the github API 😅

Taj010:
Np! thank you, I will have a look again and test the sorting too 😄

PR #2467, opened 2026-03-24
feat(ui): support numeric custom properties in Model Registry UI

Taj010:
## Description
The Model Registry UI currently filters out custom properties with numeric types (`int_value`, `double_value`), displaying only string properties. While the backend correctly stores all property types, users cannot see or edit numeric metadata in the UI.

This PR expands UI support to include INT and DOUBLE property types:

1. **Display**: Numeric properties now appear in the properties table (rendered as strings)
2. **Edit**: Type detection automatically creates properties with the correct type:
   - Integer values (`/^-?\d+$/`) → INT type
   - Decimal values (`/^-?\d+\.\d+$/`) → DOUBLE type
   - Everything else → STRING type
3. **Type transitions**: Users can edit properties and change types (e.g., "42" → "42.5" converts INT → DOUBLE)

Scope
- ✅ Added support for STRING, INT, and DOUBLE types
- ✅ Labels continue to work unchanged (STRING with empty value)
-  ⏸️ Deferred STRUCT, PROTO, BOOL types (require specialized UI)

Changes
- `utils.ts`: Added `getPropertyValue()` helper and type detection in `mergeUpdatedProperty()`
- `ModelPropertiesExpandableSection.tsx`: Updated to use value extraction helper
- Added comprehensive test coverage (20+ new tests)

## How Has This Been Tested?
- New regression tests for numeric property display
- New tests for type detection (INT, DOUBLE, STRING)
- Tests for type transitions (numeric ↔ string, int ↔ double)
- Edge case coverage (negative numbers, empty strings, scientific notation)
- Existing tests pass with numeric properties included

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/__tests__/utils.spec.ts
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/__tests__/utils.spec.ts
@@ -455,7 +455,7 @@ describe('mergeUpdatedProperty', () => {
         newPair: { key: 'version', value: '007' },
       });
       expect(result).toEqual({
-        version: { int_value: '007', metadataType: ModelRegistryMetadataType.INT },
+        version: { string_value: '007', metadataType: ModelRegistryMetadataType.STRING },
       } satisfies ModelRegistryCustomProperties);
     });
   });
--- a/clients/ui/frontend/src/app/pages/modelRegistry/screens/utils.ts
+++ b/clients/ui/frontend/src/app/pages/modelRegistry/screens/utils.ts
@@ -113,6 +113,12 @@ export const getProperties = <T extends ModelRegistryCustomProperties>(
 const INTEGER_INPUT_PATTERN = /^-?\d+$/;
 const DECIMAL_INPUT_PATTERN = /^-?\d+\.\d+$/;
 
+// "007", "01", "-01" match INTEGER_INPUT_PATTERN but should stay STRING so leading zeros are preserved.
+const hasLeadingZeroIntegerForm = (value: string): boolean => {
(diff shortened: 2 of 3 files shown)


clients/ui/frontend/src/app/pages/modelRegistry/screens/__tests__/utils.spec.ts
```diff
@@ -313,6 +373,154 @@ describe('mergeUpdatedProperty', () => {
...
+    it('should create STRING property for values with leading zeros', () => {
+      const customProperties: ModelRegistryCustomProperties = {};
+      const result = mergeUpdatedProperty({
+        customProperties,
+        op: 'create',
+        newPair: { key: 'version', value: '007' },
+      });
+      expect(result).toEqual({
+        version: { int_value: '007', metadataType: ModelRegistryMetadataType.INT },
+      } satisfies ModelRegistryCustomProperties);
```
mturley, line 459: This test name says "should create STRING property for values with leading zeros" but the assertion expects INT (`int_value: '007'`). The regex `^-?\d+$` does match `"007"`, so INT is the actual behavior.

Either:
- **Fix the test name** to match the behavior: e.g. `"should create INT property for values with leading zeros"` (if `007` → INT is intentional)
- **Or fix the behavior** to treat leading-zero strings like `"007"` as STRING (if preserving the leading zero semantics matters — `007` as an integer would normally be `7`)

I'd lean toward treating leading-zero values as STRING, since leading zeros are typically meaningful (version numbers, codes, etc.) and would be lost if round-tripped through an integer. But either way, the test name should match what's actually being asserted.

mturley:
/lgtm
/approve
Thanks!

PR #2475, opened 2026-03-25
fix(catalog): enable independent scrolling for filter panel

manaswinidas:
## Description

Enables independent scrolling for the filter panel in Model Catalog, preventing users from losing their filtering context when scrolling through model results.

Previously, the filter panel and content area scrolled together, making it difficult for users to maintain context of their filtering selections while browsing models. The filter panel now uses sticky positioning with its own scroll container.

**Changes:**
- Added sticky positioning to filter panel (SidebarPanel)
- Set max-height: 100vh with overflow-y: auto
- Filter panel now scrolls independently from content area

## How Has This Been Tested?

Manual verification:
1. Navigate to Model Catalog
2. Reduce browser window height to trigger scrolling
3. Scroll through model results
4. Verify filter panel remains visible and scrolls independently
5. With many filters applied, verify filter panel scrolls separately
6. Test at various viewport sizes

## Merge criteria:
- All the commits have been signed-off (To pass the DCO check)
- [x] Code changes follow the kubeflow contribution guidelines.
- [ ] **For first time contributors**: Please reach out to the Reviewers to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
@@ -44,8 +44,7 @@ const ModelCatalog: React.FC = () => {
         <Sidebar hasBorder hasGutter>
           <SidebarPanel
             variant="sticky"
-            width={{ default: 'width_300' }}
-            style={{ maxHeight: '100vh', overflowY: 'auto' }}
+            style={{ maxHeight: '100vh', overflowY: 'auto', width: '300px' }}
           >
             <ModelCatalogFilters />
           </SidebarPanel>
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
@@ -42,10 +42,7 @@ const ModelCatalog: React.FC = () => {
         provideChildrenPadding
       >
         <Sidebar hasBorder hasGutter>
-          <SidebarPanel
-            variant="sticky"
-            style={{ maxHeight: '100vh', overflowY: 'auto', scrollbarGutter: 'stable' }}
-          >
+          <SidebarPanel variant="sticky" style={{ maxHeight: '100vh', overflowY: 'auto' }}>
             <ModelCatalogFilters />
           </SidebarPanel>
           <SidebarContent>
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
@@ -42,14 +42,7 @@ const ModelCatalog: React.FC = () => {
         provideChildrenPadding
       >
         <Sidebar hasBorder hasGutter>
-          <SidebarPanel
-            style={{
-              position: 'sticky',
-              top: 0,
-              maxHeight: '100vh',
-              overflowY: 'auto',
-            }}
-          >
+          <SidebarPanel variant="sticky" style={{ maxHeight: '100vh', overflowY: 'auto' }}>
             <ModelCatalogFilters />
           </SidebarPanel>
           <SidebarContent>
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
@@ -42,7 +42,11 @@ const ModelCatalog: React.FC = () => {
         provideChildrenPadding
       >
         <Sidebar hasBorder hasGutter>
-          <SidebarPanel variant="sticky" style={{ maxHeight: '100vh', overflowY: 'auto' }}>
+          <SidebarPanel
+            variant="sticky"
+            width={{ default: 'width_300' }}
+            style={{ maxHeight: '100vh', overflowY: 'auto' }}
+          >
             <ModelCatalogFilters />
           </SidebarPanel>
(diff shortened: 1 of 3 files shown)


manaswinidas (commented):
/lgtm

But needs a second opinion - here's a screen recording that will help the upcoming reviewers decide

https://github.com/user-attachments/assets/0fc6d8e7-8d9a-4be7-88c1-25803c90ab04

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -42,7 +42,14 @@ const ModelCatalog: React.FC = () => {
...
         <Sidebar hasBorder hasGutter>
-          <SidebarPanel>
+          <SidebarPanel
+            style={{
+              position: 'sticky',
+              top: 0,
+              maxHeight: '100vh',
+              overflowY: 'auto',
+            }}
+          >
```
Philip-Carneiro: following the PF, it should be: <SidebarPanel variant="sticky" style={{ maxHeight: '100vh', overflowY: 'auto' }}>

Philip-Carneiro (changes requested):

manaswinidas:
Updated to use PatternFly's `variant="sticky"` prop instead of custom positioning. Thanks for the feedback\!

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -42,7 +42,7 @@ const ModelCatalog: React.FC = () => {
         provideChildrenPadding
       >
         <Sidebar hasBorder hasGutter>
-          <SidebarPanel>
+          <SidebarPanel variant="sticky" style={{ maxHeight: '100vh', overflowY: 'auto' }}>
```
Philip-Carneiro: note that scrollbar consumes some width from the actual sidebar, so a little of the border of the things inside are cut out as the image bellow:

Image

Philip-Carneiro (changes requested):

manaswinidas:
Thanks for pointing this out! The scrollbar does consume some width, but I think the current width behavior is acceptable. The sidebar layout adjusts naturally and the content remains accessible. If this becomes a UX concern during testing, we can revisit with additional width adjustments, but for now the default behavior should work well.

manaswinidas:
You're absolutely right about the width issue\! I've added `scrollbar-gutter: 'stable'` which reserves space for the scrollbar even when it's not visible. This prevents the layout shift and content cutoff you observed - the sidebar width now stays constant regardless of whether the scrollbar is showing or not. This is the best of both worlds: independent scrolling without layout shift. 👍

manaswinidas:
You're right - the previous approaches still had the scrollbar taking width. I've switched to a simpler solution: added `paddingRight` to the filter content itself. This ensures the filter components have proper spacing from the scrollbar edge, preventing any border cutoff, while keeping the implementation clean and simple. The scrollbar naturally takes its space, but content stays properly spaced. 👍

mturley (changes requested):
@manaswinidas I'm still seeing the edge of the sidebar get cut off. Interestingly, most of the contents have the correct width but it's just the perf view toggle box that is too wide. Here's a screenshot, and then another scrolled to the right:

Image

Image

manaswinidas:
Good catch\! The issue was the `ModelPerformanceViewToggleCard` had a hardcoded `minWidth: '280px'` which was causing it to overflow when the scrollbar was present. I've removed this constraint so the card naturally fits within the sidebar width. The other filter components didn't have this issue because they didn't have a fixed minimum width. Should be resolved now\! 👍

manaswinidas:
Now it looks like this(human message):
Screenshot 2026-03-26 at 5 16 01 PM

Philip-Carneiro (approved):

manaswinidas:
@mturley can you LGTM if this looks right ?

mturley:
@manaswinidas I'm so sorry for the nitpick, but can we slightly increase the width of the sidebar so the model performance view toggle text still doesn't wrap? It would look closer to the design

manaswinidas:
No worries at all\! I've increased the sidebar width to 300px using PatternFly's `width` prop. This gives the "Model performance view" toggle text enough room to display without wrapping while keeping the sidebar at a reasonable size. 👍

manaswinidas:
Fixed TypeScript error - PatternFly's `width` prop only accepts percentage values (width_25, width_33, etc.), so I've used inline style to set the fixed 300px width instead. Tests should pass now\! ✅

mturley (changes requested):
@manaswinidas sorry to continue the nitpicking - it seems like the 300px width doesn't have any effect, and even if I edit the style in my browser to be wider it's not changing the width. The toggle card still is narrow enough that its text wraps and doesn't look like it used to. Testing in mock mode PF theme:

Screenshot from your branch:
Screenshot 2026-04-03 at 4 41 23 PM

Screenshot from main, wider with the icon and text all inline with the toggle:
Screenshot 2026-04-03 at 4 41 51 PM

This might require some tinkering by hand, AI isn't great at CSS stuff. Happy to take a closer look if you want a hand troubleshooting.

manaswinidas:
@mturley you were right - all the CSS changes for nothing 😫  - just one change was enough 

This is how it looks like now

https://github.com/user-attachments/assets/5c78791f-eb6c-46a7-afdd-d2213a4f7737

mturley:
hehehe no worries. TIL about the sticky variant, good to know. Maybe we should teach our AIs to read the PF docs better, sometimes I think they're relying on old PF docs being available on the web when they were trained 😆

PR #2483, opened 2026-03-25
chore(ci): Remove trivy - model-regitsry

sameerdattav:
Removing trivy action - as instructed by @andreyvelich following https://github.com/kubeflow/katib/pull/2644


tarilabs:
/hold

I'm lost why we need to remove it from Model-Registry, please?
- we had it [pinned to latest v0.35.0](https://github.com/kubeflow/model-registry/pull/2319) at the beginning of the month, so we're not impacted.
- we are also not affected by this type of attack because @Al-Pragliola implemented the OpenSSF scorecard sha-pinning for all GitHub Actions:
👉  https://github.com/kubeflow/model-registry/pull/2370
which IMHO is a very commendable work.

sameerdattav:
@tarilabs , the removal was being done across all Kubeflow repos after the recent Trivy incident, so I followed the same approach here.

If there’s already consensus that Model Registry isn’t affected because of pinning , im happy to revert this.

@andreyvelich — could you please confirm what we should do here?

tarilabs:
thanks, as mentioned:

> * we had it [pinned to latest v0.35.0](https://github.com/kubeflow/model-registry/pull/2319) at the beginning of the month, so we're not impacted.
> * we are also not affected by this type of attack because @Al-Pragliola implemented the OpenSSF scorecard sha-pinning for all GitHub Actions:
>   👉  [chore: pin all GHAs deps #2370](https://github.com/kubeflow/model-registry/pull/2370)
>   which IMHO is a very commendable work

we Disabled the Workflow as discussed:

image


and we are waiting further instructions from @kubeflow/security-team @kubeflow/kubeflow-steering-committee

PR #2485, opened 2026-03-26
Implement cursor-based pagination for Hardware Configuration table

Taj010:
## Description

This PR removes the temporary pageSize: '99999' workaround in Performance Insights and implements proper cursor pagination for the Hardware Configuration table. The paginated flow using pageSize + nextPageToken and the load more button to fetch and append the more items is added.

## How Has This Been Tested?

Added Cypress coverage for pagination behavior in modelCatalogTabs.cy.ts

https://github.com/user-attachments/assets/b55fbdf0-b8b1-40a8-aa31-f270d96f6f50

tested by setting HARDWARE_CONFIG_PAGE_SIZE = 5 but the default is set to 20

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
+++ b/clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
@@ -132,6 +132,8 @@ export const usePaginatedCatalogPerformanceArtifacts = (
   const [isLoadingMore, setIsLoadingMore] = React.useState(false);
   const isLoadingMoreRef = React.useRef(false);
   const [loadMoreError, setLoadMoreError] = React.useState<Error | undefined>();
+  const queryVersionRef = React.useRef(0);
+  const queryVersion = queryVersionRef.current;
 
   const normalizedParams = useNormalizedPerformanceParams(params);
 
@@ -185,6 +187,9 @@ export const usePaginatedCatalogPerformanceArtifacts = (
     if (!nextPageToken || !apiAvailable || !enabled) {
       return;
     }
+    if (queryVersion !== queryVersionRef.current) {
+      return;
+    }
     if (isLoadingMoreRef.current) {
       return;
     }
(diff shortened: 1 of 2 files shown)


clients/ui/frontend/src/app/hooks/modelCatalog/useCatalogPerformanceArtifacts.ts
```diff
@@ -91,3 +97,169 @@ export const useCatalogPerformanceArtifacts = (
...
+    if (!nextPageToken || !apiAvailable || !enabled) {
+      return;
+    }
+    if (isLoadingMoreRef.current) {
+      return;
+    }
+    isLoadingMoreRef.current = true;
+    setIsLoadingMore(true);
+    setLoadMoreError(undefined);
+
```
YuliaKrimerman, line 199: The normalizedParams includes nextPageToken from the incoming props, but loadMore also reads the local nextPageToken state. When spread together ({ ...normalizedParams, nextPageToken }), local state wins — which is correct. However, if normalizedParams changes (due to sort/filter change), the old loadMore callback could fire before the effect resets nextPageToken to ''. Consider adding a guard: 
`const loadMore = React.useCallback(async () => {
  if (!nextPageToken || !apiAvailable || !enabled) {
    return;
  }
  if (isLoadingMoreRef.current) {
    return;
  }
  // ...
}, [/* deps */]);`

clients/ui/frontend/src/app/pages/modelCatalog/components/HardwareConfigurationTable.tsx
```diff
@@ -31,9 +34,26 @@ const HardwareConfigurationTable: React.FC<HardwareConfigurationTableProps> = ({
...
+    if (sortColumnField === null) {
+      return;
+    }
+
+    onSortChange?.({
+      orderBy: sortColumnField,
+      sortOrder: sortDirection === 'asc' ? SortOrder.ASC : SortOrder.DESC,
+    });
+  }, [onSortChange, sortColumnField, sortDirection]);
+
```
YuliaKrimerman, line 57: Consider whether the parent should also be notified when sort is cleared (i.e., sortColumnField becomes null). If the server should fall back to default sorting, we should send onSortChange?.({}) when sort is cleared:

`React.useEffect(() => {
  if (sortColumnField === null) {
    onSortChange?.({});
    return;
  }
  onSortChange?.({
    orderBy: sortColumnField,
    sortOrder: sortDirection === 'asc' ? SortOrder.ASC : SortOrder.DESC,
  });
}, [onSortChange, sortColumnField, sortDirection]);`

PR #2545, opened 2026-04-03
feat(ui): register and display model_type from catalog and model registry

Taj010:
## Description

This PR adds a shared Model type field (Generative / Predictive) on register-from-catalog and register-from-registry flows.  
- Catalog registration prefills the field from catalog metadata when it is generative or predictive. 
- Register-from-registry requires a selection before submit. 

Model overview and version details was also adjusted to show model type property. 

image

image

image

## How Has This Been Tested?

Manually Testing
- Open register-from-catalog with a catalog model confirm dropdown prefills
- Open MR Register model, confirm submit stays disabled until model type is chosen
- Open model and version detail pages and confirm Model type displays as expected.

Automated Testing 
- Added tests to `modelCatalogUtils.spec.ts`and `utils.spec.ts file ` and adjusted tests in `registerAndStoreFields.cy.ts`

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelTypeField.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const RegisterModelTypeField: React.FC<RegisterModelTypeFieldProps> = ({
+  modelCustomProperties,
+  onModelCustomPropertiesChange,
+  isRequired,
+}) => {
+  const stored = getModelTypeStoredValueFromCustomProperties(modelCustomProperties);
+
+  const handleChange = (key: string, isPlaceholder: boolean) => {
```
YuliaKrimerman: The handleChange callback checks both isPlaceholder and key === MODEL_TYPE_PLACEHOLDER_KEY. The isPlaceholder check alone should be sufficient — checking the key as well is redundant but harmless.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/registerModelTypeUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (!next) {
+    delete result[MODEL_TYPE_CUSTOM_PROPERTY_KEY];
+  } else {
+    result[MODEL_TYPE_CUSTOM_PROPERTY_KEY] = {
+      metadataType: ModelRegistryMetadataType.STRING,
+      // eslint-disable-next-line camelcase
+      string_value: next,
+    };
+  }
+  return result;
```
YuliaKrimerman, line 48: Please add unit tests for this utility.

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/utils.ts
```diff
@@ -200,11 +201,18 @@ export const isRegisterModelSubmitDisabled = (
   registeredModels: RegisteredModelList,
   namespaceHasAccess?: boolean,
   isNamespaceAccessLoading?: boolean,
-): boolean =>
-  !formData.modelName ||
-  isSubmitDisabledForCommonFields(formData, namespaceHasAccess, isNamespaceAccessLoading) ||
-  !isNameValid(formData.modelName) ||
-  isModelNameExisting(formData.modelName, registeredModels);
+  options?: { requireModelType?: boolean },
```
YuliaKrimerman, line 205: We've changed the default behavior here which is fine, the only thing I would maybe comment here that if in the future a caller will forget to opt-out it might be surprising . Alternatively we can default to false with explicit opt-in( just safer) not a must, you're choice

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -774,3 +774,27 @@ export const formatModelTypeDisplay = (modelTypeRaw: string | null): string => {
   // Fallback: capitalize whatever value we got
   return capitalize(modelTypeRaw.trim());
 };
+
+/**
+ * Returns model registry customProperties entries to prefill `model_type` when registering
+ * from the catalog. Only generative/predictive are copied; unknown or missing values yield {}.
+ */
+export const getCatalogModelTypePropertyForRegistration = (
```
YuliaKrimerman, line 801: In here we can reuse reuse getModelTypeStoredValueFromCustomProperties from registerModelTypeUtils.ts , or even better - buildCustomPropertiesWithModelType

YuliaKrimerman (changes requested):

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/registerModelTypeUtils.ts
Taj010: Added `registerModelTypeUtils.spec.ts`

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/utils.ts
Taj010: Adjusted to have `requireModelType` now defaults to false and adjusted the `utils.spec.ts` too

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
Taj010: Reimplemented via `getModelTypeStoredValueFromCustomProperties `and `buildCustomPropertiesWithModelType `

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
```diff
@@ -146,6 +155,11 @@ const ModelDetailsCard: React.FC<ModelDetailsCardProps> = ({
       >
         <ModelTimestamp timeSinceEpoch={rm.createTimeSinceEpoch} />
       </DashboardDescriptionListGroup>
+      <DashboardDescriptionListGroup title="Model type">
+        <Content component="p" data-testid="registered-model-model-type">
+          {formatModelTypeDisplay(modelTypeRaw)}
+        </Content>
```
mturley: `model_type` will also appear in the editable properties table rendered by `ModelPropertiesExpandableSection` below, since `getProperties()` in `screens/utils.ts` doesn't filter it out. This means the user will see it twice — here as read-only, and again as an editable property row. Consider adding `model_type` to the exclusion list in `getProperties()`.

mturley (changes requested):
Nice work! One thing to check: `model_type` is now displayed in a dedicated "Model type" field on the detail pages, but the `getProperties()` function in `screens/utils.ts` doesn't exclude `model_type` from the editable properties table. This means it will show up twice — once as the new read-only display field, and again as an editable row in the properties section. Should it be filtered out of the properties table (similar to how `_lastModified` and `_registeredFrom` are excluded)?

Also I actually think we may want a tech debt ticket to clean those two up, at least _registeredFrom isn't used anymore now that we have modelSource* properties, but that's out of scope.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
Taj010: Done! updated the `getProperties()` in `utils.ts` to exclude `model_type` and also ajsued the utils.spc.ts test file accordingly

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
```diff
@@ -117,6 +125,11 @@ const ModelVersionDetailsView: React.FC<ModelVersionDetailsViewProps> = ({
             <Sidebar hasBorder hasGutter isPanelRight>
               <SidebarContent>
                 <DescriptionList>
+                  <DashboardDescriptionListGroup title="Model type">
```
ppadti: I don't think we need to show this model type in version details card - so I think we can remove this

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
```diff
@@ -146,6 +155,11 @@ const ModelDetailsCard: React.FC<ModelDetailsCardProps> = ({
       >
         <ModelTimestamp timeSinceEpoch={rm.createTimeSinceEpoch} />
       </DashboardDescriptionListGroup>
+      <DashboardDescriptionListGroup title="Model type">
```
ppadti: I see this model type at the bottom - but design has it at the top, no?

Image

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelTypeField.tsx
```diff
@@ -0,0 +1,79 @@
...
+import FormFieldset from '~/app/pages/modelRegistry/screens/components/FormFieldset';
+import {
+  buildCustomPropertiesWithModelType,
+  getModelTypeStoredValueFromCustomProperties,
+} from './registerModelTypeUtils';
+
+const MODEL_TYPE_PLACEHOLDER_KEY = '__model-type-unset__';
+
+const MODEL_TYPE_SELECT_OPTIONS: SimpleSelectOption[] = [
+  {
```
ppadti: I see `Select model type` as one of the option - But I think we don't need that as an option

Image

ppadti (changes requested):
@Taj010 @mturley - On the Register new version form, are we not adding this field there? In this form either we will have the model name selected and disabled or drop down.

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersionDetails/ModelVersionDetailsView.tsx
Taj010: Removed it!

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
Taj010: Done! Moved model type to the top
image

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelTypeField.tsx
Taj010: Thanks for this! Adjusted it - removed the select model type option now 
image

ppadti:
Thanks @Taj010 for addressing all the review comments.
circling back to [this](https://github.com/kubeflow/model-registry/pull/2545#pullrequestreview-4106136248) in case it got missed - 
@mturley we might have add this model type field to the Register a new version form as well right?

mturley:
@ppadti I think since the model type goes into custom properties of the model and not the version, we can't include it in that form. Makes sense to me that all versions of a model are the same type

mturley:
(sorry, on mobile, I'll read all the context in a bit)

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
```diff
@@ -112,8 +117,17 @@ const ModelDetailsCard: React.FC<ModelDetailsCardProps> = ({
     />
   );
 
+  const modelTypeRaw =
+    getModelTypeRawStringFromCustomProperties(rm.customProperties) ??
+    getModelTypeRawStringFromCustomProperties(modelTypeFallbackCustomProperties);
+
   const infoSection = (
     <>
+      <DashboardDescriptionListGroup title="Model type">
```
ppadti, line 126: @Taj010 one minor changes I noticed in the mockups is that we need move model ID field to the bottom of this listGroup.

ppadti (changes requested):

clients/ui/frontend/src/app/pages/modelRegistry/screens/ModelVersions/ModelDetailsCard.tsx
Taj010: Done!

mturley:
/lgtm
I'll let others approve who have been more directly involved in reviewing here.

ppadti:
Apologies for another question here @mturley 
when we register from catalog - currently we are pre-filling the model-type from model custom properties, and letting the user to switch it to different option. Is that intended behavior? 
Screenshot 2026-04-30 at 3 57 09 PM

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
```diff
@@ -75,6 +79,11 @@ const RegisterModelDetailsFormSection = <D extends RegisterModelFormData>({
           </HelperText>
         </FormHelperText>
       </FormGroup>
+      <RegisterModelTypeField
+        modelCustomProperties={formData.modelCustomProperties}
+        onModelCustomPropertiesChange={(next) => setData('modelCustomProperties', next)}
+        isRequired={isModelTypeRequired}
```
ppadti, line 88: Looking at the mock ups both flow from model-registry and catalog - this model-type field is required. So just curios why we need this extra `isRequired` prop?

clients/ui/frontend/src/app/pages/modelRegistry/screens/utils.ts
```diff
@@ -90,7 +90,7 @@ export const getProperties = <T extends ModelRegistryCustomProperties>(
   return Object.keys(customProperties).reduce((acc, key) => {
     // _lastModified is a property that is required to update the timestamp on the backend and we have a workaround for it. It should be resolved by
     // backend team
-    if (key === '_lastModified' || /^_registeredFrom/.test(key)) {
+    if (key === '_lastModified' || key === 'model_type' || /^_registeredFrom/.test(key)) {
```
ppadti: can we use the const  `CatalogModelCustomPropertyKey.MODEL_TYPE` here instead of model_type?

ppadti (changes requested):

clients/ui/frontend/src/app/pages/modelRegistry/screens/utils.ts
Taj010: Done! updated to use `CatalogModelCustomPropertyKey.MODEL_TYPE`

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
Taj010: `RegisterModelDetailsFormSection` is shared: MR register passes `isModelTypeRequired` so the type field shows as required and matches` isRegisterModelSubmitDisabled(..., { requireModelType: true }).` Catalog register uses the same block but omits the prop and keeps `requireModelType: false` in `isRegisterCatalogModelSubmitDisabled,` so we only flag required in the MR path for now. If we want the catalog to match the mocks, we’d turn on the prop there and flip that submit check to true as well.

mturley:
@ppadti re: your earlier comment - when registering from catalog, you're right that we probably shouldn't let the user change it. we don't let them change the uri after all, this is more knowledge from the catalog that we're just passing through. cc @Taj010

Taj010:
Now from the model catalog flow - the model type is in a read-only state: 

image

clients/ui/frontend/src/app/pages/modelRegistry/screens/RegisterModel/RegisterModelDetailsFormSection.tsx
ppadti: yeah, but if you look at the mocks - model type is required field in the catalog flow as well just like MR. So we might need to pass it as required for the catalog register form as well right?

Taj010: Gotcha, Done! I’ve added `isModelTypeRequired ` and adjusted as needed.

ppadti (approved):
One small edge case is missing - when the model type is unknown, we are not handling that case for model catalog. Will create a bug for this to handle
Thanks @Taj010 !!
/lgtm 
/approve

PR #2571, opened 2026-04-10
fix(catalog): implement pagination on MCP server tools endpoint

pboyd:
## Description
Implement pagination for the tools endpoint and add the `toolLimit` parameter for the MCP server detail endpoint.

## How Has This Been Tested?
On a local dev cluster.

## Merge criteria:

### DIFF
--- a/api/openapi/src/catalog.yaml
+++ b/api/openapi/src/catalog.yaml
@@ -2015,6 +2015,7 @@ components:
         type: integer
         format: int32
         default: 10
+        minimum: 0
         maximum: 100
       in: query
       required: false
--- a/api/openapi/src/catalog.yaml
+++ b/api/openapi/src/catalog.yaml
@@ -297,6 +297,7 @@ paths:
         in: path
         required: true
       - $ref: "#/components/parameters/includeTools"
+      - $ref: "#/components/parameters/toolLimit"
   /api/mcp_catalog/v1alpha1/mcp_servers/{server_id}/tools:
     description: >-
       The REST endpoint/path used to list zero or more `MCPTool` entities for an `MCPServer`.
--- a/catalog/internal/catalog/mcpcatalog/db_mcp.go
+++ b/catalog/internal/catalog/mcpcatalog/db_mcp.go
@@ -179,7 +179,7 @@ func mergeFilterQueries(a, b string) string {
 	}
 }
 
-func (d *dbMCPCatalogImpl) GetMCPServer(ctx context.Context, serverID string, includeTools bool) (*openapi.MCPServer, error) {
+func (d *dbMCPCatalogImpl) GetMCPServer(ctx context.Context, serverID string, includeTools bool, toolLimit int32) (*openapi.MCPServer, error) {
 	id, err := apiutils.ValidateIDAsInt32(serverID, "server")
 	if err != nil {
 		return nil, fmt.Errorf("invalid server ID '%s': %w", serverID, api.ErrBadRequest)
@@ -201,6 +201,9 @@ func (d *dbMCPCatalogImpl) GetMCPServer(ctx context.Context, serverID string, in
 	if includeTools {
 		toolOptions := models.MCPServerToolListOptions{
 			ParentID: *dbServer.GetID(),
+		}
+		if toolLimit > 0 {
+			toolOptions.Pagination.PageSize = &toolLimit
 		}
 		tools, err := d.mcpServerToolRepo.List(toolOptions)
 		if err != nil {
--- a/catalog/internal/catalog/mcpcatalog/db_mcp.go
+++ b/catalog/internal/catalog/mcpcatalog/db_mcp.go
@@ -284,27 +284,21 @@ func (d *dbMCPCatalogImpl) GetMCPServerTool(ctx context.Context, serverID string
 		return nil, fmt.Errorf("server not found with ID %s: %w", serverID, api.ErrNotFound)
 	}
 
-	// List all tools for the server and find the one with matching name
+	// Filter by name at the DB level. The DB stores qualified names (serverName@version:toolName)
+	// but the API exposes only the unqualified name, so we match by suffix using LIKE.
 	toolOptions := models.MCPServerToolListOptions{
 		ParentID: id,
+		ToolName: &toolName,
 	}
 	tools, err := d.mcpServerToolRepo.List(toolOptions)
 	if err != nil {
 		return nil, fmt.Errorf("error loading tools for server %s: %w", serverID, err)
 	}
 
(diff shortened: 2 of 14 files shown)


catalog/internal/server/openapi/api_mcp_catalog_service.go
```diff
@@ -271,7 +271,24 @@ func (c *MCPCatalogServiceAPIController) GetMCPServer(w http.ResponseWriter, r *
 		var param bool = false
 		includeToolsParam = param
 	}
-	result, err := c.service.GetMCPServer(r.Context(), serverIdParam, includeToolsParam)
+	var toolLimitParam int32
+	if query.Has("toolLimit") {
+		param, err := parseNumericParameter[int32](
+			query.Get("toolLimit"),
+			WithParse[int32](parseInt32),
+			WithMaximum[int32](100),
```
chambridge: Missing WithMinimum validation. A client can send toolLimit=-5, which passes validation here and flows through as a negative LIMIT clause — behavior is DB-driver-dependent (error or undefined). Consider adding:
`WithMinimum[int32](0)`,

catalog/internal/server/openapi/api_mcp_catalog_service.go
chambridge: add `WithMinimum[int32](0)` here too to prevent negative toolLimit values from reaching the DB layer.

api/openapi/src/catalog.yaml
```diff
@@ -297,6 +297,7 @@ paths:
         in: path
         required: true
       - $ref: "#/components/parameters/includeTools"
+      - $ref: "#/components/parameters/toolLimit"
```
chambridge, line 300: on line 2018

The toolLimit parameter defines maximum: 100 but no minimum. The generated handler will accept negative values (e.g. toolLimit=-5), which would flow through as a negative LIMIT clause — behavior is DB-driver-dependent. Consider adding minimum: 0 to the schema here so the generated code rejects invalid values.

catalog/internal/catalog/mcpcatalog/db_mcp.go
```diff
@@ -292,7 +295,7 @@ func (d *dbMCPCatalogImpl) GetMCPServerTool(ctx context.Context, serverID string
 
 	// Find tool by name. DB stores tools with a qualified prefix (serverName@version:toolName)
 	// for uniqueness, but the API exposes only the unqualified tool name.
-	for _, tool := range tools {
+	for _, tool := range tools.Items {
```
chambridge: GetMCPServerTool loads all tools for the server into memory then does a linear scan by name. Now that List() returns a ListWrapper with pagination, if a default page size is ever applied, this could silently miss tools beyond the first page. More immediately, it's O(n) for something that could be a DB-level WHERE name = ? filter. Consider adding name-based filtering to MCPServerToolListOptions.

catalog/internal/catalog/mcpcatalog/db_mcp.go
```diff
@@ -265,7 +268,7 @@ func (d *dbMCPCatalogImpl) ListMCPServerTools(ctx context.Context, serverID stri
 		Items:         apiTools,
 		Size:          int32(len(apiTools)),
```
chambridge, line 269: Size is intended to be the total count, but `int32(len(apiTools))` now only reflects the number of items in the current page since pagination was added. This needs a separate count query — similar to the CountByParentIDs call used for toolCount in ListMCPServers / GetMCPServer. Something like:

```
toolCounts, err := d.mcpServerToolRepo.CountByParentIDs([]int32{id})
// ...
Size: toolCounts[id],
```

chambridge (changes requested):
Comments and a question

- Suggested API update to stop negative numbers in the limit.
- Performance call out that we could defer
- Question on whether our returned value is what we intend it to be

catalog/internal/catalog/mcpcatalog/db_mcp.go
pboyd: It's probably not the desired behavior, but it's consistent with the rest of the API:

```
$ curl -s 'http://localhost:8082/api/mcp_catalog/v1alpha1/mcp_servers?pageSize=5' | jq '. | {pageSize: .pageSize, size: .size}'
{
  "pageSize": 5,
  "size": 5
}
$ curl -s 'http://localhost:8082/api/model_catalog/v1alpha1/models?pageSize=5' | jq '. | {pageSize: .pageSize, size: .size}'
{
  "pageSize": 5,
  "size": 5
}
```

chambridge: sounds good. Might be worth a follow up issue so we actually provide a `totalSize` then.

api/openapi/src/catalog.yaml
pboyd: Fixed in the last commit.

catalog/internal/catalog/mcpcatalog/db_mcp.go
pboyd: Also fixed in the last commit.

catalog/internal/catalog/mcpcatalog/service/mcp_server_tool.go
```diff
@@ -63,10 +63,16 @@ func applyMCPServerToolListFilters(query *gorm.DB, listOptions *models.MCPServer
...
 	executionTable := utils.GetTableName(query.Statement.DB, &schema.Execution{})
 
-	return query.
+	query = query.
 		Joins(fmt.Sprintf("INNER JOIN %s ON %s.execution_id = %s.id",
 			associationTable, associationTable, executionTable)).
 		Where(fmt.Sprintf("%s.context_id = ?", associationTable), listOptions.ParentID)
+
+	if listOptions.ToolName != nil {
+		query = query.Where(fmt.Sprintf("%s.name LIKE ?", executionTable), fmt.Sprintf("%%:%s", *listOptions.ToolName))
```
chambridge: LIKE wildcard injection: toolName comes from user input (URL path param). If it contains % or _, the LIKE pattern will match unintended rows — e.g. toolName="list%" would match list_problems, list_entities, etc., and tools.Items[0] would silently return the wrong tool. Escape LIKE metacharacters before interpolation:

```suggestion
escaped := strings.NewReplacer("%", "\\%", "_", "\\_").Replace(*listOptions.ToolName)
query = query.Where(fmt.Sprintf("%s.name LIKE ?", executionTable), fmt.Sprintf("%%:%s", escaped))
```

chambridge (changes requested):
One small comment on the tool name query

chambridge (approved):

PR #2595, opened 2026-04-14
test(ui): add unit tests for useModelVersionById hook

YuliaKrimerman:
## Description

Adds unit tests for the `useModelVersionById` hook (`clients/ui/frontend/src/app/hooks/useModelVersionById.ts`), which previously had no test coverage.

The test file covers four scenarios:
- **API unavailable** — verifies the hook returns an error when `apiAvailable` is `false`
- **No modelVersionId** — verifies the hook silently fails (returns `null`, no error) when called without an ID
- **Successful fetch** — verifies the hook returns the fetched `ModelVersion` and `loaded: true`
- **Correct API arguments** — captures the callback passed to `useFetchState` and asserts `api.getModelVersion` is called with the correct `modelVersionId`

Follows the same testing patterns established by `useModelArtifactsByVersionId.spec.ts` and `useModelTransferJobs.spec.ts`.

## How Has This Been Tested?

- All 4 unit tests pass locally via `npx jest src/app/hooks/__tests__/useModelVersionById.spec.ts`
- ESLint passes with no errors on the new file

## Merge criteria:

### DIFF
--- a/clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
+++ b/clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
@@ -1,0 +1,139 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import * as React from 'react';
+import { waitFor } from '@testing-library/react';
+import { useFetchState } from 'mod-arch-core';
+import useModelVersionById from '~/app/hooks/useModelVersionById';
+import { useModelRegistryAPI } from '~/app/hooks/useModelRegistryAPI';
+import { ModelRegistryAPIs } from '~/app/types';
+import { mockModelVersion } from '~/__mocks__/mockModelVersion';
+import { testHook } from '~/__tests__/unit/testUtils/hooks';
+
+jest.mock('mod-arch-core', () => ({
+  useFetchState: jest.fn(),
+  NotReadyError: class NotReadyError extends Error {
+    constructor(message: string) {
+      super(message);
+      this.name = 'NotReadyError';
+    }
+  },
+}));
+
+global.fetch = jest.fn();
+
+jest.mock('~/app/hooks/useModelRegistryAPI', () => ({
+  useModelRegistryAPI: jest.fn(),
+}));
+
+const mockUseModelRegistryAPI = jest.mocked(useModelRegistryAPI);
+const mockUseFetchState = jest.mocked(useFetchState);
+
+const mockModelRegistryAPIs: ModelRegistryAPIs = {
+  createRegisteredModel: jest.fn(),
+  createModelVersionForRegisteredModel: jest.fn(),
+  createModelArtifactForModelVersion: jest.fn(),
+  getRegisteredModel: jest.fn(),
+  getModelVersion: jest.fn(),
+  listModelVersions: jest.fn(),
+  listRegisteredModels: jest.fn(),
+  getModelVersionsByRegisteredModel: jest.fn(),
+  getModelArtifactsByModelVersion: jest.fn(),
+  patchRegisteredModel: jest.fn(),
+  patchModelVersion: jest.fn(),
+  patchModelArtifact: jest.fn(),
+  listModelTransferJobs: jest.fn(),
+  getModelTransferJobByName: jest.fn(),
+  createModelTransferJob: jest.fn(),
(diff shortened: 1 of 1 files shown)


clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    const { result } = testHook(useModelVersionById)('version-1');
+
+    await waitFor(() => {
+      const [, , error] = result.current;
+      expect(error).toBeInstanceOf(Error);
+      expect(error?.message).toBe('API not yet available');
+    });
+  });
+
+  it('should silently fail when modelVersionId is not provided', async () => {
```
manaswinidas: Tautological: you mock `useFetchState` to return `[null, false, mockError, …]` then assert the error is exactly what you mocked. The hook's `!apiAvailable` branch is never invoked.

Capture the callback instead (same approach as test 4) and verify it rejects:
```ts
const cb = mockUseFetchState.mock.calls[0][0];
await expect(cb({})).rejects.toThrow('API not yet available');
```

clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    mockUseFetchState.mockReturnValue([null, false, undefined, jest.fn()]);
+
+    const { result } = testHook(useModelVersionById)();
+
+    await waitFor(() => {
+      const [data, , error] = result.current;
+      expect(data).toBeNull();
+      expect(error).toBeUndefined();
+    });
+  });
```
manaswinidas, line 85: Same issue — `useFetchState` is mocked to return `[null, false, undefined, …]`, then you assert `data` is null and `error` is undefined. This just validates the mock, not the hook.

To genuinely test the `!modelVersionId` branch, capture the callback and check:
```ts
const cb = mockUseFetchState.mock.calls[0][0];
await expect(cb({})).rejects.toThrow('No model version id');
```

clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    mockUseFetchState.mockReturnValue([mockedVersion, true, undefined, jest.fn()]);
+
+    const { result } = testHook(useModelVersionById)('version-1');
+
+    await waitFor(() => {
+      const [data, loaded] = result.current;
+      expect(loaded).toBe(true);
+      expect(data).toEqual(mockedVersion);
```
manaswinidas: Also tautological — asserts `data` and `loaded` equal exactly what `mockUseFetchState` was told to return. The `getModelVersion` mock is set up but never called.

This test can be removed entirely; test 4 already covers the happy path by actually invoking the captured callback and verifying `getModelVersion` receives the right args.

clients/ui/frontend/src/app/hooks/__tests__/useModelVersionById.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    testHook(useModelVersionById)('v-42');
+
+    // useFetchState receives the callback as its first argument
+    const capturedCallback = mockUseFetchState.mock.calls[0][0] as (
+      opts: unknown,
+    ) => Promise<unknown>;
+
+    await capturedCallback({});
+
+    expect(getModelVersionMock).toHaveBeenCalledWith({}, 'v-42');
```
manaswinidas: This is the only test that genuinely exercises hook logic — nice. Consider reusing the same callback-capture approach for the `!apiAvailable` and `!modelVersionId` branches so all three paths are properly covered.

manaswinidas (changes requested):
3 of the 4 tests are tautological — they mock `useFetchState`'s return value then assert exactly what was mocked, without ever invoking the hook's callback. Only test 4 (callback capture) genuinely exercises the hook logic, and it only covers the happy path.

Please rewrite tests 1–2 to capture the callback (same pattern as test 4) and verify it rejects with the expected errors. Test 3 can be dropped since test 4 already covers that path. This will give real coverage of all 3 branches.

manaswinidas (approved):
/lgtm
/approve

All review comments addressed

PR #2599, opened 2026-04-15
hide cat toogle in case only one cat is available

Philip-Carneiro:
## Description

When only one category is available for Model Catalog or MCP Catalog, the grid only show 4 cards and need to click in a link/toggle to show all the cards from that category.
These changes are made for when this happens, it shows all the cards available in that category (until the load more button appear) 

- Added `getActiveSourceLabels` utility function
- Implemented conditional rendering logic

With this changes the behaviour will be:
1. **Multiple active categories**: Normal behaviour with category toggle visible
2. **Single active category**: Toggle hidden, category auto-selected via useEffect
3. **No active categories**: Empty state displayed

Before:
- non-PF mode:
before-model-non-pf
before-mcp-non-pf

- PF mode:
before-mcp-pf
before-model-pf

After:
- non-PF mode:
after-mcp-non-pf
after-model-non-pf

- PF mode:
after-model-pf
after-mcp-pf

## How Has This Been Tested?

It was manually tested, also crated unit tests and 3 new cypress tests

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
@@ -10,11 +10,13 @@ import {
   mockCatalogPerformanceMetricsArtifact,
   mockCatalogSource,
   mockCatalogSourceList,
+  mockDefaultSources,
+  mockProviderAndCustomSources,
+  mockTwoProviderSources,
 } from '~/__mocks__';
-import type { CatalogSource } from '~/app/modelCatalogTypes';
 import { MODEL_CATALOG_API_VERSION } from '~/__tests__/cypress/cypress/support/commands/api';
 import { mockCatalogFilterOptionsList } from '~/__mocks__/mockCatalogFilterOptionsList';
-import { SourceLabel } from '~/app/modelCatalogTypes';
+import { SourceLabel, type CatalogSource } from '~/app/modelCatalogTypes';
 import { ModelRegistryMetadataType } from '~/app/types';
 
 type FilteredModelsInterceptConfig = {
@@ -90,7 +92,7 @@ const calculateExpectedCategoryCount = (sources: CatalogSource[]): number => {
 };
 
 const initIntercepts = ({
-  sources = [mockCatalogSource({}), mockCatalogSource({ id: 'source-2', name: 'source 2' })],
+  sources = mockDefaultSources(),
   modelsPerCategory = 4,
   hasValidatedModels = false,
   includeAllModelsIntercept = true,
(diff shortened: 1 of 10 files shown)


clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -397,7 +397,14 @@ describe('Performance Empty State', () => {
...
       initIntercepts({
-        sources: [mockCatalogSource({ labels: ['Provider one'] })],
+        sources: [
+          mockCatalogSource({ labels: ['Provider one'] }),
+          mockCatalogSource({
+            id: 'source-2',
+            name: 'Provider Two Source',
+            labels: ['Provider two'],
+          }),
+        ],
```
ppadti, line 407: This mocks is repeated many time - rather can we define it at the top and reuse it?

ppadti (changes requested):
Thanks @Philip-Carneiro !!
one question - @mturley @Philip-Carneiro 
when there's only 1 category the toggle is hidden, but now the user has no way to know which category they're viewing. Would it be better to show the name and description when we have 1 category or should we leave it as is, not sure.

manaswinidas:
+1 to Pushpa's suggestion - I think name and description should be there

clients/ui/frontend/src/app/pages/modelCatalog/utils/__tests__/modelCatalogUtils.spec.ts
```diff
@@ -1383,3 +1384,191 @@ describe('getModelName', () => {
     expect(result).toBe('my-model_v1');
   });
 });
+
+describe('getActiveSourceLabels', () => {
```
ppadti, line 1390: All the tests are here passing null to the catalogLabels parameter. Can we add one test which would pass catalogLabels to the getActiveSourceLabels?

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -1,19 +1,44 @@
...
+  const activeCategories = React.useMemo(
+    () => getActiveSourceLabels(catalogSources, catalogLabels),
+    [catalogSources, catalogLabels],
+  );
+
+  const isSingleCategory = activeCategories.length === 1;
+  const hasNoCategories = activeCategories.length === 0;
+
+  React.useEffect(() => {
+    if (catalogSourcesLoaded && isSingleCategory) {
```
ppadti: Nit: Consider adding `&& selectedSourceLabel !== activeCategories[0]` to the condition to skip the setState when it's already the correct value.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
```diff
@@ -270,9 +276,28 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
     );
   }
 
+  const categoryTitle = isSingleCategory
+    ? getLabelDisplayName(selectedSourceLabel || '', catalogLabels)
```
ppadti: Nit: On the very first render, selectedSourceLabel is still "All models" (not yet updated by the effect). Using activeCategories[0] from the parent directly would avoid depending on the effect timing, but the visual impact is negligible.

ppadti (changes requested):

ppadti (commented):
Sorry @Philip-Carneiro for the back and forth. The fix for the useEffect guard and the singleCategoryLabel prop looks good on the Model Catalog side — could we mirror the same changes in McpCatalog.tsx and McpCatalogGalleryView.tsx as well?

Philip-Carneiro:
> The fix for the useEffect guard and the singleCategoryLabel prop looks good on the Model Catalog side — could we mirror the same changes in McpCatalog.tsx and McpCatalogGalleryView.tsx as well?

I add the change requested

ppadti (approved):
/lgtm
/approve

PR #2603, opened 2026-04-16
Hide Empty Category if nothing is inside of the category

Philip-Carneiro:
## Description

Add changes to hide categories:
 - In Model Catalog
 - In MCP Catalog

Before the changes:
 - Non PF-mode
Screenshot 2026-04-16 at 13 29 25

 - PF-mode
 Screenshot 2026-04-16 at 13 29 00

After the changes:
 - Non PF-mode
Screenshot 2026-04-16 at 13 30 23

 - PF-mode
Screenshot 2026-04-16 at 13 30 54

## How Has This Been Tested?

It was manually tested.
To test it manually, need to edit the file `clients/ui/bff/internal/mocks/static_data_mock.go`

Adding this as part of the return object of method `func GetCatalogSourceMocks() []models.CatalogSource {`
```
		{
			Id:      "empty-models-source",
			Name:    "Empty Models Source",
			Enabled: &enabled,
			Labels:  []string{"empty models category"},
			Status:  &availableStatus,
		},
```

Adding this as part of the return object of method `func GetCatalogLabelListMock() models.CatalogLabelList {`
```
		{
			Name:        func() *string { s := "empty models category"; return &s }(),
			DisplayName: func() *string { s := "Empty Models Category"; return &s }(),
			Description: func() *string {
				s := "This category has no models — used for testing empty category hiding."
				return &s
			}(),
		},
```

Adding this as part of the return object of method `func GetMcpServerCatalogSourceMocks() []models.CatalogSource {`
```
		{
			Id:      "empty-mcp-source",
			Name:    "Empty MCP Source",
			Enabled: &enabled,
			Status:  &availableStatus,
			Labels:  []string{"empty mcp category"},
		},
```

Adding this as part of the return object of method `GetMcpServerCatalogLabelListMock() models.CatalogLabelList {`
```
		{
			Name:        func() *string { s := "empty mcp category"; return &s }(),
			DisplayName: func() *string { s := "Empty MCP Category"; return &s }(),
			Description: func() *string {
				s := "This category has no servers — used for testing empty category hiding."
				return &s
			}(),
		},
```

Run it `make build; make run;` from bff directory.

- Without this PR change, navigate to the Model Catalog and MCP Catalog and the empty category will be present at the category toggle bar
- With the PR changes, these categories will not be present

## Merge criteria:

- THIS PR MERGE IS BLOCKED TO MERGE UNTIL #2599 IS MERGED(THIS PR CONTAINS THE CHANGES FROM PR #2599)

If you have UI changes


clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
```diff
@@ -48,10 +48,24 @@ const CatalogCategorySection: React.FC<CategorySectionProps> = ({
...
-  // Get display name and description from labels API
   const categoryTitle = getLabelDisplayName(label, catalogLabels);
   const description = getLabelDescription(label, catalogLabels);
 
+  const reportTimerRef = React.useRef<ReturnType<typeof setTimeout>>();
+  React.useEffect(() => {
+    clearTimeout(reportTimerRef.current);
+    if (catalogModelsLoaded && !searchTerm) {
+      reportTimerRef.current = setTimeout(() => {
+        reportCategoryEmpty(label, catalogModels.items.length === 0);
```
YuliaKrimerman: The setTimeout(..., 100) debounce in reportCategoryEmpty is a code smell.

The 100ms timer is used to avoid re-render loops, but it introduces a race condition -- if the component unmounts and remounts within 100ms, the report could fire stale data. This pattern is duplicated identically in McpCatalogCategorySection.tsx. A cleaner approach would be to compute emptiness at the data layer (in the context itself or in getActiveSourceLabels) rather than having child components report upward with timers.

If the timer approach stays, at minimum extract the shared reporting logic into a reusable hook to eliminate the duplication

clients/ui/frontend/src/app/context/modelCatalog/ModelCatalogContext.tsx
```diff
@@ -150,6 +154,25 @@ export const ModelCatalogContextProvider: React.FC<ModelCatalogContextProviderPr
     React.useState(false);
   const [lastViewedModelName, setLastViewedModelName] = React.useState<string | null>(null);
   const [sortBy, setSortBy] = React.useState<ModelCatalogSortOption | null>(null);
+  const [emptyCategoryLabels, setEmptyCategoryLabels] = React.useState<Set<string>>(
+    () => new Set<string>(),
+  );
+  const emptyCategoryLabelsRef = React.useRef(emptyCategoryLabels);
+  emptyCategoryLabelsRef.current = emptyCategoryLabels;
+
+  const reportCategoryEmpty = React.useCallback((label: string, isEmpty: boolean) => {
```
YuliaKrimerman: The reportCategoryEmpty callback and emptyCategoryLabels state are duplicated verbatim between ModelCatalogContext and McpCatalogContext. This is the exact same code block (~15 lines) in both files.

We should extract this into a shared hook

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -1,19 +1,55 @@
...
+    emptyCategoryLabels,
+  } = React.useContext(ModelCatalogContext);
   const filtersApplied = useHasVisibleFiltersApplied();
+
+  const activeCategories = React.useMemo(
+    () => getActiveSourceLabels(catalogSources, catalogLabels),
+    [catalogSources, catalogLabels],
+  );
+
+  const effectiveActiveCategories = React.useMemo(
```
YuliaKrimerman: The effectiveActiveCategories computation, isSingleCategory/hasNoCategories flags, and the auto-select useEffect are duplicated almost line-for-line between ModelCatalog.tsx and McpCatalog.tsx. Can we extract this into a shared hook? Something like useEffectiveCategories(catalogSources, catalogLabels, emptyCategoryLabels, catalogSourcesLoaded, setSelectedLabel).

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalog.tsx
```diff
@@ -1,19 +1,55 @@
...
+    () => activeCategories.filter((c) => !emptyCategoryLabels.has(c)),
+    [activeCategories, emptyCategoryLabels],
+  );
+
+  const isSingleCategory = effectiveActiveCategories.length === 1;
+  const hasNoCategories = effectiveActiveCategories.length === 0;
+
+  React.useEffect(() => {
```
YuliaKrimerman: The useEffect that auto-selects a single category has effectiveActiveCategories in its dependency array. Since effectiveActiveCategories is a new array on every render (created by .filter()), this effect will fire on every render, not just when the actual categories change. This could cause unnecessary calls to updateSelectedSourceLabel.

You should stabilize the reference for example : 
const effectiveActiveCategoriesKey = effectiveActiveCategories.join(',');

React.useEffect(() => {
  if (catalogSourcesLoaded && isSingleCategory) {
    updateSelectedSourceLabel(effectiveActiveCategories[0]);
  }
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [catalogSourcesLoaded, isSingleCategory, effectiveActiveCategoriesKey, updateSelectedSourceLabel]);

YuliaKrimerman (commented):
Couple of small things here

clients/ui/frontend/src/app/pages/modelCatalog/screens/CatalogCategorySection.tsx
Philip-Carneiro: The duplication concern has been addressed — the shared useReportCategoryEmpty hook now replaces the identical logic that was copy-pasted in both CatalogCategorySection.tsx and McpCatalogCategorySection.tsx.

The race condition on unmount/remount has also been mitigated: the hook clears the timeout both on cleanup (return () => clearTimeout(...)) and before re-scheduling (clearTimeout at the top of the effect), so stale reports from unmounted components cannot fire.

Regarding the timer itself — the 100ms debounce is intentional and remains because of an architectural constraint in the current data flow:

Each CatalogCategorySection / McpCatalogCategorySection fetches its own data independently via per-category API calls (useCatalogModelsBySources / useMcpServersBySourceLabelWithAPI). The parent has no visibility into which categories are empty until each child completes its fetch and reports back. This is a bottom-up data flow by design — the catalog sources and labels only tell us which categories exist, not which ones have matching items after filters, search, and pagination are applied.

When multiple sections load concurrently, they each trigger reportCategoryEmpty → setEmptyCategoryLabels in the parent context. Without the debounce, synchronous state updates during the same render cycle cause cascading re-renders (parent re-renders children → children re-report → parent updates again).

Moving this to the data layer (as suggested) would require either a dedicated API endpoint returning per-category item counts (doesn't exist), or fetching all items at the parent level and computing counts locally (which defeats per-category pagination and increases load). Neither is feasible without an API-level change, so the debounce is the pragmatic solution for now.

clients/ui/frontend/src/app/hooks/useEffectiveCategories.ts
```diff
@@ -0,0 +1,44 @@
...
+  const effectiveCategoriesKey = effectiveActiveCategories.join(',');
+
+  React.useEffect(() => {
+    if (catalogSourcesLoaded && isSingleCategory) {
+      updateSelectedSourceLabel(effectiveActiveCategories[0]);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [catalogSourcesLoaded, isSingleCategory, effectiveCategoriesKey, updateSelectedSourceLabel]);
+
+  return { activeCategories, effectiveActiveCategories, isSingleCategory, hasNoCategories };
```
ppadti: This `activeCategories` is returned but never used

clients/ui/frontend/src/app/hooks/useReportCategoryEmpty.ts
```diff
@@ -0,0 +1,23 @@
...
+  isLoaded: boolean,
+  itemCount: number,
+  searchTerm: string,
+): void => {
+  const reportTimerRef = React.useRef<ReturnType<typeof setTimeout>>();
+
+  React.useEffect(() => {
+    clearTimeout(reportTimerRef.current);
+    if (isLoaded && !searchTerm) {
+      reportTimerRef.current = setTimeout(() => {
```
ppadti: What's the reason for the 100ms setTimeout in here?
wanted to understand why it wouldn't work with useEffect without the delay

clients/ui/frontend/src/app/hooks/useEffectiveCategories.ts
```diff
@@ -0,0 +1,44 @@
...
+  const isSingleCategory = effectiveActiveCategories.length === 1;
+  const hasNoCategories = effectiveActiveCategories.length === 0;
+
+  const effectiveCategoriesKey = effectiveActiveCategories.join(',');
+
+  React.useEffect(() => {
+    if (catalogSourcesLoaded && isSingleCategory) {
+      updateSelectedSourceLabel(effectiveActiveCategories[0]);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ppadti: The eslint-disable + join(',') workaround was added to avoid the effect re-firing on new array references. But I think we can just put effectiveActiveCategories in the deps directly — the if (isSingleCategory) guard already prevents unnecessary updateSelectedSourceLabel calls, so even if the effect re-runs due to a new reference, it's a no-op when there isn't exactly one category.

clients/ui/frontend/src/app/hooks/useEmptyCategoryTracking.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const useEmptyCategoryTracking = (): UseEmptyCategoryTrackingResult => {
+  const [emptyCategoryLabels, setEmptyCategoryLabels] = React.useState<Set<string>>(
+    () => new Set<string>(),
+  );
+  const emptyCategoryLabelsRef = React.useRef(emptyCategoryLabels);
+  emptyCategoryLabelsRef.current = emptyCategoryLabels;
+
+  const reportCategoryEmpty = React.useCallback((label: string, isEmpty: boolean) => {
+    const { current } = emptyCategoryLabelsRef;
```
ppadti: Just curious here - why we needed useRef here?
the functional form of setState (prev => ...) already gives access to the latest state, or am I missing a case where the ref is needed?

ppadti (changes requested):

clients/ui/frontend/src/app/hooks/useEffectiveCategories.ts
```diff
@@ -0,0 +1,45 @@
...
+  const effectiveActiveCategories = React.useMemo(
+    () => activeCategories.filter((c) => !emptyCategoryLabels.has(c)),
+    [activeCategories, emptyCategoryLabels],
+  );
+
+  const isSingleCategory = effectiveActiveCategories.length === 1;
+  const hasNoCategories = effectiveActiveCategories.length === 0;
+
+  React.useEffect(() => {
+    if (catalogSourcesLoaded && isSingleCategory) {
```
ppadti: Can we update this check to 
```suggestion
  if (
      catalogSourcesLoaded &&
      isSingleCategory &&
      selectedSourceLabel !== effectiveActiveCategories[0]
    )
```
to avoid un nessecary rendering?

clients/ui/frontend/src/app/hooks/useReportCategoryEmpty.ts
```diff
@@ -0,0 +1,21 @@
...
+  label: string,
+  isLoaded: boolean,
+  itemCount: number,
+  searchTerm: string,
+): void => {
+  React.useEffect(() => {
+    if (!isLoaded || searchTerm) {
+      return undefined;
+    }
+    const timer = setTimeout(() => {
```
ppadti: If this time out to fix the cypress failure - then i think we should not add this in the code and rather fix it in the cypress test. 
I think re-ordering the test would fix the failure?
Reorder assertions to check visible categories first and then we can assert for empty category

ppadti (changes requested):

clients/ui/frontend/src/app/hooks/useReportCategoryEmpty.ts
ppadti: In mcpCatalog.cy.ts file
if we reorder this 
   ```
 cy.findByTestId('mcp-category-title-community_mcp_servers').should('not.exist');
    cy.findByTestId('mcp-category-title-organization_mcp_servers').should('be.visible');
```
to
   ```
 cy.findByTestId('mcp-category-title-organization_mcp_servers').should('be.visible');
    cy.findByTestId('mcp-category-title-community_mcp_servers').should('not.exist');
```
wouldn't fix the test? same goes for model catalog

clients/ui/frontend/src/app/hooks/useEffectiveCategories.ts
```diff
@@ -0,0 +1,51 @@
...
+import { CatalogLabelList, CatalogSourceList } from '~/app/modelCatalogTypes';
+import { getActiveSourceLabels } from '~/app/pages/modelCatalog/utils/modelCatalogUtils';
+
+type UseEffectiveCategoriesResult = {
+  effectiveActiveCategories: string[];
+  isSingleCategory: boolean;
+  hasNoCategories: boolean;
+};
+
+const useEffectiveCategories = (
```
ppadti: This custom hook is now dead code, but I remember this was being used. Could you please double check -  if it not required now, please remove it.

ppadti (changes requested):
Sorry, this branch has got a conflict due to catalog refactor work. could you just rebase.

manaswinidas (approved):
/lgtm
/approve

manaswinidas:
One edge case we missed here is I still see "All models" and "Sample category" when I have just sample category and an empty category(a category with no models) in BFF mocks. But the happy path works fine. @ppadti @Philip-Carneiro this can be a follow up

PR #2604, opened 2026-04-16
chore: fix marking of async upload job e2e and integration tests

dbasunag:
- fix marking of async upload job e2e and integration tests
- add pytest-html, since running the tests against live cluster would require test report generation
- add support in e2e test to be able to run them against live cluster

## Merge criteria:

If you have UI changes


jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -79,8 +79,18 @@ def model_registry_client():
     host = f"{parsed.scheme}://{parsed.hostname}"
     port = parsed.port or (443 if parsed.scheme == "https" else 8080)
 
-    return ModelRegistry(host, port, author="integration-test", is_secure=False)
+    return ModelRegistry(host, port, author="integration-test", is_secure=False, user_token=user_token)
 
+@pytest.fixture(scope="session")
+def user_token() -> str:
+    return os.getenv("AUTH_TOKEN", None)  # type: ignore[arg-type]
```
jonburdo:
```suggestion
    return os.getenv("AUTH_TOKEN", "")
```
This should either return a `str` always or the return type should be `str | None`

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -79,8 +79,18 @@ def model_registry_client():
...
+
+@pytest.fixture(scope="session")
+def verify_ssl() -> bool:
```
jonburdo: This fixture is unused. If you want to avoid adding args to every function that needs this, you could just make this a function and call it where needed instead of inline:
```python
    verify_env = os.environ.get("VERIFY_SSL")
    verify = verify_env.lower() == "true" if verify_env is not None else None
```

Also it's return type is `bool` but it can return `None`

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -440,6 +452,11 @@ def _run_job_and_wait(env, tmp_path, k8s, configmap_data=None) -> JobResult:
     # Forward optional env vars from the host environment to the container
     for var in (
         "MODEL_SYNC_DESTINATION_OCI_BASE_IMAGE",
+        "MODEL_SYNC_REGISTRY_SERVER_ADDRESS",
+        "MODEL_SYNC_REGISTRY_PORT",
+        "MODEL_SYNC_REGISTRY_IS_SECURE",
+        "MODEL_SYNC_REGISTRY_AUTHOR",
```
jonburdo: We probably don't need `MODEL_SYNC_REGISTRY_AUTHOR` here unless I'm mistaken (the others make sense)

jonburdo (commented):
/lgtm
/approve
/hold

A couple comments that I think are worth address. If we're doing more work on these right away, they could be fixed in a follow-up, so feel free to merge as-is.

Feel free to unhold if you are ready to merge

jobs/async-upload/tests/integration/test_integration_async_upload.py
dbasunag: Since using it in only one place under jobs/async-upload, used it directly there without creating a utility. If this area adds more tests, then would definitely consider creating a utility function.

jobs/async-upload/tests/integration/test_integration_async_upload.py
```diff
@@ -234,7 +237,7 @@ def apply_job_with_strategic_merge(
 
 
 def wait_for_job_completion(
-    job_name: str, namespace: str, k8s_batch_client: kubernetes.client.BatchV1Api, timeout_seconds: int = 60
+    job_name: str, namespace: str, k8s_batch_client: kubernetes.client.BatchV1Api, timeout_seconds: int = 120
```
jonburdo, line 240: Good call! Should help with intermittent failures due to timeout

jonburdo (commented):
/lgtm
/approve

Looks great!

PR #2638, opened 2026-04-24
fix: enable positive_data_acceptance fuzz checks via spec tightening and test hooks

lugi0:
## Summary

Resolves [RHOAIENG-58824](https://redhat.atlassian.net/browse/RHOAIENG-58824): Schemathesis's `positive_data_acceptance` check generates schema-conformant requests and expects 2xx responses, but 53 tests fail with `RejectedPositiveData`. This PR closes the spec-server contract gaps and adds fuzz test hooks for server-side limitations that cannot be expressed in OpenAPI 3.0.

**Result: 53 → 0 stateless fuzz test failures.**

### Spec tightening (37 failures)

| Root cause | Failures | Fix |
|---|---|---|
| Path parameter IDs defined as `type: string` with no constraints; server requires numeric int32 | 19 | Added `format: int64`, `pattern: "^[1-9][0-9]{0,8}$"` to all ID path params |
| `pageSize` query param defined as `type: string` with no constraints; server parses as int32 | 8 | Added `format: int32`, `pattern: "^[1-9][0-9]{0,8}$"` (kept as `type: string` to avoid breaking API change) |
| Singleton GET endpoints mark all params optional; server requires `externalId` OR `name+parentResourceId` | 6 | Schemathesis hook enforces valid param combinations |
| `filterQuery` allows any string; server lexer only accepts ASCII grammar | 3 | Added `pattern: "^[\x20-\x7E]*$"` |
| `nextPageToken` has no format constraint; server expects base64 | 1 | Improved description (opaque cursor) |

### Spec + hook fixes (14 failures)

| Root cause | Failures | Fix |
|---|---|---|
| Required string fields accept empty string `""`; server's `IsZeroValue()` treats it as missing | 10 | Added `minLength: 1` to required string fields in Create schemas |
| Catalog server missing validation middleware for null bytes | 3 | Added `middleware.ValidationMiddleware()` to `catalog/cmd/catalog.go` |
| Artifact `oneOf` discriminator not enforced | 1 | Hook randomizes across valid artifact types for POST; ensures required fields per type |

### Additional issues found during verification

| Issue | Fix |
|---|---|
| `MetadataIntValue` spec says `format: int64` but server uses int32 (`StringToInt32`) | Changed to `format: int32`, `pattern: "^-?([1-9][0-9]{0,8}|0)$"` |
| MetadataValue types allow fuzz-generated extra properties | Added `additionalProperties: false` to all 6 MetadataValue types |
| `MetadataProtoValue` not supported by EmbedMD converter (no case in switch) | Hook replaces ProtoValue with StringValue |
| `MetadataStructValue` base64 round-trip bug — write decodes, read doesn't re-encode | Hook replaces StructValue with StringValue (see server bugs section) |
| Non-ASCII/surrogate characters in request bodies break Go JSON parser | Hook strips null bytes and surrogates via `encode("utf-8", errors="ignore")` |
| Non-ASCII characters in `customProperties` keys | Hook sanitizes keys to ASCII-only |
| Backslash/quotes in `name`/`externalId` params break server's internal filter query construction | Hook strips `\`, `'`, `"` from these params |
| Go strict JSON decoder rejects extra properties not in struct | Per-path property whitelist derived from spec, applied in `map_body` hook |
| `metadataType` field used `example` instead of `enum` | Changed to single-value `enum` for proper discriminator enforcement |
| Metric `step` is int64 in Go struct but `timestamp` is string | Hook sets `step` as integer, `timestamp` as string; clamps `step` to int64 range |
| Artifact `artifactType` is immutable but included in update schema | Hook GETs existing artifact to match type on PATCH |
| Schemathesis ignores `minLength`/`format` inherited through `allOf` | Hook replaces empty `name` with random value; validates numeric-string fields |
| Schemathesis generates empty string for path parameter IDs despite pattern constraints | Hook replaces empty/non-numeric path param IDs with `"1"` (server returns accepted 404) |
| Schemathesis explicit example replay bypasses `map_body` hook | `map_case` hook removes invalid `id` from request body as safety net |

## Server bug fix: 500 on artifact creation for missing experiment run

`UpsertExperimentRunArtifact` did not validate that the experiment run exists before creating the artifact. When the parent experiment was deleted (cascading the run), the FK violation on the attribution row surfaced as a raw 500.

Added pre-validation matching the existing pattern in `UpsertModelVersionArtifact`: call `GetExperimentRunById` first, returning 404 if not found.

**File:** `internal/core/experiment_run.go`

## `make test-fuzz` reliability fixes

| Issue | Fix |
|---|---|
| Port-forwards start in background with no readiness check; tests fail on connection refused | Added curl retry loops (30s timeout) for model-registry (`:8080`), minio (`:9000`), local registry (`:5001`) |
| `$STATUS` variable never set in `test-fuzz`; target always exits 0 | Combined shell commands to properly capture and propagate exit codes |
| Prior fuzz runs leave corrupted `customProperties` in DB; PATCH tests fail loading existing resources | Added `scripts/cleanup.sh` before both stateless and stateful test runs |
| No stateful fuzz tests for the catalog API | Added `test_catalog_stateful.py` and included in `make test-fuzz` |
| Stateful test fails with `Unsatisfiable` after thousands of successful steps | Catch `hypothesis.errors.Unsatisfiable` and emit warning instead of failing (pytest-rerunfailures not available in CI) |

## Design decisions

**Per-path property whitelist vs `additionalProperties: false`:** The Go server uses strict JSON decoding (`DisallowUnknownFields`), rejecting any property not in the struct. OpenAPI 3.0's `allOf` + `additionalProperties: false` is fundamentally broken — it evaluates per-subschema, so properties valid in the composite are rejected as "additional" in the base. Instead of fighting the spec, the `map_body` hook maintains a per-path whitelist (`_PATH_PROPERTIES`) derived from the spec.

**Hooks vs spec changes:** Some server behaviors cannot be expressed in OpenAPI 3.0 (parameter dependencies, discriminator enforcement, strict decoding). These are handled via Schemathesis hooks in `conftest.py` rather than incorrect spec annotations.

**`pageSize` kept as `type: string`:** Per reviewer feedback, changing to `type: integer` would be a breaking API change. Kept as string with `format: int32` and pattern constraint instead.

## ⚠️ ID format: int64 in spec vs int32 in server — needs discussion

Path parameter IDs use `format: int64` and `pattern: "^[1-9][0-9]{0,8}$"`. The pattern safely constrains to the int32 range, but the server validates as int32 (`ValidateIDAsInt32`). The response body `BaseResource.id` was already `format: int64` before this PR. See the [earlier discussion](https://github.com/kubeflow/hub/pull/2638#discussion_r3137612480) for options.

## Server bugs identified (not fixed, worked around)

1. **`MetadataProtoValue` not supported** — EmbedMD converter has no case for ProtoValue
2. **`MetadataStructValue` base64 round-trip bug** — write converter base64-decodes, read converter doesn't re-encode; PATCHing resources with stored StructValues crashes
3. **`name`/`externalId` filter injection** — server constructs filter queries from user input without escaping
4. **Non-ASCII in `customProperties` keys** — Go JSON decoder fails on surrogate pairs
5. **Metric `step` vs `timestamp` type inconsistency** — both declared as `type: string, format: int64`, but Go struct has `step` as `int64` (number) and `timestamp` as `string`
6. **Artifact `artifactType` immutable but in update schema** — server rejects type changes on PATCH; update schema should exclude it or mark `readOnly`

## Test plan

- [x] `make openapi/validate` — both specs pass
- [x] Stateless fuzz tests: 86 passed, 0 SUBFAILED, 63 subtests passed
- [x] Stateful fuzz tests: pass (with warning on non-deterministic `Unsatisfiable`)
- [x] E2E tests: 78 passed across all Python versions (3.10–3.14) and DB backends (MySQL, Postgres)
- [x] `Check autogenerated code is in sync` — pass
- [x] All CI checks green


api/openapi/src/lib/common.yaml
```diff
@@ -359,21 +382,27 @@ components:
...
     pageSize:
       examples:
         pageSize:
-          value: "100"
+          value: 100
       name: pageSize
       description: Number of entities in each page.
       schema:
-        type: string
+        type: integer
```
pboyd: This change might be a problem. Can it remain a `string` but have an `int32` format?

api/openapi/src/lib/common.yaml
```diff
@@ -292,6 +309,8 @@ components:
       description: The ID of resource.
       schema:
         type: string
+        format: int64
+        pattern: "^[1-9][0-9]{0,8}$"
```
pboyd, line 312: Positive `int64`'s could be up to 19 digits, there are a few instances of this.

api/openapi/src/lib/common.yaml
```diff
@@ -155,25 +160,30 @@ components:
...
+      additionalProperties: false
       required:
         - metadataType
         - int_value
       properties:
         int_value:
-          format: int64
+          format: int32
           type: string
+          pattern: "^-?[0-9]{1,9}$"
```
pboyd: Do you want to allow leading zero's here? I see the unsigned `int64` pattern prevents them.

pboyd (commented):
I added a few comments, but it's looking good.

api/openapi/src/lib/common.yaml
lugi0: For the `int64` issues, I would ask you to read through the `ID format: int64 in spec vs int32 in server` paragraph in the description - I think we should have a discussion within the team on how to handle this moving forward.

api/openapi/src/lib/common.yaml
lugi0: I believe this is due to how schemathesis would generate cases, leading to preventable failures in fuzzing tests. I can look into forcing acceptable values via schemathesis hooks, but I'm not sure why we would want to define a field as string when what we want is an integer

lugi0: Done

api/openapi/src/lib/common.yaml
lugi0: Fixed

adysenrothman:
/lgtm
Thanks for taking the time to resolve these issues! Out of curiosity, did you enable any of the other Schemathesis checks to see if these changes may have fixed some of the other test failures?

lugi0:
@adysenrothman not yet! this turned out to be much more involved than I initially anticipated, but I think with all the additional constraints via schemathesis hooks we should be in a better state overall. I will test against the other checks soon.

adysenrothman:
> @adysenrothman not yet! this turned out to be much more involved than I initially anticipated, but I think with all the additional constraints via schemathesis hooks we should be in a better state overall. I will test against the other checks soon.

I agree, this seems likely to resolve some other issues!

/approve

PR #2663, opened 2026-04-30
feat(catalog): add plugin infrastructure for extensible catalog types

Al-Pragliola:
## Description

Introduces `catalog/pkg/plugin/` — the core building blocks that allow different catalog types (models, MCP servers, datasets, etc.) to register themselves and be managed by a unified server.

This package provides:

- **`CatalogPlugin` interface** — standard contract for all catalog plugins: identity (`Name`, `Version`, `Description`), lifecycle (`Init`, `Start`, `Stop`, `Healthy`), HTTP routing (`RegisterRoutes`), and database migrations (`Migrations`).
- **Optional interfaces** — `BasePathProvider` (custom API base path), `SourceKeyProvider` (custom config key), `FlagProvider` (CLI flag registration).
- **Global registry** — thread-safe singleton with `init()`-based self-registration, deterministic iteration order, and duplicate detection (panics on conflict).
- **Config loading** — thin delegation to `basecatalog.ReadSourceConfig()`, supporting single and multi-file loading.

## Why `catalog/pkg/plugin/` instead of `pkg/catalog/plugin/`

The original plan placed the package at the repo root (`pkg/catalog/plugin/`). During implementation we hit Go's `internal` package visibility rule: code under `pkg/` at the repo root **cannot import** `catalog/internal/...` because `internal` packages are only visible to code rooted in their parent tree.

This matters because the plugin config layer needs `basecatalog.SourceConfig` and `basecatalog.ReadSourceConfig()` from `catalog/internal/catalog/basecatalog/`. Rather than re-implementing ~200 lines of config parsing, YAML schema handling, and deprecated-field merging, we moved the package into the `catalog/` module tree where the import is valid. This keeps a single source of truth for config parsing and ensures plugin tests exercise the exact same code paths as the production catalog server.

## What's NOT included (by design)

- **Server orchestration** (`server.go`) — plugin discovery, lifecycle management, and config routing are a follow-up ticket. This PR establishes the contract; the server wires it together.
- **Config decoupling** — `Config.SourceConfig` currently carries `*basecatalog.SourceConfig` directly. Full decoupling (plugin-defined config types with raw YAML routing) is a server orchestration concern, documented in the code.

## How Has This Been Tested?

```bash
go test ./pkg/plugin/ -v       # 13/13 pass
go vet ./pkg/plugin/           # clean
```

## Merge criteria:


catalog/pkg/plugin/registry.go
```diff
@@ -0,0 +1,92 @@
...
+	defer globalRegistry.mu.RUnlock()
+
+	result := make([]CatalogPlugin, 0, len(globalRegistry.order))
+	for _, name := range globalRegistry.order {
+		result = append(result, globalRegistry.plugins[name])
+	}
+	return result
+}
+
+// Get returns a plugin by name, or nil if not found.
```
pboyd:
```suggestion
// Get returns a plugin by name, or nil if not found. The returned boolean will be true if the plugin is found, and false otherwise.
```

catalog/pkg/plugin/config.go
```diff
@@ -0,0 +1,25 @@
...
+
+// LoadConfigs loads multiple sources.yaml files and returns them as
+// independent configs. Callers are responsible for any merge logic
+// (e.g., basecatalog's SourceCollection handles field-level merging).
+func LoadConfigs(paths []string) ([]*basecatalog.SourceConfig, error) {
+	configs := make([]*basecatalog.SourceConfig, 0, len(paths))
+	for _, path := range paths {
+		cfg, err := LoadConfig(path)
+		if err != nil {
+			return nil, err
```
pboyd: Unless `LoadConfig` includes it, it would be helpful to include the file that failed here.

pboyd (commented):
I had a few comments, but it looks good.

`pkg` normally implies an external interface, would we want to support external plugins? If not, maybe `catalog/internal/plugin` would be more appropriate.

pboyd (approved):
Thanks for the changes.

/lgtm
/approve

PR #2666, opened 2026-05-01
Model validation err handling

adysenrothman:
This PR fixes an existing issue with performance metrics loading for model catalog. Today, if there are two properties with the same value for a model's metrics, the loader will silently fail and not load any of the values into the db. These changes will grab the last of the duplicated values and store it while logging an error in the pod. This prevents a full failure of metrics loading and gives the user insight into issues with performance data. 

## How Has This Been Tested?
This was tested by adding 3 "mmlu" values for one of the demo AI validated models (certified/production-llm-8b) and ensuring that the metrics for that model were still loaded and logged the issue.
Screenshot 2026-05-01 at 9 58 13 AM

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/internal/catalog/modelcatalog/performance_metrics.go
+++ b/catalog/internal/catalog/modelcatalog/performance_metrics.go
@@ -505,17 +505,26 @@ func createAccuracyMetricsArtifact(evalRecords []evaluationRecord, modelID int32
 	// Properties can be empty or contain general metadata
 	properties := []models.Properties{}
 
-	// Create custom properties - simple mapping of benchmark_name to score_value
-	customProperties := []models.Properties{}
-
+	// Create custom properties - simple mapping of benchmark_name to score_value.
+	// Deduplicate by benchmark name: if multiple evaluation records share the same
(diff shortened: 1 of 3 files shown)


pboyd (commented):
/lgtm

catalog/internal/catalog/modelcatalog/performance_metrics.go
```diff
@@ -107,6 +107,11 @@ func (pr *performanceRecord) UnmarshalJSON(data []byte) error {
 	if id, ok := raw["id"].(string); ok {
 		pr.ID = id
 	}
+	if pr.ID == "" {
```
Al-Pragliola, line 110: overall looks fine, is there a reason for falling back to "config_id" if the id is missing?

adysenrothman: The input data uses config_id (not id) as the unique identifier for performance records, so the fallback ensures those records get a usable ID. Without it, perfRecord.ID would be empty, breaking deduplication and artifact naming downstream.

I came across the empty IDs in testing and it was causing issues.

PR #2678, opened 2026-05-05
fix(ci): quote output in repeat_cmd_until to prevent flaky E2E deploys

manaswinidas:
## Description

Fixes a flaky E2E test failure in the "Deploy Model Registry using manifests" step.

During a deployment rollout, when multiple pods temporarily co-exist with different images, the `repeat_cmd_until` function in `scripts/utils.sh` receives multi-value output from the jsonpath query (e.g., two image names separated by a newline). The unquoted `$output` in the `[ ]` test expands to multiple words, causing a `[: too many arguments` error. This makes the condition **never succeed**, so the deploy step spins for the full 500-second timeout before failing.

**Fix**: Switch from `[ $output $condition ]` to `[[ "$output" $condition ]]`, which properly handles multi-word values.

**Example of the failure** (from [this CI run](https://github.com/kubeflow/model-registry/actions/runs/24676129044/job/72160473099?pr=2622), attempt 1):
```
+ output='ghcr.io/kubeflow/model-registry/server:latest
ghcr.io/kubeflow/model-registry:main-'
+ '[' ghcr.io/kubeflow/model-registry/server:latest ghcr.io/kubeflow/model-registry:main- = ghcr.io/kubeflow/model-registry:main- ']'
././scripts/utils.sh: line 29: [: too many arguments
```

## How Has This Been Tested?

Verified the shell behavior locally:
```bash
# Before (fails with multi-word output):
output="val1 val2"; [ $output = val2 ]   # => bash: [: too many arguments

# After (works correctly):
output="val1 val2"; [[ "$output" = val2 ]]  # => no error, evaluates to false as expected
```

## Test Impact

No new tests needed — this fixes CI infrastructure flakiness, not application code.

Assisted-by: Claude (Anthropic)

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -26,7 +26,7 @@ repeat_cmd_until() {
 
     output=$(eval $cmd)
 
-    if [ $output $condition ]; then
+    if eval '[ "$output"' "$condition" ']'; then
       break
     else
       sleep $interval_secs


manaswinidas:
@jonburdo does this look reasonable ?

jonburdo (commented):
/lgtm
/approve


Thanks for working on this!

The issue is that if there is a terminating pod we get two images in `$output` instead of one. In such a case the `if-then` will still fail, so the test will remain flaky, but this does give us better output.

Issue #1045, opened 2025-04-29
Namespace is hardcoded in manifest files


didier-viboo:
**Describe the bug**

In order to install into a different namespace than the default, the documentation recommeds a complicated sequence of `kustomize edit` commands that would be fully unnecessary if the namespace was left out of the manifests and specified explicitly on the command line.

**To Reproduce**
1. I want to install the model registry in a non standard namespace
2. The [documentation](https://www.kubeflow.org/docs/components/model-registry/installation/) tells me that I need to run `kustomize edit set` in a bunch of files
```
for DIR in options/istio overlays/db ; do (cd $DIR; kustomize edit set namespace $PROFILE_NAME); done
```
and then run
```
kubectl apply -k overlays/db
kubectl apply -k options/istio
kubectl apply -k options/ui/overlays/istio
```
**Expected behavior**
If the namespace was simply left out of the manifests I could just run
```
kubectl apply -k overlays/db -n mynamespace
kubectl apply -k options/istio -n mynamespace
kubectl apply -k options/ui/overlays/istio -n mynamespace
```

without needing to change any configuration file.

tarilabs:
hi @didier-viboo thank you for raising this to our attention!

what are your thoughts on this one @pboyd ?

> [!CAUTION]
> we should as well consider what the Manifest repo expects during the synchronization of the manifests from _this repo_ to the kubeflow/manifests repo

from my pov, what I like about the `kustomize edit set namespace` -based approach, is that I can need to run the command only once, but potentially by leaving the `namespace` empty also in the base manifest, we can onboard both workflows per OP explanation.

pboyd:
Long story short, I think we need to keep the `namespace` in the manifests.

It predates me, but as I understand it the intention was for a kubeflow installation to have a single model registry shared between projects, in which case most users would deploy to the `kubeflow` namespace and the default namespace made a lot of sense. But now that the UI requires model registry to be in a project namespace, deploying to the `kubeflow` namespace by default is almost always incorrect (thus the admittedly "complicated sequence of `kustomize edit` commands"). With the current state, the default namespace is just in the way, and removing it would be helpful.

> we should as well consider what the Manifest repo expects during the synchronization of the manifests from this repo to the kubeflow/manifests repo

[...]

PR #2709, opened 2026-05-14
fix(manifests): add namespace: kubeflow to kustomization overlays and add namespace regression check

### LINKED ISSUE #3457: Hub / Model Registry is being deployed to the namespace `default`
### Validation Checklist

- [x] I confirm that this is a Kubeflow-related issue.
- [x] I am reporting this in the appropriate repository.
- [x] I have followed the [Kubeflow installation guidelines](https://github.com/kubeflow/manifests/blob/master/README.md).
- [x] The issue report is detailed and includes version numbers where applicable.
- [x] I have considered adding my company to the adopters page to support Kubeflow and help the community, since I expect help from the community for my issue (see [1.](https://github.com/kubeflow/community/issues/833) and [2.](https://github.com/kubeflow/community/blob/master/ADOPTERS.md#adopters-of-kubeflow-platform)).
- [x] This issue pertains to Kubeflow development.
- [ ] I am available to work on this issue.
- [x] You can join the CNCF Slack and access our meetings at the [Kubeflow Community](https://www.kubeflow.org/docs/about/community/) website. Our channel on the CNCF Slack is here [**#kubeflow-platform**](https://app.slack.com/client/T08PSQ7BQ/C073W572LA2).

### Version

master

### Detailed Description

This issue was introduced in the recent PR https://github.com/kubeflow/manifests/pull/3318
```
$ kubectl get pod -n default
NAME                                        READY   STATUS    RESTARTS      AGE
model-catalog-postgres-0                    1/1     Running   0             22h
model-catalog-server-895b9c7bb-w9hdv        2/2     Running   0             22h
model-registry-db-bf6ff894f-m9vns           1/1     Running   0             22h
model-registry-deployment-fbfc6587f-rkvjv   2/2     Running   1 (22h ago)   22h
```

- As part of the fix it would also be good to create a check from that ever happening again. We could check for the namespaces that we know that we control and error if there are any leftover pods outside of that:
https://github.com/kubeflow/manifests/blob/336ac873e50adcebae8cc35c9a5884c81fc11bff/tests/PSS_enable.sh#L11

---

First reported in https://cloud-native.slack.com/archives/C073W572LA2/p1778600481592009

### Steps to Reproduce

1. Deploy latest master to kind
2. Check running pods in namespace `default` via ` kubectl get pod -n default`

### Screenshots or Videos (Optional)

_No response_

Raakshass:
## Summary

Three of the four kustomization overlays consumed by `kubeflow/manifests` `example/kustomization.yaml` were missing the `namespace: kubeflow` directive. This causes all model-registry, istio networking, and catalog resources to deploy into the `default` namespace instead of `kubeflow` when using `kustomize build`.

This pull request adds `namespace: kubeflow` to the three affected overlays and introduces a namespace regression check script to prevent recurrence.

Fixes https://github.com/kubeflow/manifests/issues/3457  
Supersedes https://github.com/kubeflow/manifests/pull/3458 (closed per maintainer direction to fix upstream)

## Root Cause

The CI install script (`model_registry_install.sh`) passes `-n kubeflow` to `kubectl apply`, which overrides the namespace at apply-time. This masked the fact that the kustomization overlays themselves had no `namespace` directive. When users follow the documented deployment path (`kustomize build example/ | kubectl apply -f -`), there is no namespace override — all resources land in the `default` namespace.

This was confirmed by `istioctl analyze` output showing VirtualService and DestinationRule reference errors in the `default` namespace.

## Changes

| File | Change |
|---|---|
| `manifests/kustomize/overlays/postgres/kustomization.yaml` | Add `namespace: kubeflow` |
| `manifests/kustomize/options/istio/kustomization.yaml` | Add `namespace: kubeflow` |
| `manifests/kustomize/options/catalog/overlays/demo/kustomization.yaml` | Add `namespace: kubeflow` |
| `manifests/kustomize/check-namespace.sh` | New — namespace regression check script |

No change needed for `options/ui/overlays/istio/kustomization.yaml` — it already has `namespace: kubeflow`.

## Design Decisions

- **Placement:** The `namespace: kubeflow` directive is placed at the bottom of each file, consistent with the existing pattern in `options/ui/overlays/istio/kustomization.yaml`.
- **Regression check:** `check-namespace.sh` iterates over all four overlays, verifies each `kustomization.yaml` contains `namespace: kubeflow`, and checks that `kustomize build` output does not produce resources with `namespace: default`. This was explicitly requested by @jonburdo and @juliusvonkohout in https://github.com/kubeflow/manifests/pull/3458.
- **Upstream-first:** This fix is raised here in `kubeflow/hub` (the source of truth) rather than in `kubeflow/manifests`, where the `upstream/` folder is overwritten on every sync. Precedent: https://github.com/kubeflow/hub/pull/2706.

## Dependencies and Related Issues

- https://github.com/kubeflow/manifests/issues/3457 — original issue report
- https://github.com/kubeflow/manifests/pull/3458 — closed downstream pull request (maintainers directed upstream fix)
- https://github.com/kubeflow/manifests/pull/3450 — renamed `applications/model-registry` to `applications/hub` and synced v0.3.9
- https://github.com/kubeflow/hub/pull/2706 — precedent for upstream-first fix pattern

## Contributor Checklist

- [x] DCO sign-off included
- [x] AI assistance disclosed (`Assisted-by: Antigravity AI` in commit message)
- [x] Matches existing codebase pattern (`options/ui/overlays/istio/kustomization.yaml`)
- [x] Minimal scope — only the 3 files that need the fix plus 1 regression check
- [x] Root cause identified and stated in one sentence
- [x] Namespace regression check added per maintainer request

cc @tarilabs @pboyd

### DIFF
--- a/manifests/kustomize/check-namespace.sh
+++ b/manifests/kustomize/check-namespace.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# Verify that kustomize build output for all overlays sets namespace: kubeflow
+# and does not produce resources in the default namespace.
+#
+# Usage: ./check-namespace.sh
+#
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+KUSTOMIZE_DIR="${SCRIPT_DIR}"
+
+OVERLAYS=(
+  "overlays/postgres"
+  "options/istio"
+  "options/ui/overlays/istio"
+  "options/catalog/overlays/demo"
+)
+
+EXIT_CODE=0
+
+for overlay in "${OVERLAYS[@]}"; do
+  overlay_path="${KUSTOMIZE_DIR}/${overlay}"
+
+  if [ ! -d "${overlay_path}" ]; then
+    echo "SKIP: ${overlay} (directory not found)"
+    continue
+  fi
+
+  echo "Checking ${overlay}..."
+
+  output=$(kustomize build "${overlay_path}" 2>&1) || {
+    echo "FAIL: kustomize build failed for ${overlay}"
+    EXIT_CODE=1
+    continue
+  }
+
+  # Check for resources landing in the default namespace
+  default_ns_resources=$(echo "${output}" | grep -c '^\s*namespace: default$' || true)
+  if [ "${default_ns_resources}" -gt 0 ]; then
+    echo "FAIL: ${overlay} produces ${default_ns_resources} resource(s) with namespace: default"
+    EXIT_CODE=1
+  fi
+
+  # Verify namespace: kubeflow is present in the kustomization
+  if ! grep -q 'namespace: kubeflow' "${overlay_path}/kustomization.yaml"; then
+    echo "FAIL: ${overlay}/kustomization.yaml is missing 'namespace: kubeflow'"
+    EXIT_CODE=1
+  else
+    echo "PASS: ${overlay}"
+  fi
+done
+
+if [ "${EXIT_CODE}" -eq 0 ]; then
+  echo ""
+  echo "All overlays correctly set namespace: kubeflow"
+else
+  echo ""
+  echo "ERROR: namespace check failed — see above"
+fi
+
+exit "${EXIT_CODE}"
--- a/manifests/kustomize/options/catalog/overlays/demo/kustomization.yaml
+++ b/manifests/kustomize/options/catalog/overlays/demo/kustomization.yaml
@@ -70,3 +70,5 @@ patches:
             capabilities:
               drop:
                 - ALL
+
+namespace: kubeflow
--- a/manifests/kustomize/options/istio/kustomization.yaml
+++ b/manifests/kustomize/options/istio/kustomization.yaml
@@ -5,3 +5,5 @@ resources:
 - istio-authorization-policy.yaml
 - destination-rule.yaml
 - virtual-service.yaml
+
+namespace: kubeflow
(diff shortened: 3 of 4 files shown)


pboyd:
Hi @Raakshass, thanks for looking at this. Unfortunately, I don't think we can simply add the namespace because model-registry needs to actually be in the user namespace for the dashboard to work (`kubeflow-user-example-com` by default in the manifests).

The manifests actually did deploy to `kubeflow` at one time, but we removed it to address #1045. The [installation docs](https://www.kubeflow.org/docs/components/hub/installation/) provide instructions to set the namespace before applying them, which is how we've been able to get away without setting the manifest.

I'm not sure a good way to solve this, but I'll spend some time today and try to find something. Let me know if you have any ideas.

Raakshass:
youre right, i missed the user-namespace dependency and the #1045 history. hub should stay namespace-agnostic since model-registry needs to live in the profile namespace for dashboard integration
the overlay approach you proposed on the issue is the right fix set the namespace at the manifests layer instead of here. i can pick that up and raise the manifests-side overlay PR based on your diff. would be great to have your input on it given you know the deployment topology best
should i close this PR?

Raakshass:
closing this per pboyd's feedback. the hub repo should stay namespace-agnostic since model-registry needs to live in the profile namespace for dashboard integration. the fix belongs at the manifests layer via an overlay as discussed in kubeflow/manifests#3457.

pboyd:
Sorry for the delay, @Raakshass. I know we need to close this, but thank you for tackling this problem. It's been an issue for a while now.

PR #2724, opened 2026-05-19
feat(catalog): add unified plugin server 

### LINKED ISSUE #2529: [Plugin Arch] Phase 3 — Unified Catalog Server
## Summary

Create the unified catalog server entry point that hosts all registered catalog plugins, replacing the previous model-only approach with a server that discovers, initializes, and manages the lifecycle of all registered plugins.

## Motivation

Today the catalog server is tightly coupled to the model catalog. With the plugin system in place, the server should be a thin orchestrator: discover registered plugins, initialize them, wire up their HTTP handlers, and manage their lifecycle. This makes the server generic and future-proof.

## Scope

- Implement the **server entry point** that discovers all registered plugins from the registry
- Wire up plugin initialization, HTTP handler registration, and graceful shutdown
- Implement **health aggregation** across all plugins
- Ensure the server starts with zero plugins gracefully (no crash on empty registry)
- Maintain backward compatibility with existing deployment manifests

## Acceptance Criteria

- [ ] Server discovers and initializes all registered plugins at startup
- [ ] Each plugin's HTTP handlers are mounted under the correct path prefix
- [ ] Graceful shutdown stops all plugins cleanly
- [ ] Health endpoint aggregates status from all plugins
- [ ] Server starts and runs correctly with zero plugins registered
- [ ] Existing deployment manifests (Kustomize) continue to work
- [ ] `make build && make test && make lint` pass

## Dependencies

- Phase 2 #2528 

## References

- Parent: #2220

Al-Pragliola:
closes #2529

## Summary

This PR introduces a plugin server that discovers, initializes, and manages the lifecycle of all registered catalog plugins, replacing the directly-wired approach in `catalog/cmd/catalog.go`. The server is a thin orchestrator: it discovers registered plugins via the existing global registry, initializes them, mounts their HTTP handlers, and manages their lifecycle including graceful shutdown and leader election notifications.

## Motivation

The catalog server was tightly coupled to the model and MCP catalog implementations. All service creation, loader setup, route registration, and lifecycle management was hardcoded in the entry point. With the plugin system already defined (`CatalogPlugin` interface, global registry), the server should leverage it — making the entry point generic and ready for future catalog types without code changes.

## Architecture

```
catalog/cmd/catalog.go (entry point)
  |
  |-- DB init, datastore, leader election (infrastructure)
  |
  +-- plugin.Server (new)
        |-- Init()         discovers plugins from registry, loads config, calls plugin.Init()
        |-- MountRoutes()  creates chi router, calls plugin.RegisterRoutes(), adds /healthz /readyz
        |-- Start()        calls plugin.Start() in registration order
        |-- Stop()         calls plugin.Stop() in reverse order
        |-- NotifyLeader() dispatches to LeaderAware plugins in goroutines
        |
        +-- catalog plugin (registered via init())
              |-- wraps existing Loader, Services, OpenAPI controllers
              |-- RegisterRoutes: mounts model + MCP catalog routes with full paths
              |-- Start: loader.StartReadOnly()
              |-- OnBecomeLeader: registers PropertyOptionsRefresher, loader.StartLeader()
              |-- Stop: loader.Shutdown()
```

## Design decisions

**Single plugin for model + MCP catalogs** — The model catalog API's `FindSources` endpoint returns both model and MCP sources, and the `Loader` manages both with shared leader state and file watching. Splitting into separate plugins requires resolving this coupling first. The server infrastructure supports multiple plugins regardless.

**Routes use full paths** — OpenAPI-generated routes contain full patterns (e.g., `/api/model_catalog/v1alpha1/models`). Plugins mount routes directly on the root router rather than under a scoped prefix.

**MountRoutes returns error** — If any plugin fails to register routes, the server fails startup rather than running in a partially broken state.

**Dedicated shutdown context** — `Stop()` receives a fresh `context.WithTimeout(background, 10s)` instead of the already-cancelled signal context, ensuring plugins have time to flush.

**LeaderAware interface** — Optional interface plugins implement to receive leader election callbacks. Each plugin's `OnBecomeLeader` runs in its own goroutine and blocks until leadership is lost.

## Forward compatibility

The planned follow-up is: split the OpenAPI spec per plugin, refactor model catalog as a separate plugin, refactor MCP catalog as a separate plugin. This PR accommodates that:

- Server infrastructure is fully generic (iterates registry, N plugins)
- Route mounting works when each plugin has its own generated routes with distinct path prefixes
- `LeaderAware` gives each plugin independent leader notifications
- `Config.RepoSet` is read-only; multiple plugins can extract their own repos

What the split will additionally require (not in scope):
- A `DatastoreSpec()` method on plugins so each contributes its entity types
- Resolution of the `FindSources` cross-plugin coupling
- Per-plugin config sections (the `SourceKeyProvider` interface already exists for this)

## Endpoints

| Endpoint | Response |
|----------|----------|
| `GET /healthz` | Liveness probe — always `{"status":"ok"}` |
| `GET /readyz` | Readiness probe — aggregates `Healthy()` across all plugins. 200 if all healthy, 503 if any unhealthy. Zero plugins = 200. |

## How Has This Been Tested?

- 15 unit tests in `server_test.go` (zero-plugin, lifecycle, failures, health, leader, routes)
- All existing unit tests pass (`make test`, `make -C catalog test`)
- 158/158 catalog e2e tests pass (deployed to KinD with PostgreSQL)
- `make build/compile`, `make lint`, `make vet` all clean

## Merge criteria:

### DIFF
--- a/catalog/internal/plugin/server.go
+++ b/catalog/internal/plugin/server.go
@@ -1,0 +1,273 @@
+package plugin
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"sync"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/go-chi/cors"
+	"gorm.io/gorm"
+
+	"github.com/kubeflow/hub/catalog/internal/catalog/basecatalog"
+	"github.com/kubeflow/hub/internal/platform/datastore"
+)
+
+// ServerConfig holds the dependencies needed to create a plugin Server.
+type ServerConfig struct {
+	DB                     *gorm.DB
+	ConfigPaths            []string
+	PerformanceMetricsPath []string
+	RepoSet                datastore.RepoSet
+	TypeMap                map[string]int32
+	Logger                 *slog.Logger
+}
+
+// Server manages the lifecycle of catalog plugins and provides a unified HTTP server.
+type Server struct {
+	cfg     ServerConfig
+	mu      sync.RWMutex
+	plugins []CatalogPlugin
+	router  chi.Router
+}
+
+// NewServer creates a new plugin server.
+func NewServer(cfg ServerConfig) *Server {
+	if cfg.Logger == nil {
+		cfg.Logger = slog.Default()
+	}
+	return &Server{
+		cfg:     cfg,
+		plugins: make([]CatalogPlugin, 0),
+	}
+}
+
+// Init discovers all registered plugins and initializes them.
+// Fails fast on the first plugin Init error.
+func (s *Server) Init(ctx context.Context) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	registered := All()
+	if len(registered) == 0 {
+		s.cfg.Logger.Info("no plugins registered")
+		return nil
+	}
+
+	var sourceConfig *basecatalog.SourceConfig
+	if len(s.cfg.ConfigPaths) > 0 {
+		cfg, err := LoadConfig(s.cfg.ConfigPaths[0])
+		if err != nil {
+			return fmt.Errorf("loading source config: %w", err)
+		}
+		sourceConfig = cfg
+	}
+
+	for _, p := range registered {
+		basePath := computeBasePath(p)
+
+		pluginCfg := Config{
+			SourceConfig:           sourceConfig,
+			DB:                     s.cfg.DB,
+			BasePath:               basePath,
+			ConfigPaths:            s.cfg.ConfigPaths,
+			RepoSet:                s.cfg.RepoSet,
+			TypeMap:                s.cfg.TypeMap,
+			PerformanceMetricsPath: s.cfg.PerformanceMetricsPath,
+			Logger:                 s.cfg.Logger.With("plugin", p.Name()),
+		}
+
+		s.cfg.Logger.Info("initializing plugin",
+			"plugin", p.Name(),
+			"version", p.Version(),
+			"basePath", basePath,
+		)
+
+		if err := p.Init(ctx, pluginCfg); err != nil {
+			return fmt.Errorf("plugin %s init failed: %w", p.Name(), err)
+		}
+
+		s.plugins = append(s.plugins, p)
+	}
+
+	s.cfg.Logger.Info("all plugins initialized", "count", len(s.plugins))
+	return nil
+}
+
+// MountRoutes creates the HTTP router with all plugin routes and server endpoints.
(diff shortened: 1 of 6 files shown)


catalog/internal/plugin/server.go
```diff
@@ -0,0 +1,242 @@
...
+			errs = append(errs, fmt.Errorf("plugin %s: %w", p.Name(), err))
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// NotifyLeader notifies all LeaderAware plugins that this pod became leader.
+// Each plugin's OnBecomeLeader runs in its own goroutine; this method blocks
+// until all return (typically when ctx is cancelled / leadership lost).
+func (s *Server) NotifyLeader(ctx context.Context) {
```
jonburdo, line 166: Suggestion from claude fwiw:

**`NotifyLeader` holds `RLock` for the lifetime of leadership** (`server.go`)

The `RLock` is acquired at the top of `NotifyLeader` and `wg.Wait()` blocks until all `OnBecomeLeader` goroutines return — which is when leadership is lost, potentially hours later. While read locks don't block each other, they do block any write lock (`Init`), creating a latent deadlock risk. Recommend copying the plugin slice under lock, releasing it, then iterating the copy — same pattern already used in `Plugins()`:

```go
func (s *Server) NotifyLeader(ctx context.Context) {
	s.mu.RLock()
	plugins := make([]CatalogPlugin, len(s.plugins))
	copy(plugins, s.plugins)
	s.mu.RUnlock()

	var wg sync.WaitGroup
	for _, p := range plugins {
		// ...
	}
	wg.Wait()
}
```

Other minor things (non-blocking):
- `Plugin.Stop` discards the shutdown context, so the 10s deadline from the caller won't be honored if the loader supports context-aware shutdown.
- `Healthy()` always returns `true`, so `/readyz` can never surface a degraded catalog state.

jonburdo (approved):
/lgtm
/approve
/hold in case others want to review or you had other changes

I did a pass and it looks good - one suggestion from claude

catalog/internal/plugin/server.go
Al-Pragliola: Thanks for the review @jonburdo! Addressed the NotifyLeader lock concern in 83e9e240 — copied the plugin slice under lock and released it before wg.Wait(), same pattern as Plugins().

Al-Pragliola: for the other two points, the Stop in practice does not hit the 10s limit so I'd leave it as-is for now with keeping in mind that we could refactor it later if needed, the Healthy one is fine for now as we don't implement yet the two plugins (catalog and mcp)

jonburdo (commented):
/lgtm
/approve

Looks great - thanks!

PR #2780, opened 2026-06-03
fix(catalog): add health state tracking to PluginBase and negative readyz tests

dbasunag:

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/internal/plugin/base.go
+++ b/catalog/internal/plugin/base.go
@@ -3,6 +3,7 @@ package plugin
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	mapset "github.com/deckarep/golang-set/v2"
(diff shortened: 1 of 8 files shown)


jonburdo (commented):
Overall this look good to me, though @Al-Pragliola can comment better on the best approach!

Some remarks from claude:

Race condition in TestPluginBaseWatchFileLeaderOpsFailureMarksUnhealthy:

watcher.send("a.yaml") followed by an immediate assert on pb.Healthy() is racy — watchFile runs in a goroutine, and the send unblocks when the goroutine receives, not when it finishes processing (calling PerformLeaderOperations, getting the error, and storing false). Same issue with the time.Sleep(50ms) in TestPluginBaseWatchFileErrorMarksUnhealthy.

Both should use assert.Eventually to avoid flaky tests:

```go
assert.Eventually(t, func() bool { return !pb.Healthy() }, time.Second, 5*time.Millisecond)
```

catalog/internal/plugin/base.go
```diff
@@ -79,6 +89,7 @@ func (pb *PluginBase) watchFile(ctx context.Context, path string) {
 		if pb.cfg.State.ShouldWriteDatabase() {
 			allKnownSourceIDs := CollectAllSourceIDs()
 			if err := pb.cfg.Loader.PerformLeaderOperations(ctx, allKnownSourceIDs); err != nil {
+				pb.healthy.Store(false)
```
Al-Pragliola, line 97: once healthy.Store(false) is called nothing sets it back to true — a transient db error on reload permanently kills readiness even if the next reload succeeds. Shouldn't we add pb.healthy.Store(true) on the success path here and at the end of OnBecomeLeader? the FileWatcher.Path() error can stay sticky since the goroutine exits, but for leader ops it seems like we want recovery. wdyt?

catalog/internal/plugin/base.go
```diff
@@ -3,6 +3,7 @@ package plugin
 import (
```
Al-Pragliola, line 3: ReloadParsing() swallows parse errors (logs but returns nothing), so a malformed sources.yaml won't mark the plugin unhealthy — only PerformLeaderOperations failures do. is that intentional? the plugin could be serving stale config from a failed parse and still report healthy

catalog/internal/plugin/base.go
dbasunag: Thank you! I have added recovery on success for watchFile reload and OnBecomeLeader. FileWatcher.Path() error was not changed since the goroutine exits. Also added a recovery assertion to the reload test.

catalog/internal/plugin/base.go
dbasunag: Ok, I updated it. So ReloadParsing() now returns an error. Both loaders collect and return parse errors instead of swallowing them. watchFile marks the plugin unhealthy on parse failure and skips leader ops. Added TestPluginBaseWatchFileParseFailureMarksUnhealthy. Please let me know if it makes sense.

PR #2789, opened 2026-06-04
fix(controller): remove finalizer when ModelRegistry service is deleted before InferenceService

Al-Pragliola:
### Summary

  - Fix InferenceService finalizer stuck in Terminating when the ModelRegistry service is deleted before the InferenceService
  - Move deletion handling before initModelRegistryService() so the finalizer is always removed, even when the MR Service no longer exists
  - MR state update (DesiredState → UNDEPLOYED) becomes best-effort during deletion — logged and skipped if MR is unavailable
  - Add integration test reproducing the exact scenario (MR Service deleted → ISVC deleted → verify ISVC is fully cleaned up)

 ### Root Cause

  In Reconcile(), initModelRegistryService() was called before checking isvc.GetDeletionTimestamp(). When the ModelRegistry Service was already gone, getMRService() → client.Get() returned a not-found error. This
  error caused an early return, and the finalizer removal code was never reached:

  Reconcile():
```
    mrApi, err := r.initModelRegistryService(...)   // fails: Service not found
    if err != nil { return ctrl.Result{}, err }      // returns here every time
    // ...
    // finalizer removal code below is NEVER reached
    if isMarkedToBeDeleted { ... removeFinalizer ... }
```

  Every reconciliation loop hit the same error → finalizer stayed forever → InferenceService couldn't be deleted → namespace stuck in Terminating.

  ### Fix

  Moved the deletion timestamp check before initModelRegistryService(). When the ISVC is marked for deletion:

  1. Attempt to init MR service and update state (best-effort)
  2. If MR service is unavailable, log and skip
  3. Always remove the finalizer, regardless of MR availability

## How Has This Been Tested?

  - [x] New integration test: "Should remove the finalizer and allow deletion even if the ModelRegistry service no longer exists"
    - Creates namespace, MR Service, and InferenceService
    - Waits for controller to add finalizer and IS ID label
    - Deletes the MR Service
    - Deletes the InferenceService
    - Verifies the ISVC is fully deleted (not stuck in Terminating)
  - [x] Test confirmed red before fix, green after
  - [x] All 9 existing tests still pass (no regressions)

## Merge criteria:

### DIFF
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -124,6 +124,39 @@ func (r *InferenceServiceController) Reconcile(ctx context.Context, req ctrl.Req
 		mrNamespace = mrNSFromISVC
 	}
 
+	// Check if the InferenceService instance is marked to be deleted, which is
+	// indicated by the deletion timestamp being set.
+	isMarkedToBeDeleted := isvc.GetDeletionTimestamp() != nil
+
+	if isMarkedToBeDeleted {
+		if controllerutil.ContainsFinalizer(isvc, r.modelRegistryFinalizer) {
+			// Best-effort: try to update MR InferenceService state before removing the finalizer.
+			// If the MR service is unavailable (e.g., deleted before the ISVC), we still
+			// must remove the finalizer to avoid blocking deletion indefinitely.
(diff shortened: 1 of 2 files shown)


pboyd (commented):
/lgtm
/approve

PR #2815, opened 2026-06-10
feat(catalog): connect cold-start, vRAM, and container size filters to real API

manaswinidas:
## Description

- Rename cold_start_load_time_seconds to cold_start_time_to_load_seconds to align with backend field name
- Add min vRAM and image size sidebar slider filters (behind perf toggle)
- Read filter ranges from API instead of hardcoded values
- Display cold-start latency in seconds across cards, table, and filters
- Support double_value format for min_vram_gb and modelcar_image_size
- Add ColdStartMetricsArtifact type for new cold-start-metrics artifacts
- Exclude cold-start filter from artifacts endpoint (different artifact type)
- Update demo data with cold-start, vRAM, image size, and runtime commands
Screenshot 2026-06-10 at 3 40 17 PM
Screenshot 2026-06-10 at 3 40 28 PM
Screenshot 2026-06-10 at 3 40 59 PM
Screenshot 2026-06-10 at 3 41 20 PM

## How Has This Been Tested?

Fire up a kind cluster for model catalog with some performance artifacts and test whether cold start latency, min vRAM, container size, runtime command data shows up fine.

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/modelCatalog/components/SidebarSliderFilter.tsx
```diff
@@ -0,0 +1,112 @@
+import * as React from 'react';
```
ppadti, line 1: ColdStartLatencyFilter and MaxRpsFilter share the same state management pattern. Can we extract a common NumericDropdownFilter base that all three(sidebarslider, maxRPS, coldstartlatency filter) can compose from to avoid diverging behavior over time.

clients/ui/frontend/src/app/pages/modelCatalog/utils/modelCatalogUtils.ts
```diff
@@ -571,17 +580,29 @@ export const getCatalogModelTypePropertyForRegistration = (
...
-  customProperties
-    ? getCustomPropString(customProperties, CatalogModelCustomPropertyKey.MINIMUM_VRAM)
-    : '';
+): string => {
+  if (!customProperties) {
+    return '';
+  }
+  const doubleVal = getDoubleValue(customProperties, CatalogModelCustomPropertyKey.MINIMUM_VRAM);
+  if (doubleVal > 0) {
+    return `${doubleVal} GB`;
```
ppadti: Can we use `.toFixed(2)` for this util as well - we are using it in getModelSizeFromCustomProperties

clients/ui/frontend/src/app/pages/modelCatalog/components/ModelCatalogFilters.tsx
```diff
@@ -86,17 +111,21 @@ const ModelCatalogFilters: React.FC = () => {
...
-              </Content>
-            ) : undefined,
+          footer: (
+            <>
+              {hasMultiple && hasSelection && (
+                <Content component={ContentVariants.small} className="pf-v6-u-mt-sm">
+                  Showing models with all selected configurations
+                </Content>
+              )}
+              {sliderContent}
```
ppadti: We are adding these 2 new filter as footer of the validated configuration filter and in midstream we have this validated configuration behind feature flag - which would cause it to disappear when we turn off the `toolCalling` feature flag
so, can we consider adding a customContent variant and an order prop to FilterPanelItem so these can be first-class panel items rendered independently and this will make it more generic as well.

clients/ui/frontend/src/concepts/modelCatalog/const.ts
ppadti, line 496: Image

I think we need to add MIN_VRAM and IMAGE_SIZE to PERFORMANCE_FILTER_KEYS so they're cleared by getBasicFiltersOnly() when performance view is toggled off — otherwise the filter value persists in state and is sent to the API with no visible UI indicator.

ppadti (changes requested):

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -14,7 +14,9 @@ export enum ModelCatalogStringFilterKey {
 export enum ModelCatalogNumberFilterKey {
   // Performance filter key uses backend format
   MAX_RPS = 'artifacts.requests_per_second.double_value',
-  COLD_START_LATENCY = 'artifacts.cold_start_load_time_seconds.double_value',
+  COLD_START_LATENCY = 'artifacts.cold_start_time_to_load_seconds.double_value',
```
Philip-Carneiro: I think we should change this const key, to make sense, since we are removing anything related to "Cold start latency" from the code it self... Just this point, this will affect other files change you have made

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -607,6 +613,8 @@ export const MODEL_CATALOG_FILTER_CHIP_PREFIXES = {
   WORKLOAD_TYPE: 'Workload type:',
   MAX_RPS: 'Max RPS:',
   COLD_START_LATENCY: 'Cold start load time: ≤',
+  MIN_VRAM: 'Minimum vRAM <=',
+  IMAGE_SIZE: 'Container size <=',
```
Philip-Carneiro: use the ≤ instead <=
We are using it already in latency

Philip-Carneiro (changes requested):

manaswinidas:
Min vRAM and container size filters looks like this now 

https://github.com/user-attachments/assets/7b7a7544-6e34-45a0-8cae-38b488f69ba2

clients/ui/frontend/src/app/pages/modelCatalog/components/SidebarSliderFilter.tsx
manaswinidas: Will follow up later on this in another PR

clients/ui/frontend/src/concepts/modelCatalog/const.ts
```diff
@@ -466,6 +468,8 @@ export const BASIC_FILTER_KEYS: ModelCatalogFilterKey[] = [
   ModelCatalogStringFilterKey.LANGUAGE,
   ModelCatalogStringFilterKey.TENSOR_TYPE,
   ModelCatalogStringFilterKey.VALIDATED_CONFIGURATION,
+  ModelCatalogNumberFilterKey.MIN_VRAM,
+  ModelCatalogNumberFilterKey.IMAGE_SIZE,
```
ppadti: I am still seeing the filter chip with vram and container size when we turn off the perf toggle

https://github.com/user-attachments/assets/2672cee5-3ed5-45a8-9686-59e48d283b99

We just have to remove these from here and add  
```
baseSetFilterData(ModelCatalogNumberFilterKey.MIN_VRAM, undefined);
    baseSetFilterData(ModelCatalogNumberFilterKey.IMAGE_SIZE, undefined);
```
to `resetPerformanceFiltersToDefaults` call back in `ModelCatalogContext.tsx`?

ppadti (changes requested):

clients/ui/frontend/src/concepts/modelCatalog/const.ts
manaswinidas: Applied this - this works fine now. Added Cypress tests to ensure the same

PR #2817, opened 2026-06-11
fix(ui): replace Popover with Tooltip on disabled Register model button to fix focus trap

Philip-Carneiro:
## Description

Fixes an accessibility (a11y) regression on the Model Catalog Details page where keyboard focus (Tab) gets trapped on  the disabled "Register model" button.
**Root cause:** The `Popover` component with `triggerAction="hover"` combined with `isAriaDisabled` on the button causes PatternFly's internal focus management to trap the Tab cycle, preventing users from navigating past the button.
**Fix:** Replace `Popover` with `Tooltip` for the disabled "Register model" button. The `Tooltip` provides the same hover information without interfering with keyboard focus management. The `Popover` used for the "Validated" label is unaffected and remains unchanged.

### Before
<video src="https://github.com/user-attachments/assets/9e84f748-5c4d-44a5-859a-f00ad0ac8efa" controls></video>
### After
<video src="https://github.com/user-attachments/assets/fde3c0fd-3935-48cd-84b0-56a635bf8a81" controls></video>

## How Has This Been Tested?

It was manually tested as the video sample and added tests in cypress for it

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
@@ -271,6 +271,24 @@ describe('Model Catalog Details Page - Edge Cases', () => {
     cy.findAllByText('N/A').should('have.length.at.least', 1);
   });
 
+  it('should show disabled register button with tooltip when no model registries exist', () => {
+    cy.intercept('GET', '/model-registry/api/v1/model_registry*', mockModArchResponse([])).as(
+      'getEmptyModelRegistries',
+    );
+
+    setupModelCatalogIntercepts({});
+    interceptArtifactsList();
+
+    modelCatalog.visit();
+    modelCatalog.findLoadingState().should('not.exist');
+    modelCatalog.findModelCatalogDetailLink().first().click();
+    modelCatalog.findBreadcrumb().should('exist');
+
+    modelCatalog.findRegisterModelButton().should('have.attr', 'aria-disabled', 'true');
+    modelCatalog.findRegisterModelButton().trigger('mouseenter');
+    modelCatalog.findRegisterCatalogModelTooltip().should('be.visible');
+  });
+
   it('should show error alert when artifacts fail to load', () => {
     setupModelCatalogIntercepts({});
 
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
@@ -287,6 +287,7 @@ describe('Model Catalog Details Page - Edge Cases', () => {
     modelCatalog.findRegisterModelButton().should('have.attr', 'aria-disabled', 'true');
     modelCatalog.findRegisterModelButton().trigger('mouseenter');
     modelCatalog.findRegisterCatalogModelTooltip().should('be.visible');
+    cy.testA11y({ exclude: ['.pf-v6-c-tooltip'] });
   });
 
   it('should show error alert when artifacts fail to load', () => {
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
@@ -284,6 +284,7 @@ describe('Model Catalog Details Page - Edge Cases', () => {
     modelCatalog.findModelCatalogDetailLink().first().click();
     modelCatalog.findBreadcrumb().should('exist');
 
+    cy.wait('@getEmptyModelRegistries');
     modelCatalog.findRegisterModelButton().should('have.attr', 'aria-disabled', 'true');
     modelCatalog.findRegisterModelButton().trigger('mouseenter');
     modelCatalog.findRegisterCatalogModelTooltip().should('be.visible');
--- a/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsPage.tsx
+++ b/clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsPage.tsx
@@ -13,6 +13,7 @@ import {
   StackItem,
   Button,
   Popover,
+  Tooltip,
   ActionListGroup,
   Skeleton,
   Label,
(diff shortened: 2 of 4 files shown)


clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
```diff
@@ -271,6 +271,24 @@ describe('Model Catalog Details Page - Edge Cases', () => {
     cy.findAllByText('N/A').should('have.length.at.least', 1);
   });
 
+  it('should show disabled register button with tooltip when no model registries exist', () => {
```
manaswinidas, line 274: can we add a `cy.testA11y()` to double check all accessibility violations are gone?

manaswinidas (changes requested):

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
Philip-Carneiro: The cy.testA11y() call now excludes .pf-v6-c-tooltip because PatternFly v6 tooltips render with insufficient color contrast (white text #ffffff on #eeeeee background, contrast ratio 1.16:1 vs the 4.5:1 WCAG AA requirement). This is an internal PatternFly framework styling issue — the tooltip content inherits colors from PF's CSS custom properties, which we cannot override without breaking the component contract. The exclusion only applies to this specific test where the tooltip is intentionally visible during the a11y check.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogDetails.cy.ts
```diff
@@ -271,6 +271,25 @@ describe('Model Catalog Details Page - Edge Cases', () => {
...
+    cy.intercept('GET', '/model-registry/api/v1/model_registry*', mockModArchResponse([])).as(
```
manaswinidas, line 275: **Unused alias**: The `'getEmptyModelRegistries'` alias is created here but is never referenced (no `cy.wait('@getEmptyModelRegistries')` or assertion on it). Consider either:
- Removing the `.as('getEmptyModelRegistries')` since it's unused, or
- Adding a `cy.wait('@getEmptyModelRegistries')` before asserting on the button to make the test more deterministic (explicitly waits for the mock to be served before checking button state).

The test likely works today because Cypress retries `.should()` assertions, but an explicit wait is a safer pattern for intercept-dependent tests.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsPage.tsx
```diff
@@ -57,17 +58,24 @@ const ModelDetailsPage: React.FC<ModelDetailsPageProps> = ({ tab }) => {
...
+            <div>{bodyContent}</div>
+          </div>
+        ) : (
+          bodyContent
+        )
+      }
+      data-testid="register-catalog-model-tooltip"
     >
       <Button variant="primary" isAriaDisabled data-testid="register-model-button">
         Register model
```
manaswinidas, line 76: nit: **`data-testid` on Tooltip may be fragile** — PatternFly's `Tooltip` renders its popup content in a portal (appended to `document.body`). The `data-testid` prop is not part of PF's documented Tooltip API, so whether it propagates to the actual tooltip popup element is an implementation detail that could break across PF upgrades.

A more robust alternative would be to use the `id` prop (which PF explicitly supports and applies to the tooltip popup element):

```tsx
<Tooltip
  id="register-catalog-model-tooltip"
  content={...}
>
```

Then in the page object:
```ts
findRegisterCatalogModelTooltip() {
  return cy.get('#register-catalog-model-tooltip');
}
```

Non-blocking if the current approach works in CI, but worth noting for long-term maintenance.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsView.tsx
```diff
@@ -326,22 +326,22 @@ const ModelDetailsView: React.FC<ModelDetailsViewProps> = ({
...
-                      ) : (
-                        <p className={text.textColorDisabled}>No artifacts available</p>
-                      )}
+                        ) : (
+                          'No artifacts available'
+                        )}
+                      </DescriptionListDescription>
                     </DescriptionListGroup>
                     {architectures.length > 0 && (
                       <DescriptionListGroup>
```
manaswinidas, line 347: Good refactor — wrapping all branches inside a single `<DescriptionListDescription>` fixes the semantic HTML issue (previously, the error Alert, Spinner, and "No artifacts" text were siblings of `<DescriptionListDescription>` instead of inside it, which broke the `<dl>` → `<dt>`/`<dd>` contract).

Two small things to note:

1. **Visual change**: The "No artifacts available" text now renders as a plain text node instead of `<p className={text.textColorDisabled}>`. This drops the gray/disabled styling. This is actually *consistent* with how other empty values in this description list are rendered (e.g. `tensorType || 'N/A'`, `size || 'N/A'`), so I think this is a net improvement. Just want to confirm this was intentional.

2. **Scope note**: This change is tangential to the Popover → Tooltip a11y fix described in the PR title. Consider calling it out in the PR description so reviewers know it's an intentional cleanup bundled with the fix.

manaswinidas (commented):
Thanks for the update, @Philip-Carneiro! The `cy.testA11y()` addition addresses my earlier feedback. Here's a summary of the re-review:

## Checklist

| Category | Finding |
|---|---|
| **DRY violations** | None. The `registerButtonTooltip` helper properly avoids repetition across the 3 call sites. |
| **Dead code / unused imports** | None. `Popover` import is still needed for the "Validated" label and "Certified platforms" help icon. `text` import in `ModelDetailsView.tsx` is still used for "No model card". |
| **Unused alias** | `'getEmptyModelRegistries'` intercept alias is created but never waited on or referenced. See inline comment. |
| **Performance issues** | None. No new state, effects, or heavy computations introduced. |
| **Maintenance concerns** | `data-testid` on PF `Tooltip` is not a documented PF prop; tooltip popup renders in a portal, making this fragile across PF upgrades. See inline suggestion to use `id` instead. |
| **Breaking changes** | The Popover → Tooltip swap intentionally changes the visual appearance of the disabled button overlay (smaller/simpler tooltip vs structured popover). The "No artifacts available" text in `ModelDetailsView.tsx` loses its `textColorDisabled` styling — this is actually more consistent with peer fields, but is a visual change worth calling out. |
| **Tautological / repetitive tests** | None. The new test covers a unique scenario (empty registries → disabled button + tooltip) not tested elsewhere in the file. |

## Overall

The core fix (Popover → Tooltip) is correct and well-targeted. Three minor items flagged inline — none are blocking.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelDetailsView.tsx
manaswinidas: Ignore, this is intentional

manaswinidas (approved):
/lgtm
/approve

PR #2851, opened 2026-06-16
feat: set NextProtos (ALPN) on HTTP TLS configs

ugiordan:
## Summary
- Add explicit ALPN negotiation (`h2`, `http/1.1`) to all HTTP-facing TLS configurations
- BFF server TLS, BFF outbound HTTP client, InferenceService controller HTTP client
- Ensures proper HTTP/2 negotiation on OpenShift clusters where ALPN is required by the serving infrastructure

The generic `BuildTLSConfig()` in `internal/platform/tls/config.go` is intentionally NOT modified since it is shared with database connectors (MySQL, PostgreSQL) that do not use HTTP ALPN negotiation.

## Files changed
- `clients/ui/bff/cmd/main.go`: BFF server TLS config
- `clients/ui/bff/internal/integrations/httpclient/http.go`: BFF outbound HTTP client
- `pkg/inferenceservice-controller/controller.go`: InferenceService controller HTTP client

## Test plan
- [ ] Verify BFF server negotiates HTTP/2 via ALPN when TLS is enabled
- [ ] Verify BFF outbound HTTP client includes ALPN in TLS handshake
- [ ] Verify InferenceService controller HTTP client includes ALPN in TLS handshake
- [ ] Confirm database connections (MySQL, PostgreSQL) are not affected

### DIFF
--- a/clients/ui/bff/cmd/main.go
+++ b/clients/ui/bff/cmd/main.go
@@ -101,6 +101,7 @@ func main() {
 			// Configure TLS if both cert and key files are provided
 			tlsConfig := &tls.Config{
 				MinVersion: tls.VersionTLS13,
+				NextProtos: []string{"h2", "http/1.1"},
 			}
 			srv.TLSConfig = tlsConfig
 			err = srv.ListenAndServeTLS(certFile, keyFile)
--- a/pkg/inferenceservice-controller/controller.go
+++ b/pkg/inferenceservice-controller/controller.go
@@ -48,13 +48,15 @@ func NewInferenceServiceController(
 	serviceURLAnnotation,
 	defaultMRNamespace string,
 ) *InferenceServiceController {
-	httpClient := http.DefaultClient
-
+	transport := http.DefaultTransport.(*http.Transport).Clone()
 	if skipTLSVerify {
-		httpClient.Transport = &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		transport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true,
+			MinVersion:         tls.VersionTLS12,
+			NextProtos:         []string{"h2", "http/1.1"},
 		}
 	}
+	httpClient := &http.Client{Transport: transport}
 
 	return &InferenceServiceController{
 		client:                        client,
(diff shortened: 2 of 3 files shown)


pboyd:
This looks good to me.

@ederign, do you want to double-check the BFF changes?

/ok-to-test
/lgtm

ugiordan:
@pboyd needs approval I guess.

ederign:
@pboyd I'll look at it today.

clients/ui/bff/cmd/main.go
```diff
@@ -101,6 +101,7 @@ func main() {
 			// Configure TLS if both cert and key files are provided
 			tlsConfig := &tls.Config{
 				MinVersion: tls.VersionTLS13,
```
ederign, line 103: is there a reason why this is different from controller.go and don't have ForceAttemptHTTP2: true, ?

pkg/inferenceservice-controller/controller.go
ederign: I wonder if this should be httpClient := &http.Client{}

ederign: Are we sure we want to mutate the global singleton's Transport field? (but not related to this change)

ederign (changes requested):

pkg/inferenceservice-controller/controller.go
ugiordan: > I wonder if this should be httpClient := &http.Client{}

Agreed, fixed. Now clone `DefaultTransport` and create a new `http.Client` instead of touching the global singleton.

ugiordan: > Are we sure we want to mutate the global singleton's Transport field? (but not related to this change)

Fixed in the same commit. The controller now uses `http.DefaultTransport.Clone()` so the global state is never modified.

clients/ui/bff/cmd/main.go
ugiordan: ForceAttemptHTTP2 is a client-side `http.Transport` field, it doesn't exist on `http.Server`. On the server side, HTTP/2 is enabled automatically when "h2" is in `NextProtos` and the server uses TLS (https://pkg.go.dev/net/http#Server).

ederign:
@ugiordan can you please check BFF tests ?

ugiordan:
@ederign fixed, it was a gofmt alignment issue on TLSClientConfig (field alignment with ForceAttemptHTTP2). Should be green now.

ederign (approved):

ederign:
@pboyd as you lgtm, i'll merge it!

PR #2854, opened 2026-06-17
fix(ui): replace OuterScrollContainer with InnerScrollContainer for hardware config table

manaswinidas:
## Summary

- Replaces `OuterScrollContainer` with `InnerScrollContainer` from `@patternfly/react-table` to fix missing table headers
- Removes `isStickyHeader` and `hasStickyColumns` props that were conflicting with the `mod-arch-shared` Table component
- Renders toolbar/filters outside the scroll container so they stay fixed while columns scroll horizontally

## Problem

`OuterScrollContainer` combined with `isStickyHeader`/`hasStickyColumns` was causing the hardware configuration table headers to not render. The PF scroll container was incompatible with the `mod-arch-shared` Table wrapper component.

## How Has This Been Tested?

- Verified locally that table headers now render correctly
- Horizontal scrolling stays within the card boundary
- Toolbar/filters remain fixed while table content scrolls

## Test Impact

No test changes needed — this is a rendering fix only.


ppadti (approved):
/lgtm
/approve

PR #2856, opened 2026-06-17
fix(client): sanitize server address before hint logging to preserve diagnostic warnings

### LINKED ISSUE #2826: bug(client/python): Conflicting URL parsing and redundant token checks during ModelRegistry initialization
The Python Client SDK constructor for ModelRegistry accepts a server_address string alongside a boolean flag is_secure and an integer port. Currently, passing a full URL scheme (e.g., https://... or http://...) inside the server_address parameter causes conflicting internal logic paths, configuration vulnerabilities, and unhandled connection failures.

Steps to Reproduce
When initializing the client using a protocol prefix as suggested by the README.md example usage guide:

Python

# Attempting initialization with an explicit protocol scheme
registry = ModelRegistry("https://model-registry-service", author="Ada Lovelace")
Observed Behavior
Redundant Token Guardrail Blocks: If a user passes https:// within the address string, the constructor immediately triggers a security gate check, crashing with a user token must be provided for secure connection exception. This occurs even if a user explicitly passes is_secure=False to bypass token checks for local mock deployments or testing.

Protocol Race Condition: If a user passes an explicit scheme string mismatched with the security boolean flag (e.g., server_address="http://my-host" while leaving is_secure=True as a default), the codebase logs a weak warning via hint_server_address_port but proceeds to execute a secure TLS handshake over a plain HTTP socket channel, causing the connection to hang or drop silently.

Expected Behavior
The ModelRegistry constructor should cleanly handle and sanitize incoming URL strings at the public API boundary. It should dynamically infer connection security parameters if a scheme is declared, while extracting the underlying hostname/IP cleanly to pass down to the OpenAPI generation configuration layers without double-scheming conflicts.

Proposed Resolution
Introduce a robust URL sanitization block at the top of ModelRegistry.__init__ using urllib.parse.urlparse to extract clean network parameters before downstream security logic runs:

Python
from urllib.parse import urlparse

parsed = urlparse(server_address)
if parsed.scheme:
    is_secure = True if parsed.scheme == "https" else False
    server_address = parsed.hostname if parsed.hostname else parsed.netloc
    if parsed.port:
        port = parsed.port

ManvithPanyam:
### Description
This PR fixes an unhandled `StoreError` exception that occurs on local machines outside of a Kubernetes cluster environment when initializing the `ModelRegistry` client without a pre-configured token. 

Previously, passing a raw URL like `https://...` bypassed the secure state evaluation, forcing `is_secure` to `True` and triggering a hard crash when no local token was found. This patch extracts the URL scheme and sanitizes the `server_address` so local developers can instantiate the client cleanly.

### Changes Made
- Added a `urllib.parse` sanitization block in `__init__` to extract the scheme and hostname cleanly.
- Placed the sanitization block **after** the `hint_server_address_port` call to ensure the raw string is checked for diagnostic warnings (preventing test regressions).
- Simplified the boolean assignment for `is_secure` to satisfy project linting rules.

### Testing
- [x] Ran `poetry run pytest tests/test_client.py` locally (All tests pass, confirming warning hint logic is preserved).
- [x] Verified code formatting with `poetry run ruff check`.

Fixes #2826

### DIFF
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -83,7 +83,7 @@ class ModelRegistry:
         port: int = 443,
         *,
         author: str,
-        is_secure: bool = True,
+        is_secure: bool | None = None,
         user_token: str | None = None,
         user_token_envvar: str | None = None,
         custom_ca: str | None = None,
@@ -99,7 +99,7 @@ class ModelRegistry:
 
         Keyword Args:
             author: Name of the author.
-            is_secure: Whether to use a secure connection. Defaults to True.
+            is_secure: Whether to use a secure connection. Inferred from URL scheme if not provided; defaults to True.
             user_token: The PEM-encoded user token as a string.
             user_token_envvar: Environment variable to read the user token from if it's not passed as an arg. Defaults to KF_PIPELINES_SA_TOKEN_PATH.
             custom_ca: Path to the PEM-encoded root certificates as a string.
@@ -129,6 +129,21 @@ class ModelRegistry:
                 warn("User access token is missing", stacklevel=2)
 
         self.hint_server_address_port(server_address, port)
+
+        # --- START OF URL SANITIZATION BLOCK ---
+        from urllib.parse import urlparse
+
+        parsed = urlparse(server_address)
+        if parsed.scheme in ("http", "https"):
+            if parsed.port:
+                port = parsed.port
+            # Infer is_secure from scheme only if user didn't explicitly pass it
+            if is_secure is None:
+                is_secure = parsed.scheme == "https"
+        # Fall back to secure by default if still unset
+        if is_secure is None:
+            is_secure = True
+        # --- END OF URL SANITIZATION BLOCK ---
         if is_secure:
             if not custom_ca and custom_ca_envvar and (cert := os.getenv(custom_ca_envvar)):
                 logger.info(


ManvithPanyam:
@tarilabs could you take a look when you get a chance? DCO and checks are all green now.

clients/python/src/model_registry/_client.py
```diff
@@ -129,6 +129,19 @@ def __init__(  # noqa: C901
                 warn("User access token is missing", stacklevel=2)
 
         self.hint_server_address_port(server_address, port)
+
+        # --- START OF URL SANITIZATION BLOCK ---
+        from urllib.parse import urlparse
+        parsed = urlparse(server_address)
+        if parsed.scheme:
+            # Automatically infer secure state from the user's explicit URI scheme
+            is_secure = parsed.scheme == "https"
```
pboyd: This overrides the `is_secure` argument, so what happens if someone has a self-signed cert on their `https://` URL? Can they still set `is_secure=False`?

ManvithPanyam:
Hi @pboyd — good catch, and you're right that the original implementation 
unconditionally overwrote `is_secure`. This is addressed in the latest push.

`is_secure` now defaults to `None` as a sentinel, so scheme inference only 
runs when the user hasn't explicitly set it:

```python
if is_secure is None:
    is_secure = parsed.scheme == "https"
```

Passing `is_secure=False` with an `https://` URL (your self-signed cert case) 
is now fully respected — the explicit value takes priority.

pboyd:
@ManvithPanyam , thanks for the change.

/lgtm

@jonburdo, I know you were looking at this too, do you want to review this PR?

pboyd:
@ManvithPanyam, sorry I was premature. Can you look at those test failures?

/lgtm cancel

ManvithPanyam:
Found it — `urlparse("localhost:8080")` misparsed the host as a scheme 
(`scheme='localhost'`, `netloc=''`), causing the sanitization block to 
strip the server address and produce an empty string, which explains all 
48 `NonHttpUrlClientError` failures.

Fixed by scoping the block to `scheme in ("http", "https")` only — bare 
hostnames like `localhost:8080` now pass through untouched. Verified 
locally with a reproduction script and all 11 unit tests still pass.

ManvithPanyam:
@pboyd identified and fixed the root cause of the E2E failures.

The sanitization block was reassigning `server_address = parsed.hostname`, 
which stripped the scheme and gave aiohttp a bare `localhost:8080` — 
hence the `NonHttpUrlClientError` across all 48 tests.

Fixed by removing that reassignment entirely. The block now only infers 
`port` and `is_secure` from the scheme, leaving `server_address` untouched. 
All 11 unit tests pass locally.

pboyd:
Looks good now, @ManvithPanyam, thanks.

/lgtm

@jonburdo, any concerns before merging this?

ManvithPanyam:
@jonburdo no rush — just wanted to check if there's anything else 
needed from my side before this can be merged.

jonburdo (commented):
/lgtm
/approve

Thanks for the fix!

PR #2859, opened 2026-06-18
Fix empty state reset button label to always show "Reset all defaults"

ConorOM1:
## Description
The empty state reset button in the Model Catalog always showed "Reset all filters" when the Model Performance View was disabled. The hardware config toolbar consistently uses "Reset all defaults", making this a terminology mismatch. Changed both the gallery empty state button and the sidebar toolbar clear-filters button to always read "Reset all defaults".
## How Has This Been Tested?

- Unit tests pass

- Verified both button labels update correctly with performance view on and off

image

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
```diff
@@ -719,6 +719,40 @@ describe('Single Category Behavior', () => {
...
+    modelCatalog.findCategoryToggle('label-Provider one').click();
+
+    modelCatalog.findFilterShowMoreButton('Task').click();
+    modelCatalog.findFilterCheckbox('Task', 'audio-to-text').click();
+
+    modelCatalog.findModelCatalogEmptyState().should('contain.text', 'No results found');
+    modelCatalog.findEmptyStateResetFiltersButton().should('contain.text', 'Reset all defaults');
+  });
+});
+
```
manaswinidas, line 737: These two tests are tautological — the label is now a hardcoded string, so toggling `performanceView` ON vs OFF doesn't exercise any branching logic. One test asserting the static text is sufficient. Testing the same constant under different states adds no regression protection.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
```diff
@@ -285,9 +285,7 @@ const ModelCatalogGalleryView: React.FC<ModelCatalogPageProps> = ({
           description="Adjust your filters and try again."
           primaryAction={
             <Button variant="link" onClick={handleFilterReset}>
-              {performanceViewEnabled && hasPerformanceFiltersChanged
-                ? 'Reset all defaults'
-                : 'Reset all filters'}
+              Reset all defaults
```
manaswinidas: nit: Consider extracting `'Reset all defaults'` to a shared constant — this string now appears here, in `ModelCatalogSourceLabelSelector`, and in `HardwareConfigurationFilterToolbar`. A single source of truth avoids future drift if the label changes again.

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
```diff
@@ -135,7 +135,7 @@ const ModelCatalogSourceLabelSelector: React.FC<ModelCatalogSourceLabelSelectorP
           {...(onResetAllFilters && !performanceViewEnabled && hasBasicFiltersApplied
             ? {
                 clearAllFilters: handleClearAllFilters,
-                clearFiltersButtonText: 'Reset all filters',
+                clearFiltersButtonText: 'Reset all defaults',
```
manaswinidas: This is asserting that PatternFly's `<Toolbar>` renders `clearFiltersButtonText` correctly — that's PF library behavior, not application logic. What's worth testing is that we pass the right props under the right conditions (`!performanceViewEnabled && hasBasicFiltersApplied`).

clients/ui/frontend/src/app/pages/modelCatalog/screens/__tests__/ModelCatalogGalleryView.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+    Element.prototype.scrollTo = jest.fn();
+  });
+
+  describe('empty state reset button label', () => {
+    it('should show "Reset all defaults" when performance view is disabled', () => {
+      renderWithContext();
+      expect(screen.getByText('Reset all defaults')).toBeInTheDocument();
+    });
+
+    it('should show "Reset all defaults" when performance view is enabled', () => {
```
manaswinidas: Same concern as the Cypress tests — both `performanceViewEnabled: false` and `performanceViewEnabled: true` variants assert the same hardcoded string. Since the conditional was removed, these don't test different code paths. One test is enough.

Also, this duplicates the Cypress coverage — both layers now test the same static label. Pick one.

clients/ui/frontend/src/app/pages/modelCatalog/screens/__tests__/ModelCatalogGalleryView.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  catalogModelsLoaded: true,
+  catalogModelsLoadError: undefined,
+  refresh: jest.fn(),
+};
+
+const handleFilterReset = jest.fn();
+
+const renderWithContext = (contextOverrides: Partial<ModelCatalogContextType> = {}) => {
+  mockUseCatalogModelsBySources.mockReturnValue(
+    defaultHookResult as ReturnType<typeof useCatalogModelsBySources>,
```
manaswinidas: ~90 lines of context/mock boilerplate for 3 small assertions on a static string. If this file is meant as a foundation for future `ModelCatalogGalleryView` tests, noting that intent in the PR description would help reviewers understand the investment.

manaswinidas (commented):
Core change looks good — unifying the label to "Reset all defaults" makes sense. A few concerns around test coverage: the tests are largely tautological (asserting a hardcoded string under different states that no longer affect it), duplicate across Cypress and unit layers, and the boilerplate-to-value ratio in the new spec file is high. See inline comments.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalog.cy.ts
ConorOM1: good point, trimmed to a single test

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogGalleryView.tsx
ConorOM1: done

clients/ui/frontend/src/app/pages/modelCatalog/screens/ModelCatalogSourceLabelSelector.tsx
ConorOM1: noted

clients/ui/frontend/src/app/pages/modelCatalog/screens/__tests__/ModelCatalogGalleryView.spec.tsx
ConorOM1: yes good point removed

clients/ui/frontend/src/app/pages/modelCatalog/screens/__tests__/ModelCatalogGalleryView.spec.tsx
ConorOM1: removing for now

Philip-Carneiro:
tested locally, it's ok, he fixed the comments made before as well
/lgtm

ppadti:
@ConorOM1 one quick question - when the performance toggle is turned off - shouldn't it be "Reset all filters"?
because we will then apply the basic filters and they don't have defaults 
My understanding was we should show "Reset all defaults" for performance filters because they have default value and for other basic filters it will be "Reset all filters"? Let me know if I am missing something here?

ConorOM1:
> @ConorOM1 one quick question - when the performance toggle is turned off - shouldn't it be "Reset all filters"? because we will then apply the basic filters and they don't have defaults My understanding was we should show "Reset all defaults" for performance filters because they have default value and for other basic filters it will be "Reset all filters"? Let me know if I am missing something here?

@ppadti yes maybe having both was intentional ? I thought it was cleaner to standardise it, because even with perf toggle on and filters selected we show "Reset all defaults", and when selected the basic filters are also reset.

ppadti:
> @ppadti yes maybe having both was intentional ? I thought it was cleaner to standardise it, because even with perf toggle on and filters selected we show "Reset all defaults", and when selected the basic filters are also reset.

I think it was intentional to have different wordings for reset. But the point you mentioning makes sense to me - when the performance toggle is on and basic filters is applied is confusing to see reset all defaults is confusing - Maybe its better to ask UX what they want and then continue?

ConorOM1:
@ppadti addressed UX feedback, ready for re-review

ppadti (changes requested):
Thanks @ConorOM1 
One minor comment, can we also use const in MCP as well in `McpCatalogSourceLabelSelector.tsx` for consistency?

ConorOM1:
good callout @ppadti , also updated 
image

ppadti (approved):
/lgtm
/approve

PR #2690, opened 2026-05-11
Proposal for AI Hub v1

pboyd:
## Description

Proposes graduating the Model Registry API from v1alpha3 to v1 and the Catalog API from v1alpha1 to v1, with the following breaking changes:

- Fix inconsistent path parameter casing (smashed lowercase -> snake_case)
- Enforce 400 Bad Request consistently for invalid orderBy fields
- Remove experiment tracking (no UI, no controller, no known adoption)

Includes a nine-month side-by-side deprecation period before alpha removal.

## How Has This Been Tested?
N/A

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [x] The commits have meaningful messages
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).

PR #2864, opened 2026-06-18
feat(catalog): add MCP server displayName support

jonburdo:
Allow MCP catalog entries to carry an optional displayName so that the backend can preserve a human-friendly label without changing existing BFF and frontend behavior.

## Description

Additional info - summarized with Cursor:

This backend-only `displayName` addition should be backward compatible for the current stack.

Why it is not breaking:
- The backend field is optional and uses `omitempty`, so if `displayName` is not populated it is omitted from the JSON response rather than emitted as `null`.
- The BFF decodes MCP catalog responses with Go `json.Unmarshal`, which ignores unknown JSON fields by default. So if the catalog backend starts returning `displayName`, the current BFF will not fail decoding; it will simply ignore the extra field until its model is updated.
- The frontend consumes MCP catalog data through the BFF, not directly from the catalog backend. Since the BFF currently drops unknown fields, the frontend continues to receive the same effective shape it receives today.
- Current MCP catalog UI code only reads known fields like `name`, `description`, `logo`, `provider`, and `version`; it does not do strict runtime schema matching that would break on an added response field.

Current implication:
- This change is safe to merge independently as backend/API groundwork.
- It does not make `displayName` user-visible yet. A follow-up BFF/frontend ticket is still needed to plumb the field through and render it.

Also included in this backend change:
- `displayName` length is validated to a max length of 255
- Free-text MCP server search now includes `displayName`, so once the field is populated it behaves like other user-facing descriptive fields.

## Merge criteria:

If you have UI changes

### DIFF
--- a/api/openapi/catalog.yaml
+++ b/api/openapi/catalog.yaml
@@ -1631,6 +1631,10 @@ components:
             name:
               type: string
               description: Human-readable MCP server name. Must be unique within a source.
+            displayName:
+              type: string
+              description: Optional human-friendly display label for this MCP server.
+              maxLength: 255
             source_id:
               type: string
               description: Catalog source that provides this server.
--- a/api/openapi/src/plugins/mcp.yaml
+++ b/api/openapi/src/plugins/mcp.yaml
@@ -165,6 +165,10 @@ components:
             name:
               type: string
               description: Human-readable MCP server name. Must be unique within a source.
+            displayName:
+              type: string
+              description: Optional human-friendly display label for this MCP server.
+              maxLength: 255
             source_id:
               type: string
               description: Catalog source that provides this server.
--- a/catalog/internal/catalog/mcpcatalog/providers.go
+++ b/catalog/internal/catalog/mcpcatalog/providers.go
@@ -21,6 +21,7 @@ import (
 
 const (
 	yamlMCPCatalogPathKey = "yamlCatalogPath"
+	maxMCPDisplayNameLen  = 255
 )
 
 // yamlMCPParameter represents a single input parameter for an MCP tool.
@@ -104,6 +105,7 @@ type yamlMCPSecurityIndicator struct {
 // yamlMCPServer represents an MCP server definition in YAML
 type yamlMCPServer struct {
 	Name                     string                              `yaml:"name"`
+	DisplayName              *string                             `yaml:"displayName,omitempty"`
 	ExternalID               *string                             `yaml:"externalId,omitempty"`
 	Description              *string                             `yaml:"description,omitempty"`
 	Provider                 *string                             `yaml:"provider,omitempty"`
@@ -262,6 +264,16 @@ func (yp *yamlMCPProvider) emit(ctx context.Context, path string, recordChan cha
 
 // ToMCPServerProviderRecord converts a yamlMCPServer to an MCPServerProviderRecord
 func (ys *yamlMCPServer) ToMCPServerProviderRecord() MCPServerProviderRecord {
+	if ys.DisplayName != nil && len(*ys.DisplayName) > maxMCPDisplayNameLen {
+		return MCPServerProviderRecord{
+			Error: fmt.Errorf(
+				"server %q displayName exceeds maximum length of %d characters",
+				ys.Name,
+				maxMCPDisplayNameLen,
+			),
+		}
+	}
+
 	attrs := &models.MCPServerAttributes{
 		Name:       &ys.Name,
 		ExternalID: ys.ExternalID,
@@ -287,6 +299,9 @@ func (ys *yamlMCPServer) ToMCPServerProviderRecord() MCPServerProviderRecord {
 	// Convert standard properties
 	properties := []mrmodels.Properties{}
 
+	if ys.DisplayName != nil {
+		properties = append(properties, mrmodels.NewStringProperty("displayName", *ys.DisplayName, false))
+	}
 	if ys.Description != nil {
 		properties = append(properties, mrmodels.NewStringProperty("description", *ys.Description, false))
 	}
--- a/catalog/internal/catalog/mcpcatalog/service/entity_mappings_mcp_server.go
+++ b/catalog/internal/catalog/mcpcatalog/service/entity_mappings_mcp_server.go
@@ -35,6 +35,7 @@ var mcpServerProperties = map[string]filter.PropertyDefinition{
 	"lastUpdateTimeSinceEpoch": {Location: filter.EntityTable, ValueType: filter.IntValueType, Column: "last_update_time_since_epoch"},
 	"source_id":                {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "source_id"},
 	"base_name":                {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "base_name"},
+	"displayName":              {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "displayName"},
 	"description":              {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "description"},
 	"provider":                 {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "provider"},
 	"license":                  {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "license"},
--- a/catalog/internal/catalog/mcpcatalog/service/mcp_server.go
+++ b/catalog/internal/catalog/mcpcatalog/service/mcp_server.go
@@ -22,6 +22,7 @@ var (
 	ErrBaseNameContainsAtSign = errors.New("base_name cannot contain '@' character")
 	ErrBaseNameEmpty          = errors.New("base_name cannot be empty")
 	ErrBaseNameTooLong        = errors.New("base_name exceeds maximum length of 255 characters")
+	ErrDisplayNameTooLong     = errors.New("displayName exceeds maximum length of 255 characters")
 	ErrVersionTooLong         = errors.New("version exceeds maximum length of 100 characters")
 	ErrVersionContainsAtSign  = errors.New("version cannot contain '@' character")
 )
@@ -71,6 +72,7 @@ func (r *MCPServerRepositoryImpl) Save(server models.MCPServer) (models.MCPServe
 	if attr != nil && attr.Name != nil {
 		baseName := strings.TrimSpace(*attr.Name)
 		version := extractVersionProperty(server)
+		displayName := extractDisplayNameProperty(server)
 
 		// Validate base_name
 		if baseName == "" {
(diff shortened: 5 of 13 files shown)


api/openapi/src/plugins/mcp.yaml
```diff
@@ -165,6 +165,10 @@ components:
             name:
               type: string
               description: Human-readable MCP server name. Must be unique within a source.
+            displayName:
+              type: string
+              description: Optional human-friendly display label for this MCP server.
+              maxLength: 256
```
jonburdo: We'll have 256 max length for mcp registry so we align on that

pboyd:
/lgtm
/approve

I'm not sure if you're ready to merge this, so..
/hold

catalog/internal/catalog/mcpcatalog/service/mcp_server.go
```diff
@@ -22,6 +22,7 @@ var (
 	ErrBaseNameContainsAtSign = errors.New("base_name cannot contain '@' character")
 	ErrBaseNameEmpty          = errors.New("base_name cannot be empty")
 	ErrBaseNameTooLong        = errors.New("base_name exceeds maximum length of 255 characters")
+	ErrDisplayNameTooLong     = errors.New("displayName exceeds maximum length of 256 characters")
```
pboyd: Can `displayName` have a max of 255 to match `base_name`?

Alternatively, since there's no technical reason that we need to cap this at a specific size, is 255/256 the best limit here? Maybe 100 or 200 if it's arbitrary?

jonburdo: Sure. The mlflow mcp registry uses 256 for a lot of fields (following the mlflow model registry's pattern), but we already have different limits here, so maybe it's better to be consistent within kubeflow mcp catalog. 255 will at least be closer to mlflow mcp registry's limit than 100 or 200 though

pboyd (commented):
/lgtm

catalog/internal/converter/mcp_converter.go
```diff
@@ -33,6 +33,7 @@ func ConvertOpenapiMCPServerToDb(openapiServer *openapi.MCPServer) models.MCPSer
 
 	// Simple string fields
 	addStringProperty(&properties, "source_id", openapiServer.SourceId)
+	addStringProperty(&properties, "displayName", openapiServer.DisplayName)
```
Philip-Carneiro, line 36: it should follow the other standard from the repo... above this line we have source_id, here should be display_name

for suggestion: 
```
addStringProperty(&properties, "display_name", openapiServer.DisplayName)
```

catalog/internal/converter/mcp_converter.go
```diff
@@ -167,6 +168,7 @@ func convertDbMCPServerToOpenapiInternal(dbServer models.MCPServer, tools []open
 
 	// Extract simple string properties
 	openapiServer.SourceId = pa.GetStringPtr("source_id")
+	openapiServer.DisplayName = pa.GetStringPtr("displayName")
```
Philip-Carneiro, line 171: following the comment above, should be display_name

catalog/internal/plugins/mcp/plugin.go
```diff
@@ -42,6 +42,7 @@ func (p *Plugin) DatastoreEntries() []plugin.DatastoreEntry {
 			Spec: datastore.NewSpecType(mcpcatalogservice.NewMCPServerRepository).
 				AddString("source_id").
 				AddString("base_name").
+				AddString("displayName").
```
Philip-Carneiro, line 45: display_name

catalog/internal/catalog/mcpcatalog/providers.go
```diff
@@ -287,6 +299,9 @@ func (ys *yamlMCPServer) ToMCPServerProviderRecord() MCPServerProviderRecord {
 	// Convert standard properties
 	properties := []mrmodels.Properties{}
 
+	if ys.DisplayName != nil {
+		properties = append(properties, mrmodels.NewStringProperty("displayName", *ys.DisplayName, false))
```
Philip-Carneiro, line 303: display_name

catalog/internal/catalog/mcpcatalog/providers.go
```diff
@@ -104,6 +105,7 @@ type yamlMCPSecurityIndicator struct {
 // yamlMCPServer represents an MCP server definition in YAML
 type yamlMCPServer struct {
 	Name                     string                              `yaml:"name"`
+	DisplayName              *string                             `yaml:"displayName,omitempty"`
```
Philip-Carneiro, line 108:
```
DisplayName              *string                             `yaml:"display_name,omitempty"`
```

catalog/internal/catalog/mcpcatalog/service/test_helpers_test.go
```diff
@@ -22,6 +22,7 @@ func testDatastoreSpec() *datastore.Spec {
 		AddContext(testMCPServerTypeName, datastore.NewSpecType(NewMCPServerRepository).
 			AddString("source_id").
 			AddString("base_name").
+			AddString("displayName").
```
Philip-Carneiro, line 25: ```			AddString("display_name").```

catalog/internal/catalog/mcpcatalog/service/mcp_server_test.go
```diff
@@ -880,6 +896,28 @@ func TestMCPServerRepository(t *testing.T) {
...
+			Attributes: &models.MCPServerAttributes{
+				Name: new("test-server"),
+			},
+			Properties: &[]dbmodels.Properties{
+				{
+					Name:        "source_id",
+					StringValue: new("test-source"),
+				},
+				{
+					Name:        "displayName",
```
Philip-Carneiro, line 911:
```
					Name:        "display_name",
```

catalog/internal/catalog/mcpcatalog/service/mcp_server_test.go
```diff
@@ -519,21 +519,37 @@ func TestMCPServerRepository(t *testing.T) {
...
+				Attributes: &models.MCPServerAttributes{
+					Name: new("display-name-server"),
+				},
+				Properties: &[]dbmodels.Properties{
+					{
+						Name:        "source_id",
+						StringValue: new("query-source-3"),
+					},
+					{
+						Name:        "displayName",
```
Philip-Carneiro, line 532:
```
						Name:        "display_name",
```

catalog/internal/catalog/mcpcatalog/service/entity_mappings_mcp_server.go
```diff
@@ -35,6 +35,7 @@ var mcpServerProperties = map[string]filter.PropertyDefinition{
 	"lastUpdateTimeSinceEpoch": {Location: filter.EntityTable, ValueType: filter.IntValueType, Column: "last_update_time_since_epoch"},
 	"source_id":                {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "source_id"},
 	"base_name":                {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "base_name"},
+	"displayName":              {Location: filter.PropertyTable, ValueType: filter.StringValueType, Column: "displayName"},
```
Philip-Carneiro, line 38: display_name

api/openapi/catalog.yaml
```diff
@@ -1631,6 +1631,10 @@ components:
             name:
               type: string
               description: Human-readable MCP server name. Must be unique within a source.
+            displayName:
```
Philip-Carneiro, line 1634: display_name

api/openapi/src/plugins/mcp.yaml
```diff
@@ -165,6 +165,10 @@ components:
             name:
               type: string
               description: Human-readable MCP server name. Must be unique within a source.
+            displayName:
```
Philip-Carneiro, line 168: display_name

Philip-Carneiro (changes requested):

Philip-Carneiro:
Reviewed the frontend impact of this change.

**Verdict: No frontend breakage.**

The `displayName` field is safely isolated by two layers:
1. **BFF layer**: The `McpServer` struct in `clients/ui/bff/internal/models/mcp_server_catalog.go` does not have a `DisplayName` field. Go's `json.Unmarshal` silently ignores unknown fields, so `displayName` is dropped at the BFF boundary.
2. **Frontend layer**: The TypeScript `McpServer` type in `src/app/mcpServerCatalogTypes.ts` doesn't declare `displayName`. Even if it arrived, it would be silently ignored (no runtime schema validation).

I ran `npm run test` on this branch — all 410 frontend tests pass (unit + Cypress).

Bonus: the free-text search now includes `displayName`, which means frontend search results will be richer once the field is populated — without any frontend code change needed.

**One naming note**: The follow-up changes references `display_name` (snake_case) in the description, but this PR correctly uses `displayName` (camelCase) which matches the dominant convention (92.9% of MCPServer fields). There is a pre-existing inconsistency with `source_id` (snake_case), and the BFF already translates `license_link` -> `licenseLink`. When implementing the next changes, the BFF and frontend should use `displayName` (camelCase) to stay consistent.


After fix the changes I requested is safe to merge without impact the frontend

pboyd:
@Philip-Carneiro, `displayName` is more correct, in my opinion. The original plan was for path parameters to use `snake_case` and body fields to use `camelCase`. It wasn't applied consistently however, and now we have `source_id` as a sibling of `toolCount`. It's a real mess right now, but it's one of the things we're hoping to clean up with v1 of the API: [PR 2690](https://github.com/kubeflow/hub/pull/2690).

catalog/internal/converter/mcp_converter.go
jonburdo: I had originally started to use `display_name` but switched to `displayName` intentionally, because I think (at least at the API level) it will be more consistent for admin users creating yaml input. Unfortunately we ended up with a mix of snakecase (`mcp_servers`, `license_link`, `source_id`) and camelcase (most fields on an mcp server object).

But we have a proposal to converge on camelcase for response body fields: https://github.com/kubeflow/hub/pull/2690/changes#diff-72890effc6d308d960b64d1f53b48d134be2bd2e8dbd26ecddc388732c46ab58R52

We do want to pass off to mlflow mcp registry (being developed) which uses snakecase, but still I think we want to be internally consistent.

Example yaml:
```yaml
source: My Servers
mcp_servers:
  - name: my-mcp-server
    provider: My Organization
    license: apache-2.0
    license_link: https://www.apache.org/licenses/LICENSE-2.0
    description: A simple MCP server
    version: "1.0.0"
    transports:
      - http
    documentationUrl: https://docs.example.com/mcp
    repositoryUrl: https://github.com/my-org/my-mcp-server
    sourceCode: my-org/my-mcp-server
    publishedDate: "2025-03-01"
    securityIndicators:
      verifiedSource: true
      secureEndpoint: true
      sast: true
      readOnlyTools: true
    runtimeMetadata:
      defaultPort: 8080
      mcpPath: /mcp
    createTimeSinceEpoch: "1740787200000"
    lastUpdateTimeSinceEpoch: "1740787200000"
```

Philip-Carneiro:
Thanks for the clarification @jonburdo and @pboyd. I see now — the direction is camelCase for body fields (as formalized in PR #2690 's KEP-0004), and displayName aligns with that. My inline comments requesting display_name were based on matching source_id's current snake_case, but since source_id itself will be migrated to sourceId in v1, it makes sense to start clean with displayName now. I'll dismiss my changes requested review. The frontend impact analysis still holds — this is safe to merge from the UI perspective.

jonburdo:
Awesome, thanks @Philip-Carneiro !

/unhold

pboyd (commented):
/lgtm

PR #2886, opened 2026-06-25
chore(catalog): replace Postgres PVC with emptyDir

pboyd:
## Description
The catalog's Postgres database is an ephemeral cache, not a source of truth. On every leader election the leader pod reloads all data from YAML ConfigMaps and external APIs (e.g. HuggingFace), so the PVC preserves data that would be regenerated anyway.

Replace the PVC volume with an emptyDir in the StatefulSet, remove the PVC manifest and its kustomization entry, and clean up the Tilt config (catalog-pvc resource group and related resource_deps).

Docker Compose is unchanged because the shared postgres_data volume is also used by the model registry, which is stateful.

## How Has This Been Tested?
On a local dev cluster.

## Merge criteria:


pboyd:
There's a problem recovering from database pod restart now. Hold off on merging for now.

/hold

pboyd:
This is ready now.

/unhold

Al-Pragliola:
It looks like the catalog e2e tests are failing, this is happening consistently

pboyd:
@Al-Pragliola, they're passing now. I had been testing the case where it recovers from a database restart and neglected to test the more common empty start case.

jonburdo (commented):
Surfaced with cursor:

After the DB is wiped and `RunMigrations()` recreates the type rows, the process keeps using the `RepoSet` and plugin repositories that were initialized before the wipe. Those repos cache numeric `type_id` values at startup, but the recreated type IDs are not guaranteed to match the old ones, since `syncTypes()` recreates them from Go map iteration order.

I was able to reproduce this locally with a focused test: after truncating the type tables, rerunning migrations, and reusing the old repo, it still wrote rows with the stale pre-wipe `type_id`. That means the leader can recover the schema but still serve incorrect reads/writes until the pod restarts. It seems like the recovery path needs to rebuild `repoSet` / plugin services after schema recreation, or otherwise avoid caching raw type IDs across DB loss.

pboyd:
That's a good catch, thanks @jonburdo. I pushed another commit which re-builds the type IDs (and also sorts them before creating them so hopefully the IDs don't actually change).

Question for you and @Al-Pragliola.. the actual change for this PR is very simple, but recovering from the database loss is proving to be rather difficult. My original plan was to detect that the database was empty and exit, letting kubernetes restart the pod. That worked fine, everywhere except under Tilt, which detected the pod restart then updated the binary and killed the original process. It was often killing the process during the database migrations, which left it dirty and (since we don't have any mechanism to automatically recover from a failed migration) left my dev environment broken until I manually fixed it. So, given how complicated the last few commits are, is it worth it? Or should we switch to a much simpler exit and deal with manually fixing dev environments after a database restart?

jonburdo:
/lgtm
/approve

fwiw, cursor's review of latest commit. No changes need imo:

> No concrete findings.
> 
> Residual risk: the new commit appears to address the stale type_id issue correctly by both making syncTypes deterministic and refreshing plugin repositories via Reconnect before NotifyLeader. The main remaining gap is test coverage: there still isn’t an end-to-end Postgres recovery test for the full live path (RunMigrations -> Reconnect -> NotifyLeader -> API traffic), so that specific sequence is still only indirectly validated.

PR #2904, opened 2026-07-01
fix(async-upload): avoid auth override with credentials_path

jonburdo:
When async-upload passes an OCI authfile through to skopeo, stop also forwarding inline OCI credentials so save_to_oci_registry does not synthesize a competing src authfile. Keep tests focused on the authfile path taking precedence only when configured.

## Merge criteria:

If you have UI changes

### DIFF
--- a/jobs/async-upload/job/upload.py
+++ b/jobs/async-upload/job/upload.py
@@ -56,6 +56,8 @@ def _get_upload_params(config: AsyncUploadConfig) -> S3Params | OCIParams:
     elif isinstance(destination_config, OCIStorageConfig):
         push_args = []
         pull_args = []
+        username = destination_config.username
+        password = destination_config.password
         # Note: These are all skopeo args, see: https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md
         if not destination_config.enable_tls_verify:
             push_args.append("--dest-tls-verify=false")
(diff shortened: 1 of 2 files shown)


jonburdo:
Small change following https://github.com/kubeflow/hub/pull/2863

Auth was still being injected a second way by the Python client underneath.

Before this fix, if `destination-oci-credentials-path` was set, async-upload would do two things at once:

1. Pass the file directly via raw skopeo args:
   - pull: `--authfile /path/to/creds`
   - push: `--authfile /path/to/creds`

2. Also pass `oci_username` / `oci_password` into `save_to_oci_registry()`
   - those were loaded earlier from the same credentials file in `job/config.py`
   - then `clients/python/src/model_registry/utils.py` would synthesize a temp authfile
   - and convert that into:
     - pull: `--src-authfile /tmp/...json`
     - push: `--dest-authfile /tmp/...json`

So the effective args were roughly:

For pull:
```bash
--src-authfile /tmp/generated-auth.json --src-tls-verify=false --authfile /mounted/credentials.json
```

For push:
```bash
--dest-authfile /tmp/generated-auth.json --dest-tls-verify=false --authfile /mounted/credentials.json
```

That’s the “doubling up” we were talking about:
- generic `--authfile` from async-upload
- source/dest-specific authfile from the client layer

And the reason it mattered is that `--src-authfile` / `--dest-authfile` are the more specific flags, so they can override the generic `--authfile` path that `#2863` was trying to introduce.

This follow-up fixes that by saying: if async-upload is explicitly passing a `credentials_path` through to skopeo, it should **not also pass `oci_username` / `oci_password` down**, so the lower layer won’t generate the extra authfile.

pboyd (approved):
/lgtm

PR #2888, opened 2026-06-26
Add MCP catalog settings page and add/manage source subpage

Philip-Carneiro:
## Description

Adds the MCP catalog settings page to the navigation and routing, following the same structure as the existing Model catalog settings page.

### What's included:
  - New "MCP catalog settings" navigation item, placed below "Model catalog settings" in the Settings section
  - Main settings page (/mcp-catalog-settings) with an empty state and "Add a source" button
  - Manage source subpage at /mcp-catalog-settings/add-source and /mcp-catalog-settings/manage-source/:sourceId with breadcrumb navigation
  - McpCatalogSettingsContext provider as a placeholder for future data integration
  - Route utilities and URL helper functions
  - Unit tests covering component rendering, navigation behavior, routing and redirects (9 test cases)

### What's NOT included (intentionally):

  - No API integration or data fetching — this is a scaffold for future implementation
  - No table view — will be added when data consumption is implemented
  - No form in the manage source subpage — placeholder only

**Pattern reference:** All components follow the same conventions as ModelCatalogSettings, including TitleWithIcon with ProjectObjectType.mcpCatalog, ApplicationsPage layout, and route gating under isStandalone || isFederated.

Non-PF mode:
image
image

PF mode: 
image
image

## How Has This Been Tested?

It was manually tested and raised simple page tests

## Merge criteria:

- All the commits have been [_signed-off_](https://github.com/kubeflow/community/tree/master/dco-signoff-hook#signing-off-commits)  (To pass the `DCO` check)

- [x] The commits have meaningful messages
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.
- [x] Code changes follow the [kubeflow contribution guidelines](https://www.kubeflow.org/docs/about/contributing/).
- [ ] **For first time contributors**: Please reach out to the [Reviewers](https://github.com/kubeflow/hub/blob/main/OWNERS) to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR.

If you have UI changes

- [x] The developer has added tests or explained why testing cannot be added.
- [x] Included any necessary screenshots or gifs if it was a UI change.
- [ ] Verify that UI/UX changes conform the UX guidelines for Kubeflow.

PR #2908, opened 2026-07-03
Create Agents catalog 

Philip-Carneiro:
## Description

Created the base skeleton for the Agents catalog page.
image

## How Has This Been Tested?

It was manually tested, also created base tests for the new page

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalog.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalog.cy.ts
@@ -1,6 +1,11 @@
 import { agentsCatalog, agentDetailsPage } from '~/__tests__/cypress/cypress/pages/agentsCatalog';
+import { initAgentsCatalogIntercepts } from './agentsCatalogTestUtils';
 
 describe('Agents Catalog Page', () => {
+  beforeEach(() => {
+    initAgentsCatalogIntercepts();
+  });
+
   it('Agents Catalog tab should be enabled in nav', () => {
     agentsCatalog.visit();
     agentsCatalog.tabEnabled();
@@ -8,6 +13,10 @@ describe('Agents Catalog Page', () => {
 });
 
 describe('Agent Details Page', () => {
+  beforeEach(() => {
+    initAgentsCatalogIntercepts();
+  });
+
   it('should display breadcrumb with link back to catalog', () => {
     agentDetailsPage.visit('my-agent');
     agentDetailsPage.findBreadcrumbCatalogLink().should('be.visible');
--- a/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalog.cy.ts
+++ b/clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalog.cy.ts
@@ -1,0 +1,31 @@
+import { agentsCatalog, agentDetailsPage } from '~/__tests__/cypress/cypress/pages/agentsCatalog';
+
+describe('Agents Catalog Page', () => {
+  it('Agents Catalog tab should be enabled in nav', () => {
+    agentsCatalog.visit();
+    agentsCatalog.tabEnabled();
+  });
+});
+
+describe('Agent Details Page', () => {
+  it('should display breadcrumb with link back to catalog', () => {
+    agentDetailsPage.visit('my-agent');
+    agentDetailsPage.findBreadcrumbCatalogLink().should('be.visible');
+  });
+
+  it('should display agent name in breadcrumb', () => {
+    agentDetailsPage.visit('my-agent');
+    agentDetailsPage.findBreadcrumbAgentName().should('contain', 'my-agent');
+  });
+
+  it('breadcrumb link should navigate back to catalog', () => {
+    agentDetailsPage.visit('test-agent');
+    agentDetailsPage.findBreadcrumbCatalogLink().click();
+    agentsCatalog.findPageTitle().should('be.visible');
+  });
+
+  it('should redirect from :agentName to :agentName/overview', () => {
+    cy.visit('/agents-catalog/test-agent');
+    cy.url().should('include', '/agents-catalog/test-agent/overview');
+  });
+});
--- a/clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentDetailsPage.tsx
+++ b/clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentDetailsPage.tsx
@@ -1,0 +1,34 @@
+import * as React from 'react';
+import { useParams } from 'react-router';
+import { Link } from 'react-router-dom';
+import { Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
+import { ApplicationsPage } from 'mod-arch-shared';
+import { agentsCatalogUrl } from '~/app/routes/agentsCatalog/agentsCatalog';
+
+const AgentDetailsPage: React.FC = () => {
+  const { agentName: rawAgentName = '' } = useParams<{ agentName: string }>();
+  const agentName = decodeURIComponent(rawAgentName);
+
+  return (
+    <ApplicationsPage
+      breadcrumb={
+        <Breadcrumb>
+          <BreadcrumbItem>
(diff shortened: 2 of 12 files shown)


clients/ui/frontend/src/app/routes/agentsCatalog/agentsCatalog.ts
```diff
@@ -0,0 +1,10 @@
+export const agentsCatalogUrl = (sourceId?: string): string =>
+  `/agents-catalog${sourceId ? `/${sourceId}` : ''}`;
```
manaswinidas: **Dead code:** `sourceId` produces paths like `/agents-catalog/src-1`, but `AgentsCatalogRoutes.tsx` has no `:sourceId` segment — so these paths will be incorrectly captured by the `:agentName` param, causing a subtle routing bug.

Either remove `sourceId` here (and from `getAgentsCatalogDetailsRoute` below), or add a `:sourceId/:agentName` route to `AgentsCatalogRoutes.tsx`.

clients/ui/frontend/src/app/routes/agentsCatalog/__tests__/agentsCatalog.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  it('should encode special characters in agent name', () => {
+    expect(getAgentsCatalogDetailsRoute({ agentName: 'agent/with/slashes' })).toBe(
+      '/agents-catalog/agent%2Fwith%2Fslashes',
+    );
+  });
+
+  it('should include sourceId in path when provided', () => {
+    expect(getAgentsCatalogDetailsRoute({ sourceId: 'src-1', agentName: 'my-agent' })).toBe(
+      '/agents-catalog/src-1/my-agent',
+    );
```
manaswinidas: **Dead test:** This asserts the path `/agents-catalog/src-1/my-agent`, but no route in `AgentsCatalogRoutes.tsx` will ever match this shape (there's no `:sourceId/:agentName` route). This test passes but validates dead code — remove or add the corresponding route.

clients/ui/frontend/src/app/routes/agentsCatalog/__tests__/agentsCatalog.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { getAgentsCatalogDetailsRoute } from '~/app/routes/agentsCatalog/agentsCatalog';
+
+describe('getAgentsCatalogDetailsRoute', () => {
+  it('should encode agent name with spaces', () => {
+    expect(getAgentsCatalogDetailsRoute({ agentName: 'agent with spaces' })).toBe(
+      '/agents-catalog/agent%20with%20spaces',
+    );
```
manaswinidas: **Tautological:** These first two tests (`'should encode agent name with spaces'` and `'should encode special characters'`) are testing that `encodeURIComponent` (a platform built-in) works. A single case would suffice — consider testing something application-specific instead.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/agentsCatalog/agentsCatalog.cy.ts
```diff
@@ -0,0 +1,31 @@
+import { agentsCatalog, agentDetailsPage } from '~/__tests__/cypress/cypress/pages/agentsCatalog';
+
+describe('Agents Catalog Page', () => {
```
manaswinidas, line 4: **Maintenance:** Unlike the MCP catalog tests that use `beforeEach(() => { initMcpCatalogIntercepts(); })`, these tests have no API intercepts. They only work because the page renders static "coming soon" text. When real data-fetching is added, all tests will break simultaneously.

Add a `beforeEach` with stub intercepts now (even if they return empty data) so the test structure is ready for the feature implementation.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalog.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { ApplicationsPage } from 'mod-arch-shared';
+import { AGENTS_CATALOG_TITLE, AGENTS_CATALOG_DESCRIPTION } from '~/app/pages/agentsCatalog/const';
+
+// ponytail: no agent-specific ProjectObjectType exists yet — add TitleWithIcon when one is available
```
manaswinidas, line 5: **Maintenance:** Use a standard `// TODO(#issue-number):` format instead of a nickname (`ponytail`). This makes it discoverable via issue searches and assignable to track resolution.

manaswinidas (commented):
Thanks for the scaffolding! A few issues around dead code, test validity, and maintenance that should be addressed before merge.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentDetailsPage.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import { useParams } from 'react-router';
+import { Link } from 'react-router-dom';
+import { Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
+import { ApplicationsPage } from 'mod-arch-shared';
+import { agentsCatalogUrl } from '~/app/routes/agentsCatalog/agentsCatalog';
+
+const AgentDetailsPage: React.FC = () => {
+  const { agentName: rawAgentName = '' } = useParams<{ agentName: string }>();
+  const agentName = decodeURIComponent(rawAgentName);
```
manaswinidas: **Nit (perf):** `decodeURIComponent` is called on every render. Wrap in `useMemo(() => decodeURIComponent(rawAgentName), [rawAgentName])` to match how the MCP catalog handles it.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalog.tsx
manaswinidas: ignore this

manaswinidas:
Thanks for the scaffolding! Two structural gaps that should be addressed before merge — both are patterns established in the MCP settings scaffolding (#2888) and shared by every existing catalog in the codebase:

### 1. Missing Context Provider

Both existing catalogs wrap their routes in a Context Provider — even when it's empty:

- `McpCatalogRoutes` → `McpCatalogContextProvider`
- `ModelCatalogRoutes` → `ModelCatalogContextProvider`
- `McpCatalogSettingsRoutes` (from #2888) → `McpCatalogSettingsContextProvider` (literally `Record<string, never>`)

`AgentsCatalogRoutes` currently uses a bare `<Outlet />`. Please add an empty `AgentsCatalogContextProvider` (same pattern as `McpCatalogSettingsContext.tsx` from #2888) and wrap the route tree with it.

### 2. Missing Core Loader wrapper

Both catalog pages use a Core Loader component that handles loading/error/empty states before rendering child routes:

- `McpCatalogRoutes` → `McpCatalogCoreLoader`
- `ModelCatalogRoutes` → `ModelCatalogCoreLoader`

This can be a minimal pass-through for now (just render `<Outlet />`), but having the component in place avoids restructuring the route tree later.

---

**Why this matters for the next piece of work:** The upcoming feature flag + midstream routes integration will register this catalog in ODH extensions. The extension imports the Routes component directly — if it already wraps a context provider (matching the MCP/Model pattern), the ODH wrapper integration is straightforward. Without it, the downstream ODH wrapper will need to provide context externally, diverging from how every other catalog works and requiring a structural refactor of this file.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentDetailsPage.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const AgentDetailsPage: React.FC = () => {
+  const { agentName = '' } = useParams<{ agentName: string }>();
+
+  return (
+    <ApplicationsPage
+      breadcrumb={
+        <Breadcrumb>
+          <BreadcrumbItem>
+            <Link to={agentsCatalogUrl()}>Agents Catalog</Link>
```
manaswinidas, line 16: Lots of areas like this are still using hard coded strings instead of AGENTS_CATALOG_TITLE - but I won't hold up the PR for this - this is for the next PR.

manaswinidas (approved):
/lgtm
/approve

PR #2911, opened 2026-07-04
fix(csi): reject malformed model-registry URIs

immanuwell:
Bad `storageUri` values like `model-registry://`, `model-registry:///iris`, and `model-registry://iris/` slipped through CSI parsing and failed later in the flow. Kinda annoying to debug.

This patch rejects empty model and empty version path segments up front, so the CSI returns `ErrInvalidMRURI` right away.

Repro:
1. `cd cmd/csi`
2. `go test ./internal/storage -run TestParseModelVersion -v -count=1`
3. Check the malformed URI cases in the test table:
   `model-registry://`
   `model-registry:///iris`
   `model-registry://localhost:8080/`
   `model-registry://iris/`

Testing:
`go test ./... -count=1`

### DIFF
--- a/cmd/csi/internal/storage/modelregistry_provider.go
+++ b/cmd/csi/internal/storage/modelregistry_provider.go
@@ -141,8 +141,14 @@ func (p *ModelRegistryProvider) parseModelVersion(storageUri string) (string, *s
 	}
 
 	registeredModelName := tokens[0]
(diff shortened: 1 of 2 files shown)


pboyd (commented):
Thanks for the fix!

/lgtm
/approve

PR #2928, opened 2026-07-07
feat(catalog): add artifacts endpoint and template artifact type to agent plugin

Al-Pragliola:
## Description

- Adds `GET /api/agent_catalog/v1alpha1/agents/{id}/artifacts` endpoint with `artifactType` query param and pagination
- Introduces a `template-artifact` type that stores agent.yaml deployment specs as JSON-encoded strings, persisted as a first-class entity in the artifacts table (`kf.AgentTemplateArtifact`)
- The `/artifacts` endpoint serves only DB-stored artifacts; image artifacts remain inline on the Agent entity via `GET /agents/{id}`, matching how the model catalog separates inline data (`artifactCounts`) from the dedicated artifacts endpoint
- Loader saves template artifacts after the parent agent, using the MCP tool save pattern (save parent → get ID → delete old children → save new)
- Template name defaults to `agent.yaml` when not specified in YAML source data
- Returns 404 for nonexistent agents, matching the MCP catalog's behavior
- `artifactType=image-artifact` is rejected with 400 since image artifacts are not served from this endpoint
- Fixes codegen issues: `model.[]Enum` → `[]model.Enum` in api.go interface, `type_asserts` overrides for polymorphic `AgentArtifact`

## How Has This Been Tested?

- [x] `go build ./catalog/...` passes
- [x] `go test ./catalog/...` all pass
- [x] Tested on Tilt with demo data — verified:
  - `GET /agents/{id}/artifacts` returns template artifacts with proper pagination
  - `?artifactType=template-artifact` filters correctly
  - `?artifactType=image-artifact` returns 400 (not a valid type for this endpoint)
  - Nonexistent agent returns 404
  - Agent without templates returns empty list
  - Image artifacts still accessible inline via `GET /agents/{id}`

## Merge criteria:
- [x] All the commits have been signed-off
- [x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body
- [x] Code changes follow the kubeflow contribution guidelines
- [ ] For first time contributors: Please reach out to the Reviewers to ensure all tests are being run, ensuring the label `ok-to-test` has been added to the PR
- [x] The developer has added tests or explained why testing cannot be added


catalog/internal/catalog/agentcatalog/db_agent.go
```diff
@@ -148,6 +150,93 @@ func (d *DBAgentCatalog) GetAgent(ctx context.Context, agentID string) (*openapi
 	return &apiAgent, nil
 }
 
+func (d *DBAgentCatalog) GetAgentArtifacts(ctx context.Context, agentID string, artifactType []openapi.AgentArtifactTypeQueryParam, pageSize int32, orderBy openapi.OrderByField, sortOrder openapi.SortOrder, nextPageToken *string) (openapi.AgentArtifactList, error) {
```
lugi0, line 153: GetAgentArtifacts accepts pageSize, orderBy, sortOrder, and nextPageToken but ignores all of them. The response always returns every artifact, with pageSize simply echoed back and nextPageToken left empty. This violates the API contract defined in the OpenAPI spec. For now this may be acceptable if datasets are small, but it should at minimum be documented or the params should be removed from the endpoint until implemented.

catalog/internal/catalog/agentcatalog/db_agent.go
```diff
@@ -148,6 +150,93 @@ func (d *DBAgentCatalog) GetAgent(ctx context.Context, agentID string) (*openapi
...
+			ParentResourceID: &parentID,
+		}
+		templateList, err := d.agentTemplateArtifactRepo.List(listOpts)
+		if err != nil {
+			return openapi.AgentArtifactList{}, err
+		}
+		for _, tmpl := range templateList.Items {
+			items = append(items, mapDBTemplateArtifactToAPI(tmpl))
+		}
+	}
```
lugi0: When artifactType=template-artifact, the GetAgent() call is skipped entirely. The template artifact query uses the agent ID as context_id in the Attribution table — if the agent doesn't exist, it returns an empty list with 200 instead of 404. When artifactType=image-artifact or no filter is given, GetAgent() is called first, which correctly returns 404. This inconsistency could confuse API consumers.

  Suggested fix: validate agent existence at the top of GetAgentArtifacts regardless of the filter, or at least when templates-only is requested.

catalog/internal/catalog/agentcatalog/service/agent_template_artifact.go
```diff
@@ -0,0 +1,180 @@
...
+	return r.GenericRepository.List(&listOptions)
+}
+
+func (r *AgentTemplateArtifactRepositoryImpl) DeleteByParentID(parentID int32) error {
+	config := r.GetConfig()
+	return config.DB.Exec(
+		`DELETE FROM "Artifact" WHERE id IN (SELECT artifact_id FROM "Attribution" INNER JOIN "Artifact" ON "Artifact".id = artifact_id WHERE context_id = ? AND type_id = ?)`,
+		parentID, config.TypeID,
+	).Error
+}
```
lugi0, line 83: The MCP plugin's equivalent (mcp_server_tool.go:140-161) uses GORM's Table(), Select(), Joins(), and Where() with utils.GetTableName() for proper table name resolution. The raw SQL approach here hardcodes table names with double-quotes (PostgreSQL-specific quoting), making it fragile across database backends. Consider rewriting to match the MCP pattern.

catalog/pkg/openapi/model_agent_artifact.go
```diff
@@ -12,147 +12,108 @@ package openapi
...
+		if err == nil {
+			return nil // data stored in dst.AgentTemplateArtifact, return on the first match
+		} else {
+			dst.AgentTemplateArtifact = nil
+			return fmt.Errorf("failed to unmarshal AgentArtifact as AgentTemplateArtifact: %s", err.Error())
+		}
 	}
-	return o.LastUpdateTimeSinceEpoch, true
+
+	return nil
```
lugi0, line 72: If artifactType is missing or has an unrecognized value, UnmarshalJSON returns nil without error, producing a zero-value AgentArtifact (both union pointers nil). This is generated code, so it's not directly fixable here, but worth noting — the type_asserts overrides in type_asserts_overrides.go correctly handle validation as no-ops for the union type.

lugi0 (commented):
There are no tests for:
  - GetAgentArtifacts (filtering logic, empty results, mixed artifact types)
  - mapDBTemplateArtifactToAPI (field mapping correctness)
  - yamlTemplateToEntity (default name, qualified name construction)
  - DeleteByParentID (cascading delete correctness)
  - Loader's template-saving flow (save parent -> delete children -> save new)

  The MCP plugin has integration tests for its equivalent patterns (mcp_server_tool_test.go). At minimum, the repository and mapping functions should have tests.

catalog/internal/catalog/agentcatalog/db_agent.go
Al-Pragliola: fixed in the latest commit https://github.com/kubeflow/hub/pull/2928/commits/c5dcff6608221d1417bec18bcebd9ef2bcef989b

catalog/internal/catalog/agentcatalog/db_agent.go
Al-Pragliola: fixed in the latest commit https://github.com/kubeflow/hub/pull/2928/commits/c5dcff6608221d1417bec18bcebd9ef2bcef989b

catalog/internal/catalog/agentcatalog/service/agent_template_artifact.go
Al-Pragliola: the one from this PR follows the model catalog approach

catalog/pkg/openapi/model_agent_artifact.go
Al-Pragliola: ack

Al-Pragliola:
> There are no tests for:
> 
>     * GetAgentArtifacts (filtering logic, empty results, mixed artifact types)
> 
>     * mapDBTemplateArtifactToAPI (field mapping correctness)
> 
>     * yamlTemplateToEntity (default name, qualified name construction)
> 
>     * DeleteByParentID (cascading delete correctness)
> 
>     * Loader's template-saving flow (save parent -> delete children -> save new)
> 
> 
> The MCP plugin has integration tests for its equivalent patterns (mcp_server_tool_test.go). At minimum, the repository and mapping functions should have tests.

Will add test and improve coverage in a dedicated follow up PR

catalog/internal/catalog/agentcatalog/db_agent.go
lugi0: The new commit wires pageSize, orderBy, sortOrder, nextPageToken through to the template artifact repository query. However, there are remaining issues:

  - Pagination only applies to template artifacts, not image artifacts. Image artifacts come from the inline JSON blob on the agent (via GetAgent) and are always returned in full. When both types are requested (no filter), the response merges unpaginated images with paginated templates — the pageSize constraint isn't applied to the combined result.
  - nextPageToken from the template list is not propagated to the AgentArtifactList response. The return statement still just sets Size: int32(len(items)) and PageSize: pageSize with no NextPageToken. So even though templates are queried with pagination, the client can never page to the next set.

  This means pagination works at the DB query level for templates but the API response doesn't actually expose it.

Al-Pragliola: changed image artifacts to be inline only and let only the DB artifacts to live on the /artifacts endpoint so that the pagination works correctly without hacks

lugi0 (approved):
LGTM

PR #2932, opened 2026-07-08
Add Agents Catalog Foundation

Philip-Carneiro:
## Description

Implemented the complete data foundation layer for the Agents Catalog frontend. 
Followed MCP Catalog pattern.

- Non-PF
Screenshot 2026-07-08 at 10 50 49

- PF
Screenshot 2026-07-08 at 10 51 07

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentFilterOptionList.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  const mapped: AgentsCatalogFilterOptionsList['filters'] = {};
+  for (const [key, value] of Object.entries(raw.filters)) {
+    const frontendKey = BACKEND_TO_FRONTEND_AGENT_FILTER_KEY[key] ?? key;
+    if (isAgentFilterCategoryKey(frontendKey) && isFilterStringOption(value)) {
+      mapped[frontendKey] = value;
+    }
+  }
+  return { filters: mapped };
+}
```
manaswinidas, line 41: `mapBackendFilterOptions` is the only function in this PR with real branching logic (key remapping, type guards via `isFilterStringOption` using `Object.getOwnPropertyDescriptor`), yet it has zero test coverage.

Please add a unit test covering at minimum:
- Backend key remapped correctly (`tags` → `labels`)
- Unknown keys are dropped
- Non-string filter options are dropped
- `null`/`undefined` filters input returns `{ filters: undefined }`

This is where silent failures are most likely — a bug here means the filter panel renders empty with no visible error.

clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentsBySourceLabel.ts
```diff
@@ -0,0 +1,177 @@
...
+    setAllItems([]);
+    setTotalSize(0);
+    setNextPageTokenValue('');
+    setIsLoadingMore(false);
+    setLoadMoreError(undefined);
+    refetch();
+  }, [refetch]);
+
+  const paginatedData: PaginatedAgentList = {
+    items: allItems,
```
manaswinidas: This object literal is recreated on every render without `useMemo`, producing a new reference each time. Downstream consumers (e.g., `CatalogGalleryLayout`) receiving this as a prop will re-render unnecessarily.

Wrap in `useMemo` with appropriate deps, or restructure to return stable references.

clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentsBySourceLabel.ts
```diff
@@ -0,0 +1,177 @@
...
+      setLoadMoreError(undefined);
+    } catch (err) {
+      setLoadMoreError(
+        new Error(
+          `Failed to load more agents: ${err instanceof Error ? err.message : String(err)}`,
+        ),
+      );
+    } finally {
+      setIsLoadingMore(false);
+    }
```
manaswinidas, line 123: Including `isLoadingMore` in the dependency array means `loadMore`'s identity changes on every loading state toggle. When passed as a prop to gallery components, this triggers child re-renders.

Consider using a ref (`useRef`) for the guard check instead:
```ts
const isLoadingMoreRef = React.useRef(false);
```

clients/ui/frontend/src/app/agentsCatalogTypes.ts
```diff
@@ -0,0 +1,59 @@
...
+  env?: AgentEnvVar[];
+  artifacts?: AgentArtifact[];
+  customProperties?: Record<string, unknown>;
+  createTimeSinceEpoch?: string;
+  lastUpdateTimeSinceEpoch?: string;
+};
+
+export type AgentList = PaginationParams & { items?: Agent[] };
+
+export type AgentListParams = {
```
manaswinidas, line 38: nit: `pageSize` is typed as `number | string` but the service layer always does `String(listParams.pageSize)` internally. This leaks a serialization concern into the domain type. Consider typing as just `number`.

clients/ui/frontend/src/__mocks__/mockAgentsCatalogTestData.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const mockAgentList = (partial?: Partial<AgentList>): AgentList => ({
+  items: [mockAgent()],
+  pageSize: 10,
+  size: 1,
+  nextPageToken: '',
+  ...partial,
+});
+
+export const mockAgents: Agent[] = [
```
manaswinidas: nit: `mockAgents` is exported but never imported — not in this PR, and not in #2934 (which introduces `mockAgentsGallery` instead). Consider removing to avoid dead code, or aligning with the follow-up PR's naming.

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
```diff
@@ -1,3 +1,16 @@
...
   'Discover agents that are available for your organization.';
+
+export const AGENT_FILTER_KEYS: AgentFilterCategoryKey[] = ['framework', 'labels'];
+
+export const AGENT_FILTER_CATEGORY_NAMES: Record<AgentFilterCategoryKey, string> = {
+  framework: 'Framework',
+  labels: 'Labels',
+};
+
+export const BACKEND_TO_FRONTEND_AGENT_FILTER_KEY: Record<string, AgentFilterCategoryKey> = {
```
manaswinidas: This PR defines `AgentFilterCategoryKey = 'framework' | 'labels'` with a `tags → labels` mapping, but #2934 immediately replaces this with `'framework' | 'category' | 'communicationProtocol' | 'testedModels'` and empties the mapping. Consider aligning with the final schema now to avoid throwaway code.

manaswinidas (changes requested):
Good foundation work — follows the MCP catalog pattern consistently. Requesting changes for one blocking item:

**Blocking:** `mapBackendFilterOptions` needs a unit test. It's the only function with real branching logic and no coverage — a bug here silently breaks the filter panel with no visible error.

**Non-blocking suggestions** (can address in follow-up):
- `paginatedData` object should be memoized to avoid unnecessary re-renders
- `isLoadingMore` in `loadMore` deps causes identity churn — use a ref
- Align filter keys with #2934's final schema to avoid throwaway code
- Remove unused `mockAgents` export
- Type `pageSize` as `number` only

clients/ui/frontend/src/app/context/agentsCatalog/__tests__/AgentsCatalogContext.spec.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  it('updates filters via setFilters', () => {
+    const { result } = renderHook(() => React.useContext(AgentsCatalogContext), { wrapper });
+    act(() => {
+      result.current.setFilters({ framework: ['LangGraph'] });
+    });
+    expect(result.current.filters).toEqual({ framework: ['LangGraph'] });
+    act(() => {
+      result.current.setFilters((prev) => ({ ...prev, labels: ['research'] }));
+    });
```
manaswinidas, line 99: nit: `labels` is no longer a valid `AgentFilterCategoryKey` after the realignment — valid keys are now `framework | category | communicationProtocol | testedModels`. This test passes at runtime but uses an invalid key.

Please update to use a valid key (e.g., `category: ['General purpose']`) in #2934 when you touch this file.

manaswinidas (approved):
All review comments addressed — nice work. One minor nit left (stale `labels` key in the context spec) that can be fixed in #2934.

manaswinidas (approved):
/approve

PR #2934, opened 2026-07-09
Add Agents Catalog Gallery and Filters

Philip-Carneiro:
## Description

Added Agents catalog gallery and filters

Agents Catalog Gallery:
- PF
image

- Non-PF
image

Filter working:
- PF
image

- Non-PF
image

Do not merge before the #2932

## How Has This Been Tested?

It was manually tested, and added tests

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+              {hasFiltersAppliedValue && <AgentsCatalogActiveFilters />}
+            </Flex>
+          </ToolbarContent>
+        </Toolbar>
+      </StackItem>
+    </Stack>
+  );
+};
+
+export default AgentsCatalogSourceLabelSelector;
```
manaswinidas, line 140: **DRY**: This is ~95% identical to `McpCatalogSourceLabelSelector`. The only differences are test IDs, placeholder text, and which `ActiveFilters` component is rendered.

Consider extracting a shared `CatalogSourceLabelSelector` in `app/shared/components/catalog/` that accepts `searchTestId`, `placeholder`, `hasFiltersApplied`, and `renderActiveFilters` props. Both MCP and Agents variants would become ~10-line wrappers passing config.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogAllAgentsView.tsx
```diff
@@ -0,0 +1,41 @@
...
+          searchTerm={term}
+          pageSize={pageSize}
+          onShowMore={onShowMore}
+        />
+      )}
+    />
+  );
+};
+
+export default AgentsCatalogAllAgentsView;
```
manaswinidas, line 41: **DRY**: This is ~98% identical to `McpCatalogAllServersView` (same structure, same 41 lines). Only the context, `otherSectionKey`, and `CategorySection` differ.

A factory function like `createCatalogAllItemsView({ useContext, otherSectionKey, CategorySection })` would eliminate this file entirely for each catalog type.

clients/ui/frontend/src/app/pages/agentsCatalog/components/AgentsCatalogActiveFilters.tsx
```diff
@@ -0,0 +1,68 @@
...
+          >
+            {null}
+          </ToolbarFilter>
+        );
+      })}
+    </>
+  );
+};
+
+export default AgentsCatalogActiveFilters;
```
manaswinidas, line 77: **DRY**: Structurally identical to `McpCatalogActiveFilters` — same `ToolbarFilter` mapping logic, same `handleRemoveFilter`/`handleClearCategory` callbacks.

This can be a single shared `CatalogActiveFilters` component taking `filterKeys`, `filterNames`, `filters`, `setFilters`, and `testIdPrefix` as props.

clients/ui/frontend/src/app/pages/agentsCatalog/utils/agentsCatalogUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+} from '~/app/pages/agentsCatalog/const';
+import { stringFiltersToFilterQuery } from '~/app/shared/components/catalog';
+
+export const hasAgentFiltersApplied = (
+  filters: AgentsCatalogFiltersState,
+  searchQuery: string,
+): boolean => {
+  if (searchQuery && searchQuery.trim().length > 0) {
+    return true;
+  }
```
manaswinidas: **DRY**: `hasAgentFiltersApplied` is identical logic to `hasMcpFiltersApplied` in `mcpCatalogUtils.ts`. Extract a shared utility:

```ts
// app/shared/components/catalog/utils/hasFiltersApplied.ts
export function hasFiltersApplied(
  filters: Record<string, string[] | undefined>,
  filterKeys: readonly string[],
  searchQuery: string,
): boolean {
  if (searchQuery?.trim()) return true;
  return filterKeys.some((key) => {
    const value = filters[key];
    return Array.isArray(value) && value.length > 0;
  });
}
```

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
```diff
@@ -1,3 +1,50 @@
...
+  'communicationProtocol',
+  'testedModels',
+];
+
+export const AGENT_FILTER_CATEGORY_NAMES: Record<AgentFilterCategoryKey, string> = {
+  framework: 'Framework',
+  category: 'Category',
+  communicationProtocol: 'Communication protocol',
+  testedModels: 'Tested models',
+};
```
manaswinidas, line 11: **Dead code**: `BACKEND_TO_FRONTEND_AGENT_FILTER_KEY` is `{}` (empty), which means `FRONTEND_TO_BACKEND_AGENT_FILTER_KEY` (computed from it in `agentsCatalogUtils.ts`) is also empty. The `agentFiltersToFilterQuery` function passes keys through unchanged.

Either remove this if no backend key remapping is needed, or add a comment indicating it's a placeholder for future use.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  Toolbar,
+  ToolbarContent,
+  ToolbarGroup,
+  ToolbarItem,
+  ToolbarToggleGroup,
+} from '@patternfly/react-core';
+import { ArrowRightIcon, FilterIcon } from '@patternfly/react-icons';
+import { useThemeContext } from 'mod-arch-kubeflow';
+import { ThemeAwareSearchInput } from 'mod-arch-shared';
+import { RESET_ALL_FILTERS_LABEL } from '~/concepts/modelCatalog/const';
```
manaswinidas: **Cross-domain import**: `RESET_ALL_FILTERS_LABEL` is imported from `~/concepts/modelCatalog/const`. This creates a coupling between agent catalog and model catalog internals. Consider moving this constant to a shared location (e.g., `app/shared/components/catalog/const.ts`).

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogGalleryView.tsx
```diff
@@ -0,0 +1,99 @@
...
+              Reset filters
+            </Button>
+          }
+        />
+      )}
+    />
+  );
+};
+
+export default AgentsCatalogGalleryView;
```
manaswinidas, line 98: **DRY**: This gallery view is ~92% identical to `McpCatalogGalleryView`. The only meaningful differences are: the entity type (`agent` vs `McpServer`), the data hook (`useAgentsBySourceLabel` vs `useMcpServersBySourceLabel`), and the card component.

A generic `createCatalogGalleryView` factory accepting `{ useItemsBySourceLabel, CardComponent, gridSpans, pageSize, otherDisplayName, entityNoun }` would eliminate this file for each catalog type.

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalog.tsx
```diff
@@ -1,18 +1,123 @@
...
+            onResetAllFilters={handleResetAllFilters}
+          />
+        )}
+        renderAllItemsView={() => <AgentsCatalogAllAgentsView searchTerm={searchQuery} />}
+        renderGalleryView={(isSingleCategory, singleCategoryLabel) => (
+          <AgentsCatalogGalleryView
+            handleFilterReset={handleResetAllFilters}
+            isSingleCategory={isSingleCategory}
+            singleCategoryLabel={singleCategoryLabel}
+          />
```
manaswinidas, line 88: **Missing error/loading/empty states in `AgentsCatalogCoreLoader`**: The core loader (not changed in this PR but relevant to the feature) is still a bare `<Outlet />`. Both `McpCatalogCoreLoader` and `ModelCatalogCoreLoader` handle:
- API error → error alert
- Not loaded → loading spinner
- No sources configured → empty state with admin guidance

Without these guards, if the agent catalog API errors or returns no sources, the gallery will render in a broken/confusing state. Please add the same error/loading/empty-sources guards to `AgentsCatalogCoreLoader`.

clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentsBySourceLabel.ts
```diff
@@ -0,0 +1,177 @@
...
+  const [firstPageData, loaded, error, refetch] = useFetchState(
+    fetchAgents,
+    { items: [], size: 0, pageSize: 10, nextPageToken: '' },
+    { initialPromisePurity: true },
+  );
+
+  React.useEffect(() => {
+    if (loaded && !error) {
+      setAllItems(firstPageData.items ?? []);
+      setTotalSize(firstPageData.size);
```
manaswinidas, line 91: **Potential race condition**: When filter/search params change, the reset effect (line ~130) clears `allItems`, but `firstPageData` may still hold stale data momentarily. Since `loaded` remains `true` from the previous fetch, this effect can re-apply stale data before the new fetch triggers.

Consider gating on a stable request identity (e.g., a serialized params key) or adding a `fetchGeneration` counter to avoid flicker.

clients/ui/bff/internal/mocks/model_catalog_client_mock.go
```diff
@@ -466,6 +466,90 @@ func (m *ModelCatalogClientMock) GetMcpServersTools(client httpclient.HTTPClient
...
+	case "framework":
+		if agent.Framework == nil {
+			return false
+		}
+		for _, v := range values {
+			if strings.EqualFold(*agent.Framework, v) {
+				return true
+			}
+		}
+	case "category":
```
manaswinidas: Category filter is no longer part of design - please remove all category related changes

manaswinidas (changes requested):

manaswinidas:
Also, try to add some svg in the mocks logos and check whether the image renders correctly. Add a fallback icon if there's no logo

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogSourceLabelSelector.tsx
Philip-Carneiro: Good point on the similarity. I looked into extracting a shared component, but there are a few domain-specific differences that make a clean abstraction tricky right now: (1) MCP has an extra SourceLabelBlocks section that Agents doesn't need, (2) each catalog uses its own context, active filters component, and test IDs, and (3) with only 2 catalogs using this pattern, a factory/HOC would add indirection without much payoff yet. I've already extracted the shared parts that were cleanly abstractable (RESET_ALL_FILTERS_LABEL, CatalogActiveFilters, hasFiltersApplied). Happy to revisit the full extraction once we have a third catalog or the two converge further, would that work for you?

clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentsBySourceLabel.ts
Philip-Carneiro: Good catch. I traced this pattern and it's the same in useMcpServersBySourceLabel,the reset effect clears allItems but firstPageData can briefly re-apply stale results before the new fetch completes. Fixing it only in the agents hook would create an inconsistency with the MCP equivalent. I'd prefer to fix both hooks together in a follow-up to keep them in sync. Should I open a separate ticket to address the race condition in both useAgentsBySourceLabel and useMcpServersBySourceLabel?

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogAllAgentsView.tsx
Philip-Carneiro: Agreed these are structurally similar. The remaining differences are in the domain layer each view consumes, different contexts, hooks, card components, grid span configs, and display name mappings. A factory function would need to accept all of these as parameters, which essentially becomes a component with many props rather than a simplification. I think extracting the layout shell (sidebar + gallery grid + empty state) into a shared component could work, but that's a larger refactor that would also touch the MCP views. Would you be OK if I file a follow-up ticket to do that refactor across both catalogs together, rather than in this PR?

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogSourceLabelSelector.tsx
manaswinidas: including model catalog - this is the third catalog. But yeah, we can revisit this for refactor if it's not worth the trouble

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogAllAgentsView.tsx
manaswinidas: yeah feel free to track this

clients/ui/frontend/src/app/hooks/agentsCatalog/useAgentsBySourceLabel.ts
manaswinidas: yes, you can group all issues in a single ticket

manaswinidas:
Please address this comment [here](https://github.com/kubeflow/hub/pull/2932#discussion_r3558221748)

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogSourceLabelBlocks.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import { AgentsCatalogContext } from '~/app/context/agentsCatalog/AgentsCatalogContext';
+import { CatalogSourceLabelToggle, getLabelDisplayName } from '~/app/shared/components/catalog';
+import { OTHER_AGENTS_DISPLAY_NAME } from '~/app/pages/agentsCatalog/const';
+
+const ALL_AGENTS_LABEL = 'All agents';
+
+const AgentsCatalogSourceLabelBlocks: React.FC = () => {
```
ppadti, line 8: This component is not being used by any other component - we need some clarity on how should we display toggle

clients/ui/frontend/src/app/pages/agentsCatalog/AgentsCatalogCoreLoader.tsx
```diff
@@ -1,6 +1,83 @@
...
+            title="Agents catalog configuration required"
+            description={
+              isMUITheme
+                ? 'To discover agents, follow the instructions in the docs below.'
+                : 'There are no agent sources to display. Request that your administrator configure agent sources for the catalog.'
+            }
+            headerIcon={() => (
+              <img src={typedEmptyImage(ProjectObjectType.modelRegistrySettings)} alt="" />
+            )}
+            primaryAction={<WhosMyAdministrator />}
```
ppadti: Missing MUI theme handling (KubeflowDocs) in CoreLoader empty state

clients/ui/frontend/src/app/pages/agentsCatalog/AgentsCatalogCoreLoader.tsx
```diff
@@ -1,6 +1,83 @@
...
+        emptyStatePage={<Bullseye>Loading catalog sources...</Bullseye>}
+        loaded={false}
+      />
+    );
+  }
+
+  if (catalogSources?.items?.length === 0 || !hasSourcesWithModels(catalogSources)) {
+    return (
+      <ApplicationsPage
+        title={AGENTS_CATALOG_TITLE}
```
ppadti: We should `TitleWithIcon` in here - don't we have the icon finalised yet? we should update the mod-arch library first

clients/ui/frontend/src/app/pages/mcpCatalog/screens/McpCatalogSourceLabelSelector.tsx
```diff
@@ -13,7 +13,7 @@ import {
 import { ArrowRightIcon, FilterIcon } from '@patternfly/react-icons';
 import { useThemeContext } from 'mod-arch-kubeflow';
 import { ThemeAwareSearchInput } from 'mod-arch-shared';
-import { RESET_ALL_FILTERS_LABEL } from '~/concepts/modelCatalog/const';
+import { RESET_ALL_FILTERS_LABEL } from '~/app/shared/components/catalog';
```
ppadti, line 16: model catalog related files still imports in from `import { RESET_ALL_FILTERS_LABEL } from '~/concepts/modelCatalog/const';`
here we just update the MCP and agent catalog

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalog.tsx
```diff
@@ -33,16 +38,83 @@ const AgentsCatalogTitle: React.FC = () => (
...
+    selectedSourceLabel,
+    setSelectedSourceLabel,
+    catalogSources,
+    catalogLabels,
+    catalogSourcesLoaded,
+    emptyCategoryLabels,
+    setCategoryCount,
+  } = React.useContext(AgentsCatalogContext);
+
+  // ponytail: agents catalog always shows flat gallery, no category grouping
```
ppadti: We need clarification from UX on this

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
```diff
@@ -6,16 +6,43 @@ export const AGENTS_CATALOG_DESCRIPTION =
...
 ];
 
 export const AGENT_FILTER_CATEGORY_NAMES: Record<AgentFilterCategoryKey, string> = {
   framework: 'Framework',
-  category: 'Category',
   communicationProtocol: 'Communication protocol',
   testedModels: 'Tested models',
 };
 
 export const BACKEND_TO_FRONTEND_AGENT_FILTER_KEY: Record<string, AgentFilterCategoryKey> = {};
```
ppadti, line 27: I think we need the exact response from backend to make sure we don't need any mapping here

clients/ui/frontend/src/__tests__/cypress/cypress/pages/agentsCatalog.ts
```diff
@@ -19,6 +19,46 @@ class AgentsCatalog {
...
+
+  findCardDetailLink(agentId: string) {
+    return cy.findByTestId(`agent-catalog-card-detail-link-${agentId}`);
+  }
+
+  findCardDescription(agentId: string) {
+    return cy.findByTestId(`agent-catalog-card-description-${agentId}`);
+  }
+
+  findCardFrameworkLabel(agentId: string) {
```
ppadti, line 39: This page object is not being used anywhere - can we add test for this to test the labels count on card

ppadti (changes requested):
Few comments:
1. We need update the AGENT_FILTER_CATEGORY_NAMES  -   `communicationProtocol: 'Communication protocol',` -> `Tools` based on mocks
2. We are missing source Label here
Image
3. As Manaswini mentioned can we update BFF mock to have a valid logo.

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
ppadti: Started a thread on the channel for this - likewise we can update

yabbesrajan:
Clarification on Single category Catalog (Flat designs) - 
We only have one category in Agent catalog at the moment - **Agent templates**. Even custom agent templates brought in by the admin user is out of scope for 3.5. So we are good with the flat UI with only one section. It doesn't make sense to have an "All agents" or "All agent templates" section in this case. As we get more (Agents etc as well)in the future releases, we can add them as well.
@Philip-Carneiro @ppadti

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalog.tsx
Philip-Carneiro: > Clarification on Single category Catalog (Flat designs) - We only have one category in Agent catalog at the moment - **Agent templates**. Even custom agent templates brought in by the admin user is out of scope for 3.5. So we are good with the flat UI with only one section. It doesn't make sense to have an "All agents" or "All agent templates" section in this case. As we get more (Agents etc as well)in the future releases, we can add them as well. @Philip-Carneiro @ppadti

clients/ui/frontend/src/app/pages/agentsCatalog/screens/AgentsCatalogGalleryView.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { agentFiltersToFilterQuery } from '~/app/pages/agentsCatalog/utils/agentsCatalogUtils';
+import {
+  getLabelDisplayName,
+  getLabelDescription,
+  CatalogGalleryLayout,
+  EmptyCatalogState,
+} from '~/app/shared/components/catalog';
+import AgentsCatalogCard from '~/app/pages/agentsCatalog/components/AgentsCatalogCard';
+
+const PAGE_SIZE = 10;
```
ppadti: can we reuse the page size const from AGENTS_CATALOG_GALLERY.PAGE_SIZE here?

ppadti (changes requested):
@Philip-Carneiro I am still not seeing the Title of the agent like we have in mocks

Image
doesn't have to be on top search bar - but I am expecting something like we did for Model catalog and MCP catalog to show when that has one source - can we do that?
Update the BFF mock to ave just one source and render that for now

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
```diff
@@ -4,18 +4,39 @@ export const AGENTS_CATALOG_TITLE = 'Agents Catalog';
...
 
-export const AGENT_FILTER_KEYS: AgentFilterCategoryKey[] = [
-  'framework',
-  'category',
-  'communicationProtocol',
-  'testedModels',
-];
+export const AGENT_FILTER_KEYS: AgentFilterCategoryKey[] = ['framework'];
 
 export const AGENT_FILTER_CATEGORY_NAMES: Record<AgentFilterCategoryKey, string> = {
```
ppadti, line 9: Just like key mapping we need to label mapping as well, just like Model catalog
The real backend filter_options endpoint returns lowercase/kebab-case framework values ("a2a", "claude-code", "langgraph", "vanilla_python", etc.), but the UX expects properly cased display names ("A2A", "Claude Code", "LangGraph", "Python").
Update BFF mock in static_mock_data.go and add label mapping in frontend

ppadti (changes requested):
Mostly looks good to me, one small things to fix in filters

clients/ui/frontend/src/app/pages/agentsCatalog/const.ts
```diff
@@ -4,18 +4,53 @@ export const AGENTS_CATALOG_TITLE = 'Agents Catalog';
...
   framework: 'Framework',
-  category: 'Category',
-  communicationProtocol: 'Communication protocol',
-  testedModels: 'Tested models',
+};
+
+export const AGENT_FRAMEWORK_LABEL_MAPPING: Record<string, string> = {
```
ppadti, line 13: This mapping still missing 4 values are missing: langflow, llamaindex, openclaw, and opencode (looking at the backend response we missed these)- but we can make this changes in the details page PR

ppadti (approved):
One minor thing but I don't want to hold this PR - Make sure to address this comment in details page PR
Thanks @Philip-Carneiro !!
/lgtm
/approve

PR #2937, opened 2026-07-10
feat(tracking): add vendor-neutral userInteraction tracking abstraction

manaswinidas:
## Description

Add a `concepts/userInteraction` module that provides a non-opinionated tracking API via React context. This enables downstream integrations (e.g., ODH with Segment) to inject their own analytics implementation without introducing vendor-specific dependencies upstream.

### What's included

| File | Purpose |
|------|---------|
| `trackingTypes.ts` | Event property types + `TrackingOutcome` enum |
| `UserInteractionContext.ts` | React context with dev-mode console logger default |
| `useUserInteraction.ts` | Consumer hook |
| `UserInteractionProvider.tsx` | Provider for downstream to inject custom `api` |
| `index.ts` | Barrel export |

### API

| Method | Purpose |
|--------|---------|
| `trackFormEvent(name, props)` | Form submit/cancel (outcome, success, error) |
| `trackLinkEvent(name, props)` | Navigation/link clicks (from, href, to, section) |
| `trackSimpleEvent(name, props?)` | Simple named events |
| `trackPageEvent()` | Page view tracking |

### How downstream uses it

```tsx
import { useUserInteraction, TrackingOutcome } from '~/concepts/userInteraction';

const { trackFormEvent } = useUserInteraction();

// In a handler:
trackFormEvent('Model Registered', {
  outcome: TrackingOutcome.submit,
  success: true,
});
```

Downstream (e.g., ODH) injects a Segment-backed implementation via `UserInteractionProvider`:

```tsx
<UserInteractionProvider api={segmentImpl}>
  <App />
</UserInteractionProvider>
```

### Design decisions

- **No App.tsx change**: The context default (dev-mode logger) works without a provider mounted. No dead wrapper needed.
- **Reuses existing `DEV_MODE`**: Imports from `~/app/utilities/const` rather than redefining.
- **Purely additive**: 6 new files, zero modifications to existing code. No breaking changes.

## Testing

- Unit tests verify provider injection, event routing, and hot-swap behavior
- Type check passes (`npx tsc --noEmit`)
- Lint passes (`npx eslint`)

```
PASS src/concepts/userInteraction/__tests__/useUserInteraction.spec.tsx
  useUserInteraction
    ✓ calls the injected provider implementation instead of the default
    ✓ routes different event types to their respective handlers
    ✓ swaps implementation when provider is re-mounted with a different api
```

### DIFF
--- a/clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
+++ b/clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
@@ -21,12 +21,15 @@ export type FormTrackingEventProperties = {
 };
 
 export type LinkTrackingEventProperties = {
-  from?: string;
+  /** Destination URL or route path. */
   href?: string;
-  to?: string;
+  /** Kind of element clicked (e.g., "project", "model", "external"). */
   type?: string;
+  /** UI region where the click occurred (e.g., "Model Catalog", "Sidebar"). */
   section?: string;
+  /** Display label of the clicked element. */
   name?: string;
+  [key: string]: string | number | boolean | undefined;
 };
 
 export type SimpleTrackingEventProperties = {
--- a/clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
+++ b/clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
@@ -1,0 +1,34 @@
+/**
+ * Vendor-neutral tracking event types.
+ *
+ * These mirror the main tracking methods available in downstream implementations
+ * (e.g., Segment) without introducing any vendor-specific dependency upstream.
+ *
+ * Event naming convention: "What-noun Past-Tensed-verb"
+ * Examples: "Model Deployed", "Pipeline Schedule Deleted", "Workbench Created"
+ */
+
+export const enum TrackingOutcome {
+  submit = 'submit',
+  cancel = 'cancel',
+}
+
+export type FormTrackingEventProperties = {
+  outcome: TrackingOutcome;
+  success?: boolean;
+  error?: string;
+  [key: string]: string | number | boolean | undefined;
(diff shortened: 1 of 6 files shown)


clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type FormTrackingEventProperties = {
+  outcome: TrackingOutcome;
+  success?: boolean;
+  error?: string;
+  [key: string]: string | number | boolean | undefined;
+};
+
+export type LinkTrackingEventProperties = {
+  from?: string;
```
christianvogt: Tracking tools should already capture the current URL for all events. It seems redundant to have the caller supply this.

clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+  href?: string;
+  to?: string;
```
christianvogt: What's the different between `href` and `to`?

clients/ui/frontend/src/concepts/userInteraction/trackingTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+  type?: string;
+  section?: string;
+  name?: string;
```
christianvogt, line 31: These need documentation. It's not clear to me how these should differ from the event name.

manaswinidas: Added JSDoc

manaswinidas (commented):
/approve 

Thanks @christianvogt for the reviews!

PR #2963, opened 2026-07-14
fix(bff): authorize job namespace via SAR instead of requiring namespace equality

manaswinidas:
## Summary

The `ensureNamespaceMatch` check on model-transfer-job routes required `jobNamespace` to equal the model registry service namespace. This broke the legitimate cross-namespace flow where the registry lives in one namespace (e.g. `rhoai-model-registries`) and transfer jobs run in a user's project namespace.

This PR replaces `ensureNamespaceMatch` with `authorizeJobNamespace`, which performs a proper SubjectAccessReview (internal mode) or SelfSubjectAccessReview (user_token mode) against the job namespace. The caller must prove they have service-level access in the namespace they want to operate in, without requiring it to be the same as the registry namespace.

## What changed

**`clients/ui/bff/internal/api/model_transfer_job_handler.go`:**
- Removed `ensureNamespaceMatch` (static namespace equality check)
- Added `authorizeJobNamespace` method on `App` that performs a SAR/SSAR via `client.CanListServicesInNamespace` against the operated-on namespace
- All 6 handlers (list, get, create, update, delete, events) now call `authorizeJobNamespace` instead of `ensureNamespaceMatch`
- Added `context` and `kubernetes` imports for the SAR call

**`clients/ui/bff/internal/api/model_transfer_job_authz_test.go`:**
- Replaced `TestEnsureNamespaceMatch` with `TestAuthorizeJobNamespace` covering: empty namespace (defers), user has access (allowed), cross-namespace without access (rejected), same namespace with access (allowed)
- Added `TestAuthorizeJobNamespaceMissingIdentity` for the missing-identity edge case

## Security

The authorization bypass (BOLA) is still closed: in `internal` auth mode, a caller who can only `get services` in their own namespace cannot operate on jobs in another tenant's namespace because the SAR will deny it. In `user_token` mode, K8s RBAC already enforces this via the caller's own bearer token.

## Verification

- `go build ./...`: clean
- `go test ./internal/api/ -run TestAuthorizeJobNamespace -v`: all pass
- `go vet ./internal/api/`: clean

### DIFF
--- a/clients/ui/bff/internal/api/model_transfer_job_authz_test.go
+++ b/clients/ui/bff/internal/api/model_transfer_job_authz_test.go
@@ -1,47 +1,87 @@
 package api
 
-import "testing"
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"testing"
 
-// TestEnsureNamespaceMatch guards the model-transfer-job BOLA fix: the namespace
-// a handler operates on (jobNamespace query parameter or request-body namespace)
-// must not differ from the namespace authorized by the SubjectAccessReview in
-// RequireAccessToMRService.
-func TestEnsureNamespaceMatch(t *testing.T) {
+	"github.com/kubeflow/hub/ui/bff/internal/constants"
+	k8s "github.com/kubeflow/hub/ui/bff/internal/integrations/kubernetes"
+)
+
+type stubKubernetesClientForAuthz struct {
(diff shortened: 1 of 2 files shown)


ppadti (approved):
/lgtm
/approve
Tested this PR midstream to make sure the regression is resolved
Thanks @manaswinidas

PR #2966, opened 2026-07-14
fix: upgrade cosign to v3.0.4 and handle signing config conflict with explicit URLs

dbasunag:

## Description
 Update cosign from v2.0.0 to v3.0.4 in the async-upload Dockerfile. Cosign v3+ defaults --use-signing-config=true which conflicts with explicit --fulcio-url/--rekor-url flags. Auto-disable signing config when explicit service URLs are provided.

## Merge criteria:

If you have UI changes

### DIFF
--- a/clients/python/src/model_registry/signing/image_signer.py
+++ b/clients/python/src/model_registry/signing/image_signer.py
@@ -250,7 +250,7 @@ class ImageSigner:
                 token_content = f.read().strip()
             env["COSIGN_IDENTITY_TOKEN"] = token_content
 
-        if fulcio_url is not None or rekor_url is not None:
+        if fulcio_url is not None and rekor_url is not None:
             # cosign v3+ defaults --use-signing-config=true which conflicts
             # with explicit service URLs; disable it when URLs are provided
             cmd.extend(["--use-signing-config=false"])
--- a/clients/python/src/model_registry/signing/image_signer.py
+++ b/clients/python/src/model_registry/signing/image_signer.py
@@ -250,9 +250,9 @@ class ImageSigner:
                 token_content = f.read().strip()
             env["COSIGN_IDENTITY_TOKEN"] = token_content
 
-        if fulcio_url is not None and rekor_url is not None:
+        if fulcio_url is not None or rekor_url is not None:
             # cosign v3+ defaults --use-signing-config=true which conflicts
-            # with explicit service URLs; disable it when URLs are provided
+            # with any explicit service URL; disable it when URLs are provided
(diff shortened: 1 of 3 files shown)


clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -250,6 +250,11 @@ def sign(  # noqa: C901
                 token_content = f.read().strip()
             env["COSIGN_IDENTITY_TOKEN"] = token_content
 
+        if fulcio_url is not None or rekor_url is not None:
+            # cosign v3+ defaults --use-signing-config=true which conflicts
+            # with explicit service URLs; disable it when URLs are provided
+            cmd.extend(["--use-signing-config=false"])
```
jonburdo, line 256: ** 🤖 AI Review **

This changes the `cosign sign` argv, but the PR does not update the Python signing tests that hard-code that command shape. In particular, `test_sign_with_all_args` currently expects the exact argv without `--use-signing-config=false`, so this change looks likely to break that test and still leaves no focused coverage for the new conditional flag behavior. Please update the existing assertion and add a targeted test for when the flag should and should not be emitted.

clients/python/src/model_registry/signing/image_signer.py
```diff
@@ -250,6 +250,11 @@ def sign(  # noqa: C901
                 token_content = f.read().strip()
             env["COSIGN_IDENTITY_TOKEN"] = token_content
 
+        if fulcio_url is not None or rekor_url is not None:
```
jonburdo, line 253: ** 🤖 AI Review **

This now disables the initialized signing config whenever either `fulcio_url` or `rekor_url` is provided. That changes the behavior of one-sided overrides: a caller that sets only one explicit URL can no longer inherit the companion endpoint from the existing Sigstore config, which may break private deployments that rely on that fallback. Consider only disabling signing config when both URLs are provided, or rejecting/documenting partial overrides explicitly.

jonburdo: maybe the best thing is to simply error if just one of `fulcio_url` or `rekor_url` is provided. This means you have to pass both or neither.

Alternatively like Cursor suggests, maybe it would work to simply change `or` to `and`
```suggestion
        if fulcio_url is not None and rekor_url is not None:
```

dbasunag: Done.

jonburdo (commented):
/lgtm
/approve

Thanks for catching this and finding the right image!

PR #2973, opened 2026-07-16
A proposal for Skills Catalog in Kubeflow

rareddy:
This is proposal for adding Skills catalog along with Model/MCP/Agent catalogs

## Description

Add a `skill` catalog plugin that indexes AI agent skills - directories containing a `SKILL.md` per the [Agent Skills specification](https://agentskills.io/specification)

### DIFF
--- a/proposals/KEP-0005-skills-catalog/README.md
+++ b/proposals/KEP-0005-skills-catalog/README.md
@@ -13,10 +13,10 @@ Agent skills have an open specification, growing public repositories, and instal
 ### Goals
 
 - `skill` plugin; API base `/api/skill_catalog/v1alpha1`.
-- A `git-skills-plugin` source type - named after what it reads, following the existing convention (`yaml`, `hf`) - configured through a YAML file that lists repositories; SKILL.md files are parsed per the specification at sync time.
-- Versioned entries: a repository can list multiple refs (tags, releases, branches, or commits). Each ref becomes its own set of catalog entries, with the version shown in the UI.
+- A `git-skills-plugin` source type - a dedicated type for git-hosted skills, alongside the existing `yaml` and `hf` types - listing repositories whose SKILL.md files are parsed per the specification at sync time.
+- Versioned entries: a repository can list multiple refs (tags, releases, branches, or commits). Each ref becomes its own set of catalog entries, with the version shown in the UI. A branch is surfaced as `latest` (not the raw branch name), and every entry pins to the commit it resolved to, so installs are reproducible.
 - Custom metadata (trust tier, provider, category, labels) assigned in the source file; per-skill overrides; include/exclude filtering.
-- Catalog UI: gallery, detail with rendered SKILL.md, filters, install instructions; an admin settings page that manages sources (add/modify/delete source files and repo entries, with include/exclude preview before save) alongside sync status and manual sync.
+- Catalog UI: gallery, detail with rendered SKILL.md, filters, install instructions; an admin settings page that reuses the existing model/MCP catalog-settings mechanism to manage sources (add/edit/delete user-managed sources with include/exclude preview before save; shipped defaults read-only) alongside sync status and manual sync.
 - `marketplace.json` endpoint, with optional URL rewriting for deployments fronting repos with an internal git mirror.
 
 ### Non-Goals
@@ -36,28 +36,32 @@ A standard Hub catalog plugin, following the existing plugin architecture.
 
 ### 2. Skill sources
 
-Registered like any other catalog source, as `type: git-skills-plugin`; the source's configured file lists git repositories and their catalog metadata:
+Registered like any other catalog source, as `type: git-skills-plugin`. A source lists git repositories and their catalog metadata. The repository list can live in the source config **inline** (used by UI-added sources - written to the user-managed ConfigMap, editable like an `hf` source) or **via a file** (`yamlCatalogPath`, used by shipped defaults that bundle many repos). The loader accepts both.
 
 ```yaml
-source: Community Skills
-trustTier: communityContributed
-repositories:
-  - url: https://github.com/example/skills.git
-    refs: [main, v1.0]            # optional; tags/releases/branches/commits -
+# inline form (UI-added source in the user-managed ConfigMap)
+- name: Community Skills
+  id: community-skills
+  type: git-skills-plugin
+  enabled: true
+  trustTier: communityContributed
+  repositories:
+    - url: https://github.com/example/skills.git
+      refs: [main, v1.0]          # optional; tags/releases/branches/commits -
                                   #   each ref yields its own catalog entries.
                                   #   Also: scanPaths, authSecretName,
                                   #   canonicalUrl (when url is a mirror)
-    provider: Example Org
-    category: DevOps
-    labels: [community]
-    includedSkills: ["*"]
-    excludedSkills: ["*-draft"]
-    skillOverrides: [{name: deploy, category: SRE}]
+      provider: Example Org
+      category: DevOps
+      labels: [community]
+      includedSkills: ["*"]
+      excludedSkills: ["*-draft"]
+      skillOverrides: [{name: deploy, category: SRE}]
 ```
 
-Each time the catalog syncs, the plugin reads each listed repository at each listed ref (making a temporary, lightweight copy used only for parsing, then throwing it away), finds and parses `SKILL.md` files per the specification, and rebuilds the index. A skill's `version` is the ref, and the exact commit it resolved to is recorded. Skills, refs, repositories, or sources that have been removed are cleaned up. Pulling from private git repositories with securely configured credentials will be supported.
+Each time the catalog syncs, the plugin reads each listed repository at each listed ref (making a temporary, lightweight copy used only for parsing, then throwing it away), finds and parses `SKILL.md` files per the specification, and rebuilds the index. A skill's `version` is the ref (a branch is surfaced as `latest`, not the branch name), and the exact commit it resolved to is recorded as `resolvedCommit`. Skills, refs, repositories, or sources that have been removed are cleaned up. Pulling from private git repositories with securely configured credentials will be supported.
 
-Source files stored in a ConfigMap are picked up automatically whenever they change. Files that are mounted read-only appear as read-only in the UI.
+Source management reuses the same mechanism as the model and MCP catalogs: read-only defaults (shipped in a ConfigMap) merged with a user-managed ConfigMap that the settings UI writes via the BFF, with per-source auth stored as Secrets. Adding a skills git source through the UI works like adding a HuggingFace source for models. Editing the ConfigMap directly (kubectl/GitOps) works too; the file-watcher picks up either path.
 
 ### 3. Custom metadata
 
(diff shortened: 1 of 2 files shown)


proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Summary
+
+Add a `skill` catalog plugin that indexes AI agent skills - directories containing a `SKILL.md` per the [Agent Skills specification](https://agentskills.io/specification) - serving them through the standard catalog API, a catalog UI alongside the model/MCP/agent catalogs, and a plugin-marketplace endpoint so compatible agents install skills with one `marketplace add`.
+
+Skill sources are git repositories, listed in a YAML source file (ConfigMap-mountable) together with their catalog metadata. At sync, the plugin reads the repos directly and rebuilds the ephemeral database index; repos are never stored.
+
+## Motivation
+
+Agent skills have an open specification, growing public repositories, and installation tooling (`npx skills add`, Claude Code marketplaces). Organizations need what Hub already provides for models, MCP servers, and agents: a curated, filterable catalog with provenance and a governed acquisition path. Skills complete the catalog family.
```
pboyd:
```suggestion
Agent skills have an open specification, growing public repositories, and installation tooling (`npx skills add`, Claude Code marketplaces). Organizations need what Hub already provides for models, MCP servers, and agents: a curated, filterable catalog with provenance and a governed acquisition path. Skills extend the catalog family.
```

Probably not the last thing we'll add.

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Goals
+
+- `skill` plugin; API base `/api/skill_catalog/v1alpha1`.
+- A `git` source type - named per the existing convention of naming source types for what they read (`yaml`, `hf`) - configured via a YAML file listing repositories; SKILL.md parsed per the specification at sync.
```
pboyd: Is `git` the right name for this source? I realize the data comes from git, but it's also a git repo with a special layout. `git-claude-code-plugin` would perhaps be more accurate, though I think that's too wordy.

pboyd: Should we also have a source that will pull from github? We should be able to pull files from github over HTTP more efficiently than we can clone a git repo.

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+                                  #   canonicalUrl (when url is a mirror)
+    provider: Example Org
+    category: DevOps
+    labels: [community]
+    includedSkills: ["*"]
+    excludedSkills: ["*-draft"]
+    skillOverrides: [{name: deploy, category: SRE}]
+```
+
+At sync, the plugin reads each listed repository at each listed ref directly (a temporary shallow clone used only for parsing, then discarded), finds and parses `SKILL.md` files per the specification, and rebuilds the index. A skill's `version` is the ref, with the resolved commit SHA recorded. Removed skills, refs, repos, or sources are cleaned up.
```
pboyd: Is this limited to public git repos? I don't see anything about configuring auth secrets, if it's supposed to work for private repos.

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Permanent identity is `(repository, path)` - the upstream repo URL plus the skill's directory - with `version` distinguishing entries. The index is an ephemeral cache; internal IDs are not durable. This keeps external references stable across rebuilds and across deployments reading the same skill through different URLs.
+
+### 5. API surface
+
+```
+GET  /skills                    # name, q, source, sourceLabel, filterQuery, paging
+GET  /skills/{id}
+GET  /skills/filter_options
+GET  /marketplace.json
```
pboyd: Should this be `/claude/marketplace.json` or similar? I understand other tools read the same file, but I'm wondering what we'll do if opencode or pi (or whatever comes out next week) introduces a similar concept. Would we add endpoints for them at the root?

proposals/KEP-0005-skills-catalog/skill-catalog-architecture.md
```diff
@@ -0,0 +1,208 @@
...
+
+| Protocol | Input | Notes |
+|---|---|---|
+| `/plugin marketplace add <catalog-url>` | The catalog's `marketplace.json` | Claude Code + compatible agents; one add = whole curated catalog. Key feature |
+| `npx skills add <repo-url>` | A **git repo** URL, never the catalog URL | Upstream repos, or the mirror Route (Part II) |
+| Manual | `git clone` + copy | Into the agent's skills dir (`~/.claude/skills/`, `~/.agents/skills/`, ...) |
+
+The skill detail page renders all three with environment-correct, copy-paste-ready URLs.
+
+## 8. Agent Delivery - Sparse Checkout Assembly
```
pboyd: "Agent delivery" is listed as a non-goal in the KEP.

pboyd (commented):
Looks good for the most part. I left some comments.

proposals/KEP-0005-skills-catalog/README.md
rareddy: how about `git-skills-plugin` to keep the vendor name out?

rareddy: it should work with any git repos, GitHub, GitLab or private instances of git repos. Yes, during the implemention will use stickly git libaries over any github specific ones. http makes sense, SSH based protocol might bring other issues, which does not bring any value anyways.

proposals/KEP-0005-skills-catalog/README.md
rareddy: I think I mentioned in architecture, but will add comment there that it will also handle the private got repos. adding `Capability to pull from private git repositories with safe credential configuration will be supported.`

proposals/KEP-0005-skills-catalog/skill-catalog-architecture.md
rareddy: Agree; Removed this section

proposals/KEP-0005-skills-catalog/README.md
rareddy: updated the text

proposals/KEP-0005-skills-catalog/README.md
rareddy: good suggestion; added it

proposals/KEP-0005-skills-catalog/README.md
pboyd: `git-skills-plugin` works for me.

pboyd (commented):
/lgtm

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Motivation
+
+Agent skills have an open specification, growing public repositories, and installation tooling (`npx skills add`, Claude Code marketplaces). Organizations need what Hub already provides for models, MCP servers, and agents: a curated, filterable catalog with provenance and a governed acquisition path. Skills extend the catalog family.
+
+### Goals
+
+- `skill` plugin; API base `/api/skill_catalog/v1alpha1`.
+- A `git` source type - named per the existing convention of naming source types for what they read (`yaml`, `hf`) - configured via a YAML file listing repositories; SKILL.md parsed per the specification at sync.
+- Versioned entries: repo entries may list refs (tags, releases, branches, commits); each ref yields its own catalog entries, with the version shown in the UI.
```
Al-Pragliola: can you go a bit more in depth here? if sources are git repositories, do you mean here that you need different sources for different branches/tags/etc? how do we handle same skills / diff versions?

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Custom metadata (trust tier, provider, category, labels) assigned in the source file; per-skill overrides; include/exclude filtering.
```
Al-Pragliola, line 18: so we need to support metadata override at the sources level? how is the mapping between override and entries going to work?

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+                                  #   canonicalUrl (when url is a mirror)
+    provider: Example Org
+    category: DevOps
+    labels: [community]
+    includedSkills: ["*"]
+    excludedSkills: ["*-draft"]
+    skillOverrides: [{name: deploy, category: SRE}]
+```
+
+At sync, the plugin reads each listed repository at each listed ref directly (a temporary shallow clone used only for parsing, then discarded), finds and parses `SKILL.md` files per the specification, and rebuilds the index. A skill's `version` is the ref, with the resolved commit SHA recorded. Removed skills, refs, repos, or sources are cleaned up. Capability to pull from private git repositories with safe credential configuration will be supported.
```
Al-Pragliola: a clone for each ref, or clone and then fetch each ref?

proposals/KEP-0005-skills-catalog/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### 3. Custom metadata
+
+Trust tier, provider, category, and labels come from the source file and are applied to its skills at sync - never read from repo content. `trustTier` is simply a label - one of `platformProvided`, `partnerVerified`, `organizationApproved`, `communityContributed` (downstream products may map display names) - shown as a badge and filterable, with no ordering or special semantics. Sources are only manageable by editing the mounted files.
+
+### 4. Skill identity
+
+Permanent identity is `(repository, path)` - the upstream repo URL plus the skill's directory - with `version` distinguishing entries. The index is an ephemeral cache; internal IDs are not durable. This keeps external references stable across rebuilds and across deployments reading the same skill through different URLs.
```
Al-Pragliola: if we can get several entries with a version discriminator, do we expect GET  /skills/{id} (where the id is not the internal one) to return more than one entry?

proposals/KEP-0005-skills-catalog/README.md
Al-Pragliola: update: one source have an array of refs, still the latter question remains -- we treat same skill different versions are separate entries?

rareddy: Yes, that is where my thinking is; Just like LLM, a Skill can be V2 but can have entirely different content. More over with catalog we do not have tree like mechnisam to represent to match on the names in single card.

proposals/KEP-0005-skills-catalog/README.md
rareddy: This is one way I could think how one add addtional [optional] metadata for single or set of skills. But if this is anti-pattern to rest of the system, I am certainly willing to roll this back.

proposals/KEP-0005-skills-catalog/README.md
rareddy: I belive this can be implemention detail, as the data is not preistent and each ref is going to be a different entry.

proposals/KEP-0005-skills-catalog/README.md
rareddy: `version` or `ref` needs to be part of the identity. I will correct above to include ref.

rareddy:
Do you have any further questions or can I get approvals for thsi KEP?

PR #2997, opened 2026-07-20
fix(catalog): watch YAML data files for live reload in MCP and agent catalogs

pboyd:
## Description

Previously, only sources.yaml was watched for file changes. YAML data files referenced via yamlCatalogPath required a pod restart or a sources.yaml touch to pick up edits.

MCP catalog: make yamlMCPProvider.Servers() long-lived. After emitting the initial batch it sends a zero-value sentinel record, registers a file watcher on each data file via basecatalog.GetMonitor().Path(), and re-emits + re-sends the sentinel on every detected change. The loader's PerformLeaderOperations now returns immediately after launching per-source goroutines; loadServersFromProvider uses sync.OnceFunc to release the pre-registered TrackWrite slot on the first sentinel so that WaitForInflightWrites unblocks correctly. Subsequent batches (file reloads) perform orphan cleanup and status updates the same way as the initial batch.

Agent catalog: add watchAndLoadFromYAML that performs the initial load, releases its TrackWrite slot, then loops on GetMonitor().Path() events to reload on change. PerformLeaderOperations is restructured to launch goroutines and return immediately, matching the MCP and model loader patterns.

Tests updated to use t.Context() so background goroutines are cancelled when each test ends, and to break on sentinel records rather than waiting for channel close.

## How Has This Been Tested?
Verified locally on a dev cluster.

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/mcpcatalog/providers.go
+++ b/catalog/internal/catalog/mcpcatalog/providers.go
@@ -221,14 +221,14 @@ func (yp *yamlMCPProvider) Servers(ctx context.Context) <-chan MCPServerProvider
 		// Set up watchers before the initial emit so that changes arriving
 		// during or just after the read are not missed.
 		merged := make(chan struct{}, 1)
-		watchFailed := false
+		watchCount := 0
 		for _, path := range yp.paths {
 			ch, err := basecatalog.GetMonitor().Path(ctx, path)
 			if err != nil {
 				glog.Errorf("unable to watch MCP catalog file %s: %v", path, err)
-				watchFailed = true
-				break
+				continue
 			}
+			watchCount++
 			go func(c <-chan struct{}) {
 				for {
 					select {
@@ -265,8 +265,8 @@ func (yp *yamlMCPProvider) Servers(ctx context.Context) <-chan MCPServerProvider
 			return
 		}
 
-		if watchFailed {
-			// Watcher setup failed; no live updates possible.
+		if watchCount == 0 {
+			// All watcher setups failed; no live updates possible.
 			return
 		}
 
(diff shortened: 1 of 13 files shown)


catalog/internal/catalog/mcpcatalog/providers.go
```diff
@@ -207,13 +207,48 @@ func NewYamlMCPProvider(source basecatalog.MCPSource) (MCPProvider, error) {
...
 
+		// Set up watchers before the initial emit so that changes arriving
+		// during or just after the read are not missed.
+		merged := make(chan struct{}, 1)
+		watchFailed := false
+		for _, path := range yp.paths {
+			ch, err := basecatalog.GetMonitor().Path(ctx, path)
+			if err != nil {
+				glog.Errorf("unable to watch MCP catalog file %s: %v", path, err)
+				watchFailed = true
```
Al-Pragliola: one wrong path, skips all the live updates even in correct path, what do you think about skipping only the bad ones?

catalog/internal/catalog/modelcatalog/loader.go
Al-Pragliola, line 160: shouldn't we add also here .state.WaitForInflightWrites(30 * time.Second) like we have done in the other plugins?

catalog/internal/catalog/mcpcatalog/providers.go
pboyd: This is fixed in the last commit.

catalog/internal/catalog/modelcatalog/loader.go
pboyd: Also fixed in the last commit.

Issue #3014, opened 2026-07-25
Skill Catalog (KEP-0005): implementation tracking


rareddy:
Implementation tracking for the Skill Catalog plugin proposed in **KEP-0005** (#2973). Each checkbox is one deliverable landed by its own PR; check it off when that PR merges.

**Scope.** This issue tracks the upstream `kubeflow/hub` skill plugin as described in the KEP: **E1** backend, **E2** frontend, plus end-to-end testing (**E6**) and documentation (**E7**) that support it. Skill delivery into agent pods and disconnected/air-gapped support are Non-Goals of the KEP and out of scope here. Sizes: S ≈ ≤2 days, M ≈ 3–5 days, L ≈ 1–2 weeks.

Design references: KEP `proposals/KEP-0005-skills-catalog/README.md` (#2973) and the accompanying architecture document.

**E1 is organized by the plugin build path**, following [`docs/writing-a-catalog-plugin.md`](./docs/writing-a-catalog-plugin.md) (#3039): the plugin is built with the `catalog-gen` tool and the catalog skills (`/init-catalog`, `/sync-catalog`, `/catalog-add-route`, `/catalog-sample-data`). Each phase names the tool/skill that drives it; hand-written git-specific domain logic (SKILL.md parsing, repo cloning, git sync) has no generator and is called out as such. Original story IDs are kept as `SKC-nnn` tags for traceability.

---

#### Phase 0 — Spec & schema *(authored ahead of the scaffold; `catalog-gen` skips these)*

- [x] **SKC-100** Submit KEP to kubeflow/hub (S)
- [x] **SKC-101** OpenAPI spec + `skills` asset type + codegen (M) — `plugins/skill.yaml` (`Skill` schema; list/get/filter_options) [#3039 Steps 2, 4, 7.1]
- [x] **SKC-102** `git-skills-plugin` source schema (S) — `SkillCatalogs` on `SourceConfig`; both forms (inline `repositories[]` + `yamlCatalogPath`→file); repo fields `url`/`canonicalUrl`/`refs`/`scanPaths`/`authSecretName`/custom metadata/include-exclude/`skillOverrides`. Parser in `source_schema.go` so `catalog-gen` owns `sources.go` for the source collection

#### Phase 1 — Scaffold + boot — **`/init-catalog`** [#3039 Steps 1–6]

[...]

PR #3041, opened 2026-07-30
feat(catalog): add SKILL.md parser (SKC-104)

rareddy:
## Description

Add a lenient parser for SKILL.md files per the Agent Skills spec. It splits the `---`-delimited YAML frontmatter (CRLF- and BOM-tolerant) from the body and parses name, description, license, compatibility, allowed-tools, and the metadata map, plus body extraction and line count.

Lenient validation matches the spec table: name mismatch and over-length warn (and still load); missing description and invalid or unclosed frontmatter skip; a body over 500 lines warns and surfaces the line count. Skip cases use sentinel errors; non-fatal issues are collected as warnings.

Includes a fixture suite: valid, name mismatch, missing name, name too long, missing description, malformed YAML, no/unclosed frontmatter, long body, metadata map, unicode, and CRLF.

Tracking: #3014
Assisted-by: Claude Code

## How Has This Been Tested?
unit tests

- [ x] The commits have meaningful messages

### DIFF
--- a/catalog/internal/catalog/skillcatalog/skillmd_parser.go
+++ b/catalog/internal/catalog/skillcatalog/skillmd_parser.go
@@ -49,20 +49,6 @@ type ParsedSkill struct {
 	Warnings []string
 }
 
-// skillFrontmatter is the YAML frontmatter schema. Parsing is lenient: unknown
-// keys are ignored (not strict) so authors can carry extra fields.
-type skillFrontmatter struct {
-	Name          string `json:"name,omitempty"`
-	Description   string `json:"description,omitempty"`
-	License       string `json:"license,omitempty"`
-	Compatibility string `json:"compatibility,omitempty"`
-	// AllowedTools is a space-separated string per the spec, e.g.
-	// "Bash(git:*) Read". A YAML list is also accepted for leniency, so this is
-	// decoded as `any` and normalized by parseAllowedTools.
-	AllowedTools any            `json:"allowed-tools,omitempty"`
-	Metadata     map[string]any `json:"metadata,omitempty"`
-}
-
 // ParseSkillMD parses SKILL.md content leniently per the Agent Skills spec.
 // expectedName is the skill's directory name, used for the name-match warning.
 //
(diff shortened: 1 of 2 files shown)


catalog/internal/catalog/skillcatalog/skillmd_parser.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// maxSkillNameLength is the maximum name length recommended by the Agent Skills
+// specification (https://agentskills.io/specification). Longer names warn rather
+// than skip, per the lenient validation policy.
+const maxSkillNameLength = 64
+
+// utf8BOM is the UTF-8 byte-order mark, stripped from the start of a file if present.
+var utf8BOM = []byte{0xEF, 0xBB, 0xBF}
```
pboyd, line 24: Is this solving a real problem? UTF-8 defines the byte order, so I'm not sure why it would need a byte order mark.

catalog/internal/catalog/skillcatalog/skillmd_parser.go
```diff
@@ -0,0 +1,146 @@
...
+	s := strings.ReplaceAll(string(bytes.TrimPrefix(content, utf8BOM)), "\r\n", "\n")
+
+	lines := strings.Split(s, "\n")
+	if len(lines) == 0 || strings.TrimSpace(lines[0]) != "---" {
+		return "", "", ErrNoFrontmatter
+	}
+
+	closing := -1
+	for i := 1; i < len(lines); i++ {
+		if strings.TrimSpace(lines[i]) == "---" {
```
pboyd: Is `TrimSpace` correct here? I would think leading spaces should be ignored. What would a document like this do?

```
---
foo: bar
description: |
   blah
   ---
   blah
baz: qux
---
Body
```

catalog/internal/catalog/skillcatalog/skillmd_parser.go
```diff
@@ -0,0 +1,146 @@
...
+	AllowedTools  []string
+	Metadata      map[string]any
+	Body          string
+	BodyLineCount int
+	// Warnings are non-fatal issues; a skill with warnings is still indexed but
+	// its source is marked partially-available.
+	Warnings []string
+}
+
+// skillFrontmatter is the YAML frontmatter schema. Parsing is lenient: unknown
```
pboyd: Should this also be lenient for unexpected types? For instance, I believe `allowed-tools: true` or `metadata: ["x"]` would prevent the file from being loaded.

catalog/internal/catalog/skillcatalog/skillmd_parser.go
rareddy: per claude analysis, it is not byte order, some editors (Windows: Notepad, PowerShell) prepend a UTF-8 BOM as an encoding marker. Go's `strings.TrimSpace` doesn't strip `U+FEFF`, so a BOM'd `SKILL.md `reads as "\ufeff---", misses our frontmatter check, and the skill gets silently skipped. Spec doesn't require it; it's just input hardening.

catalog/internal/catalog/skillcatalog/skillmd_parser.go
rareddy: you're right. TrimSpace also strips leading spaces, so the indented --- inside the description block scalar was being treated as the closing fence. Will fix to match --- at column 0  with allowed trailing whitespace still ok. will provide a fix in few.

catalog/internal/catalog/skillcatalog/skillmd_parser.go
rareddy: Yes, `allowed-tools` was mis-interpreted ,it is a space-separated string, not a YAML list. I will correct this.

pboyd: That's not quite what I meant, although the `allowed-tools` change addresses it.

I was asking if the type checks should be as strict as they are. If `metadata` isn't a `map`, for instance, none of the front-matter can be read (it's a similar problem if `name`, `description`, etc. aren't strings).

rareddy: Good point, I did not understood it before. that was too strict. switched to decoding into a map and reading each field defensively now

PR #3043, opened 2026-08-02
feat(mcp): add serverJson field to MCP server API

dkuc:
## Summary
- Add a freeform `serverJson` field (`type: object, additionalProperties: true`) to the `MCPServer` schema so the catalog can pass through [MCP server.json](https://registry.modelcontextprotocol.io/docs#/schemas/ServerJSON) conformant metadata without coupling the API to a specific spec version
- Parse `server_json` from catalog YAML, store as a JSON string property in the database, and serve it in the REST API response
- Register the property in the datastore spec, entity mappings, and filter options exclusion list
## Why
We want to make the server json available so that it can be displayed and people can copy and paste it into the registry of their choice.  
## Test plan
- [x] Unit tests for YAML→property conversion (`TestYamlMCPServerServerJSONConversion`, `TestYamlMCPServerServerJSONNil`)
- [x] Unit tests for DB↔OpenAPI round-trip (`TestConvertDbMCPServerToOpenapi_IncludesServerJson`, `TestConvertDbMCPServerToOpenapi_NoServerJson`)
- [x] Full `catalog/internal/catalog/mcpcatalog/` and `catalog/internal/converter/` test suites pass
- [x] Manual testing: loaded production Red Hat MCP catalog data, verified `serverJson` appears correctly in `GET /api/mcp_catalog/v1alpha1/mcp_servers` response
- [x] Backwards compatible — field is optional; absent `server_json` in YAML produces no property; API omits the field when not present

### DIFF
--- a/api/openapi/catalog-v1.yaml
+++ b/api/openapi/catalog-v1.yaml
@@ -2084,6 +2084,13 @@ components:
                 Runtime metadata for deploying this MCP server via MCPServer CRDs.
                 Provides defaults, recommendations, and requirements for creating deployments.
                 This is discovery metadata, not actual deployment configuration.
+            serverJson:
+              type: object
+              description: |-
+                MCP server.json conformant structure. Opaque to this API; the
+                schema is defined by the MCP specification and may evolve independently.
+                See https://registry.modelcontextprotocol.io/docs#/schemas/ServerJSON for details.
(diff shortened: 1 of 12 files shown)


pboyd (commented):
LGTM

`make gen` should fix the failing test.

PR #3054, opened 2026-08-04
fix(catalog): reclaim stale leader lock from crashed pod

dbasunag:

## Description

When a pod crashes without releasing the leader lock, the replacement pod can never acquire it and runs in read-only mode indefinitely.
Track the lock row's version number across retries. If it hasn't changed for 2x leaseDuration (no heartbeats from the dead pod), delete the stale row so the next acquisition attempt succeeds.

## Merge criteria:

If you have UI changes

### DIFF
--- a/catalog/internal/leader/election.go
+++ b/catalog/internal/leader/election.go
@@ -3,7 +3,6 @@ package leader
 
 import (
 	"context"
-	"database/sql"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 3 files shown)


catalog/internal/leader/election.go
```diff
@@ -38,18 +43,57 @@ type lockClient interface {
 }
 
 // pglockAdapter wraps *pglock.Client to satisfy lockClient.
+// It tracks the last-seen record version number (RVN) to detect stale locks
+// left by crashed pods that never released cleanly. FailIfLocked() prevents
+// pglock's built-in two-phase takeover, so this adapter reimplements the
+// staleness check externally.
```
pboyd: Would removing `FailIfLocked()` be a simpler approach? It sort of seems like this is re-implementing that functionality.

dbasunag: It was introduced to work with dbReachable to solve the rolling update deadlock. I can look for a way to simplify that.

pboyd (commented):
Thanks for the changes. This seems like a reasonable to trade-off to me.

/lgtm
/approve

PR #3055, opened 2026-08-04
feat(catalog): git-backed skill catalog sync (SKC-105/106/107)

rareddy:
Index AI-agent skills from git repositories into the catalog datastore.

Supports #3014 

## Description
- Repo resolver (SKC-105): shallow-clone each repo at a ref into a temp dir, parse-only and discarded; immutable refs only (tags/SHAs — branches and HEAD are refused); transport allowlist, GIT_ASKPASS credential passing, clone timeout, and post-fetch size limit.
- Loader / sync orchestration (SKC-106): leader-elected drop-and-reindex per source, per-source lock, bounded FIFO resolve worker pool plus a global in-flight-clone cap, per-secret credential caching, periodic resync, and orphan/missing-source cleanup.
- Multi-version entries (SKC-107): distinct entries per (sourceID, repository, path, version); version surfaced in list/detail.
- Secret-based auth: Kubernetes Secret resolver over the in-cluster API, re-reading the SA token per request; degrades gracefully off-cluster.
- Add first-class `author` Skill field (OpenAPI + parser + datastore + API).
- Centralize skill property names/kinds in one table (datastore registration + coverage guard) so the schema can't drift from the writer/reader.
- Install git-core in the runtime image for clone-at-sync.

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

## How Has This Been Tested?
Tests included

- [ x] Automated tests are provided as part of the PR for major new functionalities; testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).


catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +146,475 @@ func (l *SkillLoader) setCloser(closer func()) {
...
+}
+
+// syncSourceLocked performs the actual drop-and-reindex from an already-parsed
+// spec. The caller must hold the source's lock. It resolves every repository at
+// every ref (bounded per sync by the worker pool, and globally by the in-flight
+// clone cap) and re-indexes what it finds; the ephemeral, rebuild-each-sync model
+// guarantees no orphans.
+func (l *SkillLoader) syncSourceLocked(ctx context.Context, sourceID string, spec *SkillSourceSpec) {
+	l.state.TrackWrite()
+	delErr := l.services.SkillRepository.DeleteBySource(sourceID)
```
pboyd: Deleting upfront and re-populating will leave it empty for as long as the sync takes. Can this do what the other catalog types do (upsert each item, then delete orphans)?

catalog/internal/catalog/skillcatalog/db_skill.go
```diff
@@ -37,11 +41,14 @@ type ListSkillsParams struct {
...
+// emptyFilterMap is the shared empty filter map returned until per-field
+// filters are implemented (SKC-108). The map must not be mutated by callers.
+var emptyFilterMap = map[string]openapi.FilterOption{}
+
 // GetFilterOptions returns the filterable fields for skills. Entity-specific
 // options are populated once fields are persisted (SKC-108).
 func (d *DBSkillCatalog) GetFilterOptions(_ context.Context) (*openapi.FilterOptionsList, error) {
-	options := make(map[string]openapi.FilterOption)
-	return &openapi.FilterOptionsList{Filters: &options}, nil
+	return &openapi.FilterOptionsList{Filters: &emptyFilterMap}, nil
```
pboyd: Do you need this change for something? I suppose it's slightly more efficient, but it seems to still be a stop-gap. There's also a new down-side that a caller could change the returned map.

catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +146,475 @@ func (l *SkillLoader) setCloser(closer func()) {
...
 			return fmt.Errorf("failed to parse skill config %s: %w", path, err)
 		}
 	}
-
-	glog.Infof("%s loader config parsing complete", "skill")
+	glog.Info("skill loader config parsing complete")
 	return nil
 }
 
 func (l *SkillLoader) PerformLeaderOperations(ctx context.Context, allKnownSourceIDs mapset.Set[string]) error {
```
pboyd, line 217: Are the syncs done by this function done concurrently or serially? Maybe I'm overlooking it, but I don't actually see anything that forks a goroutine to process them in parallel, despite all the mutexes, etc.

catalog/internal/catalog/skillcatalog/secret_resolver.go
Al-Pragliola, line 1: uhm here we are basically rewriting a simpler k8s client-go to fetch the secrets, I liked the way we avoided binding to kubernetes resources and used local files directly instead -- in your opinion can we do the same here? --git-credentials filepath, with all the credentials mapped and aggregated in a single file @pboyd @rareddy

dkuc:
I was able to run the catalog, load the skills from git, and query them from the api. It all worked very smoothly. I used this source
``` yaml
skill_catalogs:
  - name: Kubeflow Hub Skills
    id: kubeflow-hub-skills
    type: git-skills-plugin
    enabled: true
    properties:
      syncIntervalMinutes: 0
      repositories:
        - url: https://github.com/kubeflow/hub.git
          refs: [db4fda47cb7499b21ad1d0151028e304a8e02cb8]
          scanPaths: [.agents/skills/]
          provider: Kubeflow
          category: Development
```

One minor observation.  All the descriptions end in a \n and this does not appear intentional.

Example
``` json
"description": "Add an endpoint or query parameter to a catalog plugin's OpenAPI spec. Guides the user through the spec edit, then recommends running /sync-catalog to propagate changes. Does not run sync automatically — the user may want to batch multiple route additions before syncing.\n",
```

rareddy:
>One minor observation. All the descriptions end in a \n and this does not appear intentional.

@dkuc do you mean in the testcases?

dkuc:
> > One minor observation. All the descriptions end in a \n and this does not appear intentional.
> 
> @dkuc do you mean in the testcases?

No.  All the test cases use a regular scalar values for the description. This only appears to happen with a yaml "folded string. i.e when you use the `>` character

Works
``` yaml
name: catalog-add-route
description: test text
user-invocable: true
```
Extra newline
``` yaml
name: catalog-add-route
description: >
  test text
user-invocable: true
```

catalog/internal/catalog/skillcatalog/skillmd_parser.go
dkuc:
```suggestion
	return strings.TrimSpace(str)
```

catalog/internal/catalog/skillcatalog/loader.go
rareddy: Good point! using the `upsert` now and removing orphans at the end but at same time enhanced such that the re-fetch only happens when refs change or new repo added; not a blank reload each time whole catalog. We will look into force refresh all at later PRs

catalog/internal/catalog/skillcatalog/db_skill.go
rareddy: reverted this

catalog/internal/catalog/skillcatalog/loader.go
rareddy: yikes! yes, that was the intension each repo request be concurrent, reworked to do this. Many more changes to make this bounded.

catalog/internal/catalog/skillcatalog/secret_resolver.go
rareddy: good suggestion! reworked to read from the file; but at the same time mapping the secret to be loaded as file that way I can send the password file directly as suggested. worked out as much better solution

catalog/internal/catalog/skillcatalog/skillmd_parser.go
rareddy: adopted to the suggestion. 

Please re-review @dkuc @pboyd @Al-Pragliola

catalog/internal/catalog/skillcatalog/env_config.go
Al-Pragliola, line 1: We are handling everything with flags at the moment, even if we want to mix envs / flags maybe it's better to use the viper/cobra combo?

catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +181,637 @@ func (l *SkillLoader) setCloser(closer func()) {
...
+func resolveOne(ctx context.Context, job refJob, knownCommit string, sem *semaphore.Weighted, resolver refResolver, credentials func(SkillRepository) (*Credentials, error)) resolveResult {
+	if ctx.Err() != nil {
+		return resolveResult{repo: job.repo, ref: job.ref, err: ctx.Err()}
+	}
+	creds, err := credentials(job.repo)
+	if err != nil {
+		return resolveResult{repo: job.repo, ref: job.ref, err: err}
+	}
+	if knownCommit != "" {
+		if remote, rerr := resolver.RemoteCommit(ctx, job.repo, job.ref, creds); rerr == nil && remote == knownCommit {
```
dkuc, line 800: :robot: 
**Metadata edits to pinned refs are silently dropped by the immutable-ref skip optimization.**

The skip here keys only on the git commit (`remote == knownCommit`). But the indexed skill entity also embeds source-config metadata that is *not* part of the git content — `category`, `labels`, `provider`, `trustTier`, and `skillOverrides` (all applied in `buildSkillEntity`).

Failure scenario: an operator edits, say, `category` for a repo pinned to an immutable tag/SHA and reloads config. `RemoteCommit` returns the unchanged commit, so `resolveOne` returns `skipped: true`, `buildSkillEntity` never runs, and the DB keeps the old metadata forever. A full leader restart re-skips too, so the only way to force the update is to change the ref.

Consider incorporating a hash of the source-config-derived metadata into the skip key (or bypassing the skip when that config has changed since the last index), so metadata-only edits to pinned repos are re-applied.

catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +181,637 @@ func (l *SkillLoader) setCloser(closer func()) {
...
+	var errs []error
+	for _, skill := range existing {
+		attrs := skill.GetAttributes()
+		if attrs == nil || attrs.Name == nil || skill.GetID() == nil {
+			continue
+		}
+		if currentNames.Contains(*attrs.Name) {
+			continue
+		}
+		if unsafeRepos.Contains(skillProperty(skill, propRepository)) {
```
dkuc: :robot: 
**One failing ref freezes orphan cleanup for all sibling refs of the same repository.**

`unsafeRepos` protects orphan cleanup at whole-repository granularity (`skillProperty(skill, propRepository)`), but a repo can be pinned to multiple refs and a failure is recorded per-repo (`unsafeRepos.Add(res.repo.URL)`).

Failure scenario: repo `R` is pinned to `[v1, v2, v3]`; an operator removes `v2` from config while `v3` is temporarily unreachable. This sync: `v3` fails → `unsafeRepos.Add(R.URL)`. `removeOrphans` then skips *every* skill whose repository is `R`, so `v2`'s now-orphaned skills are retained until `v3` clones successfully. As long as any sibling ref of the same repo keeps failing, a removed ref's stale skills linger indefinitely.

Consider tracking unsafe state at `(repo, ref)` granularity so that a failure on one ref doesn't block cleanup of orphans belonging to other, successfully-enumerated refs of the same repository.

api/openapi/src/plugins/skill.yaml
```diff
@@ -136,6 +136,9 @@ components:
             license:
               type: string
               description: Short name of the skill's license (e.g., apache-2.0).
+            author:
```
pboyd, line 139: Can you add this to `skill-v1.yaml` too?

catalog/internal/catalog/skillcatalog/db_skill_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	skillmodels "github.com/kubeflow/hub/catalog/internal/catalog/skillcatalog/models"
+	dbmodels "github.com/kubeflow/hub/internal/platform/db/entity"
+)
+
+func strPtr(s string) *string { return &s }
+func i32Ptr(n int32) *int32   { return &n }
+func boolPtr(b bool) *bool    { return &b }
```
pboyd: These functions are unnecessary since Go 1.26. If you run `go fix` it will replace the calls to them with `new`.

catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +181,637 @@ func (l *SkillLoader) setCloser(closer func()) {
...
+// acquired so neither holds one; both are still bounded by the worker pool.
+func resolveOne(ctx context.Context, job refJob, knownCommit string, sem *semaphore.Weighted, resolver refResolver, credentials func(SkillRepository) (*Credentials, error)) resolveResult {
+	if ctx.Err() != nil {
+		return resolveResult{repo: job.repo, ref: job.ref, err: ctx.Err()}
+	}
+	creds, err := credentials(job.repo)
+	if err != nil {
+		return resolveResult{repo: job.repo, ref: job.ref, err: err}
+	}
+	if knownCommit != "" {
```
pboyd, line 799: What happens if some a category (or a label, provider, etc.) changes but there's no new commit on the repo?

catalog/internal/catalog/skillcatalog/loader.go
```diff
@@ -32,46 +181,637 @@ func (l *SkillLoader) setCloser(closer func()) {
...
+
+	existingSourceIDs, err := l.services.SkillRepository.GetDistinctSourceIDs()
+	if err != nil {
+		return fmt.Errorf("unable to retrieve existing skill source IDs: %w", err)
+	}
+
+	for oldSource := range mapset.NewSet(existingSourceIDs...).Difference(enabledSourceIDs).Iter() {
+		glog.Infof("Removing skills from source %s", oldSource)
+		l.state.TrackWrite()
+		delErr := l.services.SkillRepository.DeleteBySource(oldSource)
```
pboyd: Should this hold the source lock before deleting the source? Perhaps it doesn't matter, but if there's a chance that a sync is proceeding despite the skill being disabled, then it might be a problem.

catalog/internal/catalog/skillcatalog/repo_resolver.go
```diff
@@ -0,0 +1,554 @@
...
+
+// computeBaseEnv builds the constant portion of the git environment: prompts
+// disabled, transport allowlist applied, and system config suppressed. The
+// result is precomputed once so concurrent clones only copy a slice.
+func (r *RepoResolver) computeBaseEnv() []string {
+	allowed := make([]string, 0, len(r.allowedProtocols))
+	for p := range r.allowedProtocols {
+		allowed = append(allowed, p)
+	}
+	return append(os.Environ(),
```
pboyd, line 110: Should `os.Environ` be after the env vars you're adding? The values you're providing are likely ignored if they were also set in the pod's environment.

catalog/internal/catalog/skillcatalog/loader.go
rareddy: I was thinking about adding force-reload from the yaml as config option; but did not figure out how to do it only first time and if pod reloads do not do it again? any suggestions?

catalog/internal/catalog/skillcatalog/repo_resolver.go
rareddy: The intuition behind the comment is sound for C — getenv returns the first match — but it never gets that far, because Go collapses the duplicates first.

concern is the reverse of how Go behaves here — the current order is the one that protects the vars, and their suggested change would introduce the bug

Added a testcase

catalog/internal/catalog/skillcatalog/env_config.go
rareddy: ageed that we will not do this time.

catalog/internal/catalog/skillcatalog/loader.go
rareddy: I was thinking about adding force-reload from the yaml as config option; but did not figure out how to do it only first time and if pod reloads do not do it again?  wdyt?

catalog/internal/catalog/skillcatalog/loader.go
rareddy: fixed per suggestion to account over ref level

api/openapi/src/plugins/skill.yaml
rareddy: done

catalog/internal/catalog/skillcatalog/db_skill_test.go
rareddy: done

catalog/internal/catalog/skillcatalog/loader.go
rareddy: fixed

catalog/internal/catalog/skillcatalog/repo_resolver.go
rareddy: added a testcase for this

pboyd: True enough. I didn't know Go's exec package did that.

catalog/internal/catalog/skillcatalog/loader.go
rareddy: added a digest based repo configuration; and digest and ref will be used now for refresh. if any one of items change the refresh will run agin in next intervel.

catalog/internal/catalog/skillcatalog/loader.go
rareddy: moved the the trustTier to repo level and added the hash of SkillsRepository struct to the ref to control the reload.

dkuc (approved):

PR #3056, opened 2026-08-04
feat(catalog): add logo image endpoint for MCP servers

dkuc:
## Summary

- Adds `GET /api/mcp_catalog/v1alpha1/mcp_servers/{server_id}/logo` endpoint that serves MCP server logos as browser-loadable images
- For **base64 data URI** logos: decodes the payload and serves raw image bytes with the correct `Content-Type` and `Cache-Control: public, max-age=3600`
- For **plain URL** logos: returns a `302 Found` redirect to the original URL
- Proper error handling: `404` for missing servers/logos, `400` for invalid IDs

### Motivation

External systems like MLflow's MCP server registry require HTTP URLs for icon references (`icons[].src`). The catalog currently stores logos as inline base64 data URIs, which can't be used directly. This endpoint gives every MCP server a stable, linkable URL for its logo.

### Files changed

- `catalog/internal/plugins/mcp/logo_handler.go` — handler + data URI parser
- `catalog/internal/plugins/mcp/logo_handler_test.go` — 11 test cases
- `catalog/internal/plugins/mcp/plugin.go` — registers the custom route

## Test plan

- [x] Unit tests cover: data URI SVG/PNG decode, plain URL redirect, not-found, bad-request, malformed data URI, invalid base64, provider errors
- [x] `make build/compile` passes
- [x] `make -C catalog test` passes
- [x] `golangci-lint` clean on changed files
- [x] Manual test against live server with Red Hat MCP catalog data (base64 logos → 200, URL logos → 302, missing → 404)

### DIFF
--- a/api/openapi/catalog.yaml
+++ b/api/openapi/catalog.yaml
@@ -227,6 +227,74 @@ paths:
         required: true
       - $ref: "#/components/parameters/includeTools"
       - $ref: "#/components/parameters/toolLimit"
+  /api/mcp_catalog/v1alpha1/mcp_servers/{server_id}/logo:
+    description: >-
+      The REST endpoint/path used to fetch the logo image for an `MCPServer`.
+    get:
+      # This operation is served by a custom handler in the MCP plugin
+      # (internal/plugins/mcp/logo_handler.go): it returns raw image bytes or a
+      # 302 redirect, which the generated JSON controller cannot express.
+      # `x-internal` documents the endpoint in the published spec while excluding
+      # it from client/server code generation so the manual route is authoritative.
+      x-internal: true
+      summary: Get the logo image for an `MCPServer`.
+      description: >-
+        Returns the MCP server's logo. When the stored logo is an inline data URI, the decoded image bytes are served directly with the matching image content type (restricted to a set of known image media types). When the stored logo is a plain http(s) URL, the client is redirected to it.
+      tags:
+        - MCPCatalogService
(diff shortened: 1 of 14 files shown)


jonburdo:
A couple optional comments from AI review with claude. Possibly worth fixing 1 at least, though could also be deferred to a follow-up

1. `Non-base64 data URIs bypass the pre-decode size guard`
   - `severity`: `medium`
   - `subscription`: Both reviewers flagged that `serveDataURI` only enforces `maxLogoBytes` before decoding for base64 payloads. For percent-encoded data URIs, `url.PathUnescape(data)` runs first and the size check happens afterward, so an oversized attacker-controlled logo can still trigger large allocations and extra CPU before the handler returns `413`. Add an encoded-size bound before unescaping, or decode while counting bytes, and add a test covering an oversized percent-encoded data URI.

2. `Served logos omit an explicit Content-Disposition`
   - `severity`: `low`
   - `subscription`: One reviewer noted that the new logo endpoint serves image bytes without setting `Content-Disposition`, so direct navigation relies on browser defaults for inline rendering. This looks like a defense-in-depth gap rather than a clear exploit, but setting `Content-Disposition: inline` or `attachment` would make the intended behavior explicit, especially for `image/svg+xml`.

jonburdo:
/lgtm
/approve
/hold

Feel free to unhold when ready to merge

dkuc:
Fixed both issues.

PR #3066, opened 2026-08-09
fix(async-upload): scope S3 downloads to directory prefix

immanuwell:
## Description

S3 treats `Prefix` as a raw string. With `my-model`, a bucket containing `my-model/model.onnx` and `my-model-backup/unrelated.bin` returns both
The second path lands outside the staging dir, pretty easy to hit in shared buckets 

This adds the trailing `/` for directory listings. Keys already ending in `/` stay unchanged.

## How Has This Been Tested?

Repro:
1. Add the two keys above to MinIO.
2. Run the job with `MODEL_SYNC_SOURCE_AWS_KEY=my-model`.
3. Before this fix both objects are downloaded, now only `my-model/model.onnx` is.

Passed all 55 async upload unit tests, `make build/compile`, `make lint`, and `make vet`
A real MinIO check passed too

Broad `make test` was also run, but unrelated MySQL testcontainers timed out on startup

## Merge criteria:

- [x] All commits are signed off.
- [x] The commit has a meaningful message.
- [x] Automated tests are included.
- [x] The change was manually tested with MinIO.
- [x] Code follows the contribution guidelines.

### DIFF
--- a/jobs/async-upload/job/download.py
+++ b/jobs/async-upload/job/download.py
@@ -47,7 +47,7 @@ def download_from_s3(config: S3StorageConfig, storage_path: str):
     )
 
     bucket_name = config.bucket
-    prefix = config.key
+    prefix = config.key if config.key.endswith("/") else f"{config.key}/"
 
     # TODO: It might make sense to check if the provided key points to a single file first before assuming the directory needs to be traversed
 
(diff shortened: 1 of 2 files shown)


jobs/async-upload/job/download.py
```diff
@@ -47,7 +47,7 @@ def download_from_s3(config: S3StorageConfig, storage_path: str):
     )
 
     bucket_name = config.bucket
-    prefix = config.key
+    prefix = config.key if config.key.endswith("/") else f"{config.key}/"
```
pboyd, line 50: Thanks for the PR, @immanuwell.

One question though, what happens if `config.key` were `my-model/model.onnx`? I believe that would work today, but it would break after this change because it would become `my-model/model.onnx/`.

PR #3075, opened 2026-08-11
Feat/skc 108 skill fields

rareddy:
## Description
Propagate the skill.yaml field set through datastore entries, entity mappings, and the DB provider; complete list/get/filter_options with name/q/source/sourceLabel/filterQuery/paging. Add a git_skills_plugin demo source (mattpocock/skills) to the demo overlay.

Ref #3014

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>


catalog/internal/catalog/skillcatalog/service/entity_mappings_skill_test.go
```diff
@@ -7,14 +7,30 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// TODO: Add entity-specific properties from the OpenAPI spec to this map.
-// See catalog/internal/catalog/modelcatalog/service/entity_mappings_test.go for reference.
 var expectedSkillProperties = map[string]filter.PropertyDefinition{
```
pboyd: I didn't notice this test before, but I'd recommend removing it. It doesn't appear to be asserting anything beyond our ability to make the same change in two places.

catalog/internal/catalog/skillcatalog/service/skill.go
```diff
@@ -141,9 +145,49 @@ func mapSkillToProperties(entity models.Skill, contextID int32) []schema.Context
...
+				nonEmptySourceIDs = append(nonEmptySourceIDs, sourceID)
+			}
+		}
+	}
+
+	if len(nonEmptySourceIDs) > 0 {
+		// is_custom_property = false excludes a same-named customProperty (e.g. a
+		// SKILL.md frontmatter metadata key literally called "source_id", which
+		// customPropertiesFromMetadata does not guard against) — without it, a
+		// colliding row would duplicate the skill in the joined result set.
```
pboyd, line 191: Does the SQL for name and query need this guard also?

dkuc (approved):

catalog/internal/catalog/skillcatalog/service/entity_mappings_skill_test.go
rareddy: removed the test now @pboyd however since I rebased it showing up in single commit insted of as add on

catalog/internal/catalog/skillcatalog/service/skill.go
rareddy: yes, added the check here too now.

PR #3077, opened 2026-08-11
refactor(ui): extract shared catalog toolbar components

manaswinidas:
## Description

Extract shared browse-catalog toolbar components so Model, MCP, and Agents catalogs stop duplicating nearly identical search/filter chip UI.

- Add `CatalogSourceLabelSelector` and `CatalogActiveFilters` under `app/shared/components/catalog/`
- Migrate MCP and Agents selectors onto the shared components and delete their duplicate ActiveFilters files
- Refactor the Model selector to compose the shared shell while keeping model-only performance toolbar, sort dropdown, and alert behavior
- Gallery/AllItems views already used shared layouts; left as thin catalog wrappers

### Review follow-ups
- Show **Reset all filters** when search-only is active (not only when chips are set)
- Put performance/workload filters **above** category blocks (match prior Model layout)
- Keep search field from collapsing (`min-width: 16rem`) so placeholder/click target stay usable
- Pass `onResetAllFilters` directly to Toolbar (drop redundant guard)

No intentional user-visible behavior change beyond restoring prior Model layout/search UX.

## How Has This Been Tested?

- Unit tests for shared catalog toolbar components
- Manual: Model/MCP/Agents search, filter chips, reset, performance view order

## Merge criteria:

If you have UI changes


clients/ui/frontend/src/app/shared/components/catalog/CatalogSourceLabelSelector.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  React.useEffect(() => {
+    setInputValue(searchTerm || '');
+  }, [searchTerm]);
+
+  const handleClearAllFilters = React.useCallback(() => {
+    if (shouldShowResetAll) {
+      onResetAllFilters();
+    }
+  }, [shouldShowResetAll, onResetAllFilters]);
```
Philip-Carneiro: The `if (shouldShowResetAll)` guard is redundant. This callback is only passed to `Toolbar` when `shouldShowResetAll` is true (via `toolbarClearAllProps` at lines 96-100). The Toolbar only calls `clearAllFilters` when the button is visible, and the button only appears when `toolbarClearAllProps` is defined. The guard never prevents anything. Could simplify to passing `onResetAllFilters` directly.

Philip-Carneiro (changes requested):

clients/ui/frontend/src/app/shared/components/catalog/CatalogSourceLabelSelector.tsx
manaswinidas: Agreed — the guard was redundant since `clearAllFilters` is only passed when `shouldShowResetAll` is true. Simplified to pass `onResetAllFilters` directly.

ppadti (commented):
Few things:
1. when we do just search -  I don't see a reset filters. So question, should we have a reset button?
2. Now the search bar description truncated, and cannot click on search bar after the place holder text to write something on search bar - check the attached Screen shot
Image

ppadti (changes requested):
Also, now the perf filters on landing page is below the category section -  is that the expected behaviour? the filters should be above the category sections no?

Image

ppadti (changes requested):
One more small thing: 
when we apply both filter and search or in an empty state  - when we click on reset filter, it doesn't reset/clear the search term. Its a regression after this refactor - It works on main.

ppadti (approved):
/lgtm
/approve
tested locally the toolbar works as expected.

PR #3080, opened 2026-08-11
fix(scripts): use portable sed -i syntax in gen_type_asserts.sh

rareddy:
## Description
BSD sed (macOS) requires a suffix argument to -i; GNU sed does not. Use sed -i.bak and remove the backup file afterward so the script works on both.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>


pboyd (commented):
/approve

PR #3082, opened 2026-08-12
fix: avoid dirty DB migration state in deploy_on_kind.sh

rareddy:
## Description

kubectl apply -k creates the server deployment with the base manifest's placeholder image, then a follow-up kubectl patch swaps in the real image and triggers a second rollout. If the placeholder-image pod is already running embedded DB migrations when it gets replaced, it can be killed mid-migration, leaving golang-migrate's dirty flag set and blocking every subsequent pod from starting.

Scale the deployment to 0 before patching the image, and only scale back up once the DB is confirmed available, so only one pod with the correct image is ever created.

This is to fix build errors like this https://github.com/kubeflow/hub/actions/runs/31533721347/job/93919678902?pr=3075

### DIFF
--- a/scripts/deploy_on_kind.sh
+++ b/scripts/deploy_on_kind.sh
@@ -37,6 +37,14 @@ else
 fi
 
 kubectl apply -k manifests/kustomize/overlays/$DEPLOY_MANIFEST_DB -n "$MR_NAMESPACE"
+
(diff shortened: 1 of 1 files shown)


pboyd (approved):
/lgtm
/approve

PR #3087, opened 2026-08-14
feat(catalog): SKC-110/111/114 skill source preview, marketplace, and…

rareddy:
… hardening

## Description
- SKC-110: wire assetType: skills into the shared sources/preview endpoint; SkillPreviewer clones one repository on demand, applies include/exclude filters, and returns AssetSourcePreviewResponse
- SKC-111: GET .../claude/marketplace.json renders indexed skills as a Claude Code plugin marketplace; every ref pinned to resolvedCommit; optional URL rewrite for internal/external mirrors via SKILL_CONTENT_MIRROR env vars
- SKC-114: extract loadRepoCredentials into shared helper (loader + previewer use identical credential path); token reaches git only via GIT_ASKPASS

Update KEP-0005 to document that branches are not accepted as refs; only tags, releases, and commit SHAs are allowed (reproducible installs).

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/catalog/internal/catalog/skillcatalog/preview.go
+++ b/catalog/internal/catalog/skillcatalog/preview.go
@@ -171,8 +171,9 @@ func filterMatchName(rs ResolvedSkill) string {
 	return skillFilterName(rs.Path, frontmatter)
 }
 
-// previewSkillName is the human-facing name shown for a previewed skill: its
-// frontmatter name, falling back to the directory name.
+// previewSkillName is the human-facing name shown for a previewed skill.
+// ParseSkillMD guarantees Name is non-empty for every resolved skill, so the
+// filterMatchName fallback is a defensive guard for unexpected nil Skill values.
 func previewSkillName(rs ResolvedSkill) string {
 	if rs.Skill != nil && rs.Skill.Name != "" {
 		return rs.Skill.Name
--- a/catalog/internal/catalog/skillcatalog/preview.go
+++ b/catalog/internal/catalog/skillcatalog/preview.go
@@ -1,0 +1,181 @@
+package skillcatalog
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"golang.org/x/sync/semaphore"
+	"k8s.io/apimachinery/pkg/util/yaml"
+
+	"github.com/kubeflow/hub/catalog/internal/catalog/basecatalog"
+	model "github.com/kubeflow/hub/catalog/pkg/openapi"
+)
+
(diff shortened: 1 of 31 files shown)


catalog/internal/catalog/skillcatalog/preview.go
```diff
@@ -0,0 +1,181 @@
...
+			errs = append(errs, fmt.Sprintf("%s@%s: %v", res.repo.URL, res.ref, res.err))
+			continue
+		}
+		resolved++
+		filter := filtersByURL[res.repo.URL]
+		for i := range res.skills {
+			rs := res.skills[i]
+			previews = append(previews, model.AssetPreviewResult{
+				Name:     previewSkillName(rs),
+				Included: filter.Allows(filterMatchName(rs)),
```
pboyd, line 124: What would happen if `filter` were `nil` here?

catalog/internal/catalog/skillcatalog/preview.go
```diff
@@ -0,0 +1,181 @@
...
+	return skillFilterName(rs.Path, frontmatter)
+}
+
+// previewSkillName is the human-facing name shown for a previewed skill: its
+// frontmatter name, falling back to the directory name.
+func previewSkillName(rs ResolvedSkill) string {
+	if rs.Skill != nil && rs.Skill.Name != "" {
+		return rs.Skill.Name
+	}
+	return filterMatchName(rs)
```
pboyd, line 181: Which directory name does this fall back to? Is it relative to the repo root, or does it include the repo name itself?

Basically, I'm wondering if this fallback works correctly when `SKILL.md` is in the repo root or only if it's in a sub-directory.

catalog/internal/catalog/skillcatalog/preview.go
rareddy: Allows handles nil safely — name_filter.go:127-129:
```
  func (f *NameFilter) Allows(name string) bool {
      if f == nil {
          return true  // nil filter = no restrictions = allow all
      }
      ...
  }
```

catalog/internal/catalog/skillcatalog/preview.go
rareddy: if it is in repo root, it will fallback to adopt to the YAML front matter name.

repo_resolver.go line 362
```
func skillFilterName(relDir, frontmatterName string) string {
	if relDir == "." || relDir == "" {
		return frontmatterName
	}
	return path.Base(relDir)
}
```
However looks I did one mistake is making `name` from YAML front matter not be must have according to specification. I fixed that with next commit, where name will never be nil

PR #3106, opened 2026-08-20
feat(catalog): wire skill catalog into deployment manifests

rareddy:
## Description
Mount the skill-catalog-sources ConfigMap and the optional git-credentials Secret into the catalog Deployment, and register the third --catalogs-path. Base ships an empty skill_catalogs list, so no source is cloned by default.

The UI RBAC needed for private git repos is an opt-in kustomize component rather than part of ui/base. It grants only get/update pinned by resourceNames and pre-creates the credentials Secret, so it never needs an unnamed create on secrets.

Also fix shared CatalogSource handling: scope GetDistinctSourceIDs to the model type_id, guard status-row deletion with allKnownSourceIDs in all four loaders so one plugin cannot delete another's row, and update catalog sources by explicit column map to avoid writing the identity column.

Assisted-by: Claude Sonnet 4.6
Signed-off-by: Ramesh Reddy <rareddy@redhat.com> 


pboyd (commented):
@rareddy, would you mind splitting this up into two PRs? The loader changes are probably fine, but they're unrelated and not trivial.

PR #3107, opened 2026-08-21
fix(catalog): don't delete a source's status if another plugin still …

rareddy:
…owns it

## Description
Each loader (agent, mcp, model, skill) deleted a source's shared CatalogSource status row as soon as its own entity type stopped referencing that source, without checking whether another plugin (model, mcp, agent, skill) still configured it. Since the status row is shared across all plugin types, this could delete a source's status out from under a plugin that still owns it, whenever two plugins configured the same source ID.

Gate the delete on allKnownSourceIDs — the union of every plugin's configured source IDs — in addition to each loader's own check.

Also wires the skill catalog's filter-options materialized view to refresh after sync/removal, so provider/category/label values reflect what's actually indexed instead of drifting until an unrelated catalog happens to trigger a refresh.

Assisted-by: Claude Sonnet 5


pboyd (commented):
/lgtm
/approve

PR #3108, opened 2026-08-21
fix(catalog): scope model source-id query, reject multi-repo inline s…

rareddy:
…kill sources

## Description
GetDistinctSourceIDs queried the shared ContextProperty table for name='source_id' with no filter on entity type. All catalog types (model, mcp, agent, skill) store a source_id property in that same table, so CatalogModelRepositoryImpl.GetDistinctSourceIDs could return source IDs belonging to other catalog types whenever they share a source ID. Join back to Context and scope by type_id so only model source IDs are returned.

Also reject skill source configs that specify more than one repository inline: the settings UI only edits and writes back repositories[0], so a multi-repo inline source would be silently truncated on the next UI save. Use `yamlCatalogPath` for shipped multi-repo defaults instead.

NOTE: the identity change was removed, that was there in previous commit as that was not needed

Assisted-by: Claude Sonnet 5

### DIFF
--- a/catalog/internal/catalog/modelcatalog/service/catalog_model.go
+++ b/catalog/internal/catalog/modelcatalog/service/catalog_model.go
@@ -270,38 +270,24 @@ func (r *CatalogModelRepositoryImpl) DeleteByID(id int32) error {
 }
 
 // GetDistinctSourceIDs retrieves all unique source_id values from catalog models.
-// This method queries the ContextProperty table to find distinct string_value entries
-// where the property name is 'source_id'.
+// The query is scoped to the model entity type via type_id so that source IDs
+// belonging to other catalog types (skills, MCP servers, agents) are not returned.
 func (r *CatalogModelRepositoryImpl) GetDistinctSourceIDs() ([]string, error) {
 	config := r.GetConfig()
-
 	var sourceIDs []string
 
-	// Execute the SQL query to get distinct source_id values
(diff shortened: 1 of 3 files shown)


catalog/internal/catalog/modelcatalog/service/catalog_model.go
```diff
@@ -270,38 +270,24 @@ func (r *CatalogModelRepositoryImpl) DeleteByID(id int32) error {
 }
 
 // GetDistinctSourceIDs retrieves all unique source_id values from catalog models.
-// This method queries the ContextProperty table to find distinct string_value entries
-// where the property name is 'source_id'.
+// The query is scoped to the model entity type via type_id so that source IDs
+// belonging to other catalog types (skills, MCP servers, agents) are not returned.
 func (r *CatalogModelRepositoryImpl) GetDistinctSourceIDs() ([]string, error) {
```
pboyd, line 275: I believe this is the fourth identical copy of this function. What do you think about combining them?

pboyd (commented):
One thing, but it's not a deal-breaker. /unhold if you want to merge

/lgtm
/approve
/hold

catalog/internal/catalog/modelcatalog/service/catalog_model.go
rareddy: I will get this as enhancement in separate PR. I was just following precedence for sure. /unhold

rareddy: /hold cancel

PR #3109, opened 2026-08-21
feat(catalog): add skill catalog manifests and KEP examples

rareddy:
## Description

Wire the skill catalog source into the catalog deployment: seed an empty skill-catalog-sources ConfigMap, mount it at /skill-catalog, pass an extra --catalogs-path, and optionally mount the
skill-catalog-git-credentials Secret. Declare an empty skill_catalogs section in the ODH default sources.

Add a skill-catalog-settings UI component with its RBAC. The ConfigMap permissions match what the model and MCP catalog settings pages need and stay within that shape: `get` on the shared default-catalog-sources (defaults are read to merge with user sources, never written) and `get`/`update` on skill-catalog-sources alone. Every rule is pinned with resourceNames and none grants `create`, since Kubernetes does not evaluate resourceNames for `create` and an unnamed `create` on secrets would let the ServiceAccount mint a service-account-token Secret and read it back; both objects the settings page writes are pre-created by the manifests. **One caveat carries over to the other catalogs:** skill-catalog-sources is kustomize-generated, so on a GitOps deployment the next sync reverts UI-added sources, this is accepted risk as other catalogs. An operator model solves this issue.

Also document the file form of a skill source (yamlCatalogPath) in KEP-0005 alongside the existing inline example.

Assisted-by: Claude Code

### DIFF
--- a/manifests/kustomize/options/catalog/base/deployment.yaml
+++ b/manifests/kustomize/options/catalog/base/deployment.yaml
@@ -32,6 +32,13 @@ spec:
       - name: mcp-sources
         configMap:
           name: mcp-catalog-sources
+      - name: skill-sources
+        configMap:
+          name: skill-catalog-sources
+      - name: skill-git-credentials
+        secret:
+          secretName: skill-catalog-git-credentials
+          optional: true
       containers:
         - name: catalog
           env:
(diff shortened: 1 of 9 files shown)


pboyd (commented):
/lgtm

I'll wait to approve this to give @dkuc a chance to respond.

PR #3111, opened 2026-08-22
refactor(catalog): hoist GetDistinctSourceIDs into GenericRepository

rareddy:
## Description

The model, MCP, skill, and agent repositories each carried a byte-identical copy of GetDistinctSourceIDs. Move it onto GenericRepository, deriving the entity and property table names from the type parameters and the join column from PropertyFieldName, and drop the four copies. All four repositories embed GenericRepository, so they satisfy their interfaces unchanged.

Addresses review feedback on #3108.

Assisted-by: Claude Code

### DIFF
--- a/internal/platform/db/repository/generic_repository.go
+++ b/internal/platform/db/repository/generic_repository.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kubeflow/hub/internal/platform/db/filter"
 	"github.com/kubeflow/hub/internal/platform/db/schema"
 	"github.com/kubeflow/hub/internal/platform/db/scopes"
+	"github.com/kubeflow/hub/internal/platform/db/utils"
 	platformerrors "github.com/kubeflow/hub/internal/platform/errors"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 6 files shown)


internal/platform/db/repository/generic_repository.go
```diff
@@ -534,6 +535,29 @@ func (r *GenericRepository[TEntity, TSchema, TProp, TListOpts]) GetConfig() Gene
 	return r.config
 }
 
+// GetDistinctSourceIDs retrieves all unique source_id property values for this
+// repository's entities. The query is scoped to the repository's own type_id so
+// that source IDs belonging to other catalog types are not returned.
+func (r *GenericRepository[TEntity, TSchema, TProp, TListOpts]) GetDistinctSourceIDs() ([]string, error) {
```
pboyd: `GenericRepository` is used in more places than the original four copies of this method, where it really makes no sense (e.g. `RegisteredModel` in the registry). Could this be in a wrapper around `GenericRepository` shared by the four models? Or perhaps a helper function called by a thin wrapper on those models?

rareddy: ok, reworked the PR to move to helper function now. Please review @pboyd

pboyd (approved):
Looks good to me. Thanks, @rareddy.

/lgtm
/approve

PR #3136, opened 2026-08-30
feat(ui-bff)!: switch BFF proxy targets to v1 APIs

### LINKED ISSUE #2994: Update UI BFF to target v1 API
Update the UI Backend-for-Frontend (BFF) to proxy requests to v1 API endpoints instead of alpha. Per KEP-0004: "The Python library and BFF will switch to the v1 endpoints as soon as they are available."

## Changes

**1\. Proxy target paths** — Update BFF proxy routes to target /api/model_registry/v1/, /api/model_catalog/v1/, /api/mcp_catalog/v1/

**2\. Parameter names** — Any BFF-constructed URLs use v1 path parameter names (e.g. registered_model_id)

**3\. Frontend** — Check if React frontend constructs API paths directly; update if needed

## Key files

  * clients/ui/bff/ — Go BFF server

  * clients/ui/bff/api/ — API proxy handlers

  * clients/ui/frontend/ — React frontend

myesmin:
## Summary
- Switches the UI BFF (`clients/ui/bff`) to proxy `/api/model_registry/v1/`, `/api/model_catalog/v1/`, `/api/mcp_catalog/v1/`, and `/api/agent_catalog/v1/` instead of the deprecated `v1alpha3`/`v1alpha1` paths, per KEP-0004.
- Adds `pkg/openapi-v1`, a new generated Go client for the v1 model-registry API, used only by the BFF. `pkg/openapi` (v1alpha3) is left untouched — it's the shared model-type package for the server internals (`internal/core`, `internal/converter`, `pkg/api`), the CSI plugin, and the inference-service controller, none of which are switching yet, and repointing it would break all of them.
- Swaps 13 BFF files from `pkg/openapi` to `pkg/openapi-v1`. No path/query parameter renames were needed in the BFF itself: it builds URLs by substituting ID values directly rather than by named parameters, so the KEP's `registeredmodelId` → `registered_model_id` style renames don't surface in this code.
- Fixes a real correctness gap found along the way: the catalog `source_id` → `sourceId` JSON field rename (per the KEP) wasn't reflected in the BFF's `Agent`/`CatalogModel`/`MCPServer` struct tags — left as-is, that field would have silently gone empty once traffic hit v1.
- Frontend needs no changes — confirmed it never constructs registry/catalog API paths directly, everything goes through the BFF's own separately-versioned `/api/v1/...` prefix.

Closes #2994

## Test plan
- [x] `make build/compile` (root) — clean
- [x] `make vet` (root) — clean (matches the project's own definition, which excludes a pre-existing unrelated warning in `internal/*/filter/parser.go`)
- [x] `golangci-lint run` (root) — 0 new issues; 79 pre-existing findings, all in `catalog/`, a directory untouched by this PR
- [x] `pkg/openapi-v1` — builds and vets clean standalone
- [x] BFF: `make build && make vet && make lint && make test` — all clean, full suite passing including real `envtest` (kubebuilder control plane), not skipped
- [x] `go mod tidy` clean at both root and BFF module level
- [ ] Manual verification against a live v1 model-registry/catalog server (not available in this environment)


clients/ui/Dockerfile
```diff
@@ -33,6 +33,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 COPY pkg/openapi /usr/src/app/pkg/openapi
+COPY pkg/openapi-v1 /usr/src/app/pkg/openapi-v1
```
ppadti, line 36: Same pkg/openapi-v1 copy is needed in Dockerfile.standalone — standalone builds use that file and will fail without it.

clients/ui/bff/internal/models/catalog_model_list.go
```diff
@@ -29,7 +29,7 @@ type CatalogModel struct {
 	Name                     string                            `json:"name"`
 	Provider                 *string                           `json:"provider,omitempty"`
 	Readme                   *string                           `json:"readme,omitempty"`
-	SourceId                 *string                           `json:"source_id,omitempty"`
+	SourceId                 *string                           `json:"sourceId,omitempty"`
```
ppadti, line 32: +1, to Paul's [comment here ](https://github.com/kubeflow/hub/pull/3136#issuecomment-5499622669)

ppadti (changes requested):
We also need to update the `local-deployment-guide.md` file - it still references v1alpha3 curl example

clients/ui/bff/internal/mocks/static_data_mock.go
```diff
@@ -225,6 +226,47 @@ func newCustomProperties() map[string]openapi.MetadataValue {
 	}
 }
 
+func newCustomPropertiesV1() map[string]openapiv1.MetadataValue {
```
manaswinidas, line 229: `newCustomProperties` / `newCustomPropertiesV1` are near-identical. Prefer one shared fixture + thin typed wrappers so keys don't drift.

clients/ui/bff/internal/api/artifacts_handler.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/kubeflow/hub/ui/bff/internal/integrations/httpclient"
 
 	"github.com/julienschmidt/httprouter"
-	"github.com/kubeflow/hub/pkg/openapi"
+	"github.com/kubeflow/hub/pkg/openapi-v1"
```
manaswinidas, line 12: Import is `pkg/openapi-v1` but symbols stay `openapi.*` (both packages named `openapi`). Alias as `openapiv1` everywhere (or generate with `--package-name openapiv1`) so the client in use is obvious.

clients/ui/bff/internal/repositories/model_registry.go
```diff
@@ -138,10 +138,10 @@ func (m *ModelRegistryRepository) ResolveServerAddress(clusterIP string, httpPor
 	// In federated mode, if external address is available, use it
 	if isFederatedMode && externalAddressRest != "" {
 		// External address is assumed to be HTTPS
-		url := fmt.Sprintf("%s://%s/api/model_registry/v1alpha3", protocol, externalAddressRest)
+		url := fmt.Sprintf("%s://%s/api/model_registry/v1", protocol, externalAddressRest)
```
manaswinidas, line 141: Path `/api/model_registry/v1` inlined twice. Match catalog: extract `ModelRegistryAPIPath` const.

clients/ui/bff/internal/models/catalog_model_list.go
manaswinidas, line 32: `SourceId` here vs `SourceID` on Agent/McpServer for the same `sourceId` JSON field — pick one Go name.

Also: `CustomProperties` still typed from `pkg/openapi` (v1alpha3) while this proxies catalog `v1`. Confirm `MetadataValue` JSON is identical, else silent decode drift.

clients/ui/frontend/src/__tests__/cypress/cypress/tests/mocked/modelCatalog/modelCatalogAllModelsView.cy.ts
```diff
@@ -62,7 +62,7 @@ const initIntercepts = ({
                 mockCatalogModel({
                   name: `${label.toLowerCase()}-model-${i + 1}`,
                   // eslint-disable-next-line camelcase
```
manaswinidas, line 64: Dead `eslint-disable-next-line camelcase` — key is already `sourceId`. Drop it (same on L94).

clients/ui/bff/internal/repositories/model_catalog.go
```diff
@@ -10,9 +10,9 @@ import (
 
 const (
 	ModelCatalogServiceName = "model-catalog"
-	ModelCatalogAPIPath     = "/api/model_catalog/v1alpha1"
-	McpCatalogAPIPath       = "/api/mcp_catalog/v1alpha1"
-	AgentCatalogAPIPath     = "/api/agent_catalog/v1alpha1"
+	ModelCatalogAPIPath     = "/api/model_catalog/v1"
```
manaswinidas, line 13: v1 path swap has no unit assert on these consts (registry covered only via fixture URL strings). Small table test on path consts / `ResolveServerAddress` would lock the cutover.

Same for BFF `sourceId` JSON tags — a marshal round-trip test would catch `source_id` regressions.

manaswinidas (commented):
Non-blocking follow-ups (DRY, import alias clarity, naming consistency, dead eslint, a couple test gaps). Nothing merge-blocking in the diff itself — please rebase first (`mergeable: dirty`).

manaswinidas:
Also, there are merge conflicts

PR #3159, opened 2026-09-01
feat(catalog): add endpoint to read and clear a source's persisted status

pboyd:
## Description

The catalog backend only refreshes a source's status when it reloads
config from the mounted configmap, which can lag a frontend save by up
to two minutes. Until then, FindSources keeps returning the old status,
so a user who just fixed (or broke) a source configuration sees a
stale "available" instead of the outcome of their change.

Add GET and DELETE on /api/model_catalog/v1/sources/{source_id}/status
so the frontend can read the persisted status and clear it right after
saving, rather than showing a stale value while waiting on the next
reload.

## How Has This Been Tested?
Unit tests, end-to-end tests in this PR, and manually tested on a dev cluster.

## Merge criteria:


adysenrothman (approved):
/lgtm
/approve

PR #3198, opened 2026-09-10
fix(catalog): stop paginate from panicking on an empty artifact list

### LINKED ISSUE #3197: Performance artifact pagination panics on an empty list when recommendations=true and nextPageToken is set
**Describe the bug**

`PerformanceArtifactService.paginate` in `catalog/internal/catalog/modelcatalog/performance_artifact_service.go` runs only when `recommendations=true`, because recommendations are computed over the full list and paged in memory afterwards. It locates the cursor with a `for index = range list.Items` loop and then slices `list.Items[index+1:]`. When the list is empty the loop body never runs, `index` stays 0, and the slice is `[1:0]`, which panics.

The list is empty whenever the Pareto filter removed every artifact, or the model has none, so a client that follows a `nextPageToken` from an earlier response and lands on such a list gets a dropped connection instead of an empty page. There is no recover middleware in the catalog server, so `net/http` logs the panic and closes the connection.

Two smaller things sit on the same lines. A token whose ID is not in the list returns an empty page, which is fine but accidental (the loop leaves `index` at the last element). An undecodable token returns the first page, as the comment above it intends.

**To Reproduce**

1. Build `&dbmodels.ListWrapper[sharedmodels.CatalogMetricsArtifact]{Items: []sharedmodels.CatalogMetricsArtifact{}}`.
2. Call `service.paginate(list, 10, &token)` with `token := scopes.CreateNextPageToken(1, "")`.
3. Panic: `runtime error: slice bounds out of range [1:0]`.

Over HTTP: `GET /api/model_catalog/v1/sources/{source_id}/models/{model_name}/artifacts/performance?recommendations=true&nextPageToken=<any valid token>` against a model whose artifacts are all filtered out.

**Expected behavior**

An empty page with no `nextPageToken`, the same result a cursor past the end of a non-empty list already produces.

**Environment**

- *Kubernetes version*: not involved, reproduces in a unit test on `main` (`48594f24`)

**Additional context**

Fix is to start after the cursor only when it is found and otherwise return nothing, which also removes the accidental behaviour for a missing ID. PR with a reproducing test follows.

DsThakurRawat:
## Description

`PerformanceArtifactService.paginate` slices `list.Items[index+1:]` after a `for index = range list.Items` loop. When the list is empty the loop never runs, `index` stays 0, and the slice is `[1:0]`, which panics. That path is reached by `GET .../artifacts/performance?recommendations=true&nextPageToken=<token>` whenever the Pareto filter has removed every artifact, and the catalog server has no recover middleware, so the client gets a closed connection instead of an empty page.

The loop now records where the cursor was found and starts after it; a cursor that is not in the list, including the empty list, yields no items. That is what a cursor past the end of a non-empty list already returned, so no observable behaviour changes apart from the panic. Undecodable tokens still return the first page, as the existing comment intends.

Fixes #3197

## How Has This Been Tested?

`TestPerformanceArtifactService_PaginateCursor` covers four cases: cursor in the list, cursor over an empty list (this one panicked before the change, checked by stashing the fix and running the test), cursor not in the list, and an undecodable token. `make -C catalog test`, `make test`, `make vet` and `make lint` pass locally on Go 1.27 with golangci-lint v2.12.2 from `make bin/golangci-lint`. The catalog module has no lint target, so I ran the same binary on `catalog/internal/catalog/modelcatalog/...` with `--new-from-rev origin/main`: 0 issues on the changed lines.

## Merge criteria:

### DIFF
--- a/catalog/internal/catalog/modelcatalog/performance_artifact_service.go
+++ b/catalog/internal/catalog/modelcatalog/performance_artifact_service.go
@@ -400,14 +400,16 @@ func (s *PerformanceArtifactService) paginate(list *dbmodels.ListWrapper[sharedm
 	}
 
 	if cursor != nil {
-		var index int
-		for index = range list.Items {
-			id := list.Items[index].GetID()
+		// A cursor that is not in this list, or an empty list, has nothing after it.
+		start := len(list.Items)
+		for i := range list.Items {
+			id := list.Items[i].GetID()
 			if id != nil && *id == cursor.ID {
+				start = i + 1
 				break
 			}
 		}
-		list.Items = list.Items[index+1:]
+		list.Items = list.Items[start:]
 	}
 
 	list.NextPageToken = ""
--- a/catalog/internal/catalog/modelcatalog/performance_artifact_service_test.go
+++ b/catalog/internal/catalog/modelcatalog/performance_artifact_service_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubeflow/hub/catalog/internal/catalog/modelcatalog/models"
 	sharedmodels "github.com/kubeflow/hub/catalog/internal/db/models"
 	dbmodels "github.com/kubeflow/hub/internal/platform/db/entity"
+	"github.com/kubeflow/hub/internal/platform/db/scopes"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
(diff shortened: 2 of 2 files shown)


pboyd (approved):
Thanks for the PR, @DsThakurRawat!

/lgtm
/approve

PR #3203, opened 2026-09-11
fix(catalog): stop preview pagination repeating or looping on duplicate names

### LINKED ISSUE #3202: Preview pagination repeats or never advances when two entries share a name
**Describe the bug**

`filterAndPaginate` in `catalog/internal/server/openapi/v1/api_model_catalog_service_service.go` (same code in `v1alpha1`) pages preview results with the in-memory `paginator`. Its cursor identifies the last item of a page by `SortValue(ORDERBYFIELD_ID)`, and for `ModelPreviewResult` and `AssetPreviewResult` that is the `Name`. Preview names are not unique, so when the item at a page boundary has an identical neighbour, `Paginate` resumes after the first item with that name rather than after the one it actually served. With `pageSize` smaller than the run of duplicates the token never changes and a client that follows `nextPageToken` loops forever; with a larger page the duplicate is served again on the next page.

Three configurations produce duplicate names in a preview, as far as I can read the code: a YAML catalog with two entries under the same name; a Hugging Face source whose `includedModels` lists both `org/*` and `org/model`, since `FetchModelsForPreview` appends the results of each pattern without deduplicating; and a skill repository previewed at more than one ref, since each ref's skills are reported under the frontmatter name while the loader indexes them separately under the composite name from `skillEntityName`.

**To Reproduce**

1. Write a YAML catalog with two models named `acme/granite-3b` and one named `acme/llama-7b`.
2. Call `PreviewCatalogSource` with `type: yaml`, that catalog as `catalogData`, and `pageSize=1`, then keep passing back the `nextPageToken` it returns.

Observed on `main` (`b15fb0f2`), through the handler in a unit test:

```
--- pageSize=1
request 1: status=200 size=1 items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 2: status=200 size=1 items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 3: status=200 size=1 items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
--- pageSize=2
request 1: status=200 size=2 items=[acme/granite-3b acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 2: status=200 size=2 items=[acme/granite-3b acme/llama-7b] nextPageToken=""
```

With `pageSize=1` the third model is never reached. With `pageSize=2` the second granite entry is returned on both pages, so a client sees four items where the catalog has three.

**Expected behavior**

Every preview entry is returned exactly once and the walk ends, whatever the page size.

**Environment**

- *Kubernetes version*: not involved, reproduces in a unit test on `main` (`b15fb0f2`)

**Additional context**

`FindSources` uses the same paginator but keys on source IDs, which are unique, so it is not affected.

Two places the fix could live. The in-memory cursor could carry the item's position among equal names so `Paginate` resumes in the right spot for every caller, at the cost of a token format change. Or the preview builders could collapse repeated names before paging, which matches what the loader does for YAML and Hugging Face models but would under-report skills, since those are indexed once per ref. Happy to send a PR with a reproducing test for whichever you prefer.

DsThakurRawat:
## Description

Fixes #3202.

The in-memory `paginator` in `catalog/internal/server/openapi/v1/pagination.go` (and its `v1alpha1` copy) identifies the last item of a page by its sort value and `SortValue(ORDERBYFIELD_ID)`. For `ModelPreviewResult` and `AssetPreviewResult` that ID is the name, and preview names repeat: a YAML catalog can list a model twice, a Hugging Face source can match a model through both `org/*` and `org/model`, and a skill repository previewed at two refs reports every skill under its frontmatter name once per ref. `Paginate` resumed after the first item with the cursor's name, so with a page smaller than the run of duplicates the token never changed and a client following `nextPageToken` looped forever; with a larger page the duplicate was served on two pages.

The cursor now also carries `Skip`, the number of identical items that precede the item it points at, and `Paginate` resumes after that exact one. The token becomes `skip:value:id`, with the id last so stored names such as `sourceId:modelName` still round-trip. The previous `value:id` form is still decoded, so a client mid-walk during an upgrade keeps working instead of getting a 400 from the new replica.

`FindSources` uses the same paginator but keys on source IDs, which are unique, so its behaviour is unchanged apart from the token format.

## How Has This Been Tested?

New tests, added to both `v1` and `v1alpha1`:

- `TestPaginateDuplicateIDs` walks three items with a repeated name through the paginator at page sizes 1, 2 and 3 and requires every item exactly once, with the token advancing on each page.
- `TestPreviewCatalogSourceDuplicateNames` does the same through `PreviewCatalogSource` with a YAML catalog, at page sizes 1 and 2.
- `TestDecodeStringCursorLegacyFormat` covers the old `value:id` token and a current token whose id contains a colon.

Before the fix, the handler walk on `main` (`b15fb0f2`) produced:

```
--- pageSize=1
request 1: items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 2: items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 3: items=[acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
--- pageSize=2
request 1: items=[acme/granite-3b acme/granite-3b] nextPageToken="OmFjbWUvZ3Jhbml0ZS0zYg=="
request 2: items=[acme/granite-3b acme/llama-7b] nextPageToken=""
```

Run locally on this branch: `make -C catalog test` (all 17 packages ok, against the Postgres test container), `go vet ./internal/server/openapi/...`, and `golangci-lint run ./internal/server/openapi/... --new-from-rev origin/main` with 0 issues.

This change was written with the assistance of Claude Code (noted in the commit trailer); the fix and the tests were reviewed and run by hand.

## Merge criteria:

### DIFF
--- a/catalog/internal/server/openapi/v1/api_model_catalog_service_service_test.go
+++ b/catalog/internal/server/openapi/v1/api_model_catalog_service_service_test.go
@@ -2690,3 +2690,44 @@ func TestClearSourceStatus(t *testing.T) {
 func strPtr(s string) *string {
 	return new(s)
 }
+
+func TestPreviewCatalogSourceDuplicateNames(t *testing.T) {
+	service := newTestServiceWithSources(map[string]*model.CatalogModel{})
+	catalogData := `
+models:
+  - name: "acme/granite-3b"
+    description: "first entry"
+  - name: "acme/granite-3b"
+    description: "second entry, same name"
+  - name: "acme/llama-7b"
+    description: "third entry"
+`
+	want := []string{"acme/granite-3b", "acme/granite-3b", "acme/llama-7b"}
+
+	for _, pageSize := range []string{"1", "2"} {
+		t.Run("pageSize="+pageSize, func(t *testing.T) {
+			var got []string
+			token := ""
+			for range 10 {
+				configFile := writeTempYAML(t, "config", "type: yaml\n")
+				catalogDataFile := writeTempYAML(t, "catalogdata", catalogData)
+
+				resp, err := service.PreviewCatalogSource(context.Background(), configFile, pageSize, token, "", catalogDataFile)
+				require.NoError(t, err)
+				require.Equal(t, http.StatusOK, resp.Code)
+				body, ok := resp.Body.(model.CatalogSourcePreviewResponse)
+				require.True(t, ok, "expected CatalogSourcePreviewResponse, got %T", resp.Body)
+
+				for _, item := range body.Items {
+					got = append(got, item.Name)
+				}
+				if body.NextPageToken == "" {
+					break
+				}
+				require.NotEqual(t, token, body.NextPageToken, "nextPageToken did not advance")
+				token = body.NextPageToken
+			}
+			assert.Equal(t, want, got)
+		})
+	}
+}
--- a/catalog/internal/server/openapi/v1/pagination.go
+++ b/catalog/internal/server/openapi/v1/pagination.go
@@ -116,13 +116,16 @@ func (p *paginator[T]) Token() string {
 func (p *paginator[T]) Paginate(items []T) ([]T, *paginator[T]) {
 	startIndex := 0
 	if p.cursor != nil {
+		seen := 0
 		for i, item := range items {
-			itemValue := item.SortValue(p.OrderBy)
-			id := item.SortValue(model.ORDERBYFIELD_ID)
-			if id != "" && id == p.cursor.ID && itemValue == p.cursor.Value {
+			if !p.cursor.matches(item, p.OrderBy) {
+				continue
+			}
+			if seen == p.cursor.Skip {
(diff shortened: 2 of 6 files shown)


pboyd (commented):
Thanks for the fix, @DsThakurRawat.

/ok-to-test

pboyd (approved):
Now that I revisit this code, I think we would have a problem if `Value` ever had a colon, but we don't have any I'm aware of today and the problem existed before this change. So, it looks good to me.

/lgtm
/approve

PR #3213, opened 2026-09-13
fix(catalog): keep float precision in pagination cursors

### LINKED ISSUE #3212: Cursor pagination on a double-valued property repeats or skips items and can loop forever
**Describe the bug**

`CreateNextPageToken` in `internal/platform/db/scopes/paginate.go` encodes a `*float64` sort value with `fmt.Sprintf("%.15f", ...)`. A float64 can need 17 significant digits to round-trip, so any value with more precision than that is rounded inside the token. `buildWhereClause` then resumes the walk with `(value > cursor OR (value = cursor AND id > lastId))`, and because the rounded cursor no longer equals the stored value, the boundary row is served again or skipped, together with every row whose value sits between the two numbers.

The catalog takes this path whenever models or artifacts are ordered by a custom double property (`artifacts.<name>.double_value` or `<name>.double_value`). That is what the frontend sends for the "lowest cold start" sort on the gallery (`ModelCatalogGalleryView.tsx`) and for the cold start and latency sorts on the performance artifacts of a card (`ModelCatalogCardBody.tsx`). Those are measured numbers with full float precision, for example `0.8523489932885906`, whose token value is `0.852348993288591`.

Three shapes, measured on `main` (`e43c096d`) through `CatalogModelRepository.List` against the Postgres test container:

- `DESC`, four models with 16-digit values, `pageSize=2`: page 1 is right, then every following page returns the same two models and `nextPageToken` never empties. A client that follows the token loops forever.
- `ASC`, same data: one model is never returned.
- Three models sharing one computed value (`0.7999999999999999`, the mean of 0.7, 0.8 and 0.9), `pageSize=1`: the token carries `0.800000000000000`, page 2 is empty, and two of the three models are never returned.

Values with a few decimals, such as `0.9512`, page correctly because `%.15f` reproduces them exactly, which is why the existing sort tests pass.

**To Reproduce**

1. Create four catalog models, each with a model artifact whose `accuracy` property is `0.8523489932885906`, `0.8523489932885907`, `0.7345678901234567` and `0.9123456789012345`.
2. `GET /api/model_catalog/v1/models?orderBy=artifacts.accuracy.double_value&sortOrder=DESC&pageSize=2`, then keep passing back `nextPageToken`. The router does not validate `orderBy` against the enum, so the custom property reaches the repository as typed.

Output of the repository-level walk (models are named `m<case>-<index>`, the loop was cut at 20 pages):

```
DESC pageSize=2: page1=[m1-3 m1-1] page2=[m1-1 m1-0] page3=[m1-1 m1-0] ... page20=[m1-1 m1-0]   m1-2 never returned, no page is last
ASC  pageSize=2: page1=[m0-2 m0-0] page2=[m0-3]                                                    m0-1 never returned
shared mean, ASC pageSize=1: page1=[m2-0] page2=[]                                                 m2-1 and m2-2 never returned
```

`GetAllModelArtifacts` builds its token through the same helper in `catalog/internal/db/service/catalog_artifact.go` when `orderBy` is `<property>.double_value`, so the artifacts endpoint pages the same way.

**Expected behavior**

Every row is returned exactly once and the walk ends, for any float value the database can store.

**Environment**

- *Kubernetes version*: not involved, reproduces in a unit test on `main` (`e43c096d`)

**Additional context**

Only the catalog passes a `*float64` into `CreateNextPageToken`; the core repositories build their tokens from ids and timestamps and are not affected. The `%.15f` formatting arrived with #1818.

The fix is to encode the float with `strconv.FormatFloat(v, 'g', -1, 64)`, the shortest string that parses back to the same float64, which Postgres reads back exactly in the `WHERE` clause. A PR with a round-trip test on the token and a page-walk test on the model repository follows.

DsThakurRawat:
## Description

Fixes #3212.

`CreateNextPageToken` in `internal/platform/db/scopes/paginate.go` wrote a `*float64` sort value into the token with `%.15f`. A float64 can need 17 significant digits, so a value such as `0.8523489932885906` became `0.852348993288591` in the token, and the `WHERE` clause that resumes the walk (`value > cursor OR (value = cursor AND id > lastId)`) no longer matched the row the token pointed at. Ordering by a double custom property then repeated or skipped rows at each page boundary, and in descending order the token never emptied, so a client following `nextPageToken` looped forever. Values with a few decimals, such as `0.9512`, were unaffected, which is why the existing sort tests passed.

The token now carries `strconv.FormatFloat(v, 'g', -1, 64)`, the shortest string that parses back to the same float64, which Postgres reads back exactly. Integer and string values are encoded as before. Only the catalog passes floats into this helper (models and artifacts ordered by `<property>.double_value`); the core repositories build their tokens from ids and timestamps and are not affected.

## How Has This Been Tested?

New tests:

- `TestCreateNextPageTokenFloatRoundTrip` in `internal/platform/db/scopes` encodes nine values, decodes the token and requires the parsed float to equal the original. On `main` it fails for `0.8523489932885906`, `0.7999999999999999` and `-0.8523489932885906`.
- `TestListDoublePropertyOrderingPagination` in `catalog/internal/catalog/modelcatalog/service` walks every page of `CatalogModelRepository.List` ordered by `artifacts.accuracy.double_value`, ascending and descending, with four full-precision values at `pageSize=2` and three models sharing `0.7999999999999999` at `pageSize=1`, and requires every model exactly once and the walk to end. On `main` the ascending case drops a model and the descending case never ends.

Run locally on this branch: `go vet` on both packages, `golangci-lint run ... --new-from-rev origin/main` with 0 issues, and `go test ./internal/platform/db/... ./catalog/...` with all 21 packages passing against the Postgres test container.

This change was written with the assistance of Claude Code (noted in the commit trailer); the fix and the tests were reviewed and run by hand.

## Merge criteria:


pboyd (commented):
Thanks for the PR, @DsThakurRawat.

/lgtm
/approve

PR #3221, opened 2026-09-13
feat(catalog): add RFC 8594 deprecation headers to alpha catalog APIs

### LINKED ISSUE #3218: Catalog alpha endpoints send no Deprecation or Sunset headers, unlike the registry proxy
**Describe the bug**

KEP-0004 §5 says both alpha versions are served alongside v1 for nine months and that "during this period, alpha responses include `Deprecation` and `Sunset` HTTP headers per RFC 8594, and the server logs a warning on every alpha endpoint call."

The registry side does this: `internal/server/middleware/deprecation.go` sets `Deprecation`, `Sunset` and a `Link: rel="successor-version"` header on `/api/model_registry/v1alpha3/` responses and logs a rate-limited warning, and `cmd/proxy.go` mounts it when `--alpha-sunset-date` is given.

The catalog side does not. `catalog/cmd/catalog.go` wraps the router only in `ValidationMiddleware`, `MountRoutes` in `catalog/internal/plugin/server.go` adds only the chi logger and CORS, and nothing under `catalog/` references `Deprecation` or `Sunset`. So `/api/model_catalog/v1alpha1/`, `/api/mcp_catalog/v1alpha1/` and `/api/agent_catalog/v1alpha1/` answer with no hint that they are deprecated, and there is no log line to find alpha callers with.

**To Reproduce**

1. Run the catalog from `main` (`e43c096d`).
2. `curl -i http://localhost:8080/api/model_catalog/v1alpha1/sources` and compare with `curl -i` against the registry proxy started with `--alpha-sunset-date 2027-06-01` on `/api/model_registry/v1alpha3/registered_models`.

The registry response carries the three headers; the catalog response carries none.

**Expected behavior**

The same headers and warning on the catalog's alpha paths, with `Link` pointing at each plugin's v1 base path (`/api/model_catalog/v1/`, `/api/mcp_catalog/v1/`, `/api/agent_catalog/v1/`; the skill plugin has no alpha).

**Environment**

- *Kubernetes version*: not involved

**Additional context**

The middleware in `internal/server/middleware/deprecation.go` hard-codes the registry prefix and successor. Generalising it to a list of alpha-to-successor prefix pairs lets both servers share it; the catalog would mount it from `MountRoutes` behind a `--alpha-sunset-date` flag (or an env var, matching how the catalog reads its leader settings). One side effect worth stating: the UI BFF still targets the alpha catalog paths until #3136 lands (#2994), so its calls would carry the headers and produce one warning per minute in the catalog log, which is harmless and arguably the point. I will send the PR.

DsThakurRawat:
## Description

Fixes #3218.

KEP-0004 §5 says both alpha API versions carry `Deprecation` and `Sunset` headers during the deprecation window. The registry proxy does this behind `--alpha-sunset-date`; the catalog server did not, so `/api/model_catalog/v1alpha1/`, `/api/mcp_catalog/v1alpha1/` and `/api/agent_catalog/v1alpha1/` responses gave no hint that they are deprecated.

The deprecation middleware moves from `internal/server/middleware` to the shared `internal/platform/server/middleware` and takes a list of deprecated prefixes with their successors instead of a hard-coded registry pair. The proxy keeps passing its single pair, so its behaviour and its tests are unchanged apart from the import. The catalog plugin server derives one pair per plugin whose `Version()` is an alpha (`/api/<name>_catalog/v1alpha1/` to `/api/<name>_catalog/v1/`) and mounts the middleware in `MountRoutes` when `--alpha-sunset-date` (or `CATALOG_ALPHA_SUNSET_DATE`) is set. The skill plugin is v1 only and contributes nothing. Without the flag nothing changes.

## How Has This Been Tested?

- `TestDeprecationMiddleware_MultiplePaths` and `_NoPaths` cover the generalised middleware; the existing registry tests were kept and now pass the registry pair explicitly.
- `TestAlphaDeprecatedPaths` checks the derivation for a default-path plugin, a `BasePathProvider` plugin, and a v1-only plugin.
- `TestMountRoutesAlphaDeprecationHeaders` mounts a v1alpha1 test plugin through the real `MountRoutes` and asserts `Deprecation: true`, `Sunset`, and `Link: </api/test_catalog/v1/>; rel="successor-version"` on its route, none of them on `/healthz`, and none at all when the date is unset.
- `go build ./...`, `go vet` and `golangci-lint run --new-from-rev origin/main` on the touched packages report nothing; `catalog --help` lists the flag and `--alpha-sunset-date 13-09-2026` fails at startup with the format message.

One side effect worth knowing: the UI BFF still targets the alpha catalog paths until #3136 lands, so with the flag set its calls carry the headers and produce one warning line per minute in the catalog log.

This change was written with the assistance of Claude Code (noted in the commit trailer); the code and the tests were reviewed and run by hand.

## Merge criteria:


pboyd (commented):
@DsThakurRawat, thanks for the PR.

/lgtm
/approve

DsThakurRawat:
Thanks for  merge :)

PR #3222, opened 2026-09-13
fix: reject an orderBy the v1 spec does not declare with 400

### LINKED ISSUE #3217: v1 APIs accept an unknown orderBy and silently sort by the default, where KEP-0004 requires 400
**Describe the bug**

KEP-0004 §2 says: "Sorting on an unsupported field must return `400 Bad Request`. Some endpoints currently fall back to the default sort order, which masks client bugs and makes behavior unpredictable. The `400` response is correct; enforce it everywhere." On `main` (`e43c096d`) the v1 endpoints of both APIs still fall back.

The generated v1 routers cast the query string straight into `OrderByField` without checking it (`internal/server/openapi/v1/api_model_registry_service.go`, `catalog/internal/server/openapi/v1/api_model_catalog_service.go`, and the MCP, agent and skill routers; none calls `IsValid` or `NewOrderByFieldFromValue`), the service layers pass the string through, and `PaginateWithOptions` in `internal/platform/db/scopes/paginate.go` maps anything outside its allowed-columns map to the default column. So a typo in a client sorts by id and returns 200.

**To Reproduce**

Through the repositories the v1 handlers call, with two rows named `b-model` and `a-model`:

```
catalog CatalogModelRepository.List on Postgres
  orderBy="NAME"              err=<nil> order=[a-model b-model]
  orderBy="BOGUS"             err=<nil> order=[b-model a-model]
  orderBy="id; DROP TABLE x"  err=<nil> order=[b-model a-model]

core RegisteredModelRepository.List on MySQL
  orderBy="BOGUS"             err=<nil> order=[b-model a-model]
```

Or `GET /api/model_catalog/v1/models?orderBy=BOGUS` and `GET /api/model_registry/v1/registered_models?orderBy=BOGUS`: both return 200 with the default ordering.

(The injection-shaped value is not a security problem, the map lookup fails and the default column is used; it is here to show the value is never inspected.)

**Expected behavior**

A `400` with a message naming the field, on every v1 list endpoint, for any `orderBy` value outside the endpoint's enum. For the catalog that enum has to include the documented custom-property forms `<property>.<double_value|int_value|string_value>` and `artifacts.<property>.<type>`, which the frontend sends today (`ModelCatalogGalleryView.tsx`, `ModelCatalogCardBody.tsx`), and `RECOMMENDED` where it applies. The alpha endpoints keep their current behaviour for the deprecation window, as the KEP describes.

**Environment**

- *Kubernetes version*: not involved, reproduces in a unit test on `main` (`e43c096d`)

**Additional context**

The check belongs in the v1 service layer rather than the generated routers: `buildListOption` in `internal/server/openapi/v1/api_model_registry_service_service.go` for the registry, and a shared validator for the catalog v1 services. Two side effects to plan for: the v1 fuzz fixtures being migrated in #3165 may expect 200 for arbitrary `orderBy`, and the catalog Python client's docstring lists `ACCURACY` as an `order_by` value that the API does not know. I will send the PR.

DsThakurRawat:
## Description

Fixes #3217.

KEP-0004 §2 says v1 must answer `400` when asked to sort on an unsupported field. Both servers still fell back to the default column: the generated routers cast `orderBy` without checking it, the services passed it through, and the pagination scope substituted the default for anything outside its column map.

- **Registry v1**: `buildListOption` rejects a value outside the `OrderByField` enum (`CREATE_TIME`, `LAST_UPDATE_TIME`, `ID`).
- **Catalog v1**: a small validator (`order_by.go`) checks each list endpoint against what its spec declares. Models accept the shared enum, `ACCURACY` (which `ListModels` maps to `artifacts.overall_average.double_value`), `<property>.<value_column>` and `artifacts.<property>.<value_column>`, exactly the forms the repository honours and the frontend sends. MCP servers and tools, agents and skills accept the shared enum; skills and agent artifacts keep their case-insensitive match.
- The existing `Unsupported orderBy field` case in the v1 model test flips from 200 to 400; the alpha copy of that test is untouched.
- The catalog Python client docstring said `ACCURACY` was an ordinary value; it now says which values are standard and gives a property example.

Two places are deliberately left as they are, and I would like a maintainer call on them:

1. **Artifact listings** (`GetAllModelArtifacts`, `GetAllModelPerformanceArtifacts`): the v1 spec's own `artifactOrderBy` description says "if an invalid format is used (e.g., `accuracy` without `.value_type`), it falls back to ID ordering", and the hardware configuration table in the UI sends bare metric names (`ttft_mean`, `replicas`) to the performance endpoint today. Enforcing 400 there would contradict the spec text and break that table once the BFF moves to v1 (#3136). If the KEP wins, the spec text and the frontend need to change together.
2. **The shared enum** includes `RECOMMENDED`, which only models honour. MCP, agent and skill endpoints now accept it because their spec declares it; narrowing the enum per plugin is a spec change I can follow up with if wanted.

The alpha APIs keep their fallback for the deprecation window, as the KEP describes.

## How Has This Been Tested?

- `TestBuildListOptionOrderBy` (registry): every enum value passes, `BOGUS`, `name`, `id` and an injection-shaped string return `ErrBadRequest`.
- `TestOrderBySpecValidate` (catalog): a table over the model, enum and case-insensitive specs, including `ACCURACY`, both property forms, a dashed property name, an unknown value column, an empty property name, four segments, and lower-case field names.
- `TestFindModelsOrderByValidation`, `TestFindMCPServersRejectsUnknownOrderBy`, `TestFindSkillsRejectsUnknownOrderBy` go through the handlers with the existing mocks.
- `go test` on `catalog/internal/server/openapi/v1`, `internal/server/openapi/v1`, and both alpha packages passes; `golangci-lint run --new-from-rev origin/main` on the touched packages reports nothing.
- The fuzz suite (`schemathesis.toml` has `positive_data_acceptance` on) generates only declared enum values for enum-typed `orderBy` parameters, so it is unaffected; the free-string artifact parameters were left out for the reason above.

This change was written with the assistance of Claude Code (noted in the commit trailer); the code and the tests were reviewed and run by hand.

## Merge criteria:


DsThakurRawat:
This looks good to me 
waiting for Reviewers

pboyd (commented):
Thanks, @DsThakurRawat.

/lgtm
/approve

PR #3238, opened 2026-09-14
Add storage to MCP runtimeMetadata (catalog server + BFF DTO)

dkuc:
## Description

MCP servers with a read-only root filesystem need writable paths (e.g. `/tmp`, `/app/logs`), but the catalog had no way to declare them, so operators had to patch each `MCPServer` by hand after deployment. The catalog server also silently dropped any unknown `runtimeMetadata` field, so the value never reached consumers.

This change:

- Adds a `storage` array to the `MCPRuntimeMetadata` OpenAPI schema (`catalog-v1.yaml` / `catalog.yaml` and the `mcp-v1.yaml` / `mcp.yaml` source plugins).
- Regenerates the catalog Go model, adding the new `MCPStorageMount` / `MCPStorageSource` types.
- Mirrors the field in the UI BFF catalog DTO (`McpRuntimeMetadata.Storage` with `McpStorageMount` / `McpStorageSource`).

`runtimeMetadata` is persisted as a single JSON blob, so the field round-trips with **no DB migration**. Consumers (e.g. the RHOAI `MCPServer` converter) can then map these mounts onto `spec.config.storage` so writable scratch space is provisioned automatically while the root filesystem stays read-only.

This is one of three coordinated changes; the downstream `odh-dashboard` consumer and the `model-metadata-collection` catalog data are updated in companion PRs.

## How Has This Been Tested?

- Regenerated the OpenAPI Go model and confirmed the build passes (`make build` in `catalog/`).
- Verified the new `storage` field round-trips through the catalog server as part of the `runtimeMetadata` JSON blob.

## Merge criteria:

- All the commits have been _signed-off_ (To pass the `DCO` check)


pboyd (commented):
/lgtm

@manaswinidas, do you want to review the BFF changes in this PR?

clients/ui/bff/internal/models/mcp_server_catalog.go
```diff
@@ -1,6 +1,9 @@
 package models
 
-import "github.com/kubeflow/hub/pkg/openapi"
+import (
+	"github.com/kubeflow/hub/pkg/openapi"
+	corev1 "k8s.io/api/core/v1"
+)
```
manaswinidas, line 6: Couples catalog DTOs to k8s API version. Prefer free-form maps / thin local structs here; keep `corev1` at the K8s convert boundary.

clients/ui/bff/internal/models/mcp_server_catalog.go
```diff
@@ -129,6 +132,21 @@ type McpRuntimeMetadata struct {
 	Capabilities                 *McpRuntimeMetadataCapabilities    `json:"capabilities,omitempty"`
 	McpPath                      *string                            `json:"mcpPath,omitempty"`
 	Prerequisites                *McpPrerequisites                  `json:"prerequisites,omitempty"`
+	Storage                      []McpStorageMount                  `json:"storage,omitempty"`
```
manaswinidas, line 135: BFF adds `storage`, but `mcpServerCatalogTypes.ts` and `mod-arch.yaml` do not. Sync FE + mod-arch, or note deferral to the companion PR.

clients/ui/bff/internal/models/mcp_server_catalog.go
```diff
@@ -129,6 +132,21 @@ type McpRuntimeMetadata struct {
...
+}
+
+// McpStorageMount matches an MCPServer spec.config.storage entry.
+type McpStorageMount struct {
+	Path        string           `json:"path"`
+	Permissions string           `json:"permissions,omitempty"`
```
manaswinidas, line 141: Catalog uses `*string` enum; plain `string` drops null/unset. Prefer `*string` or a typed enum pointer like sibling fields.

clients/ui/bff/internal/models/mcp_server_catalog.go
```diff
@@ -129,6 +132,21 @@ type McpRuntimeMetadata struct {
...
+	Source      McpStorageSource `json:"source"`
+}
+
+type McpStorageSource struct {
+	Type      string                        `json:"type"`
+	EmptyDir  *corev1.EmptyDirVolumeSource  `json:"emptyDir,omitempty"`
+	ConfigMap *corev1.ConfigMapVolumeSource `json:"configMap,omitempty"`
+	Secret    *corev1.SecretVolumeSource    `json:"secret,omitempty"`
```
manaswinidas, line 149: Catalog OpenAPI is free-form `object` (`map[string]interface{}`). Typed `corev1.*VolumeSource` silently drops unknown keys and can remashal `sizeLimit` differently. Match catalog wire shape, or tighten the OpenAPI schema.

clients/ui/bff/internal/models/mcp_server_catalog.go
```diff
@@ -129,6 +132,21 @@ type McpRuntimeMetadata struct {
...
+
+// McpStorageMount matches an MCPServer spec.config.storage entry.
+type McpStorageMount struct {
+	Path        string           `json:"path"`
+	Permissions string           `json:"permissions,omitempty"`
+	Source      McpStorageSource `json:"source"`
+}
+
+type McpStorageSource struct {
+	Type      string                        `json:"type"`
```
manaswinidas, line 146: Nit: bare `string` while this file uses typed enums elsewhere. Consider `McpStorageSourceType` consts.

manaswinidas (commented):
BFF DTO review (`clients/ui` only) — a few contract/maintenance notes.

clients/ui/bff/internal/models/mcp_server_catalog.go
dkuc: Removed the corev1 types from the BFF DTO; storage sources now use local free-form maps.

clients/ui/bff/internal/models/mcp_server_catalog.go
dkuc: Added storage and its mount/source types to both the frontend types and mod-arch.yaml.

clients/ui/bff/internal/models/mcp_server_catalog.go
dkuc: Changed permissions to an optional typed pointer so an unset value stays unset.

clients/ui/bff/internal/models/mcp_server_catalog.go
dkuc: Switched volume options to free-form maps and added a round-trip test for emptyDir: {}, sizeLimit, and an unknown key.

clients/ui/bff/internal/models/mcp_server_catalog.go
dkuc: Added McpStorageSourceType constants for EmptyDir, ConfigMap, and Secret.

manaswinidas (approved):
/lgtm
/approve 

From the FE side

PR #3241, opened 2026-09-15
prevent loading unpermitted gated models into catalog

adysenrothman:
## Merge criteria:

If you have UI changes


catalog/internal/catalog/modelcatalog/hf_catalog.go
```diff
@@ -574,6 +574,17 @@ func (p *hfModelProvider) getModelsFromHF(ctx context.Context) ([]ModelProviderR
 			continue
 		}
 
+		// Check if model is gated without access granted - block from loading
+		if p.shouldBlockGatedModel(ctx, modelInfo) {
```
dkuc: Could we distinguish an access denial from a failed check here? `checkGatedAccess` also returns `false` for 429s, 5xx responses, and network errors. Skipping those models without returning an error makes orphan cleanup delete their existing catalog entries and marks the source `available`. I reproduced this with 429 and 503 responses. We should report these failures and preserve the affected records until access can be checked successfully.

catalog/internal/catalog/modelcatalog/hf_catalog.go
```diff
@@ -656,12 +667,40 @@ func (p *hfModelProvider) fetchModelInfo(ctx context.Context, modelName string)
...
+	accessType := deriveHFAccessType(modelInfo)
+	
+	// Only block if model is gated
+	if !strings.HasPrefix(accessType, "gated_") {
+		glog.V(3).Infof("Model %s is not gated (access_type=%s), will not block", modelInfo.ID, accessType)
+		return false
+	}
+	
+	// Check if access has been granted
+	hasAccess := p.checkGatedAccess(ctx, modelInfo.ID)
```
dkuc: Could we reuse this result when building the model metadata? `populateFromHFInfo` calls `checkGatedAccess` again, so allowed gated models now make two requests. I reproduced a 200 followed by a 403: the model was still returned, but with `hf_gated_access_granted=false`. Using one result for filtering and metadata would keep them consistent and avoid the extra request.

catalog/internal/catalog/modelcatalog/hf_catalog.go
```diff
@@ -574,7 +593,28 @@ func (p *hfModelProvider) getModelsFromHF(ctx context.Context) ([]ModelProviderR
...
-		record := p.convertHFModelToRecord(ctx, modelInfo, modelName)
+		// Check if model is gated without access granted - block from loading.
+		// Transient failures (429, 5xx, network errors) are treated as failed
+		// models so existing catalog entries are preserved until the check succeeds.
+		// gatedAccess is forwarded to convertHFModelToRecord so the
+		// hf_gated_access_granted property uses the same result.
+		shouldBlock, gatedAccess, gatedErr := p.shouldBlockGatedModel(ctx, modelInfo)
+		if gatedErr != nil {
+			glog.Errorf("Failed to check gated access for %s: %v", modelName, gatedErr)
+			failedModels = append(failedModels, modelName)
```
dkuc, line 604: This still removes existing models when a sync partially succeeds. `FailedModels` changes the status, but orphan cleanup only keeps names from successful records. I reproduced this with one public model loading and a gated model returning 429/503: the gated record was deleted and the source marked `partially-available`.

Could we preserve the failed model names during cleanup, or skip cleanup for an incomplete batch? The new test needs to exercise the loader to catch this.

catalog/internal/catalog/modelcatalog/hf_catalog.go
```diff
@@ -574,7 +593,28 @@ func (p *hfModelProvider) getModelsFromHF(ctx context.Context) ([]ModelProviderR
...
+			continue
```
dkuc, line 605: Could we keep retries running when these checks fail during the initial load? If all selected models hit this branch, `Models` receives zero records plus an error and returns before starting its ticker.

I reproduced this with an initial 503: no provider channel is created, so a healthy next response is never tried. The source stays in error until a reload, restart, or leadership change. We should report the initial failure while keeping periodic syncs active.

dkuc (approved):

PR #3252, opened 2026-09-18
fix(controller): prevent nil pointer dereference in InferenceService URL reconciliation

### LINKED ISSUE #3251: fix(controller): nil pointer dereference panic in InferenceServiceController.Reconcile
### Bug
During reconciliation of an `InferenceService` custom resource, `InferenceServiceController.Reconcile` attempts to retrieve the current URL property from Model Registry in `pkg/inferenceservice-controller/controller.go`:

```go
mrCurrentIvcUrl := mrIs.CustomProperties["url"].MetadataStringValue.GetStringValue()

### PR DESCRIPTION
**Pull request by vjkumar2756**
Fixes #3251

During reconciliation of an `InferenceService` CR, line 202 in `pkg/inferenceservice-controller/controller.go` retrieved the URL custom property from Model Registry directly:

```go
mrCurrentIvcUrl := mrIs.CustomProperties["url"].MetadataStringValue.GetStringValue()

### CODE REVIEW

**Comment by vjkumar2756** [comment]
@fege  @Philip-Carneiro  could you please review this pr

File: pkg/inferenceservice-controller/controller.go [hunk 1]
```diff
@@ -199,7 +199,12 @@ func (r *InferenceServiceController) Reconcile(ctx context.Context, req ctrl.Req
 			return ctrl.Result{}, fmt.Errorf("unable to find InferenceService with id %s in model registry: %w", mrIsvcId, err)
 		}
 
-		mrCurrentIvcUrl := mrIs.CustomProperties["url"].MetadataStringValue.GetStringValue()
+		var mrCurrentIvcUrl string
+		if mrIs.CustomProperties != nil {
+			if urlMeta, ok := mrIs.CustomProperties["url"]; ok && urlMeta.MetadataStringValue != nil {
+				mrCurrentIvcUrl = urlMeta.MetadataStringValue.GetStringValue()
+			}
+		}
```
pboyd, line 207: Are you sure these guards are necessary? If the `CustomProperties` map is `nil` then `CustomProperties["url"]` will return the zero value for `MetadataValue` and not panic. The zero value for `MetadataValue` has a `nil` `MetadataStringValue`, but since `GetStringValue()` returns `""` on an `nil` instance that doesn't appear to be a problem either.

Was this based on a crash you experienced? There might be something else going on.

pboyd (commented):
If you still want to keep this PR, can you rebase it from `main` and remove the unrelated commits?

vjkumar2756:
@pboyd 
Hi
I well resolve the conflicts in this PR. Could you please take a quick look for this pr.
thanks for review

pboyd:
> @pboyd Hi I well resolve the conflicts in this PR. Could you please take a quick look for this pr. thanks for review

@vjkumar2756, thanks for cleaning up the commit history, it's much easier to review now.

Please see my earlier comment. This change seems unnecessary to me, but please let me know if I'm wrong.

vjkumar2756:
Hi @pboyd 
You are completely right  after double checking, I realize this change was unnecessary . already safe from any nil pointer panics.
if any suggestions ?

pboyd:
> Hi @pboyd You are completely right after double checking, I realize this change was unnecessary . already safe from any nil pointer panics. if any suggestions ?

@vjkumar2756, thank you for checking it. Since it's not addressing a real problem, I'm going to close this PR.

